PR #30, opened 2021-08-06
Test sv2 ffi interface

rrybarczyk:


protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
```diff
@@ -227,3 +227,15 @@ impl<'a, T> From<Seq064K<'a, T>> for Vec<T> {
         v.0
     }
 }
+
+impl<'a, T> From<Vec<T>> for Seq0255<'a, T> {
```
Fi3, line 231: This is not safe as it can create invalid sequences, should be TryFrom and should check the input len or should stay under `#[cfg(feature = "prop_test")]` .

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
```diff
@@ -227,3 +227,15 @@ impl<'a, T> From<Seq064K<'a, T>> for Vec<T> {
...
+    fn from(v: Vec<T>) -> Self {
+        Seq0255(v, PhantomData)
+    }
+}
+
+impl<'a, T> From<Vec<T>> for Seq064K<'a, T> {
```
Fi3, line 237: This is not safe as it can create invalid sequences, should be TryFrom and should check the input len or should stay under `#[cfg(feature = "prop_test")]` .

protocols/v2/binary-sv2/no-serde-sv2/codec/src/lib.rs
```diff
@@ -120,7 +120,7 @@ impl<'a> From<Vec<u8>> for EncodableField<'a> {
 }
```
Fi3, line 120: There is no need to change this file. This file should remain exactly equal to the one that is the main branch.

protocols/v2/binary-sv2/serde-sv2/src/primitives/byte_arrays/b016m.rs
```diff
@@ -4,7 +4,7 @@ use alloc::vec::Vec;
 use core::convert::TryFrom;
```
Fi3, line 4: There is no need to change this file. This file should remain exactly equal to the one that is the main branch.

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
```diff
@@ -11,7 +11,9 @@ pub struct Seq0255<'s, T: Serialize + TryFromBSlice<'s> + Clone> {
     seq: Option<Seq<'s, T>>,
     data: Option<Vec<T>>,
 }
+// impl<'s> Eq for Seq064K<'s, B016M<'s>> {}
```
Fi3, line 14: There is no need to change this file. This file should remain exactly equal to the one that is the main branch.

protocols/v2/subprotocols/common-messages/src/lib.rs
```diff
@@ -20,3 +27,77 @@ pub extern "C" fn _c_export_channel_endpoint_changed(_a: ChannelEndpointChanged)
...
+}
+
+#[cfg(feature = "prop_test")]
+impl SetupConnection<'static> {
+    pub fn from_random(g: &mut Gen) -> Self {
+        let protocol = setup_connection::Protocol::MiningProtocol;
+        // let protocol = setup_connection::Protocol::JobDistributionProtocol;
+        // let protocol = setup_connection::Protocol::TemplateDistributionProtocol;
+        // let protocol = setup_connection::Protocol::JobNegotiationProtocol;
+        let mut endpoint_host = Gen::new(255);
```
Fi3, line 47: Not sure if generate a new random generator inside `from_random`  is the right way to go or if the passed generator `g` should be used. Is not clear from quickcheck documentation we should ask somewhere which is the right pattern and put a comment with a reference to the answer.

protocols/v2/subprotocols/template-distribution/src/new_template.rs
```diff
@@ -1,6 +1,8 @@
 #[cfg(not(feature = "with_serde"))]
 use alloc::vec::Vec;
 #[cfg(not(feature = "with_serde"))]
+use alloc::vec;
```
Fi3, line 4: Remove warnings

protocols/v2/subprotocols/template-distribution/src/new_template.rs
```diff
@@ -45,16 +47,18 @@ pub struct NewTemplate<'decoder> {
...
     pub coinbase_tx_locktime: u32,
     /// Merkle path hashes ordered from deepest.
     #[cfg_attr(feature = "with_serde", serde(borrow))]
     pub merkle_path: Seq0255<'decoder, U256<'decoder>>,
 }
 
+
 #[repr(C)]
 #[cfg(not(feature = "with_serde"))]
+#[derive(Clone, Debug)]
```
Fi3, line 61: No need to add this. Remove it.

protocols/v2/subprotocols/template-distribution/src/new_template.rs
```diff
@@ -69,6 +73,36 @@ pub struct CNewTemplate {
     merkle_path: CVec2,
 }
 
+impl CNewTemplate {
```
Fi3, line 76: This block should be removed.

Fi3 (changes requested):
There are 2 issues related to the PR that should be added as they are not a priority but still very
relevant:

### First issue:
The below should compile:
cargo test --features prop_test --features with_serde 

### Second issue
When implementing Arbitrary we should also implement shrink.

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
rrybarczyk: I added this in to resolve this error I was getting:
```
error[E0277]: the trait bound `Seq0255<'decoder, U256<'decoder>>: core::cmp::Eq` is not satisfied
   --> protocols/v2/subprotocols/template-distribution/src/new_template.rs:56:5
    |
56  |     pub merkle_path: Seq0255<'decoder, U256<'decoder>>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `core::cmp::Eq` is not implemented for `Seq0255<'decoder, U256<'decoder>>`
```
Is there an alternative to fix this error?

protocols/v2/subprotocols/template-distribution/src/new_template.rs
rrybarczyk: Delete this entire CNewTemplate impl block?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
Fi3: I tried to remove it and I wasn't  getting errors. But if you get that error leave it there.

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
rrybarczyk: I replaced the `From` impl with the `TryFrom` impl. Can you check out the impl and make sure it is correct? If it isn't, just lmk how I should change it.

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
rrybarczyk: I replaced the `From` impl with the `TryFrom` impl. Can you check out the impl and make sure it is correct? If it isn't, just lmk how I should change it.

protocols/v2/subprotocols/common-messages/src/lib.rs
rrybarczyk: Is the below explanation accurate? If so, I will create an issue on quickcheck. If not, just lmk how I should modify it.

### Question: Idiomatic way to specify the length of an arbitrary vector


I have a struct with a field containing a data type that represents a vector of 255 bytes:
```
struct Foo {
  bar: B0255
}
```

Where

```
struct B0255<'b>(Inner<'b>);

enum Inner<'a> {
   Ref(&'a [u8]),
   Owned(Vec<u8>),
}
```

To test this struct, I have created a new struct and implemented `Arbitrary`, which calls a function `from_random` which is implemented for `Foo`.

```
struct RandomFoo(Foo);
impl Arbitrary from RandomFoo {
  fn arbitrary(g: &mut Gen) -> Self {
    RandomFoo(Foo::from_random(g))
  }
}
```

Currently, I am passing in the generator to the `from_random` function and generating the 255 byte vectors as:

```
impl Foo {
  fn from_random(g: &mut Gen) -> Self {
    let bar: B0255 = Vec<u8>::arbitrary(g).try_into().unwrap();
    Foo { bar }
  }
}
```


Alternatively, I can implement `from_random` as:

```
impl Foo {
  fn from_random(g: &mut Gen) -> Self {
    let mut bar: B0255 = Gen::new(255);
    bar: B0255 = Vec::<u8>::arbitrary(&mut bar).try_into().unwrap();
    Foo { bar }
  }
}
```

Which, if either, `from_random` implementation is the idiomatic way to generate a vector of a specific size?

Fi3: I would change the first example with:
```rust
impl Foo {
  fn from_random(g: &mut Gen) -> Self {
    
    let mut bar = Vec<u8>::arbitrary(g);
    bar.truncate(255);
    let bar: B0255 = bar.try_into().unwrap();
    Foo { bar }
  }
}
```
I truncate the vec to be sure to not panic with a too big vec. Testing if `try_into` return the right error when called with a too big value would be the best approach but not sure if it is possible as `arbitrary` need to return `Self` (I would ask also that in the issue) .

I would also add that:
1. we need `from_random` cause `Arbitrary` need to be implemented in another crate
2. `Arbitrary` need to be implemented for structs that can contains a lot different primitives (max bytes 64K, max bytes 255, exactly 32 bytes ecc ecc)
3. When `from_random` is implemented for fixed size primitives (eg U256) instead of `truncate` we use `resize(255, 0)` is that correct?

Fi3:
> When I run `$ cargo test --features prop_test`, I get the following warning for every `decoder.next_frame();`:
> 
> ```
> warning: unused `Result` that must be used
>    --> protocols/v2/sv2-ffi/src/lib.rs:644:9
>     |
> 644 |         decoder.next_frame();
>     |         ^^^^^^^^^^^^^^^^^^^^^
>     |
>     = note: `#[warn(unused_must_use)]` on by default
>     = note: this `Result` may be an `Err` variant, which should be handled
> ```
> 
> Do you have a suggestion on how I can fix these warnings?

Did you try `let _ = decoder.next_frame();` If it do not work just put `#[allow(unused_must_use)]` no need to handle the result as we know that it is an error and we know that is the correct behavior.

protocols/v2/subprotocols/common-messages/src/lib.rs
rrybarczyk: https://github.com/BurntSushi/quickcheck/issues/297

rrybarczyk:
> > When I run `$ cargo test --features prop_test`, I get the following warning for every `decoder.next_frame();`:
> > ```
> > warning: unused `Result` that must be used
> >    --> protocols/v2/sv2-ffi/src/lib.rs:644:9
> >     |
> > 644 |         decoder.next_frame();
> >     |         ^^^^^^^^^^^^^^^^^^^^^
> >     |
> >     = note: `#[warn(unused_must_use)]` on by default
> >     = note: this `Result` may be an `Err` variant, which should be handled
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you have a suggestion on how I can fix these warnings?
> 
> Did you try `let _ = decoder.next_frame();` If it do not work just put `#[allow(unused_must_use)]` no need to handle the result as we know that it is an error and we know that is the correct behavior.

Yes that fixed it! Ty.

PR #119, opened 2021-12-14
[WIP] Misc warning fixes

rrybarczyk:


rrybarczyk:
This PR
- Adds a `toolchain.toml` file in the root directory specifying the repo's toolchain, version 1.53.0
- Fixes some warnings generated from more recent cargo clippy versions, NOT toolchain version 1.53.0
- Allows dead-code to pass in GH A clippy. This is bad. This is done because the version of clippy in GH Actions `rust.yaml` kept resulting in dead-code warnings, when it should not. Running `cargo clippy` with the 1.53.0 toolchain locally does not yield these warnings. I thought that adding the `toolchain: 1.53.0` in the `steps` block of the `rust.yaml` GH Actions file would fix this, it did not. I then thought adding the `toolchain.toml` in the root directory would fix this, it did not. So for now, we ignore dead-code, but this should be changed eventually to get clippy working the right way in GH Actions.

PR #139, opened 2022-03-29
Style: Rename free_encoder to flush_encoder

### LINKED ISSUE #135: Rename free_encoder to flush_encoder
While working on the template provider in C, the [free_encoder ffi](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/sv2-ffi/src/lib.rs#L312) name is slightly confusing because my first thought was, this is used to free memory but its used to free up the encoder to allow the bytes in the buffer to be overwritten.

A better name imho would be `flush_encoder` as its more aligned with C/Linux APIs: https://linux.die.net/man/3/fflush.

Eunoia1729:
Fixes #135


pavlenex:
Hey @Eunoia1729 thanks for your contributions to the project 🙏 
We have a Discord channel https://discord.gg/fsEW23wFYs where we all connect and discuss Stratum V2. If you'd like to feel free to join us, would be great to connect with you outside of GitHub.

PR #164, opened 2022-06-10
Use FFI-safe errors + fix prop test errs

rrybarczyk:
Fix property based testing errors.

Run command:
```
% cargo test --features prop_test
```

- Fixed `TryInto` error
- Added `BinaryError` and `CodecError` variants to `Sv2Error` enum
- Added `NonPrimitiveTypeCannotBeEncoded`, `ValueExceedsMaxSize`, and `SeqExceedsMaxSize` to `binary_sv2` `Error` enum and removed the `Todo` variant
- Fixed `template-distribution` test by replacing `256` with `32` for `merkle_path_inner`

TODO ([issue here](https://github.com/stratum-mining/stratum/issues/167)):
Create a `CError` enum that matches the `binary_sv2` `Error` enum, but is FFI safe. Use this `CError` in `sv2-ffi` `lib.rs` in the `pub unsafe extern "C" fn encode` function.

### DIFF
--- a/protocols/v2/sv2-ffi/sv2.h
+++ b/protocols/v2/sv2-ffi/sv2.h
@@ -416,14 +416,6 @@ void free_submit_solution(CSubmitSolution s);
 #include <ostream>
 #include <new>
 
-enum class Sv2Error {
-  MissingBytes,
-  EncoderBusy,
-  Todo,
-  Unknown,
-  InvalidSv2Frame,
-};
-
 struct DecoderWrapper;
 
 struct EncoderWrapper;
@@ -500,6 +492,32 @@ struct CSv2Message {
     SetupConnection_Body setup_connection;
     SetupConnectionError_Body setup_connection_error;
     SetupConnectionSuccess_Body setup_connection_success;
+  };
+};
+
+struct Sv2Error {
+  enum class Tag {
+    BinaryError,
+    CodecError,
+    EncoderBusy,
+    InvalidSv2Frame,
+    MissingBytes,
+    PayloadTooBig,
+    Unknown,
+  };
+
+  struct BinaryError_Body {
+    CError _0;
+  };
+
+  struct CodecError_Body {
+    Error _0;
+  };
+
+  Tag tag;
+  union {
+    BinaryError_Body binary_error;
+    CodecError_Body codec_error;
   };
 };
 


rrybarczyk:
@Fi3, the tests compile after updating the calls to the `from_message` fn in `framing_sv2` to pass the `channel_msg` variable, all the tests compile.

There is one test that is failing in `sv2_ffi/lib.rs`, which is [`tests::encode_with_c_new_template_id`](https://github.com/stratum-mining/stratum/blob/537562494ed00bdccd21d20600fbffcc31d73ccd/protocols/v2/sv2-ffi/src/lib.rs#L708).

The error is: 
```
---- tests::encode_with_c_new_template_id stdout ----
thread 'tests::encode_with_c_new_template_id' panicked at 'called `Result::unwrap()` on an `Err` value: Todo', protocols/v2/subprotocols/template-distribution/src/lib.rs:75:84
```

This tells me that an `unwrap` is being called somewhere that is returning an error that is marked as a `todo`. Perhaps it is the [`Sv2Error::Todo` variant](https://github.com/stratum-mining/stratum/blob/537562494ed00bdccd21d20600fbffcc31d73ccd/protocols/v2/sv2-ffi/src/lib.rs#L276)?

To try and determine which `unwrap` is leading to the unimplemented error, I began commenting out the test code. I ended up commenting out the entire fn and just returning a `true`. As such:
```
#[quickcheck_macros::quickcheck]
 fn encode_with_c_new_template_id(message: RandomNewTemplate) -> bool {
    true
}
```
However, this still results in the same error. So it does not seem to be anything in this particular test, but rather something with how we are setting up the prop tests? 

I am unsure how to continue. Any ideas?

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/mod.rs
```diff
@@ -50,7 +50,7 @@ impl<'a> B016M<'a> {
     }
 }
 
-use core::convert::{TryFrom, TryInto};
```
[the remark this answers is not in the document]
rrybarczyk: Fixed by changing `256` -> `32` for  `merkle_path_inner`.

protocols/v2/sv2-ffi/src/lib.rs
```diff
@@ -271,12 +272,48 @@ pub enum CResult<T, E> {
...
 pub enum Sv2Error {
+    BinaryError(binary_sv2::CError),
+    CodecError(codec_sv2::Error),
+    /// Errors on oversized payload
+    /// TODO: Make `PayloadTooBig(String)` and pass in the `message` (which is type `T`)
+    PayloadTooBig,
     MissingBytes,
     EncoderBusy,
-    Todo,
     Unknown,
```
Fi3, line 282: Can we remove it?

protocols/v2/sv2-ffi/src/lib.rs
```diff
@@ -271,12 +272,48 @@ pub enum CResult<T, E> {
...
+            MissingBytes => write!(f, "Missing expected bytes"),
+            EncoderBusy => write!(f, "Encoder is busy"),
+            Unknown => write!(f, "Unknown error occurred"),
+        }
+    }
+}
+
+impl From<binary_sv2::CError> for Sv2Error {
+    fn from(e: binary_sv2::CError) -> Sv2Error {
+        Sv2Error::BinaryError(e.into())
```
Fi3, line 307: Remove the into() as in unecessary

protocols/v2/sv2-ffi/src/lib.rs
Fi3: cloned it and checked I confirm that we can remove the Unknown

PR #188, opened 2022-07-26
[WIP] Remove serde dependencies

joshuaahuang:


pavlenex:
@joshuaahuang  seems this pr would de-clutter code a bit, @xraid pointed it would be useful to have it but seems it has conflicts and the tests are stuck in loop, do you plan to get back to this one? cc @rrybarczyk

rrybarczyk:
Closing because out of date and because we want to continue to keep the `serde` dependency for the `message-generator` crate

PR #191, opened 2022-08-02
Update JobNegotiation mining_job_token

4ss0:
Use u32 instead of B0_255 to semplify implementation, 2^32 tokens per
proxy are enough.


ccdle12:
I agree that 2^32 is enough but isn't this against the specs? 

B0_255 has a specific requirement for serialization and deserialization. You could have a length of bytes anywhere between 0-255. As u32 it will always serialize to the length of a u32.

```
1-byte length L, unsigned integer 8-bits, followed by a sequence of L bytes. Allowed range of length is 0 to 255.
```

PR #195, opened 2022-08-17
Error handling for `codec_sv2` + `noise_sv2`

rrybarczyk:


protocols/v2/codec-sv2/Cargo.toml
```diff
@@ -10,6 +10,7 @@ license = "MIT"
 
 [dependencies]
 serde = { version = "1.0.89", default-features = false, optional = true }
+snow = "0.8"
```
Fi3, line 13: code- sv2 should depend on snow only when is compiled with the with noise flag. You can make it optional or you could reexport snow from noise-sv2 I think that the latter is a better option

utils/network-helpers/src/noise_connection_tokio.rs
```diff
@@ -156,8 +156,11 @@ impl Connection {
         let fourth_message = receiver_incoming.recv().await.unwrap();
         let mut fourth_message: HandShakeFrame = fourth_message.try_into().unwrap();
         let fourth_message = fourth_message.payload().to_vec();
+        dbg!(&fourth_message);
```
Fi3, line 159: remove or comment

protocols/v2/codec-sv2/Cargo.toml
rrybarczyk: this dependency was added only to support the new `codec_sv2::error::Error::SnowError` variant and i was wondering if there was a better alternative than importing the package here. i like the idea of re-exporting from `noise-sv2`, and will change to do that

Issue #71, opened 2021-11-17
add test for the other examples


Fi3:

Fi3:
Add tests that build and verify the correct execution of what is in ./examples
This test should be called by github actions

darricksee:
More discussion about tests:
1.) for the ping pong examples they just infinite loop - I'm thinking maybe we add an argument -test which makes them only ping/pong a set number of times and maybe panic if they don't receive a ping/pong?
2.) I see that the github actions builds the 2 ping pong examples and the sv1-client-and-server example - did you want to build all the maintained (non-deprecated) examples (ping/pong, interop-cpp x 2, sv1-client-and-server)
3.) For the interop-cpp example(s) that also infinite loops - would you recommend a similar approach as I proposed for the ping pong examples?

fi3 — Today at 12:40 PM
yep just a test on CI that build it run it for lets say 10 sec and if it do not panic it pass.

rockcoolsaint:
Is anybody working on this issue? I would love to give it a go @darricksee

GitGab19:
Closing this as stale

Issue #70, opened 2021-11-17
add test for interop cpp (it should try to build it and run the example)


Fi3:

Fi3:
This test should be called by github actions

rrybarczyk:
@darricksee did you tackle this one?

darricksee:
Yeah! I have it in my branch

darricksee:
Can you assign it to me

pavlenex:
Done @darricksee I assume none of the opened PR's that you have currently are related to this specific issue?

darricksee:
Yeah - #220 contains a test for interop cpp

GitGab19:
Closing this as stale

PR #233, opened 2022-09-16
71 - Added github actions for running examples

darricksee:
This fulfills both #70 and #71.

### DIFF
--- a/examples/ping-pong-without-noise/src/main.rs
+++ b/examples/ping-pong-without-noise/src/main.rs
@@ -5,23 +5,27 @@ use async_std::{
     prelude::*,
     task,
 };
-use std::time;
+use std::{env, net::SocketAddr, time};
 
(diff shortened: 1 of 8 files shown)


darricksee:
These run locally using `act` but we probably shouldn't merge this to `main` until we get a 👍 from github that they won't shut down our repo for mining. They don't seem to like the `sv1-client-and-server` example.

examples/ping-pong-without-noise/src/main.rs
```diff
@@ -5,48 +5,70 @@ use async_std::{
     prelude::*,
     task,
 };
-use std::time;
+use std::{env, net::SocketAddr, time};
 
-const ADDR: &str = "127.0.0.1:34254";
+//Pick any unused port
+const ADDR: &str = "127.0.0.1:0";
```
darricksee, line 11: To avoid port collisions let's just pick an available port.

PR #240, opened 2022-09-20
Fixes the deadlock in the ping/pong test

darricksee:
This fixes the [deadlock in the ping/pong test](https://github.com/stratum-mining/stratum/actions/runs/3089282863).

I also am removing the binaries in `deps` that aren't needed


rrybarczyk:
#241 replaces this

PR #241, opened 2022-09-21
 Fixes the deadlock in the ping/pong test

rrybarczyk:
@darricksee is unable to run GitHub Actions. Opening this PR for to replace their #240 PR, but will run GitHub Actions


rrybarczyk:
Creates Issue #249

PR #247, opened 2022-09-23
71 - Attempt to fix client/server deadlock example in CI

darricksee:
I think the problem with github is the lack of cpu power that we are running these tests with - the server/client example doesn't sleep between lock drop/acquire - I think there is just 1 thread that always gets the lock so nothing progresses and the cpu is pegged which makes github angry.

It'd be great if someone who doesn't have github actions block could run this action to see if it fixes the issue.


rrybarczyk:
Closed w #248

PR #255, opened 2022-09-26
253 fix client server ex

rustynirvana:
An attempt to get the client/server example running in ubuntu based on the ping/pong fix.


protocols/v1/src/error.rs
```diff
@@ -6,6 +6,7 @@ pub enum Error {
     Method(MethodError),
     InvalidJsonRpcMessageKind,
     InvalidSubmission,
+    InvalidState(String),
```
rustynirvana, line 9: I added this for debugging - LMK if you don't want to keep it or if you had something else in mind.

PR #299, opened 2022-10-06
translator: use base58 auth key string

### LINKED ISSUE #297: TProxy pub key in config should be string instead of array
Currently `upstream_authority_pubkey` is given in the config as an array of bytes [here](https://github.com/stratum-mining/stratum/blob/914ca6eb4a2576deb94b5bea6d193e4012a9af8f/roles/translator/proxy-config.toml#L4).
Needs to be read in as a string like the sv2-proxy config [here](https://github.com/stratum-mining/stratum/blob/914ca6eb4a2576deb94b5bea6d193e4012a9af8f/roles/v2/mining-proxy/proxy-config.toml#L1).

ccdle12:
fixes #297

Replaces the auth pubkey decimal array in the proxy config file with the base58 encoded string and deserializes the key.

### DIFF
--- a/roles/translator/proxy-config.toml
+++ b/roles/translator/proxy-config.toml
@@ -1,7 +1,7 @@
 # Upstream connection information
 upstream_address = "127.0.0.1"
 upstream_port = 34254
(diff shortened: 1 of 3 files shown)


ccdle12:
@rrybarczyk was this already fixed in?: a1472c00c562c506aa4ee379264f0ddeb240ea31

pavlenex:
> @rrybarczyk was this already fixed in?: [a1472c0](https://github.com/stratum-mining/stratum/commit/a1472c00c562c506aa4ee379264f0ddeb240ea31)

Confirmed with Rachel we actually added this in already, hence we agreed to close this one.

Issue #207, opened 2022-08-25
Follow rust-lightnings custom logging pattern to support logging in roles crate


rrybarczyk:
- The `roles` binaries should have logging capabilities
- Rust lightning has created their own custom logging to gain logging functionality without using an external crate
- The rust-lightning logging logic is found [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/util/logger.rs) and [here](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/util/macro_logger.rs)
- Need to pull this logic in so it can be used in `roles`
- Need to verify it is non-blocking

darricksee:
This was closed with https://github.com/stratum-mining/stratum/pull/303

PR #303, opened 2022-10-10
207 add logging trait

rustynirvana:
This PR fulfills [issue 207 ](https://github.com/stratum-mining/stratum/issues/207) - which copies the logging trait/implementation used by the [lightning rust library](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/util/logger.rs). 

This PR is a POC to show how it works with a simple logger - a more complex logger will be implemented once the approach is approved. This approach allows any user of this library to implement their own logger to do whatever they want. This ensures that there will be no logger implementation lock in by using the sv2 library.


utils/logging/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "logging"
```
Fi3, line 2: Name should be logging_sv2 so we can publish it without conflict

utils/logging/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+    Debug,
+    /// Designates useful information
+    Info,
+    /// Designates hazardous situations
+    Warn,
+    /// Designates very serious errors
+    Error,
+}
+
+impl PartialOrd for Level {
```
Fi3, line 38: this seems to be equal to the derived PartailOrd for a generic enum. Is that the case? If yes why you coded it instead of just derive it?

utils/logging/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    #[inline]
+    fn ge(&self, other: &Level) -> bool {
+        *self as usize >= *other as usize
+    }
+}
+
+impl Ord for Level {
+    #[inline]
```
Fi3, line 66: this seems to be equal to the derived PartailOrd for a generic enum. Is that the case? If yes why you coded it instead of just derive it?

utils/logging/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+            file,
+            line,
+        }
+    }
+}
+
+/// A trait encapsulating the operations required of a logger
+pub trait Logger {
+    /// Logs the `Record`
+    fn log(&self, record: &Record);
```
Fi3, line 125: I thought that we wanted an async logger @rrybarczyk ? If that is the case this function should be async or we should have 2 function one async and one not.

utils/logging/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+}
```
Fi3, line 126: I would also implementd a logger so that we can use it in the roles without reimplemting it everythime something like:
```rust
enum Destination {
    Std,
    Costum {path_info: String, path_err: String },
}

pub struct MyLogger {
    destination: Destination
}

impl Logger for MyLogger {
    fn log(&self, record: &Record) {
        match self.destination {
            Destination::Std => {
                match record.level {
                    Level::Info => println!("{}",record.args),
                    //Level::Error => printerr!("{}",record.args),
                    _ => (),
                }
            }
            Destination::Costum { path_info, path_err } => todo!(),
            }
        }

    }
}

```

utils/logging/src/lib.rs
darricksee: Yes - it's true that we want an async logger. This is really a POC to get approval on the design. I'll be writing an async logger for the pool role next.

Fi3: Ok perfect, I like the POC

utils/logging/src/lib.rs
darricksee: Agreed - I pulled a lot of this code from the LN rust library - they implemented this back in 2018 - maybe the default implementation was different back then. I derived PartialOrd and Ord and all the tests still pass.

utils/logging/src/lib.rs
darricksee: Further - you can't currently use `async` in traits - I could pull in [async-trait crate from crates.io](https://github.com/dtolnay/async-trait) but in the rust docs it says "Note that using these trait methods will result in a heap allocation per-function-call. This is not a significant cost for the vast majority of applications, but should be considered when deciding whether to use this functionality in the public API of a low-level function that is expected to be called millions of times a second."

I feel like the `log` method is probably a low level function and might be called frequently.

Fi3: yes but if we want to write log to a db for example I guess that we want it to be async. If we do not want to be async we could implement the logger in a way that log just save the Record in internal state, and just spwan a task that read the internal state and write the new records where we want. We should have at least one StandardLogger for tokio and one for async_std

Fi3: utils is not part of core IMO

darricksee: Well, since we don't have a core yet I suppose that can be moved when core gets created?

Fi3: I think that everything under ./protocol/ is core ./utils/ are not part of core in fact nothing under ./protocols depend on ./utils/

darricksee: There are lots of `buffer_sv2` dependencies in protocol/v2 - codec-sv2, framing-sv2, noise-sv2, binary-sv2, no-serde-sv2. What am I missing?

Fi3: yep but it should always be optional, if not we should move it out

rustynirvana:
Pulling in a log or tracing crate is the better way to go here.

PR #323, opened 2022-10-20
207 logging via tracing

darricksee:
This pr introduces the tokio tracing crate to the library and applications to be used for logging. The applications use a simple tracing_subscriber crate to print logging out to stdout but we can later decide to use a different crate to log in different formats or different outputs.

### DIFF
--- a/roles/translator/src/main.rs
+++ b/roles/translator/src/main.rs
@@ -20,8 +20,6 @@ use std::{
 };
 use v1::server_to_client;
 
-use tracing_subscriber;
-
 use tracing::{error, warn};
 
 /// Process CLI args, if any.
--- a/roles/translator/src/main.rs
+++ b/roles/translator/src/main.rs
@@ -24,7 +24,6 @@ use tracing_subscriber;
 
 use tracing::{error, warn};
 
-
 /// Process CLI args, if any.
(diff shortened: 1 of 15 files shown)


darricksee:
Interesting discussion about this PR - https://discord.com/channels/950687892169195530/958814900770205739/1032952476476903464

roles/translator/src/main.rs
```diff
@@ -35,8 +37,10 @@ fn process_cli_args() -> ProxyResult<ProxyConfig> {
 
 #[async_std::main]
 async fn main() {
+    tracing_subscriber::fmt::init();
+
     let proxy_config = process_cli_args().unwrap();
-    println!("PC: {:?}", &proxy_config);
+    warn!("PC: {:?}", &proxy_config);
```
rrybarczyk, line 43: Think this should just be `info` instead of `warn`

rrybarczyk (changes requested):
Looks great. Perhaps some documentation in the `README` of the project's root dir would be good?

roles/translator/src/main.rs
darricksee: Yeah - I agree! I just tried to keep it printing always like it was - maybe I should have taken some artistic liberties!

rrybarczyk (approved):

PR #336, opened 2022-11-02
binary-sv2: add inner_as_ref() to Seq064K

ccdle12:
Adds  a method to borrow the inner vector of T for Seq064K. This is specifically required when the job_creator receives a NewTemplate with coinbase_tx_outputs and will need to deserialize each item into a TxOut.

### DIFF
--- a/protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
+++ b/protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
@@ -109,6 +109,12 @@ impl<'a, T: 'a> Seq064K<'a, T> {
             Err(Error::SeqExceedsMaxSize)
         }
     }
+
+    /// Return a reference to the inner vector of T with the same lifetime
+    /// guarantees as the struct.
+    pub fn inner_as_ref(&'a self) -> &'a Vec<T> {
+        &self.0
+    }
 }
 
 impl<'a, T: GetSize> GetSize for Seq064K<'a, T> {
--- a/protocols/v2/binary-sv2/binary-sv2/src/lib.rs
+++ b/protocols/v2/binary-sv2/binary-sv2/src/lib.rs
@@ -273,6 +273,15 @@ mod test {
 
             assert_eq!(deserialized, expected);
         }
+
+        #[test]
+        fn test_inner_as_ref() {
+            let mut input = [1, 2, 9];
+            let b: B064K = (&mut input[..]).try_into().unwrap();
+
+            let borrow = b.inner_as_ref();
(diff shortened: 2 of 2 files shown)


Fi3:
Hi @ccdle12 can you please also add  test here https://github.com/ccdle12/stratum/blob/2022.11.02-Seq064K-inner-as-ref/protocols/v2/binary-sv2/binary-sv2/src/lib.rs ? Ty

ccdle12:
> Hi @ccdle12 can you please also add test here https://github.com/ccdle12/stratum/blob/2022.11.02-Seq064K-inner-as-ref/protocols/v2/binary-sv2/binary-sv2/src/lib.rs ? Ty

@Fi3 added basic test and rebased

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
```diff
@@ -109,6 +109,12 @@ impl<'a, T: 'a> Seq064K<'a, T> {
             Err(Error::SeqExceedsMaxSize)
         }
     }
+
+    /// Return a reference to the inner vector of T with the same lifetime
+    /// guarantees as the struct.
+    pub fn inner_as_ref(&'a self) -> &'a Vec<T> {
```
ccdle12, line 115: @Fi3 do you have a suggestion of renaming this method?

As far I can tell, the implementation of `inner_as_ref()` for `Seq0255` gets all inner T and returns a vector of all bytes of all T in sequence. So the iterator would be stepping by every individual byte instead of stepping by each `T`.

In this case, I want to get a `Vec<T>` and get each iteration of `T` and deserialize to a `Tx`: https://github.com/ccdle12/stratum/blob/2022.11.03-fix-cb-height-bug/protocols/v2/roles-logic-sv2/src/job_creator.rs#L220

Fi3: What about something like `map_inners` so that we can do `self.coinbase_outputs.map_inners(TxOut::deserialize)`

If you go with this solution please add it also for `Seq0255`, and (fi possible) for the respective types in the serde serializer/deserializer https://github.com/stratum-mining/stratum/tree/main/protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences

Also this method should be tested here https://github.com/stratum-mining/stratum/blob/main/protocols/v2/binary-sv2/binary-sv2/src/lib.rs

Fi3: Also please note the  `inner_as_ref()` for `Seq0255` do not return a vectors of bytes but return a vector of slices. So I guess that if you want implement it exactly as we did for the `Seq0255` it will works for your use case.

Issue #317, opened 2022-10-18
Pool: Bug with extranonce length panic


ccdle12:
## Overview

I was testing with the Template Provider and found the Pool panics on deserializing a coinbase transaction. It seems to be due to missing an extra 16 bytes in the extra nonce.

Below are some log prints:

- 116 bytes coinbase serialized, taken from here: https://github.com/stratum-mining/stratum/blob/main/protocols/v2/roles-logic-sv2/src/job_creator.rs#L85

    - *DEBUG: serialized coinbase: [2, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 34, 1, 19, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 1, 0, 242, 5, 42, 1, 0, 0, 0, 22, 0, 20, 83, 18, 96, 170, 42, 25, 158, 34, 140, 83, 125, 250, 66, 200, 43, 234, 44, 124, 31, 77, 0, 0, 0, 0]*

- The  prefix, suffix and extranonce are from here (before passing to `merkle_root_from_path()`): https://github.com/stratum-mining/stratum/blob/main/roles/v2/pool/src/lib/mining_pool/mod.rs#L54

    - *DEBUG: prefix: [2, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 34, 1, 19, 0, 0]
    DEBUG: suffix: [255, 255, 1, 0, 242, 5, 42, 1, 0, 0, 0, 22, 0, 20, 83, 18, 96, 170, 42, 25, 158, 34, 140, 83, 125, 250, 66, 200, 43, 234, 44, 124, 31, 77, 0, 0, 0, 0]
[...]

PR #337, opened 2022-11-02
Segwit compatibility in the Pool

ccdle12:
Built on top of #336, the changes relevant to this PR begin at e688285.

This change allows the Pool to handle NewTemplates that contain Segwit inputs. The Pool can receive and process multiple coinbase_tx_outputs. This will enable the Template Provider to send the coinbase witness commitment (a TxOut) to the Pool and the Pool can submit work for valid Segwit blocks.

Things to consider:

- The Pool can theoretically calculate their own coinbase witness commitment using `DoubleSHA256(witness root hash | witness reserved value)`

- If we should/do need to send the coinbase witness commitment, maybe consider making this explicit for the additional outputs field and no other type of outputs are allowed?

- `WITNESS_RESERVE_VALUE` is currently hardcoded, there is a proposal in the spec for the TemplateProvider to send this value in a NewTemplate message

- Testing on Regtest, these changes now remove the `AcceptBlock FAILED (unexpected-witness, ContextualCheckBlock : unexpected witness data found)` error, but high hash errors still persist when attempting to validate a block with many transactions. I suspect it could be an issue with hardcoding the difficulty in the proxy or a bug in parsing coinbase_prefix/suffix not taking into account the extra 2 byte flag for segwit or it could be adding the work to the coinbase tx is buggy as reported in #317, further investigation required

### DIFF
--- a/protocols/v2/binary-sv2/binary-sv2/src/lib.rs
+++ b/protocols/v2/binary-sv2/binary-sv2/src/lib.rs
@@ -273,6 +273,15 @@ mod test {
 
             assert_eq!(deserialized, expected);
         }
+
+        #[test]
+        fn test_inner_as_ref() {
+            let mut input = [1, 2, 9];
+            let b: B064K = (&mut input[..]).try_into().unwrap();
+
+            let borrow = b.inner_as_ref();
+            assert_eq!(&borrow[..], &[1, 2, 9]);
+        }
     }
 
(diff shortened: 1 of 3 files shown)


pavlenex:
Done in #440 so will close this one. @ccdle12 feel free to re-open the issue or report back in case you find it's not been addressed properly.

PR #364, opened 2022-12-02
262 - Error handling connections

darricksee:
This pr fixes the following error conditions with the translation proxy/pool connections:

1. The translator connects successfully to the pool, but then disconnects - previously the translator couldn't reconnect as the pool got into a bad state
2. The translator tries to connect to the unencrypted port of the pool - previously the translator wouldn't exit, it'd just hang and it wasn't clear what the problem was.

I also added logging around these cases to make it more clear what the problems are.

### DIFF
--- a/roles/translator/src/upstream_sv2/upstream.rs
+++ b/roles/translator/src/upstream_sv2/upstream.rs
@@ -29,7 +29,11 @@ use roles_logic_sv2::{
     utils::{get_target, Mutex},
 };
 use std::{net::SocketAddr, sync::Arc};
-use tracing::{debug, info, trace, warn};
+use std::thread::sleep;
+use std::time::Duration;
+use tracing::{debug, error, info, trace, warn};
+use crate::error::Error::CodecNoise;
+
 
 /// Represents the currently active mining job being worked on.
 #[allow(dead_code)]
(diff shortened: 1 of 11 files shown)


roles/translator/src/upstream_sv2/upstream.rs
```diff
@@ -141,12 +142,24 @@ impl Upstream {
...
+        let socket = loop {
+            match TcpStream::connect(address).await {
+                Ok(socket) => break socket,
+                Err(e) => {
+                    error!(
+                        "Failed to connect to Upstream role at {}, retrying in 5s: {}",
+                        address, e
+                    );
+
+                    sleep(Duration::from_secs(5));
```
Fi3, line 155: maybe we want to parametrize the sleep duration?

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -8,23 +8,10 @@ use roles_logic_sv2::{
...
+    utils::{hash_rate_to_target, Mutex},
 };
 use std::{convert::TryInto, sync::Arc};
-use tracing::trace;
-// [h/s] Expected hash rate of the device (or cumulative hashrate on the
-// channel if multiple devices are connected downstream) in h/s.
-// Depending on server’s target setting policy, this value can be used for
-// setting a reasonable target for the channel. Proxy MUST send 0.0f when
-// there are no mining devices connected yet.
-pub fn hash_rate_to_target(_hs: f32) -> U256<'static> {
```
Fi3, line 20: finally :+1: . This could brake POC2 (pool <-> proxy <-> sv2 cpu miner). Btw if that is the case this should be fixed in my next PR so I think that is fine

utils/network-helpers/src/plain_connection_tokio.rs
```diff
@@ -43,6 +50,12 @@ impl PlainConnection {
                     Ok(_) => {
                         if let Ok(x) = decoder.next_frame() {
                             sender_incoming.send(x.into()).await.unwrap();
+                        } else {
+                            error!("Error parsing incoming frame");
+                            if strict {
```
Fi3, line 55: I expect `decoder.next_frame` to return an Error at least half of the time that we call it. Cause:
1. decoder internal buffer is void and writable() return 6 
2. decoder internal buffer is filled with 6 bytes (the sv2 frame)
3. decoder deocde the frame and get payload len
4. writable() return the payload len
5. we fill the internal buffer with payload len
6. next_frame() return Ok(internal buffer)

For that I don't see where we could use the strict option

roles/translator/src/upstream_sv2/upstream.rs
darricksee: I'm going to skip this for now. This is going to be a pretty corner case where the pool isn't already running, the only thing I could think of doing more is to make the retries back off - so it starts at 1s and then fades back with every try - both really feel like overkill for this corner case.

utils/network-helpers/src/plain_connection_tokio.rs
darricksee: Updated and fixed/changed to be more specific to noise handshake

PR #368, opened 2022-12-06
[WIP] Move channel logic into roles logic

Fi3:
In order to implement extended channel into the proxy, all the logic relative to Sv2 channels is moved from pool and proxy into roles-logic-sv2.
We need this cause an extended channel in the proxy and a pool share a lot of functions having only one implementation instead of two is better to avoid having to fix bugs 2 times.
Another advantage of moving the logic into roles-logic-sv2 is that right now is in random places in the pool and in the proxy having it all in one place hopefully it will make it easier to follow the "flow".
Also having this logic under a specific struct that do "only" that is easier to test.

This PR will:
1.  Fix a bug in Target, this is needed by the channel logic in order to work
2. Upgrade Extranonce, add methods to Extranonce needed by the channel logic
3. Upgrade subprtocols/mining, add methods that make it easier to work with Sv2 errors this make the channel logic implementation less complex
4. move the channel logic: it create a new module in roles-logic-sv2 this module will export ChannelFactory used by the pool to open downstream channels and by the proxy that have extended channel opened upstream to open downstream channels
5. refactor proxy and pool in order to use the newly crated channel logic
6. remove all the old code into roles-logic-sv2 that it is obsolete after channel logic
7. add unit tests
8. add interop tests
9. add documentation

Where we are:
1. 100%
2. 100%
3. 100%
4. 100%
5. 100%
6. 100%
7. 30%
8. 0%
9. 70%


coleFD:
I noticed the `handle_new_extended_mining_channel()` logic was converted to `todo()!`. Does work still need to be done here?

protocols/v2/binary-sv2/no-serde-sv2/derive_codec/src/lib.rs
```diff
@@ -290,12 +325,37 @@ pub fn decodable(item: TokenStream) -> TokenStream {
...
+        parsed_struct.name,
+        derive_static_fields,
+
     );
 
     // Never executed at runtime it ok to panic
     result.parse().unwrap()
 }
 
+fn get_static_genercis(gen: &str) -> &str {
```
coleFD, line 351: looks like `generics` is misspelled

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -0,0 +1,1075 @@
...
+use template_distribution_sv2::{NewTemplate, SetNewPrevHash as SetNewPrevHashFromTp};
+
+use bitcoin::hashes::{sha256d::Hash, Hash as Hash_};
+
+/// Rapresent the action that needs to be done when a new share is received.
+#[derive(Debug, Clone)]
+pub enum OnNewShare {
+    /// Used when the received is malformed, is for an inexistent channel or do not meet downstream
+    /// target.
+    SendErrorDowsntream(SubmitSharesError<'static>),
```
coleFD, line 26: typo on Downstream

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -0,0 +1,1075 @@
...
+        }
+    }
+
+    pub fn on_new_prev_hash_from_tp(
+        &mut self,
+        m: &SetNewPrevHashFromTp<'static>,
+    ) -> Result<u32, Error> {
+        let job_id = self.job_creator.on_new_prev_hash(m).unwrap_or(0);
+        let new_prev_hash = SetNewPrevHash {
+            channel_id: 0,
```
coleFD, line 209: why is channel_id always 0?

Fi3: Good catch. So, here we set the channel to 0 cause this prev hash i valid for every channel created. I used 0 cause is kinda (internally) reserved cause channels start from 1.
A better solution could be use another type with only the PrevHash field that we need like:
```rust
struct ReducedPHash {
   job_id: u32,
   ...
}
```
then `ChannelFactory::new_prevhash` will be `Option<ReducedPHash>`
then `ChannelFactory::on_new_prev_hash` will take a `ReducedPHash`
then `ProxyExtendedChannelFactory::on_new_prev_hash` will transform the `SetNewPrevHash` in a `ReducedPHash` before calling `ChannelFactory::on_new_prev_hash` 

And when we need a `SetNewPrevHash` we use the right channel id. Right now you can set `SetNewPrevHash::channel_id` at whatever number you want and everything should keep working cause is not checked anywhere (not sure if it should be, not sure also if that field have any practical utility)

I will fix the code to use the ReducedPHash struct and set the right channel_id 

Same thing for `new_extended_job()` in `job_creators`  we are creating an extended job that will be used by all the channels so we put 0 as id I will modify the code in a way similar to the one that I proposed above also here.

coleFD: Your call. if you think we can just add comments to explain 0 is the general case, I am good with that too.

If you think the changing to a new type improves stability I have a couple questions:
1. would it make sense to create a separate issue for this?
2. would an `Option<u32>` be easier than a new type? Then the during serialization the `Some/None` checks can be enforced. 
3. If the new Type is the best way, what do you think about changing `Reduced` to `Staged` (or some other word that implies channel_id needs to be added to make a SNPH)? "Reduced" gives the connotation fields have been removed for different reasons.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -0,0 +1,1075 @@
...
+
+    fn new_standard_channel_for_hom_downstream(
+        &mut self,
+        request_id: u32,
+        downstream_hash_rate: f32,
+        id: u32,
+    ) -> Result<Vec<Mining>, Error> {
+        let hom_group_id = 0;
+        let mut result = vec![];
+        // first unwrap: is ok to unwrap safe_lock results
```
darricksee, line 262: I don't see what unwraps these lines are referring to.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: 1. I would prefer do it in this PR
2. The SetNewPrevHash is a Sv2 message and channel_id is not optional make it optional would mean changing the spec so, using Option<u32> and changing the spec would harder then just internally  defining a new type.
3. Agree

* the new type is just used internally in the lib what is been sent is SetNewPrevHash with the right channel id (or group id when we send it to proxy that groups channels)

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: ops I removed the unwraps but not the comments

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -0,0 +1,1075 @@
...
+    pub fn new_extended_channel(
+        &mut self,
+        request_id: u32,
+        hash_rate: f32,
+        min_extranonce_size: u16,
+    ) -> Option<Vec<Mining>> {
+        let extended_channels_group = 0;
+        let max_extranonce_size = self.extranonces.get_range2_len() as u16;
+        if min_extranonce_size <= max_extranonce_size {
+            // SECURITY is very unlikely to finish the ids btw this unwrap could be used by an attaccher that
```
darricksee, line 165: Yeah - I could see this potentially causing a problem - it looks like the id would just wrap and start using/reusing/stepping on other ids. I made an issue to look into this at some point - https://github.com/stratum-mining/stratum/issues/379

Note: it'd take 136 years if we were using 1 id/second to reach max :)

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -0,0 +1,1075 @@
...
+            .coinbase_tx_prefix
+            .as_ref();
+        let coinbase_tx_suffix = self
+            .last_valid_job
+            .as_ref()
+            .ok_or(Error::ShareDoNotMatchAnyJob)?
+            .0
+            .coinbase_tx_suffix
+            .as_ref();
+        let merklet_path = self.last_valid_job.as_ref().unwrap().0.merkle_path.to_vec();
```
darricksee, line 652: merklet_path -> merkle_path

protocols/v2/roles-logic-sv2/src/channel_logic/proxy_group_channel.rs
```diff
@@ -0,0 +1,173 @@
...
+    ) -> Result<Vec<Mining<'static>>, Error> {
+        let channel = StandardChannel {
+            channel_id: m.channel_id,
+            group_id: m.group_channel_id,
+            target: m.target.clone().into(),
+            extranonce: m.extranonce_prefix.clone().into(),
+        };
+        let channel_id = m.channel_id;
+        let mut res = vec![];
+        //res.push(Mining::OpenStandardMiningChannelSuccess(m));
```
darricksee, line 97: Should this be removed?

protocols/v2/roles-logic-sv2/src/channel_logic/proxy_group_channel.rs
```diff
@@ -0,0 +1,173 @@
...
+            .ok_or(Error::ImpossibleToCalculateMerkleRoot)?;
+            res.push(Mining::NewMiningJob(standard_job));
+        };
+
+        self.hom_downstreams.insert(channel_id, channel);
+
+        Ok(res)
+    }
+    fn update_new_prev_hash(&mut self, m: &SetNewPrevHash) {
+        while let Some(job) = self.future_jobs.pop() {
```
darricksee, line 129: This is iterating and popping the last off each time - which means that this will set `last_valid_job` to the first valid job - i think you want to use `pop_front` here right?

protocols/v2/roles-logic-sv2/src/channel_logic/proxy_group_channel.rs
```diff
@@ -0,0 +1,173 @@
...
+            if job.job_id == m.job_id {
```
darricksee, line 130: Do you want to pop all the jobs out of the vec even if the job_ids don't match?

rrybarczyk (changes requested):
Looking great @Fi3. When I run the tproxy off this branch, I am still getting the `"Failed to decode noise frame!"` error from [this line](https://github.com/stratum-mining/stratum/blob/05eba37a45823a607f53d397e62897a28dd45c68/utils/network-helpers/src/noise_connection_async_std.rs#L64). I can't remember now, was that something that this PR was going to fix?

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: ok for now we have an issue for that and we leave it like as it is in this PR

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: :+1:

protocols/v2/roles-logic-sv2/src/channel_logic/proxy_group_channel.rs
Fi3: yes

protocols/v2/roles-logic-sv2/src/channel_logic/proxy_group_channel.rs
Fi3: yes on new prev hash we reset the futures job

protocols/v2/roles-logic-sv2/src/channel_logic/proxy_group_channel.rs
Fi3: is fine like that the last future is likely the one that we want to use

roles/v2/pool/src/main.rs
```diff
@@ -18,12 +18,10 @@ pub type Message = PoolMessages<'static>;
 pub type StdFrame = StandardSv2Frame<Message>;
 pub type EitherFrame = StandardEitherFrame<Message>;
 
-const HOM_GROUP_ID: u32 = u32::MAX;
-
 const PRIVATE_KEY_BTC: [u8; 32] = [34; 32];
 const NETWORK: Network = Network::Testnet;
 
-const BLOCK_REWARD: u64 = 625_000_000_000;
+const BLOCK_REWARD: u64 = 5_000_000_000;
```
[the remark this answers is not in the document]
Fi3: this is for using it in regtest I will put this value in config file so we can easly change it

Fi3:
> Looking great @Fi3. When I run the tproxy off this branch, I am still getting the `"Failed to decode noise frame!"` error from [this line](https://github.com/stratum-mining/stratum/blob/05eba37a45823a607f53d397e62897a28dd45c68/utils/network-helpers/src/noise_connection_async_std.rs#L64). I can't remember now, was that something that this PR was going to fix?

this is due to an error log that should be removed I can delete it in this PR

roles/v2/pool/src/main.rs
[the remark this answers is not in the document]
darricksee: Is there a GH issue for this? Wouldn't we rather calculate it, having it in the config is not the right solution.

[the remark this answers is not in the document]
Fi3: no can you open one? if not I will open it just let me know on discord. Thanks

Issue #365, opened 2022-12-05
Inconsistent Match Cases of SV1 Message Params


coleFD:
# Description
While testing the Tproxy with the pool role and the SV1 Device `/examples/sv1-mining-device`, I ran into an issue where the SV1 Mining Device is unable to convert a `Response` into a `Subscribe`. It looks like the bug is due to the inability to find a valid match for `response.result` since the ordering of the incoming `response.result` array may not be deterministic. By adding the quick fix (line 392  `/protocols/v1/src/methods/server_to_client.rs` in the attached image) I was able to successfully subscribe to the TProxy role. The errors below were the result before the quick fix. 

# Question
There are other instances in the `/protocols/v1` crate that match against arrays that also do not consider the ordering of the values in `result` / `params`. Should each instance be rewritten or is the ordering actually deterministic and this specific case is just a typo?

# Source of Bug
![image](https://user-images.githubusercontent.com/105945999/205643147-aa3cd9e9-12ae-46e8-a372-5c1c32653796.png)

# Tproxy Logs:
```
2022-12-05T12:34:25.931895Z  INFO translator::downstream_sv1::downstream: PROXY SERVER - ACCEPTING FROM DOWNSTREAM: 127.0.0.1:45592
2022-12-05T12:34:25.932126Z  INFO translator::downstream_sv1::downstream: Receiving from Mining Device: "{\"id\":\"1670243665931878063\",\"method\":\"mining.configure\",\"params\":[[\"version-rolling\"],{}]}"
2022-12-05T12:34:25.932227Z  INFO translator::downstream_sv1::downstream: Down: Configuring
2022-12-05T12:34:25.932372Z  INFO translator::downstream_sv1::downstream: Sending to Mining Device: "{\"id\":\"1670243665931878063\",\"error\":null,\"result\":{\"minimum-difficulty\":false,\"version-rolling\":true,\"version-rolling.mask\":\"ffffffff\",\"version-rolling.min-bit-count\":\"00000000\"}}\n"
2022-12-05T12:34:25.932846Z  INFO translator::downstream_sv1::downstream: Receiving from Mining Device: "{\"id\":\"1670243665931878063\",\"method\":\"mining.subscribe\",\"params\":[\"80\",\"08000002\"]}"
2022-12-05T12:34:25.932927Z  INFO translator::downstream_sv1::downstream: Down: Subscribing
[...]

PR #369, opened 2022-12-07
V1 Bug Fix: Match Case Ordering of mining.subscribe

coleFD:
closes https://github.com/stratum-mining/stratum/issues/365.

Note: there is an open question on the linked issue asking if this needs to be considered in other array match cases in V1

### DIFF
--- a/protocols/v1/src/methods/server_to_client.rs
+++ b/protocols/v1/src/methods/server_to_client.rs
@@ -393,7 +393,7 @@ impl<'a> TryFrom<&Response> for Subscribe<'a> {
             ParsingMethodError::ImpossibleToParseResultField(Box::new(msg.clone()))
         })?;
         let (extra_nonce1, extra_nonce2_size, subscriptions_) = match &params[..] {
-            [JString(a), JNumber(b), JArrary(d)] => (
+            [JArrary(d), JString(a), JNumber(b)] => (
                 // infallible
                 Extranonce::try_from(hex::decode(a)?)?,
                 b.as_u64().ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


Fi3:
Hi, I think that the issue is in how the message is encoded, (here https://github.com/stratum-mining/stratum/blob/main/protocols/v1/src/methods/server_to_client.rs#L376) not how is decoded. That cause the translator work with both S9 and cpuminer, so is more likely that the bug is in the examples/sv1-mining-device.

coleFD:
> Hi, I think that the issue is in how the message is encoded, (here https://github.com/stratum-mining/stratum/blob/main/protocols/v1/src/methods/server_to_client.rs#L376) not how is decoded. That cause the translator work with both S9 and cpuminer, so is more likely that the bug is in the examples/sv1-mining-device.



> Hi, I think that the issue is in how the message is encoded, (here https://github.com/stratum-mining/stratum/blob/main/protocols/v1/src/methods/server_to_client.rs#L376) not how is decoded. That cause the translator work with both S9 and cpuminer, so is more likely that the bug is in the examples/sv1-mining-device.

@Fi3 I closed the PR but ran into this https://en.bitcoin.it/wiki/Stratum_mining_protocol#mining.subscribe. Not sure if this is a good source, but is there the possibility that miners could handle either case?

Fi3:
yep we should dig more into it. Also we can not have a value serialized in way and de serialized in another way so we have to change something

Fi3:
Ok sorry disregard what I my first message, is no sense. The part used by the translator is line 376 and is the working one.
The part used by the examples/sv1-mining-device is the one that you changed and is the not working one.

So I reopen it. But before merging it we need to change the variable names at line 389.

Fi3:
this names:
```rust
        let (extra_nonce1, extra_nonce2_size, subscriptions_) = match &params[..] {
 ```
 are not what you are mathcing against they should be
 ```rust
         let (subscriptions_, extra_nonce1, extra_nonce2_size) = match &params[..] {
```

rrybarczyk (approved):

PR #383, opened 2022-12-19
Mocking template provider

lorbax:


utils/message-generator/debug
```diff
@@ -0,0 +1,55 @@
+  0,
```
darricksee, line 1: Where is this new file used?

protocols/v2/subprotocols/common-messages/Cargo.toml
```diff
@@ -14,7 +14,8 @@ binary_sv2 = {version = "0.1.*", path = "../../../../protocols/v2/binary-sv2/bin
 const_sv2 = {version = "0.1.*", path = "../../../../protocols/v2/const-sv2"}
 quickcheck = { version = "1.0.3", optional=true }
 quickcheck_macros = { version = "1", optional=true }
+serde_repr = {version= "0.1.10", optional=true}
```
darricksee, line 17: Can you explain why we need this?

lorbax: I had problem in parsing type Protocol, so I used serve_repr
https://github.com/stratum-mining/stratum/blob/main/protocols/v2/subprotocols/common-messages/src/setup_connection.rs#L287

lorbax: https://github.com/stratum-mining/stratum/pull/383/commits/db1936bc365dd6be0ab13fc13e52de31763eea2c

utils/message-generator/debug
lorbax: I needed that file to fix a bug of the MG, I don't know how it is included in the commit. I'll remove it.

Issue #259, opened 2022-09-27
SV1/SV2 Translator: Check Downstream Shares Meet Upstream Target Before Submission


rrybarczyk:
before relay share to upstream we neede to check that downstream share meet upstream target (maybe we already do that I didn't check I just put this point here in case we are not doing it) 
Upon receiving a `mining.submit` from a Downstream (SV1 Miner), the `Bridge` (in `bridge.rs`) should check that the received share meets the pdiff set by the Upstream (SV2 Pool) BEFORE translating the `mining.submit` into a `SubmitSharesExtended` for the `Upstream`.

ccdle12:
Is this related to this TODO: https://github.com/stratum-mining/stratum/blob/main/examples/sv1-mining-device/src/client.rs#L91

pavlenex:
Done in #391

PR #391, opened 2022-12-27
Fix translator compatibility with pool

Fi3:


coleFD:
looks like this closes https://github.com/stratum-mining/stratum/issues/259

Fi3:
@coleFD  yep

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -127,9 +134,55 @@ impl ParseDownstreamMiningMessages<(), NullDownstreamMiningSelector, NoRouting>
...
+                roles_logic_sv2::channel_logic::channel_factory::OnNewShare::SendSubmitShareUpstream(_) => unreachable!(),
+                roles_logic_sv2::channel_logic::channel_factory::OnNewShare::RelaySubmitShareUpstream => unreachable!(),
+                roles_logic_sv2::channel_logic::channel_factory::OnNewShare::ShareMeetBitcoinTarget((share,t_id,coinbase)) => {
+                    info!("Found share that meet bitcoin target");
+                    let solution = SubmitSolution {
+                        template_id: t_id,
+                        version: share.get_version(),
+                        header_timestamp: share.get_n_time(),
+                        header_nonce: share.get_nonce(),
+                        coinbase_tx: coinbase.try_into().unwrap(),
```
coleFD, line 157: can this fail?

roles/translator/src/proxy/bridge.rs
```diff
@@ -95,49 +137,74 @@ impl Bridge {
         let tx_sv2_submit_shares_ext = self_
             .safe_lock(|s| s.tx_sv2_submit_shares_ext.clone())
             .unwrap();
+        let target = self_.safe_lock(|s| s.target.clone()).unwrap();
         task::spawn(async move {
             loop {
-                let (sv1_submit, extranonce) = rx_sv1_submit.clone().recv().await.unwrap();
+                let upstream_target: [u8; 32] =
+                    target.safe_lock(|t| t.clone()).unwrap().try_into().unwrap();
```
coleFD, line 144: can this fail?

roles/v2/pool/src/lib/mining_pool/message_handler.rs
Fi3: do not remember max size of a bitcoin tx if is less than 64K bytes this can not fail.

roles/translator/src/proxy/bridge.rs
```diff
@@ -95,49 +137,74 @@ impl Bridge {
...
                 let channel_sequence_id =
                     self_.safe_lock(|s| s.channel_sequence_id.next()).unwrap() - 1;
-                let sv2_submit: SubmitSharesExtended =
-                    Self::translate_submit(&self_, channel_sequence_id, sv1_submit, &extranonce)
-                        .unwrap();
-                tx_sv2_submit_shares_ext.send(sv2_submit).await.unwrap();
+                let sv2_submit: SubmitSharesExtended = self_
+                    .safe_lock(|s| {
+                        s.translate_submit(channel_sequence_id, sv1_submit, extrnonce)
+                            .unwrap()
```
coleFD, line 156: Potential poison lock

coleFD: I received a poison error while running a local sim on this branch. Not sure where the origin of the error was though

roles/translator/src/proxy/bridge.rs
Fi3: no as soon as we pass a 32 byte vector from upstream.rs

roles/translator/src/proxy/bridge.rs
Fi3: yep better to get the result out of the lock  and then unwrap there

PR #395, opened 2023-01-03
Feat: TProxy Status Channel

coleFD:
The goal of this PR is to add a status channel to notify the main thread in the TProxy of status updates (errors/logs) in the other components (Downstream, Bridge, Upstream). This will allow clients to easily spin up another Upstream and switch to backup pools in the case the connected pool instance fails as well as other custom handling of TProxy components based on their status

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/handlers/mining.rs
+++ b/protocols/v2/roles-logic-sv2/src/handlers/mining.rs
@@ -247,6 +247,7 @@ pub trait ParseUpstreamMiningMessages<
 
         // Is fine to unwrap on safe_lock
         match (message_type, payload).try_into() {
+            //PANICS
             Ok(Mining::OpenStandardMiningChannelSuccess(mut m)) => {
                 let remote = match routing_logic {
                     MiningRoutingLogic::None => None,
@@ -274,6 +275,7 @@ pub trait ParseUpstreamMiningMessages<
                         .unwrap(),
                 }
             }
+            // Returns         Ok(SendTo::None(Some(m)))
             Ok(Mining::OpenExtendedMiningChannelSuccess(m)) => match channel_type {
                 SupportedChannelTypes::Standard => Err(Error::UnexpectedMessage),
                 SupportedChannelTypes::Extended => self_mutex
@@ -284,6 +286,7 @@ pub trait ParseUpstreamMiningMessages<
                     .safe_lock(|s| s.handle_open_extended_mining_channel_success(m))
                     .unwrap(),
             },
+            // Panics
             Ok(Mining::OpenMiningChannelError(m)) => match channel_type {
                 SupportedChannelTypes::Standard => self_mutex
                     .safe_lock(|x| x.handle_open_mining_channel_error(m))
@@ -298,6 +301,7 @@ pub trait ParseUpstreamMiningMessages<
                     .safe_lock(|x| x.handle_open_mining_channel_error(m))
                     .unwrap(),
             },
+            // Panics
             Ok(Mining::UpdateChannelError(m)) => match channel_type {
                 SupportedChannelTypes::Standard => self_mutex
                     .safe_lock(|x| x.handle_update_channel_error(m))
@@ -312,6 +316,7 @@ pub trait ParseUpstreamMiningMessages<
                     .safe_lock(|x| x.handle_update_channel_error(m))
                     .unwrap(),
             },
+            // Panics
             Ok(Mining::CloseChannel(m)) => match channel_type {
                 SupportedChannelTypes::Standard => {
                     self_mutex.safe_lock(|x| x.handle_close_channel(m)).unwrap()
@@ -326,6 +331,7 @@ pub trait ParseUpstreamMiningMessages<
                     self_mutex.safe_lock(|x| x.handle_close_channel(m)).unwrap()
                 }
             },
+            //Panics
             Ok(Mining::SetExtranoncePrefix(m)) => match channel_type {
                 SupportedChannelTypes::Standard => self_mutex
                     .safe_lock(|x| x.handle_set_extranonce_prefix(m))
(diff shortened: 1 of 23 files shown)


roles/translator/src/status.rs
```diff
@@ -0,0 +1,46 @@
...
+    Shutdown(Error<'a>),
+    Healthy(String),
+}
+
+#[derive(Debug)]
+pub struct Status<'a> {
+    pub state: State<'a>,
+}
+
+/// # Description
```
darricksee, line 14: Great description!

roles/translator/src/upstream_sv2/upstream.rs
```diff
@@ -495,14 +530,12 @@ impl ParseUpstreamMiningMessages<Downstream, NullDownstreamMiningSelector, NoRou
     fn handle_open_extended_mining_channel_success(
         &mut self,
         m: roles_logic_sv2::mining_sv2::OpenExtendedMiningChannelSuccess,
-    ) -> Result<roles_logic_sv2::handlers::mining::SendTo<Downstream>, roles_logic_sv2::errors::Error>
-    {
-        let extranonce_size_bytes = m.extranonce_size / 8;
```
darricksee, line 500: Here this no longer needs to be divided by 8 - it's already in `bytes`

protocols/v2/roles-logic-sv2/src/handlers/mining.rs
```diff
@@ -247,6 +247,7 @@ pub trait ParseUpstreamMiningMessages<
 
         // Is fine to unwrap on safe_lock
         match (message_type, payload).try_into() {
+            // Can Panic
```
Fi3, line 250: Not sure why do you want communicate with this comment, can you expand it?
Also note that if the `Mutex` that hold the routing logic (or Self) is poisoned, is very likely that we are in a non recoverable situation. 

This review apply also to all the other same comment in this file.

roles/translator/src/status.rs
```diff
@@ -0,0 +1,46 @@
...
+/// # Example
+/// ```
+/// let (tx_status: Sender<Status>, rx_status) = async_channel::unbounded();
+/// let variable = handle_result!(
+///     tx_status,
+///     type_.try_into()
+/// );
+/// ```
+#[macro_export]
+macro_rules! handle_result {
```
Fi3, line 28: Still have to deeper review it but this seem like something that should stay in `/utils/` so that we can have a consistent way to manage this kind of issues across all the SRI codebase.

roles/translator/src/status.rs
```diff
@@ -0,0 +1,46 @@
...
+            Err(e) => {
+                tracing::error!("Error: {:?}", e);
+                // send status
+                $sender
+                    .send($crate::status::Status {
+                        state: $crate::status::State::Shutdown(e.into()),
+                    })
+                    .await
+                    .unwrap();
+                // implement error specific handling here instead of panicking
```
Fi3, line 41: Maybe is better to inject error handling here

roles/translator/src/status.rs
```diff
@@ -0,0 +1,46 @@
+use crate::error::Error;
+
+#[derive(Debug)]
+pub enum State<'a> {
+    Shutdown(Error<'a>),
+    Healthy(String),
```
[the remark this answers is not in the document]
coleFD: maybe it could be used to create a single point to send logs to a monitoring service. Cant really think of any other reason

roles/translator/src/status.rs
```diff
@@ -0,0 +1,46 @@
...
+macro_rules! handle_result {
+    ($sender:expr, $res:expr) => {
+        match $res {
+            Ok(val) => val,
+            Err(e) => {
+                tracing::error!("Error: {:?}", e);
+                // send status
+                $sender
+                    .send($crate::status::Status {
+                        state: $crate::status::State::Shutdown(e.into()),
```
Fi3, line 37: We should be able to:
1. handle the error in the context where it has been generated 
2. handle the error at "main" level and maybe shutdown the thread/context that generate the error
3. handle the error at "main" level and exit

To me it seem that right now we have only (3) is that right? Maybe (1) is not necessary but I think that (2) is very necessary.

roles/translator/src/status.rs
Fi3: We should modify `network_helpers` so that we can use this macro when we are trying to read/write from the socket and in case of an error fail. So now that we have it I think that we can remove the panics from `network_helpers`

protocols/v2/roles-logic-sv2/src/handlers/mining.rs
darricksee: So when I was looking for possible panics throughout this match to see what we have to look out for - so these comments just specify which of these matches can panic. This was just my way of keeping track of which ones I had investigated - they can be removed if you don't think they provide any value.

roles/translator/src/status.rs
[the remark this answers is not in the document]
darricksee: So my idea here was that in the future these can be a way to message back to the main thread various healthy state messages. I'm fine with removing this for now until we have a solid use case for it.

protocols/v2/roles-logic-sv2/src/handlers/mining.rs
Fi3: Yep I think that we can remove them as it explicitly said in the code so no need for a comment IMO

roles/translator/src/status.rs
darricksee: For the examples that we have now they all do shutdown - but my thinking was eventually the main thread could potentially recover a task based on the State/Status/Error sent. Right now we haven't figured out what tasks can be restarted or which ones should be restarted so for now we are just shutting everything down "gracefully". You are correct though in saying that we should be able to do all 3 eventually; that's the goal.

Fi3:
after last call I think that we are ready to merge it so that we can then merge #395

darricksee (approved):

Issue #397, opened 2023-01-05
Pool role sending initial messages out of order


darricksee:
[According to the spec ](https://github.com/stratum-mining/sv2-spec/blob/main/img/4.3.b-Mining-Protocol-Messages-Mining-on-Standard-Channel.png)after a OpenStandardMining the pool should send a NewMiningJob and then a PrevHash message  to the miner. When I tested the bosminer_erupter->pool the pool was sending the SetNewPrevHash immediately after the OpenStandardMiningChannelSuccess which caused the bosminer_eruptor to crash:

```2023-01-04T20:50:18.757856Z TRACE stratum-v2: bosminer::client::stratum_v2: incoming=OpenStandardMiningChannelSuccess { request_id: 10, channel_id: 1, target: Uint256Bytes([0, 0, 0, 0, 18, 83, 63, 220, 16, 3, 144, 103, 124, 18, 124, 69, 243, 64, 124, 247, 90, 75, 19, 90, 163, 209, 0, 151, 46, 136, 231, 204]), extranonce_prefix: VarSeq("[0, 0,... 32 items]"), group_channel_id: 0 }
tokio-runtime-worker2023-01-04T20:50:18.757913Z TRACE stratum-v2: bosminer::client::stratum_v2: incoming=SetNewPrevHash { channel_id: 1, job_id: 0, prev_hash: Uint256Bytes([0, 35, 145, 112, 86, 218, 218, 189, 29, 87, 112, 22, 111, 223, 99, 203, 24, 168, 74, 133, 185, 61, 28, 177, 196, 7, 96, 102, 0, 0, 0, 0]), min_ntime: 1672865219, nbits: 545259519 }
' panicked at 'TODO: requested job ID not found', bosminer/src/client/stratum_v2.rs:535:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted```

After changing the pool to send the NewMiningJob first I got a similar error:

[...]

PR #398, opened 2023-01-05
Send NewMiningJob before SNPH on start with OpenStandardChannel

darricksee:
This fixes https://github.com/stratum-mining/stratum/issues/397

according to the spec - https://github.com/stratum-mining/sv2-spec/blob/main/img/4.3.b-Mining-Protocol-Messages-Mining-on-Standard-Channel.png

It also syncs the SNPH message going out to the downstream with the proper job_id


pavlenex:
@darricksee This one seems to be out of date with the base branch, can you rebase?

darricksee:
Made the spec more specific

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -409,28 +413,28 @@ impl ChannelFactory {
             // If we have just a prev hash we need to send it after the SetupConnectionSuccess
             // message
             (Some((prev_h, _)), None, true) => {
-                let prev_h = prev_h.into_set_p_hash(channel_id);
+                let prev_h = prev_h.into_set_p_hash(channel_id, None);
                 result.push(Mining::SetNewPrevHash(prev_h.clone()));
                 Ok(())
             }
             // If we have a prev hash and a last valid job we need to send before the prev hash and
```
Fi3, line 420: This comment should be changed

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -409,28 +413,28 @@ impl ChannelFactory {
...
             // the the valid job
             (Some((prev_h, _)), Some(job), true) => {
-                let prev_h = prev_h.into_set_p_hash(channel_id);
-                result.push(Mining::SetNewPrevHash(prev_h.clone()));
+                let prev_h = prev_h.into_set_p_hash(channel_id, Some(job.job_id));
                 result.push(Mining::NewMiningJob(job));
```
Fi3, line 424: make sure future_job is set to true and NPH job id is the one of the job

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -409,28 +413,28 @@ impl ChannelFactory {
...
+                result.push(Mining::SetNewPrevHash(prev_h.clone()));
                 Ok(())
             }
             // If we have everything we need, send before the prev hash and then all the jobs
             (Some((prev_h, _)), Some(job), false) => {
-                let prev_h = prev_h.into_set_p_hash(channel_id);
-                result.push(Mining::SetNewPrevHash(prev_h.clone()));
+                let prev_h = prev_h.into_set_p_hash(channel_id, Some(job.job_id));
```
Fi3, line 430: in this case you need to send first the last_valid_job (make sure that you set future=true) then the set_new_prev_hash (make sure that job_id is the one of the previously sented job) the all the other future job that are in self.future_jobs (these job must be sented cause they could be referenced by a future prev_hash)

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -409,28 +413,28 @@ impl ChannelFactory {
...
             // the the valid job
             (Some((prev_h, _)), Some(job), true) => {
-                let prev_h = prev_h.into_set_p_hash(channel_id);
-                result.push(Mining::SetNewPrevHash(prev_h.clone()));
+                let prev_h = prev_h.into_set_p_hash(channel_id, Some(job.job_id));
                 result.push(Mining::NewMiningJob(job));
+                result.push(Mining::SetNewPrevHash(prev_h.clone()));
                 Ok(())
             }
             // If we have everything we need, send before the prev hash and then all the jobs
```
Fi3, line 428: change this comment

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -120,10 +120,14 @@ pub struct StagedPhash {
 }
 
 impl StagedPhash {
-    pub fn into_set_p_hash(&self, channel_id: u32) -> SetNewPrevHash<'static> {
+    pub fn into_set_p_hash(
+        &self,
+        channel_id: u32,
+        new_job_id: Option<u32>,
```
Fi3, line 126: Why we need to have an option here? What changed? Why after these PR the Option wasn't needed and now it is?

darricksee: So the problem is here:
bosminer::client::stratum_v2: incoming=NewMiningJob { channel_id: 1, job_id: 2, future_job: true) }

bosminer::client::stratum_v2: incoming=SetNewPrevHash { channel_id: 1, job_id: 0, }

before this change when I just swapped the messages the SNPH's job_id didn't match the NMJ message that proceeded it. This change injects the NMJ's job_id into the StagedPhash so that when the SNPH message goes out the job_ids match.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
darricksee: So the code above is the thing that makes sure the SNPH and NMJ job_ids match. future_job is set true otherwise this case wouldn't match (Some(job)) comes from the `last_valid_job` which originates in the `future_jobs` vec.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
darricksee: I'm thinking this is already happening. The `last_valid_job` comes from the `future_jobs` list so I'm pretty sure `future_job` would already = true.

My above code will have injected the `last_valid_job`'s id into the SNPH via this `into_set_p_hash(.., Some(job.job_id))` call

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: `last_valid_job` can originate from a future job/template or from a non future job/template and is always non future per definition. A valid job is a job that need to be mined so not future.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: Here we have:
1. a SNPH
2. a valid job (non future per definition)
3. a list of future job that could be referenced by the next SNPH

Most natural thing to do here would have been to send the SNPH then the non future job then the future jobs. But we added the condition that  after OpenChannel.Success first message received MUST be a future job.

Cuase of that condition what we  need to do is:
1. take the valid job and transform it in a future job and send it as first message
2. send the SNPH (before sending it change the field job_id to match the one of the job sented in (1))
3. send all the future jobs (we still need them cause they are valid future job that can be referenced by a future SNPH)

Fi3: if the future_jobs are sented before the SNPH the downstream will discard all of them but the last_valid_job but this is not what we want

darricksee: done

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
darricksee: done

Fi3 (approved):

Issue #400, opened 2023-01-10
Running the tproxy->pool with cpu miner results in a lot of Submit Share errors difficulty-too-low


darricksee:
Running the cpu_miner->TProxy->Pool configuration I am seeing a lot of 
`2023-01-09T13:11:07.423849Z ERROR translator::proxy::bridge: Submit share error difficulty-too-low` errors.

pavlenex:
Fixed in #401

PR #401, opened 2023-01-10
Target compare bug

darricksee:
This fixes the target compare bug - https://github.com/stratum-mining/stratum/issues/400

I also added a unit test to verify (it breaks with the old code).

### DIFF
--- a/protocols/v2/subprotocols/mining/src/lib.rs
+++ b/protocols/v2/subprotocols/mining/src/lib.rs
@@ -147,10 +147,11 @@ pub use submit_shares::{
 pub use update_channel::{UpdateChannel, UpdateChannelError};
 const MAX_EXTRANONCE_LEN: usize = 32;
 
+/// Target is a 256-bit unsigned integer in little-endian
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Target {
-    head: u128,
-    tail: u128,
+    head: u128, // least significant bits
+    tail: u128, // most significant bits
 }
 
 impl Target {
--- a/protocols/v2/subprotocols/mining/src/lib.rs
+++ b/protocols/v2/subprotocols/mining/src/lib.rs
@@ -1005,21 +1005,12 @@ pub mod tests {
 
     #[test]
     fn test_ord_with_equal_head_tail() {
-        let target_1 = Target {
-            head: 1,
-            tail: 1,
-        };
-        let target_2 = Target {
-            head: 1,
-            tail: 2,
-        };
+        let target_1 = Target { head: 1, tail: 1 };
+        let target_2 = Target { head: 1, tail: 2 };
         assert!(target_1 < target_2);
 
         //also test with equal tails
-        let target_3 = Target {
-            head: 2,
-            tail: 2,
-        };
+        let target_3 = Target { head: 2, tail: 2 };
         assert!(target_2 < target_3);
     }
 
--- a/protocols/v2/subprotocols/mining/src/lib.rs
+++ b/protocols/v2/subprotocols/mining/src/lib.rs
@@ -203,7 +203,7 @@ impl Ord for Target {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
         if self.tail == other.tail && self.head == other.head {
             core::cmp::Ordering::Equal
-        } else if self.head != other.head {
+        } else if self.head == other.head {
             self.tail.cmp(&other.tail)
         } else {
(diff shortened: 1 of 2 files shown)


Sjors:
I still get `Submit share error difficulty-too-low` errors after a few minutes.

There's only one `Sending to Mining Device: "{\"method\":\"mining.set_difficulty\",\"params\":[0.07142857142857142]}\n"` log entry right after authorization, it doesn't seem to adjust after that.

Maybe squash the three commits?

protocols/v2/subprotocols/mining/src/lib.rs
```diff
@@ -203,7 +203,7 @@ impl Ord for Target {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
         if self.tail == other.tail && self.head == other.head {
             core::cmp::Ordering::Equal
-        } else if self.head != other.head {
+        } else if self.tail != other.tail {
```
Sjors, line 206: Shouldn't this also check head equality?

```rs
} else if self.head == other.head && self.tail != other.tail {
```

(which is equivalent to `} else if self.head == other.head {`

Sjors:
Actually there seem to be two `mining.set_difficulty` messages in quick succession, one in hex and the other in decimal:

```
2023-01-11T12:32:31.314644Z  INFO translator::upstream_sv2::upstream: Up: New Extended Mining Job
2023-01-11T12:32:31.314887Z  INFO translator::upstream_sv2::upstream: Up: Set New Prev Hash
2023-01-11T12:32:34.168660Z  INFO translator::downstream_sv1::downstream: PROXY SERVER - ACCEPTING FROM DOWNSTREAM: 192.168.x.x:33464
2023-01-11T12:32:34.169238Z  INFO translator::downstream_sv1::downstream: Receiving from Mining Device: "{\"id\": 1, \"method\": \"mining.subscribe\", \"params\": [\"cpuminer/2.5.1\"]}"
2023-01-11T12:32:34.169607Z  INFO translator::downstream_sv1::downstream: Down: Subscribing
2023-01-11T12:32:34.170157Z  INFO translator::downstream_sv1::downstream: Sending to Mining Device: "{\"id\":\"1\",\"error\":null,\"result\":[[[\"mining.set_difficulty\",\"ae6812eb4cd7735a302a8a9dd95cf71f\"],[\"mining.notify\",\"ae6812eb4cd7735a302a8a9dd95cf71f\"]],\"316506011af64a0001\",6]}\n"
2023-01-11T12:32:34.170903Z  INFO translator::downstream_sv1::downstream: Receiving from Mining Device: "{\"id\": 2, \"method\": \"mining.authorize\", \"params\": [\"provoost.cpu\", \"anything123\"]}"
2023-01-11T12:32:34.171178Z  INFO translator::downstream_sv1::downstream: Down: Authorizing
2023-01-11T12:32:34.171597Z  INFO translator::downstream_sv1::downstream: Sending to Mining Device: "{\"id\":\"2\",\"error\":null,\"result\":true}\n"
2023-01-11T12:32:34.171834Z  INFO translator::downstream_sv1::downstream: Sending to Mining Device: "{\"method\":\"mining.set_difficulty\",\"params\":[0.07142857142857142]}\n"
```

protocols/v2/subprotocols/mining/src/lib.rs
darricksee: I don't think so. I think that the tail contain the most significant bits so if they are != then the head bits don't matter. We only need to compare the head bits if the tail bits are equal.

Sjors: Would be good to leave a comment above `pub struct Target` about whether this thing is big or little endian.

Sjors: Looks like U256 values in v2 are all little endian:
* https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md#532-openstandardminingchannel-client---server
* https://github.com/stratum-mining/sv2-spec/blob/main/03-Protocol-Overview.md#31-data-types-mapping

So that means `tail` bytes are the most significant. So your version is correct.

Sjors:
Running it again with your version. It seems to no longer throw this low difficulty message.

Though it does seem weird that it's not telling the CPU miner to increase its difficulty. I don't know what the optimal rate of submitting shares is from the miner to the translator. It seems to do one per second.

Sjors:
Spoke too soon, it just threw the message...

PR #422, opened 2023-01-30
382 use correct coinbase output value

darricksee:
This splits up the remaining coinbase reward into each of the outputs. 

I also removed a "test" that looked like a config file.


protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -275,6 +275,7 @@ fn coinbase(
```
darricksee, line 275: All of the changes below fix various unused warnings that were spamming the output when running tests.

coleFD (approved):

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -822,12 +822,31 @@ impl PoolChannelFactory {
...
             self.job_creator
                 .on_new_template(m, true, self.pool_coinbase_outputs.clone())?;
         self.inner.on_new_extended_mining_job(new_job)
     }
 
+    /// From the spec the coinbase_tx_value_remaining is:
+    /// The value, in satoshis, available for spending in coinbase outputs
+    /// added by the client. Includes both transaction fees and block subsidy.
+    /// Divide this amount by all outputs
+    pub fn split_outputs(pool_coinbase_outputs: &mut Vec<TxOut>, coinbase_tx_value_remaining: u64) {
```
djozis, line 840: I see the output amounts are calculated by dividing integers evenly - what about remainder, if it is not a perfect division?

djozis: Looks good!

djozis (approved):

Fi3:
# Protocol

The template distribution protocol let the TP add some outputs to the coinbase of the template.
This is needed for 2 main reasons:
1. adding output with the witness commitment hash see https://github.com/stratum-mining/sv2-spec/pull/15
2. in case of third party TPAsAService add an output the pay that service provider

`coinbase_tx_value_remaining` it is used to communicate 2 information:
1. the sum of the fee that are in the provided template
2. the  sum of the output values added by the TP (I expect it to be 0 in the majority of the cases)

# SRI

In SRI when we initialize a channel factory we pass the coinbase outputs that the pool want to add https://github.com/stratum-mining/stratum/blob/24ec4d6e66c5650b2d2d73222f000214fca72869/roles/v2/pool/src/lib/mining_pool/mod.rs#L402 right now only one. This output is already initialized with value=coinbase_reward but this make no sense as the pool can not know in advance the fees and if TP reserve some reward for itself. It would be better if the pool would add an output with value 0 and then we will add the `coinbase_tx_value_remaining` to it. To make sure that we are mining valid job I think that channel factory should not assume that pool creating an output with value=0 and just replace value with `coinbase_tx_value_remaining`

## What if pool want to add more than on outputs
We could slightly modify the `[channel factory]::new` signature and instead of taking `pool_coinbase_outputs: Vec<TxOut>` we can take `pool_coinbase_outputs: Vec<(TxOut,u32)>`  so that we specify which percentage of the  `coinbase_tx_value_remaining` should be taken by each output.

## Bugs
There are 3 main issue 2 with the channel factory implementation and one with this PR.
1. `[channel factory].pool_coinbase_outputs` is a misleading name cause it contains both the pool outputs AND the TP outputs
2. `[channel factory].on_new_template(..)` call `JobCreator.on_new_template(..)` that add the TP outputs to `pool_coinbase_outputs` https://github.com/stratum-mining/stratum/blob/24ec4d6e66c5650b2d2d73222f000214fca72869/protocols/v2/roles-logic-sv2/src/job_creator.rs#L73 and it seems to me that the TP outputs are never removed. Se we can end up with enormous coinbases in little time if this is the case this is a high priority issue.
3. cause (1)  this PR will split the reward not only between the pool added outputs but also to the TP added outputs

It would be good with some one can double check the `Protocol` part of this message maybe @jakubtrnka ?

darricksee:
Bugs:
1.) In the `[channel factory].pool_coinbase_outputs` this only includes the pool's outputs because of 2.)
2.) Here we add the TP outputs to **a clone** of the pool_coinbase_outputs
3.) Because it's a clone the original `pool_coinbase_outputs` never gets altered so no bug here. 

I don't feel like we need to get fancy at this point wrt specifying percentages - that can be a later feature if required.

Fi3:
yep missed the clone, so no bug! About splitting reward not sure if needed without percentage but maybe yes, I would like to answer the below questions before going with it:
Pool need to add more then one output, and if yes why the correct behavior is to split the reward equally between all the added outputs?

Fi3:
Maybe is better on `on_new_template` in `JobCreator` to take an `&Vec<TxOut>` and clone it in the function so that is impossible to modify the external one?

darricksee:
> Maybe is better on `on_new_template` in `JobCreator` to take an `&Vec<TxOut>` and clone it in the function so that is impossible to modify the external one?

I'm confused by this... currently the code does:
```
let new_job =
            self.job_creator
                .on_new_template(m, true, self.pool_coinbase_outputs.clone())?;
```
We are cloning the vec before passing it to `on_new_template` - this already prevents that method from modifying the external vec. It doesn't make sense to clone it in the function.

darricksee:
> yep missed the clone, so no bug! About splitting reward not sure if needed without percentage but maybe yes, I would like to answer the below questions before going with it: Pool need to add more then one output, and if yes why the correct behavior is to split the reward equally between all the added outputs?

I'm totally fine with just having 1 output and no splitting. It's just a feature I thought could be useful but probably will be rarely used.

jakubtrnka:
> the sum of the output values added by the TP (I expect it to be 0 in the majority of the cases)

What is meant by this? What is "output value added by the TP".

In my view coinbase_tx_value_remaining is used to communicate only one single thing:
Total amount of satoshis available for the additional output(s) that would be added by pool.

Fi3 (approved):

PR #423, opened 2023-01-30
Pool: Status Channel 

coleFD:
This PR implements a status channel in the pool role for error handling


roles/v2/pool/src/lib/template_receiver/mod.rs
```diff
@@ -130,26 +131,30 @@ impl TemplateRx {
...
-                .await
-                .expect("Failed to send solution!");
+                    .try_into();
+            match sv2_frame_res {
+                Ok(frame) => {
+                    handle_result!(status_tx, Self::send(self_.clone(), frame).await);
+                }
+                Err(_e) => {
+                    // return submit error
+                    todo!()
```
djozis, line 155: Was this intentionally left?

roles/v2/pool/src/status.rs
```diff
@@ -0,0 +1,115 @@
...
+    Upstream(async_channel::Sender<Status>),
+}
+
+impl Sender {
+    /// used to clone the sending side of the status channel used by the TCP Listener
+    /// into individual Sender's for each Downstream instance
+    pub fn listener_to_connection(&self) -> Self {
+        match self {
+            Self::DownstreamListener(inner) => Self::Downstream(inner.clone()),
+            _ => unreachable!(),
```
djozis, line 19: I'm probably missing context on what makes this unreachable, but should this be a different panic of some kind, describing the issue?

coleFD: Possibly. For context, the `Sender<Status>` is wrapped in an enum to let the main thread know which component encountered the error. This function is used to clone the `Sender` sent to the TCP listener component on creation into a new `Sender` for each `Downstream` that gets created by the listener. Since this is the only time there is a Parent -> Child relationship regarding the `Sender<Status>` type, it is only needed for this specific use case.

roles/v2/pool/src/lib/template_receiver/mod.rs
coleFD: Yes, this should return a SubmitShares.Error message defined by the sv2 spec. it is in our backlog to implement this

roles/v2/pool/pool-config.toml
```diff
@@ -1,6 +1,6 @@
 listen_address = "0.0.0.0:34254"
-# tp_address = "127.0.0.1:8442"
-tp_address = "75.119.150.111:8442"
+tp_address = "127.0.0.1:8442"
+# tp_address = "75.119.150.111:8442"
```
darricksee, line 3: Let's keep this pointed at the hosted one - the INTEROPERABILITY.md says that the pool connects directly to the hosted TP.

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -52,7 +52,7 @@ impl ParseDownstreamMiningMessages<(), NullDownstreamMiningSelector, NoRouting>
                     Err(e) => Err(e),
                 }
             })
-            .unwrap()?;
+            .map_err(|_| roles_logic_sv2::Error::PoisonLock)??;
```
darricksee, line 55: Just double checking that `??` was intended.

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -89,7 +91,7 @@ impl ParseDownstreamMiningMessages<(), NullDownstreamMiningSelector, NoRouting>
         let res = self
             .channel_factory
             .safe_lock(|cf| cf.on_submit_shares_standard(m.clone()))
-            .unwrap();
+            .map_err(|_| roles_logic_sv2::Error::PoisonLock)?;
```
darricksee, line 94: Why not pass the _ into PoisonLock?

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -104,9 +106,9 @@ impl ParseDownstreamMiningMessages<(), NullDownstreamMiningSelector, NoRouting>
                         version: share.get_version(),
                         header_timestamp: share.get_n_time(),
                         header_nonce: share.get_nonce(),
-                        coinbase_tx: coinbase.try_into().unwrap(),
+                        coinbase_tx: coinbase.try_into()?,
                     };
-                    // TODO we can block everything with the below
+                    // TODO we can block everything with the below (looks like this will infinite loop??)
```
darricksee, line 111: Yeah - looks like it'll just spam the `solution_sender` if that try_send fails... @Fi3 said he's implementing this branch now so hopefully it'll resolve itself soon.

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -218,7 +220,7 @@ impl ParseDownstreamMiningMessages<(), NullDownstreamMiningSelector, NoRouting>
                     self_.get_downstream_mining_data(),
                 )
             })
-            .unwrap();
+            .map_err(|_| roles_logic_sv2::Error::PoisonLock)?;
         // Is fine to unwrap on safe_lock
```
darricksee, line 224: This comment is obsolete now.

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -81,28 +84,40 @@ impl Downstream {
         task::spawn(async move {
             debug!("Starting up downstream receiver");
             loop {
-                let receiver = cloned.safe_lock(|d| d.receiver.clone()).unwrap();
+                let receiver_res = cloned
```
darricksee, line 87: Can we get away with moving this outside the loop? Seems silly to clone it on every iteration right?

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -178,78 +197,99 @@ impl Pool {
...
                 .safe_lock(|s| {
                     s.channel_factory
                         .safe_lock(|f| f.on_new_prev_hash_from_tp(&new_prev_hash))
-                        .unwrap()
+                        .map_err(|e| PoolError::PoisonLock(e.to_string()))
                 })
-                .unwrap();
+                .map_err(|e| PoolError::PoisonLock(e.to_string()));
+            let x_res = handle_result!(status_tx, x_res);
+            let x = handle_result!(status_tx, x_res);
```
darricksee, line 284: nitpick - can you change the name of this variable to something descriptive of its content - maybe job_id_res

roles/v2/pool/src/lib/mining_pool/message_handler.rs
coleFD: yep this is another case where the function inside the safe_lock returns a result inside the safe_lock result. The double ?? handles both since they are both compatible with the return type of the function

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -273,36 +317,55 @@ impl Pool {
...
                 "New template received, creating a new mining job(s): {:?}",
                 new_template
             );
 
-            let channel_factory = self_.safe_lock(|s| s.channel_factory.clone()).unwrap();
-            let mut messages = channel_factory
-                .safe_lock(|cf| cf.on_new_template(&mut new_template).unwrap())
-                .unwrap();
-            let downstreams = self_.safe_lock(|s| s.downstreams.clone()).unwrap();
+            let channel_factory = self_
```
darricksee, line 327: Do you think we need to clone the channel_factory on every  message - or can this be outside the loop?

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -178,78 +197,99 @@ impl Pool {
...
-                .unwrap();
+                .map_err(|e| PoolError::PoisonLock(e.to_string()));
+            let x_res = handle_result!(status_tx, x_res);
+            let x = handle_result!(status_tx, x_res);
+
             match x {
                 Ok(job_id) => {
-                    let downstreams = self_.safe_lock(|s| s.downstreams.clone()).unwrap();
+                    let downstreams = self_
+                        .safe_lock(|s| s.downstreams.clone())
```
darricksee, line 289: My spidy sense tells me that this is a race condition. We are cloning the list of downstreams and then doing something important with that vec. It's possible for a new downstream to be added to the list as we are sending out the new prevhash. I'm not sure if that's an issue or not - it depends on how the downstream gets its first prevhash... I'll add an issue to investigate... https://github.com/stratum-mining/stratum/issues/430

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -331,27 +393,29 @@ impl Pool {
             new_template_processed: false,
             channel_factory,
             last_prev_hash_template_id: 0,
+            status_tx,
         }));
 
         let cloned = pool.clone();
         let cloned2 = pool.clone();
-        let cloned3 = pool.clone();
+        let cloned3 = pool;
```
darricksee, line 401: This one isn't cloned though...

roles/v2/pool/src/main.rs
```diff
@@ -135,19 +137,23 @@ async fn main() {
...
         config.tp_address.parse().unwrap(),
         s_new_t,
         s_prev_hash,
         r_solution,
         r_message_recv_signal,
+        status::Sender::Upstream(status_tx.clone()),
     )
-    .await;
+    .await
+    .unwrap();
```
darricksee, line 156: Can you add a comment here along the lines of "unwrap() ok here because we want the pool to die if it can't connect to the TP"?

roles/v2/pool/src/status.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    pub async fn send(&self, status: Status) -> Result<(), async_channel::SendError<Status>> {
```
darricksee, line 23: This is currently never used so can you add a `_` to it - I'd love to get to inbox(0) on these types of warnings.

utils/error-handling/src/lib.rs
```diff
@@ -7,7 +7,7 @@
 /// 1. can only be used within async functions since status needs to be send over async channel
 /// 2. The macro must be used within a loop since it calls `continue` on error. If `unwraps/expects` are used within a function
 ///     without a lopo, you should make the function return a result and handle the result within a main loop
-/// 3. The macromust be able to reference the user defined function `crate::status::handle_error(T, U);` where U is the output
+/// 3. The macromust be able to reference the user defined function `crate::status::handle_error(T, U) -> ErrorBranch;` where U is the output
```
darricksee, line 10: macromust -> macro must

roles/v2/pool/src/status.rs
djozis: Might be good to at least document in the doc comment that it panics if not called on a Self::DownstreamListener. This feels weird to me, but I think I'm missing too much context still to suggest an alternative at this time.

roles/v2/pool/src/main.rs
xraid: > Can you add a comment here along the lines of "unwrap() ok here because we want the pool to die if it can't connect to the TP"?

i don't think pool should die even if its dead in one sense, should keep trying connect to TP in list
but also need some way signal Ops is needed . trigger a `alert_TP_down.sh` that ops then can choose use text mail phone etc. for instant notification

darricksee: In general I agree. I added this gh issue that should resolve this - https://github.com/stratum-mining/stratum/issues/427 This PR isn't about fixing all of the issues so I'm ok with letting this one get merged as long as we have a gh issue documenting the problem and we can fix it later.

xraid: > In general I agree. I added this gh issue that should resolve this - #427 This PR isn't about fixing all of the issues so I'm ok with letting this one get merged as long as we have a gh issue documenting the problem and we can fix it later.

should We introduce a backlog.md for will make complete later TODO 's

xraid: > > In general I agree. I added this gh issue that should resolve this - #427 This PR isn't about fixing all of the issues so I'm ok with letting this one get merged as long as we have a gh issue documenting the problem and we can fix it later.
> 
> should We introduce a backlog.md for will make complete later TODO 's

or just in code have the TODO 's

becouse this GH issues and PR's is messy . ppl close issues to reopen in own name and reference issues to new issues that gets to be spagetti in end ...

a issue should be kept as a issue and further the issue in the issue . agree ?

darricksee: I don't entirely understand. I think that if GH issues and PRs are messy then we should get better and organizing them. I wouldn't want to have to maintain a backlog.md in addition to GH issues.

darricksee (approved):

PR #426, opened 2023-01-31
Improved stream reading including detection of end of stream and multiple messages per packet, resolves #399

### LINKED ISSUE #399: Bug: TProxy TcpStream Read Infinite Loop
The TProxy CPU will increase rapidly and only handle a few connections because due to the nature of `async_std::io::BufReader`, `Lines.next()` will return `None` when there is no data in the buffer, so the `while` loop starting on line 93 (https://github.com/stratum-mining/stratum/blob/main/roles/translator/src/downstream_sv1/downstream.rs#L87-102) only loops when it receives a message. All other times the task is stuck looping in the parent. 

I was able to 14 connections before adding a 5 millis sleep in the parent loop. Then I was able to connect 10k connections at 50% cpu. Sleeping is probably shouldnt be the long term solution. This websocket interface maybe a good thing to take inspiration from (https://docs.rs/websocket/0.26.5/websocket/async/struct.TcpStream.html#method.poll_read_ready)

Also something to keep in mind:
Currently, when a miner drops its connection, the TProxy does not close the TCP file descriptors so CPU is never freed. Because of this, a malicious client could repetitively connect and disconnect to overload the TProxy

djozis:


roles/translator/src/downstream_sv1/downstream.rs
```diff
@@ -104,26 +104,30 @@ impl Downstream {
         // role, or the message is sent upwards to the Bridge for translation into a SV2 message
         // and then sent to the SV2 Upstream role.
         let _socket_reader_task = task::spawn(async move {
+            let mut messages = BufReader::new(&*socket_reader).lines();
```
djozis, line 107: I don't think it's new with this commit, but I believe this is vulnerable to DOS attacks since `lines()` will read indefinitely, storing all read bytes if no line terminator is found (potential out-of-memory).

pavlenex:
Hey @djozis, thanks for your PR 🎉 
I am not sure if you are on our Discord, but if you aren't you should join in there, a lot of SV2 contributors hang around there and it's great place to ask questions, get feedback on PR's etc.

I'll let @coleFD and @darricksee review this as they're most familiar with tProxy. I just approved the tests to run.

roles/translator/src/downstream_sv1/downstream.rs
darricksee: Isn't it bounded by the capacity from the BufReader::new() which has a default capacity of 1024?

pub fn new(inner: R) -> BufReader<R> {
        BufReader::with_capacity(DEFAULT_BUF_SIZE, inner)
    }

roles/translator/src/downstream_sv1/downstream.rs
```diff
@@ -104,26 +104,30 @@ impl Downstream {
...
+                            Some(Ok(incoming)) => {
+                                debug!("Receiving from Mining Device: {:?}", &incoming);
+                                let incoming: json_rpc::Message = handle_result!(tx_status_reader, serde_json::from_str(&incoming));
+                                // Handle what to do with message
+                                let res = Self::handle_incoming_sv1(self_.clone(), incoming).await;
+                                handle_result!(tx_status_reader, res);
+                            }
+                            Some(Err(error)) => {
+                                debug!("Error reading from Mining Device connection: {:?}", &error);
+                                break;
```
darricksee, line 125: I'm thinking in these 2 cases we should use `handle_result` to remain consistent. Something like handle_result!(tx_status_reader, Err(error));

roles/translator/src/downstream_sv1/downstream.rs
darricksee, line 107: By moving this outside the loop it'll only read from the socket once and then the task will die once it completes processing all the messages.

djozis: In my tests, the stream from `.lines()` blocks appropriately when calling `next()` until it reads end of line or reads EOF (ie. the underlying reader returns `Ok(0)`) - and correctly awaits the underlying reader when it doesn't immediately have data available. In my tests, `TcpStream` also only returned `Ok(0)` when the connection had been remotely terminated, which makes sense, since it is the EOF value - and I checked the underlying std variation and examples which supported this. When I put it all together it worked correctly in my test harness - let me know if/where you're seeing what you're describing and I'll check it out!

coleFD: In my local test everything worked as expected

djozis: The capacity will make it so the Lines aggregator will only read 8 kb at a time, but the aggregator itself has no cap as far as I can tell:
https://github.com/async-rs/async-std/blob/main/src/io/buf_read/lines.rs#L64
https://github.com/async-rs/async-std/blob/main/src/io/buf_read/mod.rs#L242-L266

roles/translator/src/downstream_sv1/downstream.rs
djozis: I went ahead and updated the error case.

In the case where the connection was terminated, do you have a suggestion for the appropriate "Error" to be passing to handle_result?

darricksee: This one seems reasonable -     Error::Io(std::io::Error) - though you'd have to make an Error.

darricksee (approved):

PR #431, opened 2023-02-03
WIP Add share handler for jn proxy

Fi3:


protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -52,12 +52,59 @@ pub enum OnNewShare {
...
     ShareMeetBitcoinTarget((Share, u64, Vec<u8>)),
     /// Indicate that the share meet downstream target, in the case we could send a success
     /// response dowmstream.
     ShareMeetDownstreamTarget,
 }
 
+impl OnNewShare {
+    pub fn into_extended(&mut self, extranonce: Vec<u8>) {
+        match self {
+            OnNewShare::SendErrorDowsntream(_) => (),
```
darricksee, line 65: Dowsntream -> Downstream

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -43,7 +43,7 @@ pub struct PartialSetCustomMiningJob {
 pub enum OnNewShare {
     /// Used when the received is malformed, is for an inexistent channel or do not meet downstream
     /// target.
-    SendErrorDowsntream(SubmitSharesError<'static>),
+    SendErrorDownstream(SubmitSharesError<'static>),
```
djozis, line 46: On line 1481 in this file, SendErrorDowsntream needs to be updated (tests don't compile).

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -52,12 +52,59 @@ pub enum OnNewShare {
...
+                    let share = SubmitSharesExtended {
+                        channel_id: up_id,
+                        sequence_number: share.sequence_number,
+                        job_id: share.job_id,
+                        nonce: share.nonce,
+                        ntime: share.ntime,
+                        version: share.version,
+                        extranonce: extranonce.try_into().unwrap(),
+                    };
+                    let mut self_ = Self::SendSubmitShareUpstream(Share::Extended(share));
```
djozis, line 78: `*self = Self::Send...` might be easier to read, unless I missed a reason to do it this way (same on lines 95 - 100).

protocols/v2/roles-logic-sv2/src/selectors.rs
```diff
@@ -65,14 +65,56 @@ impl<Down: IsMiningDownstream> DownstreamMiningSelector<Down>
...
             .remove(&channel_id)
-            .unwrap_or_default()
+            .unwrap_or_default();
+        for d in &downs {
+            self._remove_downstream(d.clone());
+        };
+        downs
     }
 
+    fn _remove_downstream(&mut self, d: Arc<Mutex<Down>>) {
```
djozis, line 77: Retain simplifies this.
```
    fn _remove_downstream(&mut self, d: Arc<Mutex<Down>>) {
        self.request_id_to_remotes.retain(|k, v| {
            !Arc::ptr_eq(v, &d)
        });
        self.channel_id_to_downstream.retain(|k, v| {
            !Arc::ptr_eq(v, &d)
        });
    }
```

protocols/v2/roles-logic-sv2/src/selectors.rs
```diff
@@ -65,14 +65,56 @@ impl<Down: IsMiningDownstream> DownstreamMiningSelector<Down>
     }
 
     fn remove_downstreams_in_channel(&mut self, channel_id: u32) -> Vec<Arc<Mutex<Down>>> {
```
djozis, line 67: It looks like `remove_downstream` and `_remove_downstream` have been added to this trait solely for the internal implementation of this function - basing that assessment on the other implementation of this trait invoking `unreachable!` in the implementation, and these functions not being invoked anywhere else in this pull request.

If that's the case, perhaps the functions might have appropriately restricted visibility by either declaring them inside of `remove_downstreams_in_channel` (ie. nested functions), or declaring them in a `impl<Down: IsMiningDownstream> ProxyDownstreamMiningSelector<Down> {` block outside the trait implementation?

Let me know if I missed something...

protocols/v2/roles-logic-sv2/src/selectors.rs
```diff
@@ -65,14 +65,56 @@ impl<Down: IsMiningDownstream> DownstreamMiningSelector<Down>
...
+            if Arc::ptr_eq(v, &d) {
+                ks.push(k.clone())
+            }
+        };
+        for k in ks {
+            self.channel_id_to_downstream.remove(&k).unwrap();
+        };
+    }
+
+    fn remove_downstream(&mut self, d_: Arc<Mutex<Down>>) {
```
djozis, line 100: This and `_remove_downstream` look like they can both accept `&Arc<Mutex<Down>>` instead of `Arc<Mutex<Down>>` (results in not having to increase/decrease the atomic reference count unnecessarily).

roles/v2/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -306,6 +313,23 @@ impl DownstreamMiningNode {
     }
 
     pub fn exit(self_: Arc<Mutex<Self>>) {
+        if let Some(up) = self_.safe_lock(|s| s.upstream.clone()).unwrap() {
+            super::upstream_mining::UpstreamMiningNode::remove_dowstream(up,self_.clone());
+            //if let Some(id) = self_.safe_lock(|d| {
```
djozis, line 318: Were these comments intentionally left here? If so, perhaps some context for what they might be for could be useful.

roles/v2/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -419,22 +415,30 @@ impl
...
-                // sending them upstream If that is the case it should be
-                // done by GroupChannel not here
                 let message = Mining::SubmitSharesStandard(m);
                 Ok(SendTo::RelayNewMessageToRemote(remote.clone(), message))
             }
+            DownstreamMiningNodeStatus::ChannelOpened(Channel::DowntreamHomUpstreamExtended {
+                ..
+            }) => {
+                let remote = self.upstream.as_ref().unwrap();
+                let res = UpstreamMiningNode::handle_std_shr(remote.clone(), m).unwrap();
```
djozis, line 435: Is this unwrap safe? If it is and it's obvious, it might just be my limited knowing being newer to the code base... but otherwise, a comment explaining why could be helpful.

roles/v2/mining-proxy/src/lib/job_negotiator/mod.rs
```diff
@@ -33,22 +33,26 @@ use setup_connection::SetupConnectionHandler;
...
+        config: Config,
     ) -> Arc<Mutex<Self>> {
         let stream = TcpStream::connect(address).await.unwrap();
         let initiator = Initiator::from_raw_k(authority_public_key).unwrap();
         let (mut receiver, mut sender): (Receiver<EitherFrame>, Sender<EitherFrame>) =
             Connection::new(stream, HandshakeRole::Initiator(initiator)).await;
 
-        let config_file: String = std::fs::read_to_string("proxy-config.toml").unwrap();
-        let config: Config = toml::from_str(&config_file).unwrap();
+        //let config_file: String = std::fs::read_to_string("proxy-config.toml").unwrap();
```
djozis, line 54: What are these two commented lines intended for?

roles/v2/mining-proxy/src/lib/job_negotiator/mod.rs
```diff
@@ -91,17 +103,19 @@ impl JobNegotiator {
...
-                            .unwrap();
                         let coinbase_output_max_additional_size = CoinbaseOutputDataSize {
                             coinbase_output_max_additional_size: m
                                 .coinbase_output_max_additional_size,
                         };
-                        sender
+                        let out_script = self_mutex
+                            .safe_lock(|s| s.last_coinbase_out.clone())
+                            .unwrap()
+                            .unwrap();
```
djozis, line 113: Is this (second) `unwrap` safe? If so, I'm curious, what makes it safe (ie. what guarantees that `last_coinbase_out` is not `None`)?

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -35,6 +38,105 @@ pub type StdFrame = StandardSv2Frame<Message>;
 pub type EitherFrame = StandardEitherFrame<Message>;
 pub type ProxyRemoteSelector = Prs<DownstreamMiningNode>;
 
+use std::sync::atomic::AtomicBool;
+/// USED to make sure that if a future new_temnplate and a set_new_prev_hash are received together
+/// the future new_temnplate is always handled before the set new prev hash.
+pub static IS_NEW_TEMPLATE_HANDLED: AtomicBool = AtomicBool::new(true);
```
djozis, line 44: Instead of being static, should this be an `Arc<AtomicBool>` or similar in a field in a struct somewhere?

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -35,6 +38,105 @@ pub type StdFrame = StandardSv2Frame<Message>;
...
+                let factory = ProxyExtendedChannelFactory::new(
+                    group_id,
+                    extranonces,
+                    None,
+                    downstream_share_per_minute,
+                    kind,
+                    Some(vec![]),
+                    up_id,
+                );
+                let mut initialized = Self::Extended(Some(factory));
```
djozis, line 108: `*self = Self::Extended...` might be simpler.

djozis: (On line 123 as well.)

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -90,14 +192,28 @@ pub struct UpstreamMiningNode {
     /// The `request_id` from the downstream is NOT guaranteed to be unique, so it must be changed.
     request_id_mapper: RequestIdMapper,
     downstream_selector: ProxyRemoteSelector,
-    pub group_channels: GroupChannels,
-    channel_factory: Option<ProxyExtendedChannelFactory>,
+    //pub group_channels: GroupChannels,
+    //channel_factory: Option<ProxyExtendedChannelFactory>,
```
djozis, line 196: What's intended with these commented lines?

protocols/v2/roles-logic-sv2/src/selectors.rs
Fi3: You are right no need to have it in the trait, I put it there trying to make things works and then left there.

roles/v2/mining-proxy/src/lib/upstream_mining.rs
Fi3: Why do you think that would be better to have it in a field in a struct somewhere? Maybe it would be a little bit more safe but not so much cause the visibility need to be almost global. We could initialize it in main and pass it around that would make the codebase more complex not sure if it is worth it.

roles/v2/mining-proxy/src/lib/template_receiver/mod.rs
```diff
@@ -64,7 +67,9 @@ impl TemplateRx {
         .try_into()
         .unwrap();
         Self::send(self_mutex.clone(), sv2_frame).await;
+        let cloned = self_mutex.clone();
 
+        tokio::task::spawn(async { Self::on_new_solution(cloned, solution_receiver).await });
```
djozis, line 72: Looks like this could be simplified to `tokio::task::spawn(Self::on_new_solution(cloned, solution_receiver));`.

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -671,6 +852,96 @@ impl UpstreamMiningNode {
...
+
+    pub fn handle_std_shr(
+        self_: Arc<Mutex<Self>>,
+        share_: SubmitSharesStandard,
+    ) -> Result<Mining<'static>, Error> {
+        if self_.safe_lock(|s| s.channel_kind.is_extended()).unwrap() {
+            let share = self_
+                .safe_lock(|s| {
+                    let factory = s.channel_kind.get_factory();
+                    // TODO remove this unwrap
```
djozis, line 867: Which unwrap? Not seeing adjacent to this comment.

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -671,6 +852,96 @@ impl UpstreamMiningNode {
...
+                            channel_id: share_.channel_id,
+                            last_sequence_number: share_.sequence_number,
+                            new_submits_accepted_count: 1,
+                            new_shares_sum: 1,
+                        };
+                        let message = Mining::SubmitSharesSuccess(success);
+                        Ok(message)
+                    }
+                    Share::Standard(_) => {
+                        unreachable!()
```
djozis, line 929: What makes this unreachable? I tried to figure it out, but it wasn't so easy for me.

message-generator-tests.sh
```diff
@@ -3,7 +3,12 @@ message_generator_dir="./utils/message-generator/"
 
 cd $message_generator_dir
 
+command=""
```
darricksee, line 6: I don't get the point of this change. My guess is it's no longer needed?

roles/v2/mining-proxy/proxy-config.toml
```diff
@@ -1,11 +1,11 @@
...
+    { channel_kind = "Group", address = "127.0.0.1", port = 44254, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL", jn_values = { address = "127.0.0.1", port = 44264, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"}},
+    { channel_kind = "Group", address = "127.0.0.1", port = 54254, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL", jn_values = { address = "127.0.0.1", port = 54264, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"}},
 ]
-upstreams_jn = [{ address = "127.0.0.1", port = 34264, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"}]
 tp_address = "127.0.0.1:8442" 
 listen_address = "127.0.0.1"
 listen_mining_port = 34255
 max_supported_version = 2
 min_supported_version = 2
+downstream_share_per_minute = 0.01
```
darricksee, line 11: Why so low?

Fi3: this is supposed to be used on regetes and with my PC seems a good value. Btw I can change it but not sure about the value to use

roles/v2/mining-proxy/proxy-config.toml
```diff
@@ -1,11 +1,11 @@
 upstreams = [
-    { channel_kind = "Group",  address = "127.0.0.1", port = 34254, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"},
-    { channel_kind = "Group",  address = "127.0.0.1", port = 44254, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"},
-    { channel_kind = "Group",  address = "127.0.0.1", port = 54254, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"}
+    { channel_kind = "ExtendedWithNegotiator", address = "127.0.0.1", port = 34254, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL", jn_values = { address = "127.0.0.1", port = 34264, pub_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"}},
```
darricksee, line 2: I would love to see documentation on these config values and what they do.

roles/v2/mining-proxy/proxy-config.toml
darricksee: Ideally the way this would work is you'd specify something like 1 share per minute and the proxy would calculate the correct difficulty to send to the miner such that they submit about 1 share per minute. In my mind .01 shares per minute would mean that the miner will only send 1 share every 100 minutes.

roles/v2/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -342,62 +352,34 @@ impl
...
-                            m.group_channel_id,
-                        );
-                    }
+            .safe_lock(|up| up.channel_kind.is_extended())
+            .unwrap()
+        {
+            let messages = up
+                .as_mut()
+                .unwrap()
+                .safe_lock(|up| {
```
darricksee, line 365: Instead of locking a 2nd time why not just pull the messages out in the above `safe_lock` call?

roles/v2/mining-proxy/src/lib/job_negotiator/message_handler.rs
```diff
@@ -6,14 +6,22 @@ use roles_logic_sv2::{
...
+use roles_logic_sv2::{bitcoin::TxOut, errors::Error};
 
 impl ParseServerJobNegotiationMessages for JobNegotiator {
     fn handle_set_coinbase(&mut self, message: SetCoinbase) -> Result<SendTo, Error> {
         info!(
             "Received allocate mining job token success message: {:?}",
             message
         );
+        let txout = TxOut {
+            value: 5_000_000_000,
```
darricksee, line 19: Does this get overwritten somewhere?

roles/v2/mining-proxy/src/lib/job_negotiator/mod.rs
```diff
@@ -77,6 +81,12 @@ impl JobNegotiator {
...
     pub fn on_upstream_message(self_mutex: Arc<Mutex<Self>>) {
         task::spawn(async move {
+            let sender_max_size = self_mutex
+                .safe_lock(|s| s.sender_coinbase_output_max_additional_size.clone())
+                .unwrap();
+            let sender_out_script = self_mutex
+                .safe_lock(|s| s.sender_coinbase_out.clone())
+                .unwrap();
             loop {
                 let receiver = self_mutex.safe_lock(|d| d.receiver.clone()).unwrap();
```
darricksee, line 91: Do you know why we need to clone the receiver on every iteration of the loop?

roles/v2/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -419,22 +401,32 @@ impl
...
+                ..
+            }) => {
+                // Safe unwrap is channel have been opened it means that the dowsntream is paired
+                // with an upstream
+                let remote = self.upstream.as_ref().unwrap();
+                let res = UpstreamMiningNode::handle_std_shr(remote.clone(), m).unwrap();
+                Ok(SendTo::Respond(res))
+            }
             _ => {
                 info!("Rceived share TODO send it somwhere");
```
darricksee, line 427: I'd love to understand more about when this happens. Can you write up a github issue around this panic so we can understand it and have a record to fix later.

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -131,32 +243,72 @@ impl UpstreamMiningNode {
...
             recv_ph,
             request_ids,
             channel_ids,
+            downstream_share_per_minute,
+            solution_sender,
+            recv_coinbase_out,
+            tx_outs: HashMap::new(),
+            first_ph_received: false,
+            job_up_to_down_ids: HashMap::new(),
+            //is_new_template_handled: true,
```
darricksee, line 260: Is this needed any longer?

roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -131,32 +243,72 @@ impl UpstreamMiningNode {
             channel_id_to_job_dispatcher: HashMap::new(),
             request_id_mapper,
             downstream_selector,
-            group_channels: GroupChannels::new(),
-            channel_factory: None,
-            channel_kind,
+            //group_channels: GroupChannels::new(),
```
darricksee, line 246: Can these now be removed?

darricksee (approved):

roles/v2/mining-proxy/proxy-config.toml
Fi3: Is like that but because when sending `OpenStandardMiningChannel` the test mining-device put a value not related with  the real hash-rate of the cpu (5.4) to make things usable I need to put 0.01 share per minute cause the proxy use that non sense hash rate value to calculate the target. Maybe in the future we should enhance the CPU mining device so that is able to communicate a right hash rate.

roles/v2/mining-proxy/src/lib/downstream_mining.rs
Fi3: I thought that it was more readable in that way and that the compiler would have been smart enough to not lock 2 times but I'm changing it to lock only one time.

roles/v2/pool/pool-config.toml
```diff
@@ -1,6 +1,6 @@
 listen_address = "0.0.0.0:34254"
-# tp_address = "127.0.0.1:8442"
-tp_address = "75.119.150.111:8442"
```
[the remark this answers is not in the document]
Fi3: Not sure maybe we can discuss it during next call for now I leave it like that. I prefer to default to self hosted TP and opt out for the hosted one.

PR #432, opened 2023-02-06
Added handling of interrupt signal

djozis:


bcral:
Pulled and tested locally - really helpful for closing processes gracefully instead of panicking.  LGTM.

roles/translator/src/main.rs
```diff
@@ -161,9 +162,26 @@ async fn main() {
...
-        let task_status = rx_status.recv().await.unwrap();
+        let task_status = select! {
+            task_status = rx_status.recv().fuse() => task_status,
+            interrupt_signal = interrupt_signal_future => {
+                match interrupt_signal {
+                    Ok(()) => {
+                        println!("Interrupt received!");
+                    },
+                    Err(err) => {
+                        eprintln!("Unable to listen for interrupt signal: {}", err);
```
darricksee, line 177: eprintln -> println

roles/translator/src/main.rs
```diff
@@ -161,9 +162,26 @@ async fn main() {
...
+
     // Check all tasks if is_finished() is true, if so exit
     loop {
-        let task_status = rx_status.recv().await.unwrap();
+        let task_status = select! {
+            task_status = rx_status.recv().fuse() => task_status,
+            interrupt_signal = interrupt_signal_future => {
+                match interrupt_signal {
+                    Ok(()) => {
+                        println!("Interrupt received!");
```
darricksee, line 174: These should use the typical `info!("Interrupt received");` log output

roles/translator/src/main.rs
darricksee, line 177: use `error!()` here

roles/translator/src/main.rs
```diff
@@ -161,9 +162,26 @@ async fn main() {
         b,
     );
 
+    let mut interrupt_signal_future = Box::pin(tokio::signal::ctrl_c().fuse());
+
     // Check all tasks if is_finished() is true, if so exit
     loop {
-        let task_status = rx_status.recv().await.unwrap();
+        let task_status = select! {
+            task_status = rx_status.recv().fuse() => task_status,
```
darricksee, line 170: Can you add a comment describing what this select! is doing? Just something simple like "this code is waiting for either a value to be received on the rx_status channel or for an interrupt signal to be received and then breaking the main loop to end the process if appropriate"

djozis: Hmm... to me, comments are about explaining why code is doing what it is doing, when it isn't clear why. I'm happy to defer to you if you feel it's valuable here, though.

darricksee: Yeah, that's fair! This is pretty self explanatory so I'm fine with no comment.

darricksee (approved):

PR #434, opened 2023-02-06
Version Control

coleFD:
I left this as a WIP since there are a couple things that could possibly change based on team discussion. 

As it stands, this PR creates a Job that will run only on pushes to `main`. The Action auto detects changes to crates, bumps the patch version if there are changes, creates a new commit and pushes the commit to the `bot/versioning` branch. Currently, publishing and changelog generation will require a manual usage of the `sv2-publish.sh` script. Changing major/minor versions according to Semver is still not supported, so when making major/minor changes, we will still need to manually set the crate version.

Discussion topics:
1. should we have the bot push the version commit directly to `main` or should it push to the `bot/versioning` branch and auto PR 
2. Github secrets to allow publishing from a manual CI job, and possibly a GPG key that allows the autoversion Job to force push to main

# Edit 1:
The `autoversion` workflow will now submit the PR and will not run on merges into main from this PR. Just need to handle publishing releases


.github/workflows/release.yaml
```diff
@@ -0,0 +1,249 @@
+# This workflow is used to publish SV2 crates to cratesio as well as releasing binaries of the roles
+# (translator, pool, mining_proxy) to github. The github releases will fail if there are no tags available
+# so the `autoversion` workflow must be ran and the resulting PR must be merged at least once before this will 
+# release the binaries to github. 
+
+# If the binary releases do fail to due to not having tags, force run the `autoversion` workflow
+# on the main branch and merge the resulting PR to create the tags and move them to the main branch.
+
+# This should be disabled for forks
```
darricksee, line 9: Is it disabled for forks or is there something every fork needs to do?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,249 @@
...
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: 
+        run: echo TEST
+      
+      # - name: Install cargo-smart-release
```
darricksee, line 39: Can you remove this commented out block or add another comment as to why we want to keep it here?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,249 @@
...
+  github_release_translator:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # This defines what environments we will build on
+        include:
+          - os: ubuntu-latest
+            artifact_name: translator_sv2
+            asset_name: translator_sv2-linux-amd64
+          # - os: windows-latest
```
darricksee, line 66: Im assuming this is commented out because it doesnt work on windows?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,249 @@
...
+
+      # # PR updates
+      # - name: Create pull request
+      #   uses: peter-evans/create-pull-request@v4
+      #   with:
+      #     base: main
+      #     branch: bot/versioning
+
+  github_release_translator:
+    runs-on: ${{ matrix.os }}
```
darricksee, line 58: We should probably block windows here if it doesnt work.

.github/workflows/release.yaml
coleFD: Yep. I will make a ticket once this gets merged

.github/workflows/release.yaml
coleFD: windows-latest is commented out from the `matrix.include` block so it will not run on windows

darricksee (approved):
This is approved - I'll wait for @fi3's token->github work to happen before merging.

darricksee (approved):

PR #438, opened 2023-02-08
used shrink for prop tests

Sakshya2308:
1.  Implemented `shrink` while implementing `Arbitrary`.
2. Added `CONTRIBUTING.md` file so that people continue to use this pattern.

### DIFF
--- a/protocols/v1/src/tempCodeRunnerFile.rs
+++ b/protocols/v1/src/tempCodeRunnerFile.rs
@@ -1,0 +1,1 @@
+Arbitrary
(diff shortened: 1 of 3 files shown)


protocols/v1/src/tempCodeRunnerFile.rs
```diff
@@ -0,0 +1 @@
+Arbitrary
```
darricksee, line 1: What is this file for?

Sakshya2308: It was there in the tempCodeRunnerFile.rs before, I thought it'd not be a good idea to touch it, since i just started rust.

pavlenex:
@Sakshya2308 do you still intend to work on this one, if so, could you rebase, if not could you close it?

PR #441, opened 2023-02-14
Feat/coinbase config

coleFD:
This PR allows us to set coinbase payout addresses in the pool config file. 

Notes:
1. The message generator test does not support the ability to check multiple fields for a single message so the msg gen test added only checks the `coinbase_tx_suffix` as of now. The test still passes if the second value is removed (only check `coinbase_tx_prefix`
2. I am not sure why the values in the message gen test require `Seq64K` instead of `B064K` as defined in the spec. Is this a message generator bug?


roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -391,10 +391,18 @@ impl Pool {
...
+            .iter()
+            .map(|pub_key_wrapper| {
+                TxOut {
+                    // value will be updated by the addition of `ChannelFactory::split_outputs()` in PR #422
+                    value: crate::BLOCK_REWARD,
+                    script_pubkey: Script::new_p2pk(&pub_key_wrapper.pub_key),
+                }
+            })
+            .collect();
+        println!("PUB KEY: {:?}", pool_coinbase_outputs);
```
darricksee, line 405: This should use info!()

darricksee (approved):

darricksee (approved):

PR #443, opened 2023-02-16
JN in tProxy

4ss0:
JN in the translator proxy for the tests and demos with sv1 mining devices for MVP2 release


pavlenex:
LET'S GOOOOOOOOOO! 🚀

pavlenex:
Continued in #471

PR #445, opened 2023-02-16
Hot fix for mg action

bcral:
Adds a quick patch to force the MG Github Action to fail if any one of the tests panics.

Also fixes a few configs to match new requirements.


bcral:
The message generator check should fail - that's the main reason for this PR.

bcral:
No longer relevant - Please delete.

PR #446, opened 2023-02-20
Add systemd example for Translator

Sjors:
Create a user `stratum`, install rust and clone the repo. And then use this systemd service to have it run automatically.

### DIFF
--- a/examples/systemd/sv2-translator.service
+++ b/examples/systemd/sv2-translator.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=Stratum v2 Translator
+After=sv2-translator.service
+
+[Service]
+# Assumes a user stratum with Rust installed and the repository cloned into ~/src
+# cargo build -p translator --release
+WorkingDirectory=/home/stratum/src/stratum/roles/translator
+ExecStart=/home/stratum/src/stratum/target/release/translator
+User=stratum
+Group=stratum
+Type=simple
+KillMode=process
+TimeoutSec=60
+Restart=always
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target


examples/systemd/sv2-translator.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Stratum v2 Translator
+After=sv2-translator.service
+
+[Service]
+# Assumes a user stratum with Rust installed and the repository cloned into ~/src
+WorkingDirectory=/home/stratum/src/stratum/roles/translator
+ExecStart=/home/stratum/.cargo/bin/cargo run -p translator
```
Sjors, line 8: I assume there's a way to install the binary and avoid `cargo run`?

Sjors: Updated to us `cargo build -p translator --release` and point to the target dir.

It would be nice if the working directory could just be some random temp directory, but it won't start when I do that.

xraid: $USER/src/stratum/target/release/translator 

?

Sjors: `/usr/local/bin` is not accesible for my `stratum` user (no sudo rights).

`~/.cargo/bin` would be a better location, but I couldn't get the something like `cargo install` to work. Manually copying the binary feels like a hack, but it's an option.

Sjors: Not sure if $USER works inside a systemd service file?

xraid:
maybe add a installTranslarorProxy.sh to do the dance ?

Sjors:
Isn't there a more Rust'y way to do that?

xraid:
https://doc.rust-lang.org/cargo/commands/cargo-install.html

Sjors:
I usually manually install `systemd` services. Using a script is not that convenient, since it would need `sudo` rights.

I made a separate issue for `cargo install`: #450

Sjors:
I wrote:

> It would be nice if the working directory could just be some random temp directory, but it won't start when I do that.

See #451.

Sjors:
Note to self: update instructions to put `proxy-config.toml` somewhere else and point to it with `-c ~/.stratum/proxy-config.toml`

marathon-gary:
Is this PR still current/active/desired?

rrybarczyk:
Closing as doing `systemd` services feels premature and this PR has been stale for a long time.

marathon-gary:
https://github.com/stratum-mining/stratum/issues/1057

PR #449, opened 2023-02-20
Add coverage from message generator to coverage action

bcral:
This change:

1.) Adds a new Message Generator test to test/message-generator/cov for generating a coverage report
2.) Runs this test in .github/workflows/coverage.yaml to generate report, and upload it to codecov with the existing tarpaulin report
3.) Includes information on the cargo-llvm-cov package in the Message Generator README

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,5 +1,5 @@
-# Performs test coverage of project's libraries using cargo-tarpaulin and generates results using
-# codecov.io.
+# Performs test coverage of project's libraries using cargo-tarpaulin and the message-generator, 
+# and generates results using codecov.io.
 # The follow flags are used when executing cargo-tarpaulin:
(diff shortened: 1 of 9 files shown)


bcral:
It's worth noting that, currently, the MG coverage test only tests roles/v2/pool, not the translator proxy.  This, and more detailed coverage for the pool, can be added later.

.github/workflows/coverage.yaml
```diff
@@ -35,7 +37,62 @@ jobs:
...
+      - name: Log data from rustc
+        run: rustc -Vv
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Run mg codecov tests
+        run: sh ./mg-codecov-tests.sh
+
+      - name: Open file
```
darricksee, line 71: We probably don't need this step any longer right?

darricksee (approved):

PR #453, opened 2023-02-23
MG allow match on multiple fields

### LINKED ISSUE #448: Allow MG actions to check more than one value when using match_message_field
Currently, when checking the return value of an action with match_message_field, you can only check one key/value pair in the entire object that is returned.  To check a second value, you would either need to send the message again, or re-run the test to check the second key/value.

It would be extremely beneficial to be able to pass multiple keys/values, fail if they are not all correct, and pass only if they are all correct.

This issue was mentioned in #441

bcral:
Resolves #448 

Allow the MG to check multiple fields per message.

Added some error messages and slight doc changes.

Updated all existing MG tests to prevent a breaking change, increased timeout to pass GH actions, and changed MG action test to fail if any test panics.


utils/message-generator/src/executor.rs
```diff
@@ -397,7 +390,7 @@ impl Executor {
                     ActionResult::MatchExtensionType(ext_type) => {
                         if header.ext_type() != *ext_type {
                             println!(
-                                "WRONG EXTENCTION TYPE expcted: {} received: {}",
+                                "WRONG EXTENSION TYPE expcted: {} received: {}",
```
darricksee, line 393: might as well fix both spelling errors: expcted -> expected

utils/message-generator/src/executor.rs
```diff
@@ -339,50 +331,51 @@ impl Executor {
...
                                         let msg = serde_json::to_value(&m).unwrap();
-                                        check_msg_field(msg,&field_name,&value_type,field);
+                                        check_each_field(msg,field_data);
                                     }
                                 },
                                 Err(e) => panic!("err {:?}",e),
                             }
                         } else {
-                            panic!()
+                            panic!("match_message_field subprotocol not valid")
```
darricksee, line 376: Would be useful to spit out the subprotocol value in the panic to make debugging easier.

PR #469, opened 2023-03-02
Add `measure_hashrate` function and use it to set `nominal_hash_rate` in `OpenStandardMiningChannel`

aniket2405:
As part of `Summer of Bitcoin 2023` competency test, add `measure_hashrate` function and use it to set `nominal_hash_rate` in `OpenStandardMiningChannel`


darricksee:
Can you fix the above lnt/fmt actions?

aniket2405:
Screenshot from 2023-03-02 21-54-27

This particular line causes the panick: `let stream = TcpStream::connect(address).await.unwrap();`. Even though we are using async-await, the `stream` get a NULL value and the `unwrap()` panicks. I can console `address` and it displays the correct IPv4 address but somehow `stream` does not catch it. I tried changing port numbers too, but the same thing persists. I reckon it is an async-await issue. Do you think this might be a reason?

coleFD:
> Screenshot from 2023-03-02 21-54-27
> 
> This particular line causes the panick: `let stream = TcpStream::connect(address).await.unwrap();`. Even though we are using async-await, the `stream` get a NULL value and the `unwrap()` panicks. I can console `address` and it displays the correct IPv4 address but somehow `stream` does not catch it. I tried changing port numbers too, but the same thing persists. I reckon it is an async-await issue. Do you think this might be a reason?

I am not getting the error, do you have a local pool running?

coleFD:
> No, I don't have a local pool running.
> 
> I am using this command to run the rust code: `cargo run --bin mining-device`.
> 
> Are you using something else? The error maybe because of incorrect build.

The TcpStream tries to connect to 127.0.0.1:35255 which is the standard listening address/port of the pool role. Try running `cargo run -p pool_sv2 -- -c roles/v2/pool/pool-config.toml` before you start up the mining-device and see if that works

aniket2405:
image


Hi, this error pops up when I run the command `cargo run -p pool_sv2 -- -c roles/v2/pool/pool-config.toml`

pavlenex:
Hey @aniket2405, I assume you're trying to connect to a local template provider, we do have a hosted one, so you should just comment out local template provider and point to `tp_address = "75.119.150.111:8442"`in roles > v2 > pool > `pool-config.toml`

If for some reason you want to run a local template provider we have it documented [here](https://github.com/stratum-mining/stratum/blob/main/INTEROPERABILITY-TESTS.md#optional-to-locally-run-the-template-provider-tp).

You have instructions on how to run basic stack also [documented on the website.](https://stratumprotocol.org/getting-started.html)  which is really similar to the one linked above.

Also make sure to join our Discord, it's much quicker to get answer to qucik questions in #newbie-questions channel.

Issue #447, opened 2023-02-20
Add better definition to error messages for failed MG tests.


bcral:
Right now the message generator has a very generic panic with no message whenever it fails.  We should have specific messages for:
1. When the MG test times out, ex. "Test failed due to timeout"
2. When the MG test case receives an unexpected message, ex. "STDERR did not match expected case: <Test case>"

These need to be implemented for STDOUT and STDERR, specifically the check_std_out() and check_std_err() functions in message-generator/src/external_commands.rs.

aniket2405:
hey, I can work on this and post a PR soon.

plebhash:
@lorbax is this still relevant?

lorbax:
> @lorbax is this still relevant?

partially relevant

GitGab19:
Closing this as stale

PR #475, opened 2023-03-05
Add better definition to error messages for failed MG tests (#447)

aniket2405:
For the `check_std_out()` function, we first check if the error is a `TimeoutError`, and if it is, we print the `"Test failed due to Timeout"` message.

For the `check_std_err()` function, we check if the error is a `StringMismatchError` and print the `"STDERR did not match expected case"` message along with the `expected` and `actual` values of the error.

If the `get_warn_no_panic()` function returns true, we return an empty `Err value`, else we panic.


roles/v2/pool/pool-config.toml
```diff
@@ -5,7 +5,7 @@ listen_jn_address = "127.0.0.1:34264"
 authority_public_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"
 authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
-test_only_listen_adress_plain =  "0.0.0.0:34250"
```
darricksee, line 8: This can't change to `34255` because [the tproxy is already listening on this port](https://github.com/stratum-mining/stratum/blob/main/roles/translator/proxy-config.toml#L13).

darricksee: I'd rather you changed the miner to connect on port 34250

darricksee (approved):

aniket2405:
Hi, why is the MG Test failing repeatedly? The `details` do not provide a clear picture.

darricksee:
It looks like compile errors:

warning: `roles_logic_sv2` (lib) generated 3 warnings
   Compiling message_generator_sv2 v0.1.0 (/home/runner/work/stratum/stratum/utils/message-generator)
error: expected `{`, found `=>`
   --> src/external_commands.rs:234:45
    |
234 |                 if let error = TimeoutError => {
    |                 --                          ^^ expected `{`
    |                 |
    |                 this `if` expression has a condition, but no block

error: expected `{`, found `=>`
   --> src/external_commands.rs:264:45
    |
264 |                 if let error = TimeoutError => {
    |                 --                          ^^ expected `{`
    |                 |
    |                 this `if` expression has a condition, but no block

darricksee (changes requested):
Fix compile errors.

aniket2405:
hi, fixed the changes

pavlenex:
@aniket2405  do you still intend to work on this one, if so, could you rebase, if not could you close it?

PR #480, opened 2023-03-06
MG TProxy test + tests' path (draft)

lorbax:
- New TProxy test
 - New format for message generator addresses 
 - Improved documentation


message-generator-tests.sh
```diff
@@ -5,5 +5,5 @@ cd $message_generator_dir
 
 for entry in `ls $search_dir`; do
     echo $entry
-    cargo run -- ../$search_dir$entry || { echo 'mg test failed' ; exit 1; }
+    cargo run -- $search_dir$entry
```
bcral, line 8: Why was this changed?  Don't we want to know when the tests fail?

protocols/v2/noise-sv2/src/lib.rs
```diff
@@ -342,7 +342,11 @@ impl handshake::Step for Responder {
                     from_bytes(&mut in_msg);
                 match negotiation_message {
                     Ok(negotiation_message) => {
+                        println!("CCCCCCC");
```
bcral, line 345: Were these just for testing, or do we want these in for production?

lorbax: Yes the are for testing and shouldn't be there, I don't know what happened

Sjors:
You may also want to remove WIP from the 64a06b39d3fd2ee124bbcd4ee62fe1ee765fc4c3 commit message, and/or squash that commit.

PR #482, opened 2023-03-07
Extend MG coverage by using existing GH Action

bcral:
This change pulls the message-generator-check GH action into the coverage action, merging the two into just one test.  It now both runs the test and pushes coverage reports from the pool and mining-proxy to codecov.

The mining-proxy currently does not report any coverage in codecov.  I think this is because the process panics when it is closed, and will likely need the same treatment the pool and TProxy got from @djozis in #432 

This also includes a few functional additions to the MG - like waiting for each operation to close at the end of each test before continuing execution.


roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -485,7 +485,10 @@ impl UpstreamMiningNode {
                     Err(crate::lib::error::Error::UpstreamNotAvailabe(address))
                 }
             },
-            None => todo!(),
+            None => {
+                println!("No connection was found.");
```
darricksee, line 489: this should be logging with `tracing::error!`

darricksee (approved):

PR #483, opened 2023-03-08
SV1 fixes from client to real pool testing

krisztiankurucz:
Found two issues with the SV1 protocol testing the SV1 client + server example as well as client + real pools:
1. The JSON pointer (https://www.baeldung.com/json-pointer) syntax requires a `/` at the beginning of the string to correctly parse configure extensions received from the client such as version rolling mask.
2. Even though the protocol spec would suggest that `version-rolling.min-bit-count` should be returned by the server in the response to `mining.configure`, in practice with real pools it is often not included, since this should be implicit in the mask that is returned. The current SV1 protocol requires all 3 values returned, or it will fail parsing. This change makes the `version-rolling.min-bit-count` optional so that the client is able to successfully parse real pool responses.


protocols/v1/src/methods/server_to_client.rs
```diff
@@ -501,29 +501,39 @@ impl TryFrom<&Response> for Configure {
...
                 .try_into()?;
-            let version_rolling_min_bit_count: HexU32Be = version_rolling_min_bit_count
-                .unwrap()
-                .as_str()
-                .ok_or_else(|| ParsingMethodError::UnexpectedObjectParams(params.clone()))?
-                .try_into()?;
+
+            // version-rolling.min-bit-count is often not returned by stratum servers,
+            // but min-bit-count should be taken into consideration in the returned mask
+            let version_rolling_min_bit_count: HexU32Be = match version_rolling_min_bit_count {
```
darricksee, line 519: Would love it if you could add tests to cover these changes!

protocols/v1/src/methods/server_to_client.rs
```diff
@@ -501,29 +501,39 @@ impl TryFrom<&Response> for Configure {
         // Deserialize version-rolling response.
         // Composed by 3 required fields
```
darricksee, line 502: This comment will be wrong now right?

krisztiankurucz: Yup, I'll update!

protocols/v1/src/methods/server_to_client.rs
krisztiankurucz: Added `configure_response_parsing_all_fields` and `configure_response_parsing_no_vr_min_bit_count` in `protocols/v1/src/methods/server_to_client.rs` which should test these changes.

protocols/v1/src/methods/server_to_client.rs
krisztiankurucz: Updated comment

darricksee (approved):
This looks good to me - will merge once all the checks pass!
Thanks!

PR #485, opened 2023-03-09
Fixing race condition between roles after doing handshake

darricksee:
Currently there is a race condition after doing the handshake between upstream/downstream roles. Here's the situation roughly:

1. Pool sends final handshake message
2. Pool loops and waits for the message to go out
3. Proxy receives final handshake message 
4. Proxy immediately sends SetupConnection message
5. Pool receives SetupConnection message
6. Pool hasn't yet transitioned from Handshake mode to Transport mode so it thinks the incoming message is another Handshake message and errors out.

This fix removes step 2 in favor of the following flow

1. Pool encodes final handshake message
2. Pool transitions from Handshake mode to Transport mode
3. Pool sends final handshake message
4. Proxy receives final handshake message
5. Proxy immediately send SetupConnection message
6. Pool receives SetupConnection message
7. Pool is already in Transport mode so properly decrypts message and handles it.

### DIFF
--- a/roles/v2/mining-proxy/src/lib/upstream_mining.rs
+++ b/roles/v2/mining-proxy/src/lib/upstream_mining.rs
@@ -39,8 +39,8 @@ pub type EitherFrame = StandardEitherFrame<Message>;
 pub type ProxyRemoteSelector = Prs<DownstreamMiningNode>;
 
 use std::sync::atomic::AtomicBool;
-/// USED to make sure that if a future new_temnplate and a set_new_prev_hash are received together
-/// the future new_temnplate is always handled before the set new prev hash.
+/// USED to make sure that if a future new_template and a set_new_prev_hash are received together
+/// the future new_template is always handled before the set new prev hash.
 pub static IS_NEW_TEMPLATE_HANDLED: AtomicBool = AtomicBool::new(true);
 
 #[derive(Debug)]
@@ -200,7 +200,7 @@ pub struct UpstreamMiningNode {
     #[allow(dead_code)]
     tx_outs: HashMap<u64, Vec<TxOut>>,
     first_ph_received: bool,
-    // When a future job is recieved from an extended channel this is transformed to severla std
+    // When a future job is received from an extended channel this is transformed to severla std
     // job for HOM downstream. If the job is future we need to keep track of the original job id and
     // the new job ids used for the std job and also which downstream received which id. When a set
     // new prev hash is received if it refer one of these ids we use this map and build the right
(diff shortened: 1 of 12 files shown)


roles/v2/mining-proxy/src/lib/upstream_mining.rs
```diff
@@ -414,15 +414,25 @@ impl UpstreamMiningNode {
...
         sv2_frame: StdFrame,
     ) -> Result<(), crate::lib::error::Error> {
-        let (has_sv2_connetcion, mut connection) = self_mutex
-            .safe_lock(|self_| (self_.sv2_connection.is_some(), self_.connection.clone()))
+        let (has_sv2_connection, mut connection, address) = self_mutex
+            .safe_lock(|self_| {
+                (
+                    self_.sv2_connection.is_some(),
+                    self_.connection.clone(),
+                    self_.address,
```
darricksee, line 422: Just using address for logging purposes

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -60,10 +60,12 @@ impl Downstream {
         pool: Arc<Mutex<Pool>>,
         channel_factory: Arc<Mutex<PoolChannelFactory>>,
         status_tx: status::Sender,
+        address: SocketAddr,
```
darricksee, line 63: Passing address through for logging purposes.

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -233,6 +236,7 @@ impl Pool {
             config.listen_address
         );
         while let Ok((stream, _)) = listener.accept().await {
+            let address = stream.peer_addr().unwrap();
```
coleFD, line 239: is it safe to unwrap this?

utils/network-helpers/src/noise_connection_tokio.rs
```diff
@@ -31,6 +31,8 @@ impl Connection {
         Receiver<StandardEitherFrame<Message>>,
         Sender<StandardEitherFrame<Message>>,
     ) {
+        let address = stream.peer_addr().unwrap();
```
coleFD, line 34: same as above

Fi3 (approved):

roles/v2/pool/src/lib/mining_pool/mod.rs
darricksee: yeah - it looks safe. In the example code they are using `?` - I followed the call down to the bottom and it looks safe to me.

utils/network-helpers/src/noise_connection_tokio.rs
darricksee: same.

PR #488, opened 2023-03-10
Fix: Share Hash Calcs and Other Bugs

coleFD:
### 2. Handles downstream extranonce1 tracking in the Tproxy:
before this, we were creating a new "channel" for each downstream connection in the TProxy for easier tracking, but we were opening every channel with channel_id 1, which caused each share submission to check shares with the wrong extranonce. Now we are storing a separate channel_id for each connection in the `Downstream` struct, and a new wrapper type `SubmitSharesWithId` was added that gets sent to the `Bridge` so it can properly look up the correct channel.
### 3. Segwit support for miners without Segwit support:
According to BIP-141 the txid is the same for segwit and legacy transactions because the txid remains as the SHA256d hash of the serialized legacy transaction (excludes segwit the specific fields `marker`, `flag` and `witness` data). The shares created by a cpuminer would not match that of the miner because we were sending segwit data in the `coinbase_tx_prefix` and `coinbase_tx_suffix`, and the cpuminer would just hash what it was given without checks. To fix this, a helper function was added `extended_job_to_non_segwit()` that removes the segwit data in a `NewExtendedMiningJob`. Currently this conversion is only done in the TProxy but if this data is sent to a miner with segwit support, it should still produce the same hashes. Note: The pool still submits the coinbase tx with segwit data included to the Template Provider regardless if the miner is hashing the stripped coinbase or not.

Note: if testing with local instances of the roles, you will notice that after a few jobs arrive, the shares start to fail checks. This is because job management is out of whack and should be fixed in another issue

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
+++ b/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
@@ -779,9 +779,6 @@ impl ChannelFactory {
         println!("COINBASE SUFFIX: {:?}\n", &coinbase_tx_suffix);
         println!("EXTRANONCE: {:?}\n", &extranonce);
 
-        
-        
-
         if hash <= bitcoin_target {
             let coinbase = [coinbase_tx_prefix, &extranonce[..], coinbase_tx_suffix]
                 .concat()
--- a/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
+++ b/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
@@ -698,27 +698,28 @@ impl ChannelFactory {
                 upstream_target, ..
             } => upstream_target.clone(),
         };
-        let last_job_id = self
+        let _last_job_id = self
             .last_valid_job
             .as_ref()
             .ok_or(Error::ShareDoNotMatchAnyJob)?
             .0
             .job_id;
-        if m.get_job_id() < last_job_id {
-            let error = SubmitSharesError {
-                channel_id: m.get_channel_id(),
-                sequence_number: m.get_sequence_number(),
-                // Infallible unwrap we already know the len of the error code (is a
-                // static string)
-                error_code: SubmitSharesError::stale_share_error_code()
-                    .to_string()
-                    .try_into()
-                    .unwrap(),
-            };
-            return Ok(OnNewShare::SendErrorDownstream(error));
-        } else if m.get_job_id() > last_job_id {
-            return Err(Error::JobNotUpdated(m.get_job_id(), last_job_id));
-        }
+        // *** TODO: uncomment below after mining proxy job management is fixed
+        // if m.get_job_id() < last_job_id {
+        //     let error = SubmitSharesError {
+        //         channel_id: m.get_channel_id(),
+        //         sequence_number: m.get_sequence_number(),
+        //         // Infallible unwrap we already know the len of the error code (is a
+        //         // static string)
+        //         error_code: SubmitSharesError::stale_share_error_code()
+        //             .to_string()
+        //             .try_into()
+        //             .unwrap(),
+        //     };
+        //     return Ok(OnNewShare::SendErrorDownstream(error));
+        // } else if m.get_job_id() > last_job_id {
+        //     return Err(Error::JobNotUpdated(m.get_job_id(), last_job_id));
+        // }
         let (downstream_target, extranonce) = self
             .get_channel_specific_mining_info(&m)
             .ok_or(Error::ShareDoNotMatchAnyChannel)?;
--- a/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
+++ b/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
@@ -246,7 +246,7 @@ impl ChannelFactory {
                 .safe_lock(|ids| ids.new_channel_id(extended_channels_group))
                 .unwrap();
             self.channel_to_group_id.insert(channel_id, 0);
-            let target = crate::utils::hash_rate_to_target_le(hash_rate, self.share_per_min);
+            let target = crate::utils::hash_rate_to_target(hash_rate, self.share_per_min);
             let extranonce = self
                 .extranonces
                 .next_extended(max_extranonce_size as usize)?;
@@ -291,7 +291,7 @@ impl ChannelFactory {
         let hom_group_id = 0;
         let mut result = vec![];
         let channel_id = id;
-        let target = crate::utils::hash_rate_to_target_le(downstream_hash_rate, self.share_per_min);
+        let target = crate::utils::hash_rate_to_target(downstream_hash_rate, self.share_per_min);
         let extranonce = self
             .extranonces
             .next_standard()
@@ -334,7 +334,7 @@ impl ChannelFactory {
             .safe_lock(|ids| ids.new_channel_id(group_id))
             .unwrap();
         let complete_id = GroupId::into_complete_id(group_id, channel_id);
-        let target = crate::utils::hash_rate_to_target_le(downstream_hash_rate, self.share_per_min);
+        let target = crate::utils::hash_rate_to_target(downstream_hash_rate, self.share_per_min);
         let extranonce = self
             .extranonces
             .next_standard()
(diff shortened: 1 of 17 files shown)


protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -172,7 +172,7 @@ fn reduce_path<T: AsRef<[u8]>>(coinbase_id: [u8; 32], path: &[T]) -> [u8; 32] {
 /// [3] https://en.wikipedia.org/wiki/Negative_hypergeometric_distribution
 /// bdiff: 0x00000000ffff0000000000000000000000000000000000000000000000000000
 /// https://en.bitcoin.it/wiki/Difficulty#How_soon_might_I_expect_to_generate_a_block.3F
-pub fn hash_rate_to_target(h: f32, share_per_min: f32) -> U256<'static> {
+fn _hash_rate_to_target(h: f32, share_per_min: f32) -> [u8; 32] {
```
Fi3, line 175: not sure that we need `hash_rate_to_target_be`.
Here we should only have the function `hash_rate_to_target` and it should return an `U256`. We should also remove the possibility to do `U256` from `[u8;32]` and `U256` try from `Vec<u8>` 

The issue here is that we treat `U256` like a generic byte array but in the spec is defined as a number.  So the only way to construct an `U256` should be using the methods `from_le_bytes(b: [u8;32]) -> Self` and `from_be_bytes(b: [u8;32]) -> Self`

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -285,11 +285,124 @@ fn coinbase(
     }
 }
 
+/// Helper type to strip a segwit data from the coinbase_tx_prefix and coinbase_tx_suffix
+/// to ensure miners are hashing with the correct coinbase
+pub fn extended_job_to_non_segwit(
+    job: NewExtendedMiningJob<'static>,
```
Fi3, line 291: this is an only Sv1 feature, Sv2 software should not depend on upstream removing segwit data from the tx. Also it return a `NewExtendedMiningJob` that do not complain with the spec. If possible I would put this function somewhere else maybe in `utils/something` or in `protocols/sv1`, and I would also modify the signature so that it do not return a `NewExtendedMiningJob` but something else maybe just `(prefix,suffix)` ?

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -893,7 +917,8 @@ impl PoolChannelFactory {
         &mut self,
         m: &SetNewPrevHashFromTp<'static>,
     ) -> Result<u32, Error> {
-        let job_id = self.job_creator.on_new_prev_hash(m).unwrap_or(0);
+        // since we have staged phash 0 is no longer reserved and 1 is the default
```
Fi3, line 920: What do you mean?

coleFD: I noticed the first SNPH getting sent out was with job_id=0 whereas the first NEMJ getting sent was with job_id=1 so the tproxy was having issues (the first job getting sent to miners was with job_id=0 and we were checking shares against job_id=1. iirc, the `StagedPrevHash` was added so we dont have to have 0 reserved anymore for when a prev hash arrives and there is no template.

protocols/v2/roles-logic-sv2/src/utils.rs
coleFD: Okay, I can change this. I added this because in the `difficulty_from_target` function in the Tproxy is based on BE targets and in the tproxy difficulty management PR I just kept it that way. Will change back and handle it in the tproxy in the the diff management PR

protocols/v2/roles-logic-sv2/src/job_creator.rs
coleFD: Okay, I can do this, but first I would like to confirm that  all SV2 miner firmware is segwit compatible. If not we will have issues later on, and if that is the case, this function could be used in the pool before broadcasting to ensure support for either cases. Assuming you mean its not compatible with the spec since the job is getting altered after the pool sends it out?

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
coleFD: it may be better to get the JN in Tproxy PR merged first so I can rebase and confirm

protocols/v2/roles-logic-sv2/src/job_creator.rs
Fi3: ok maybe make more sense to apply your patch  everywhere both on Sv1 and Sv2. In the extended job we send the tx_outputs so no sense to send also the witness.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -724,6 +747,7 @@ impl ChannelFactory {
         };
         let hash_ = header.block_hash();
         let hash = hash_.as_hash().into_inner();
+        println!("\nSHARE HASH: {:?}", &hash);
```
darricksee, line 750: Use debug!() or trace!()

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -1068,7 +1093,8 @@ impl ProxyExtendedChannelFactory {
         m: &SetNewPrevHashFromTp<'static>,
     ) -> Result<Option<PartialSetCustomMiningJob>, Error> {
         if let Some(job_creator) = self.job_creator.as_mut() {
-            let job_id = job_creator.on_new_prev_hash(m).unwrap_or(0);
+            // since we have staged phash 0 is no longer reserved and 1 is the default
+            let job_id = job_creator.on_new_prev_hash(m).unwrap_or(1);
```
darricksee, line 1097: While we are in this file can you fix the below warnings (line 1422,1423) - i see them on so many PRs.

roles/translator/src/downstream_sv1/downstream.rs
```diff
@@ -447,7 +453,11 @@ impl IsServer<'static> for Downstream {
             let mut downstream_part: Vec<u8> = request.extra_nonce2.clone().into();
             tproxy_part.append(&mut downstream_part);
 
-            let to_send = (request.clone(), tproxy_part);
+            let to_send = SubmitShareWithChannelId {
+                channel_id: self.connection_id,
+                share: request.clone(),
+                extranonce: tproxy_part,
+            };
```
darricksee, line 460: While we are in this file can you fix these warnings (line 540) - i see them on so many PRs.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
coleFD, line 750: will remove this

darricksee (changes requested):

roles/translator/src/downstream_sv1/diff_management.rs
```diff
@@ -227,9 +229,9 @@ mod test {
     #[tokio::test]
     async fn test_diff_management() {
         let downstream_conf = DownstreamDifficultyConfig {
-            min_individual_miner_hashrate: 0.0,  // updated below
-            miner_num_submits_before_update: 15, // update after 5 submits
-            shares_per_minute: 100.0,            // 10 shares per minute
+            min_individual_miner_hashrate: 0.0,   // updated below
+            miner_num_submits_before_update: 150, // update after 5 submits
+            shares_per_minute: 1000.0,            // 10 shares per minute
```
darricksee, line 234: These comments look wrong now (granted they looked wrong before too)

darricksee (approved):

Sjors:
Testing this with my CPU miner against Braiinspool I very quickly get `Submit share error Ok("difficulty-too-low")`.

coleFD:
> Testing this with my CPU miner against Braiinspool I very quickly get `Submit share error Ok("difficulty-too-low")`.

I just noticed this as well,  but it's not a major issue. So to get that error the share has to fail the upstream target and downstream target check. Because the we are solely handling difficultly management in the downstream component and are not currently updating with the Bridge component (which handles the check), the downstream target is just set as the upstream difficulty and never updates after that. If you change your downstream difficulty config to match that of what Braiins is using (not sure what that is), it wont happen as ofter. For instance, if braiins is expecting 1 share/min at `h` hashrate. you can set your downstream difficulty config to 1 share/min and `x` hash rate and it will start checking against the miner's difficulty setting. 

This would just be a temporary fix so you dont see the error, but I will add an issue for it so we can see a clean `Share meets downstream target` log instead of the error

Sjors:
Thanks for clarifying.

PR #491, opened 2023-03-13
Docs: Channel Factory

coleFD:


protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -252,6 +260,7 @@ impl ChannelFactory {
             };
             self.extended_channels.insert(channel_id, success.clone());
             let mut result = vec![Mining::OpenExtendedMiningChannelSuccess(success)];
+            // should SNPH be pushed after NEMJ?
```
darricksee, line 263: This does look like a bug as the NEMJ should always be first... Can you create a bug to investigate this line?

coleFD: @fi3 fixed this in a his local fork of the share hash fix PR. I can remove this

darricksee (approved):

Issue #429, opened 2023-02-01
There is a DOS vulnerability in the downstream reader


darricksee:
During this bug fix user djozis found a vulnerability - when reading from the downstream connection we basically try to read everything that is put on the pipe. We should read in chunks and process and if in the processing we discover that it's garbage we should drop the connection.

Context:
https://github.com/stratum-mining/stratum/pull/426#discussion_r1092283449

Fi3:
This is not done. #504 solve it but is very old, maybe make sense to close it and open a new PR

PR #497, opened 2023-03-18
Limit acceptable line length to avoid DoS

douglaz:
Fixes https://github.com/stratum-mining/stratum/issues/429

Open questions:
1) What `MAX_LINE_LENGTH` should be?
2) Should we use `LimitedLines` in other places? Will probably require it to move to another crate (where?)


douglaz:
> I took a very brief look to see if there's something standard in Tokio we can use for this. Should we use LinesCodec from Tokio instead?: https://docs.rs/tokio-util/latest/tokio_util/codec/struct.LinesCodec.html#method.new_with_max_length
Yeah, I didn't see that. It's certainly seems a good option.

> 
> Something like:
> 
> ```
>          let mut reader = FramedRead::new(socket_reader, LinesCodec::new_with_max_length(MAX_LENGTH));
>          while let Some(frame) = reader.next().await {
>              match frame {
>                  Ok(line) => { ...
> ```
> 
> I haven't tested the above for correctness, just wanted to share my finding that perhaps we needn't roll our own in this case, but didn't take it further than that.

It won't work without using/creating an adaptor or changing the stream/channels from `async-std` to `tokio`

douglaz:
> It won't work without using/creating an adaptor or changing the stream/channels from `async-std` to `tokio`
Created another PR using an adaptor: https://github.com/stratum-mining/stratum/pull/504

coleFD:
> > It won't work without using/creating an adaptor or changing the stream/channels from `async-std` to `tokio`
> 
> Created another PR using an adaptor: #504

can we close this PR @darricksee ?

PR #501, opened 2023-03-20
S9 fixes

coleFD:
The main issue why CGminer miners are not submitting shares is because CGminer does not support an extranonce2 space over 8 bytes, but it only enforces this limit just before submitting shares, so the miner will keep working. 

Other issues found:
1. we were not calculating the version correctly for cases when miners are version rolling. see (https://en.bitcoin.it/wiki/BIP_0310#Changes_in_request_.22mining.submit.22)
2. we did not support cases when the miner extranonce2 space was any value other than 14 bytes and the tproxy extranonce1 space was 2 bytes. This is now supported for any value as long as the extranonce2 space specified in the proxy config is supported by the miner, and this config value is less than max supported value of the pool extranonce2 size since we need a total extranonce2 space of  `proxy_extranonce1 + miner_extranonce2`


pavlenex:
Holly smokes! I am excited to test this out tomorrow, thanks for tracking all of these down @coleFD 🔥

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -654,4 +654,39 @@ pub mod tests {
...
+            26, 0, 0, 0, 0, 0, 0, 0, 0, 38, 106, 36, 170, 33, 169, 237, 103, 66, 68, 105, 2, 55,
+            65, 241, 216, 46, 82, 223, 150, 0, 97, 103, 2, 82, 186, 233, 145, 90, 210, 231, 35,
+            100, 107, 52, 171, 233, 50, 200, 0, 0, 0, 0,
+        ];
+        let extranonce = vec![0_u8; 15]; // braiins pool requires 15 bytes for extranonce
+        encoded.extend_from_slice(coinbase_prefix);
+        let mut encoded_clone = encoded.clone();
+        encoded_clone.extend_from_slice(&extranonce[..]);
+        encoded_clone.extend_from_slice(coinbase_suffix);
+        // let mut i = 1;
```
darricksee, line 681: Do we need to keep this? If so add comment for why.

roles/translator/src/proxy/bridge.rs
```diff
@@ -411,14 +408,27 @@ impl Bridge {
...
                     extranonce,
+                    extranonce2_len,
+                    version_rolling_mask,
                 } = handle_result!(tx_status, rx_sv1_submit.clone().recv().await);
                 let channel_sequence_id = self_
                     .safe_lock(|s| s.channel_sequence_id.next())
                     .map_err(|_| PoisonLock);
                 let channel_sequence_id = handle_result!(tx_status, channel_sequence_id) - 1;
+
+                // let channel_extranonce1_len = self_
```
darricksee, line 419: remove if not needed.

roles/translator/src/proxy/bridge.rs
```diff
@@ -497,16 +510,30 @@ impl Bridge {
...
-            None => self
-                .channel_factory
-                .last_valid_job_version()
-                .ok_or(Error::RolesSv2Logic(RolesLogicError::NoValidJob))?,
+        let last_version = self
+            .channel_factory
+            .last_valid_job_version()
+            .ok_or(Error::RolesSv2Logic(RolesLogicError::NoValidJob))?;
+        let version = match (sv1_submit.version_bits, version_rolling_mask) {
+            (Some(vb), Some(mask)) => (last_version & !mask.0) | (vb.0 & mask.0),
```
darricksee, line 521: It would be good to add this link as a comment as to why we are doing this masking here - https://github.com/slushpool/stratumprotocol/blob/master/stratum-extensions.mediawiki#changes-in-request-miningsubmit

PR #502, opened 2023-03-20
Fix: diff-too-low error in TProxy

coleFD:
This PR converts the SubmitShare channel to allow for general messaging from the Downstreams to the Bridge. With that we can now notify the Bridge of hashrate updates from downstream to update each downstream's targets when changed


coleFD:
while rebasing the diff-too-low PR, i found out the pooler/cpuminer submits its extranonce backwards, whereas CGminer doesnt, so when the cpuminer does decide to roll its extranonce (it does so infrequently), it will be incompatible with our tproxy and we will still get the "diff-too-low" error, but in this case, the error is semi-valid. 

if we want to allow support for the cpuminer we would need to set it in a config, but i dont think this is crucial. better to focus on supporting the actual asics. 

Note CGminer is compatible as is, this is just regarding incompatibility with the pooler/cpuminer

roles/translator/src/main.rs
```diff
@@ -54,7 +54,7 @@ async fn main() {
     // `tx_sv1_submit` sender is used by `Downstream` to send a `mining.submit` message to
```
darricksee, line 54: These comments need to be updated.

darricksee (approved):

PR #503, opened 2023-03-21
Remove future field in job messages and replace it with min_ntime

Fi3:


protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
```diff
@@ -321,3 +343,75 @@ impl<'a, const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const
...
 }
+
+/// The liftime is here only for type compatibility with serde-sv2
+#[repr(C)]
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Sv2Option<'a, T>(pub Vec<T>, PhantomData<&'a T>);
+
+// TODO add test for that and implement it also with serde!!!!
+impl<'a, const SIZE: usize> Sv2Option<'a, super::inner::Inner<'a, true, SIZE, 0, 0>> {
+    pub fn to_option(&self) -> Option<Vec<u8>> {
```
coleFD, line 354: is there a reason not to consume self here?

protocols/v2/const-sv2/src/lib.rs
```diff
@@ -39,7 +39,8 @@ pub const MESSAGE_TYPE_SET_COINBASE: u8 = 0x50;
 // MINING PROTOCOL MESSAGES TYPES
 pub const MESSAGE_TYPE_CLOSE_CHANNEL: u8 = 0x18;
 pub const MESSAGE_TYPE_NEW_EXTENDED_MINING_JOB: u8 = 0x1f;
-pub const MESSAGE_TYPE_NEW_MINING_JOB: u8 = 0x1e;
+pub const _MESSAGE_TYPE_NEW_MINING_JOB_OLD: u8 = 0x1e;
```
coleFD, line 42: any reason to keep the old constant?

coleFD:
i ran the test diff management test 1000 iterations last night and it never failed, it seems like some cpu may have gotten bottlenecked during the run. it should pass if run again

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -1235,7 +1232,7 @@ impl ProxyExtendedChannelFactory {
                     Some(custom_mining_job),
                 ));
             }
-            if new_job.future_job {
+            if new_job.is_future() {
```
coleFD, line 1235: converting to `else if` would be a little more efficient

coleFD (approved):

coleFD (changes requested):
approved after the minor changes recommended get resolved

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
Fi3: Yes. The `to` prefix in a function eg `to_option` usually in rust mean that we take a reference. Here we take a reference cause is more likely  to have a reference when we need to use it (we usually have views on byte arrays that rapresent the received message).

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
Fi3: changed first condition exclude the second one so we can safely use else if

protocols/v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/seq_inner.rs
coleFD: its not letting me resolve so I will give a thumbs up for resolved comments

coleFD (approved):

PR #506, opened 2023-03-26
Pool & Translator Proxy - conf refactoring 

GitGab19:
As discussed in the last dev calls, tracked config files for both SRI pool and tProxy can generate some issues during testing phase. So, also to accomplish better future externals who will test our MVP2 after its release, following the advice of @coleFD on Discord chat, I propose this solution to stop tracking custom configs. 
For context, discussion about roles pre-configurations can be found here:
https://discord.com/channels/950687892169195530/1076193338677538857/1088091760514826340   

Basically (using the same words of @coleFD) the structure is:
1. Just have a singe [role]-config-example.toml in each roles directory that has enough descriptions of each field and shows possible values
2. Dont allow changes other than comments and documentation to this file (live values stay the same)
3. in the README.md for each role we specify that you need to copy this in to the /conf directory where /conf/[role]-config.toml runs by default and you as an individual can make as many configurations as you want but they dont get tracked on the main repo. So you can still add a JN-hosted.toml  as well and run with with the -- -c [path to conf]
4. we gitignore all files in the /conf directories so they remain an empty folder on the SRI repo

Looking for feedback and opinions about this proposal! 🙏

PS: if acked, I'll begin the refactoring of the Getting Started page of website, which will exploit this new roles structure


roles/v2/pool/README.md
```diff
@@ -0,0 +1,41 @@
...
+|  | SV1 Mining Device | <-> | Translator Proxy | <------> | SV2 Pool Server |  |
+|  +-------------------+     +------------------+   |  |   +-----------------+  |
+|                                                   |  |                        |
++---------------------------------------------------+  +------------------------+
+
+```
+
+## Setup
+### Configuration File
+The `pool-config-example.toml` is is a configuration example which can be modified by the party that is running the SV2 Pool (most
```
coleFD, line 22: copy/paste into the `conf/` directory as well for custom config

roles/translator/tproxy-config-example.toml
```diff
@@ -23,10 +23,16 @@ min_supported_version = 2
 min_extranonce2_size = 8
 coinbase_reward_sat = 5_000_000_000
 
-# optional jn config, if set the tproxy start on JN mode
-# [jn_config]
-# jn_address = "127.0.0.1:34264"
-# tp_address = "127.0.0.1:8442"
+# JN config (optional), if set the tproxy starts on JN mode
```
darricksee, line 26: IMO jn_config should be disabled by default.

roles/v2/pool/pool-config-example.toml
```diff
@@ -1,11 +1,23 @@
...
 # list of compressed or uncompressed pubkeys for coinbase payout (only supports 1 item in the array at this point)
 coinbase_outputs = [
     "04466d7fcae563e5cb09a0d1870bb580344804617879a14949cf22285f1bae3f276728176c3c6431f8eeda4538dc37c865e2784f3a9e77d044f33e407797e1278a",
 ]
+
+# SRI Pool JN config
+# local TP (must run a TP locally for this to run)
+tp_address = "127.0.0.1:8442"
+# hosted testnet TP 
+# tp_address = "89.116.25.191:8442"
```
darricksee, line 16: My vote would be that this should be the default. Then we can still point to this config in the easy startup guide docs.

roles/v2/pool/pool-config-example.toml
```diff
@@ -1,11 +1,23 @@
...
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
+listen_address = "0.0.0.0:34254"
 # list of compressed or uncompressed pubkeys for coinbase payout (only supports 1 item in the array at this point)
 coinbase_outputs = [
     "04466d7fcae563e5cb09a0d1870bb580344804617879a14949cf22285f1bae3f276728176c3c6431f8eeda4538dc37c865e2784f3a9e77d044f33e407797e1278a",
 ]
+
+# SRI Pool JN config
+# local TP (must run a TP locally for this to run)
```
darricksee, line 13: The TP in this case doesn't relate to JN - the pool connects directly to the TP and opens up a listen port for the JN to connect to. So you can move the `SRI Pool JN config` comment down to the listen_jn_address as that is the only JN pool config field.

roles/v2/pool/README.md
```diff
@@ -0,0 +1,41 @@
...
+typically the pool service provider) to address the most preferred customization.
+
+The configuration file contains the following information:
+
+1. The SRI Pool information which includes the SRI Pool authority public key 
+   (`authority_pubkey`), the SRI Pool authority secret key (`authority_secret_key`), along with its certificate validity (`cert_validity_sec`). In addiction to this, it contains also the address which it use to listen new connections from downstream roles (`listen_address`) and the list of uncompressed pubkeys for coinbase payout (`coinbase_outputs`).
```
darricksee, line 28: addiction -> addition

coleFD:
I will make a PR to fix the test failure

roles/translator/tproxy-config-example.toml
GitGab19: Done.
Even if I think this should be prioritize, since it's the biggest enhancement of our MVP2. But we'll try to emphasize it in docs 👍

roles/v2/pool/pool-config-example.toml
GitGab19: Agree on this.
Rn it's the most prioritize in docs since it's well explained how to run a TP locally (which works on linux only).
Btw I also updated the comment like the one you suggested for the tProxy config above.

PR #510, opened 2023-03-28
Add error handling around panic in tproxy

darricksee:
1.) Added error handling around a panic
2.) Added a MG test to test error
3.) Added the ability to mock mining protocol messages in MG.
4.) Moved all the tproxy init functionality in a separate task so that signals and statuses can be handled during the init phase of startup.


roles/v2/pool/src/error.rs
```diff
@@ -7,6 +7,7 @@ use std::{
 #[derive(std::fmt::Debug)]
 pub enum PoolError {
     Io(std::io::Error),
+    InvalidExtranonce(String),
```
coleFD, line 10: is this variant needed somewhere? here (https://github.com/stratum-mining/stratum/pull/513/files#diff-a2101e207a5d63e663b5e05e4fb90216cc3973ef8027c011c0a58417464cb73b) I added an error code relating to this to return in an `OpenMiningChannelError` which may cover the case you had in mind

coleFD (approved):

test/message-generator/test/translation-proxy.json
```diff
@@ -53,6 +53,7 @@
        {
             "command": "cargo",
             "args": [
+                        "llvm-cov",
```
bcral, line 56: This should also include a "--no-report" flag

Fi3:
LGTM

Fi3 (approved):

PR #512, opened 2023-03-28
Update and extend MG coverage tests

bcral:
MG coverage test structure changed to reduce errors during file upload action.
Test case added with bad share submitted to test ShareSubmitError response.


test/message-generator/test/interop-proxy-with-multi-ups-extended-jn.json
```diff
@@ -68,6 +68,11 @@
                 }
             }
         },
+        {
+            "command": "sleep",
+            "args": ["1000"],
```
darricksee, line 73: Why do we need to sleep for 1000s?

test/message-generator/test/interop-proxy-with-multi-ups-extended.json
```diff
@@ -68,6 +68,11 @@
                 }
             }
         },
+        {
+            "command": "sleep",
+            "args": ["1000"],
```
darricksee, line 73: same?

test/message-generator/test/interop-proxy-with-multi-ups-extended-jn.json
bcral: Short answer: I'm not sure.
I believe this was included on main when I started work, was since deleted, and when I rebased, it was just added in from what I had locally.  I can delete them.

bcral:
Tests updated to reflect SRI spec message orders - specifically the order of New Mining Job -> Set New Previous Hash

darricksee (approved):

PR #513, opened 2023-03-28
(Draft) Feat: Client Auth Trait Func in Upstream Nodes

coleFD:
Added trait function `ParseDownstreamMiningMessages::downstream_is_authorized` for Upstream nodes to implement client authorization services if needed

PR #514, opened 2023-03-29
Feat: Client Auth Trait Func in Upstream Nodes 

coleFD:
extension of https://github.com/stratum-mining/stratum/pull/513

This PR adds a commit to the linked PR to ensure dropping downstream connections in the Pool when a Downstream gets dropped. Since the Pool keeps track of Downstream in order to relay new SNPH and Templates, we need to drop the Downstream from the downstream mapping state in the Pool. To do this, this PR notifies the main thread through the Status channel and calls a new function `Pool::drop_connection` to do so.

An alternative method would be to remove the downstream mapping and utilize a broadcast channel to notify downstreams. This method may be more efficient since we would not need to loop through downstreams to broadcast similar to the Bridge -> Downstream relationship in the TProxy, but it would require a larger refactor. If we want to go this route, we can close this PR and I will implement in the linked draft PR


test/message-generator/test/pool-sri-test-reject-auth.json
```diff
@@ -0,0 +1,141 @@
...
+            "message": {
+                "type": "OpenStandardMiningChannel",
+                "request_id": 89,
+                "user_identity": "",
+                "nominal_hash_rate": 10,
+                "max_target": [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]
+            },
+            "id": "open_standard_mining_channel"
+        },
+        {
```
bcral, line 22: This block can be removed, since it isn't used in this test.

Issue #520, opened 2023-04-03
Deadlock in SRI when connecting to TP


ccdle12:
## Description

When reviewing and refactoring the TP code ready for Bitcoin Core PR concept ACK, I noticed that the TP sends the incorrect order of messages when receiving a COINBASE_OUTPUT_DATA_SIZE message.

Current order:

```
Client  -> COINBASE_OUTPUT_DATA_SIZE
Server  -> SET_NEW_PREV_HASH
Server  -> NEW_TEMPLATE
```

Expected order:

```
Client  -> COINBASE_OUTPUT_DATA_SIZE
Server  -> NEW_TEMPLATE
Server  -> SET_NEW_PREV_HASH
```

The correct order is referenced here in the specs: https://github.com/stratum-mining/sv2-spec/blob/main/07-Template-Distribution-Protocol.md

This led to identifying a deadlock in the SRI code when sending the *expected order* of messages from the TP.

### Deadlock

Currently, even using the code in the TP that has the messages out of order, it still leads to a Deadlock in the SRI code that requires a new block to be generated in regtest or testnet after any proxy connects to the mining pool.

Below is an overview on the Deadlock:

1. When a TemplateReceiver receives a `NewTemplate` message, it enters [this conditional path in TemplateRx::start_templates().](https://github.com/stratum-mining/stratum/blob/main/roles/v2/mining-proxy/src/lib/template_receiver/mod.rs#L106) It latches [IS_NEW_TEMPLATE_HANDLED (AtomicBool)](https://github.com/stratum-mining/stratum/blob/main/roles/v2/mining-proxy/src/lib/upstream_mining.rs#L44) to false. The `NewTemplate` message is sent across a channel using [self.send_new_tp.](https://github.com/stratum-mining/stratum/blob/main/roles/v2/mining-proxy/src/lib/template_receiver/mod.rs#L109)

[...]

PR #522, opened 2023-04-04
Updated SRI codebase to work with fixed TP, Fix minor issues

Fi3:
TP has been fixed so that SNPH and NEMJ messages order is no more inverted, this updated the SRI codebase to work with the new TP. Also fix JN issues in the tproxy and the minig-prox.


ccdle12:
Thanks, I tested this branch and these changes seem to have fixed the deadlock issue reported in #520

ccdle12 (approved):

PR #539, opened 2023-04-13
Whats miner fix

darricksee:


protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -746,23 +747,26 @@ impl ChannelFactory {
...
+            debug!("Bitcoin target: {:?}", bitcoin_target);
             let upstream_target: binary_sv2::U256 = upstream_target.clone().try_into().unwrap();
             let mut upstream_target = upstream_target.to_vec();
             upstream_target.reverse();
             debug!("Upstream target: {:?}", upstream_target.to_vec().to_hex());
-            let mut hash = hash;
-            hash.reverse();
-            debug!("Hash: {:?}", hash.to_vec().to_hex());
+            let mut print_hash = hash.clone();
+            print_hash.reverse();
```
darricksee, line 756: Seems like we shouldn't reverse the hash before it's being used to check the `bitcoin_target` right?

Fi3: [u8;32] is copy so it reverse just a copy

darricksee: yep - i figured that out :) - clippy warned about it - i changed it back.

Fi3 (approved):

PR #541, opened 2023-04-14
Update .codecov.yaml

tusharv01:
This script will attempt to upload the coverage data using the Codecov Bash uploader, and if it fails (indicated by a non-zero exit status), it will retry up to 5 times with a 10 second delay between each attempt. If the upload is successful at any point during the retry loop, the loop will be terminated early.

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -2,3 +2,20 @@ coverage:
   status:
     project: off
     patch:   off
+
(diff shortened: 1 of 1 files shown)


Fi3 (approved):

Fi3:
@tusharv01 thanks!

PR #544, opened 2023-04-19
Auth trait

darricksee:
extension of https://github.com/stratum-mining/stratum/pull/513

This PR adds a commit to the linked PR to ensure dropping downstream connections in the Pool when a Downstream gets dropped. Since the Pool keeps track of Downstream in order to relay new SNPH and Templates, we need to drop the Downstream from the downstream mapping state in the Pool. To do this, this PR notifies the main thread through the Status channel and calls a new function Pool::drop_connection to do so.

An alternative method would be to remove the downstream mapping and utilize a broadcast channel to notify downstreams. This method may be more efficient since we would not need to loop through downstreams to broadcast similar to the Bridge -> Downstream relationship in the TProxy, but it would require a larger refactor. If we want to go this route, we can close this PR and I will implement in the linked draft PR


protocols/v2/roles-logic-sv2/src/handlers/mining.rs
```diff
@@ -189,6 +205,14 @@ pub trait ParseDownstreamMiningMessages<
 
     fn is_work_selection_enabled(&self) -> bool;
 
+    /// returns None if the user is authorized and Open
+    fn downstream_is_authorized(
```
Fi3, line 209: change function's name in `is_downstream_authorized`

roles/translator/src/upstream_sv2/upstream.rs
```diff
@@ -717,25 +724,29 @@ impl ParseUpstreamMiningMessages<Downstream, NullDownstreamMiningSelector, NoRou
     /// Handles the SV2 `OpenExtendedMiningChannelError` message (TODO).
     fn handle_open_mining_channel_error(
         &mut self,
-        _: roles_logic_sv2::mining_sv2::OpenMiningChannelError,
+        m: roles_logic_sv2::mining_sv2::OpenMiningChannelError,
     ) -> Result<roles_logic_sv2::handlers::mining::SendTo<Downstream>, RolesLogicError> {
-        todo!()
+        Ok(SendTo::None(Some(Mining::OpenMiningChannelError(
```
Fi3, line 729: `SendTo::None(Some(Mining::OpenMiningChannelError)` is not matched in `match next_message_to_send` this lead to a panic

roles/v2/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -203,157 +210,4 @@ impl ParseDownstreamMiningMessages<(), NullDownstreamMiningSelector, NoRouting>
         };
         Ok(SendTo::Respond(Mining::SetCustomMiningJobSuccess(m)))
     }
-
```
Fi3, line 206: Nice! Don't know why I left it here but it redundant. Only thing the we should add to the function defined in the trait is the below:
```rust
                    (SupportedChannelTypes::Group, true) => self_mutex
                        .safe_lock(|self_| self_.handle_set_custom_mining_job(m))
                        .map_err(|e| roles_logic_sv2::Error::PoisonLock(e.to_string()))?,
``` 
This should be added in the trait definition in the default `handle_message_mining` where we match `Ok(Mining::SetCustomMiningJob(m))`

Fi3: No looked at it better and version of this function that we have in the trait is the right one so no need to add the above. That's cause we can not have a supported channel type that is Group only with work selection enabled is just not possible under Sv2

roles/v2/pool/src/lib/mining_pool/mod.rs
```diff
@@ -523,6 +537,11 @@ impl Pool {
                 error!("Downstream shutdown and Status Channel dropped");
             }
         });
+        cloned3
+    }
+
+    pub fn drop_downstream(&mut self, downstream_id: u32) {
```
Fi3, line 543: Why use a function if the only thing that we do is calling remove? Also Arc<Downstream> can be cloned are we sure that removing it from the self.downstream HashMap is sufficient for actually drop every reference to downstream? With the code as it today seems so but maybe in the future will not be anymore the case, maybe (if possible) we should add a test that make sure that drop downstream remove every reverence to downstream.

Possibe thing that could work just thinking out loud:
```rust
    pub fn drop_downstream(&mut self, downstream_id: u32) {
        let d = self.downstreams.remove(&downstream_id).unwrap();
        // we could fail early in debug enviornment this do not assure anything btw
        debug_assert!(Arc::strong_count(&d) == 1);
        // or we could drop the value inside the safe lock not sure what will happen
        d.safe_lock(|d| drop(d)).unwrap();
    }
```

roles/translator/src/upstream_sv2/upstream.rs
darricksee: Am I missing something? That case was handled in the line changes above at 449.

roles/v2/pool/src/lib/mining_pool/mod.rs
darricksee: The drop_downstream is used to access the downstreams hashmap which isn't public. This makes sense to me. Here is the bigger issue that I found/fixed:

If a downstream gets dropped it gets removed from the downstreams list... but because we clone the downstreams there is a race condition like this:

 

1.) New template comes in and we clone the downstreams with 100 downstreams

2.) downstream #6 disconnects and is removed from downstreams

3.) as we are iterating the clone of downstreams from step 1 downstream #6 still exists in the clone

4.) we hit #6 and try to send the template to them - this fails as they aren't actually there anymore

5.) the handle_result (i think) is a break on this error and so the iteration doesn't complete and downstreams 7-100 never get their new templates

Fi3: Don't you think that we could have memory leaks? What worry me is that in some future version of the code someone call `drop_dowstream` and think to drop it but maybe some clone of the `Arc` is still around. Btw I'm ok of merging it as it is maybe just change the function name or put the `debug_assert`. What do you think?

roles/translator/src/upstream_sv2/upstream.rs
```diff
@@ -717,25 +724,29 @@ impl ParseUpstreamMiningMessages<Downstream, NullDownstreamMiningSelector, NoRou
...
     }
 
     /// Handles the SV2 `CloseChannel` message (TODO).
     fn handle_close_channel(
         &mut self,
-        _m: roles_logic_sv2::mining_sv2::CloseChannel,
+        m: roles_logic_sv2::mining_sv2::CloseChannel,
     ) -> Result<roles_logic_sv2::handlers::mining::SendTo<Downstream>, RolesLogicError> {
-        todo!()
+        Ok(SendTo::None(Some(Mining::CloseChannel(m.as_static()))))
```
Fi3, line 749: `SendTo::None(Some(Mining::CloseChannel)` is not matched in match next_message_to_send this lead to a panic

Fi3: This  was the one missing, I reviewed the code on vim then jumped here and wrote that comment for the wrong handler sorry

roles/v2/pool/src/lib/mining_pool/mod.rs
darricksee: There is no way to guarantee that a future dev won't introduce memory leaks. This additional way to drop downstreams doesn't make the opportunity for memory leaks greater. I'm also fine with adding the debug_assert but I don't think it adds any value - it's perfectly reasonable for a downstream to be removed from the main version of the map but still reside (temporarily) in a cloned version. Currently all the cloned versions are only in scope for a short amount of time and once they go out of scope they are freed.

Fi3: agree what worry me is the name if I call drop_something I expect to drop the actual thing not just the Arc. The debug assert IMHO in the case of a future dev that call drop thinking of dropping downstream and not just the Arc will be maybe notified during the tests that the assumption is not correct, not much value btw.

sutt0n: If you wanted to drop the actual thing, could you expose the the actual thing's mutex guard via the mutex wrapper's `to_remove()`?

Fi3 (approved):

PR #577, opened 2023-05-21
WIP start token signature for JDS

4ss0:
Signature that uses the ed25519 schema to sign the merkle path in the CommitMiningJobSuccess message.

The following are the things that need to be done:
    -Use the pool's keys. Currently, the code is using example keys that are only for testing purposes.
    -Complete the tests by let the pool use the verification code for the signed data


roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -1,27 +1,30 @@
...
-use tracing::info;
+use tracing::{debug, info};
 
 #[derive(Debug)]
 pub struct JobDeclaratorDownstream {
     sender: Sender<EitherFrame>,
     receiver: Receiver<EitherFrame>,
     // TODO this should be computed for each new template so that fees are included
     coinbase_output: Vec<u8>,
+    config: Configuration,
```
Fi3, line 27: We do not need to have all the config data here. Also best practice is to not derive Debug on struct that contains sensible data (like the pool private key). You can either implement Debug manually and avoid to log the secret or using something like https://docs.rs/secrecy/latest/secrecy/struct.Secret.html

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -118,6 +126,72 @@ fn get_random_token() -> B0255<'static> {
...
+            .fold(vec![], |mut acc, bs| {
+                for b in bs {
+                    acc.push(b)
+                }
+                acc
+            });
+
+    // Sign message
+    let signature: Signature = keypair.sign(&message);
+    println!("signature is: {:?}", signature);
```
Fi3, line 161: use info! or debug! not println!

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -118,6 +126,72 @@ fn get_random_token() -> B0255<'static> {
...
+    signature.to_bytes().to_vec().try_into().unwrap()
+}
+#[allow(dead_code)]
+pub fn verify_token(
```
Fi3, line 165: This should be in the pool not in the job negotiator server

roles/jd-server/src/lib/job_declarator/mod.rs
4ss0: I'll give a look to the struct.Secret and implement this solution

roles/jd-server/src/lib/job_declarator/mod.rs
4ss0: yup, forgot to change it here, thanks

4ss0:
Move the signature to JD_logic branch at PR #593. This PR will be soon close and not merged

PR #583, opened 2023-06-15
Mg replacefield struct and code tests

lorbax:
FIRST COMMIT:
    ReplaceField struct

    The messages came along with a vector of (String, String), where the
    first entry is the field_name to be replaced and the second is the
    keyword associated. This is replaced with a much more human readable
    struct Replacefield {
        field_name: String,
        keyword: String
    }

SECOND COMMIT:
    improve code tests

    - one test did't pass and now it is fixed
    - add a second test that checks on a sv2 message that the composition
    (serialization and then deserialization) is the identity

THIRD COMMIT:
    SaveField string

    In GetMessageField, the field "fields" that represents the fields that
    must be saved from the message is Vec<(String, Strin)>, where first
    entry is the the field_id and the second is the keyword to retrieve the
    first. Now a more suitable struct
    SaveField {
       field_id: String,
       keyword: String,
    }
    is introduced.

FOURTH COMMIT:
    fix bug in function change_value_of_serde_field

    This function (in executor.rs) worked only for MiningMessages. Now it returns the
    modified message as a String and can be used in everywhere in
    change_fields function (also in executor.rs)

FIFTH COMMIT
    fix replace field bug

    Before, if we tried to use ReplaceField in a message with a field that
    is not a u32, it would generate an error because the function
    change_value_of_serde_field was set to modify only fields that are u32.
    Now it is fixed. You can see it by running the translation-proxy.json,
    where several fields' types in messages are replaced.

### DIFF
--- a/test/message-generator/test/translation-proxy.json
+++ b/test/message-generator/test/translation-proxy.json
@@ -12,18 +12,27 @@
         "You can use any miner you want, by changing the execution command or by launching it separately"
     ],
     "common_messages": [
+        {
+            "message": {
+                "type": "SetupConnectionSuccess",
+                "flags": 300,
+                "used_version": 180
+            },
+            "replace_fields": [["flags", "setup_connection_success_flag"],["used_version", "setup_connection_success_used_version"]],
+            "id": "setup_connection_success_tproxy"
+        }
     ],
     "mining_messages": [
         {
             "message": {
                 "type": "OpenExtendedMiningChannelSuccess",
-                "request_id": 666666,
+                "request_id": 666,
                 "channel_id": 1,
-                "target": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 255, 255, 255, 255],
+                "target": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 55, 55, 55, 55, 25, 25, 0, 0],
                 "extranonce_size": 16,
                 "extranonce_prefix": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
             },
-            "replace_fields": [["request_id","variable_name"]],
+            "replace_fields": [["request_id","open_extended_mining_channel_success_request_id"], ["target", "open_extended_mining_channel_success_target"]],
             "id": "open_extended_mining_channel_success"
         },
         {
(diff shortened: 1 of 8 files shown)


test/message-generator/test/translation-proxy.json
```diff
@@ -12,18 +12,27 @@
...
-                "request_id": 666666,
+                "request_id": 666,
                 "channel_id": 1,
-                "target": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 255, 255, 255, 255],
+                "target": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 55, 55, 55, 55, 25, 25, 0, 0],
                 "extranonce_size": 16,
                 "extranonce_prefix": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
             },
-            "replace_fields": [["request_id","variable_name"]],
+            "replace_fields": [["request_id","variable_name"], ["target", "daisy"]],
```
Fi3, line 35: As var name it would be better to use something meaningful eg: open_extended_mining_channel_response_request_id or open_extd_m_channel_req_id

This is valid for all the variable_name in this PR

lorbax: ok now the labels should be ok!

utils/message-generator/src/executor.rs
```diff
@@ -758,19 +759,21 @@ impl Executor {
...
+    let field_name = next.field_name;
     let value = values
         .get(&keyword)
         .expect("value not found for the keyword");
 
+    let message_as_string = serde_json::to_string(&m.clone()).unwrap();
+
     match m.clone() {
         AnyMessage::Common(m) => {
             let path = match m {
```
Fi3, line 779: You do not need the path string. You can do:
```rust
            let message_as_serde_value = serde_json::to_value(&m).unwrap().as_object().unwrap().values().next().unwrap().clone();
```
 This will take the serde::Value that you want. You can use it everywhere in this function instead of using the path var

utils/message-generator/src/executor.rs
```diff
@@ -779,8 +782,9 @@ fn change_fields<'a>(
                 CommonMessages::SetupConnectionError(_) => "SetupConnectionError",
                 CommonMessages::SetupConnectionSuccess(_) => "SetupConnectionSuccess",
             };
-            let mut message_as_serde_value = serde_json::to_value(&m).unwrap();
+            let message_as_serde_value = serde_json::to_value(&m).unwrap();
             let m_ = change_value_of_serde_field(message_as_serde_value, path, value, field_name);
+            let m_ = into_static(AnyMessage::Common(serde_json::from_str(&m_).unwrap()));
             if replace_fields.len() == 0 {
```
Fi3, line 788: Use `.is_empty()` instead of `.len() == 0`

Fi3:
Also if the MG is not tacked by the github actions, before a PR please do cargo clippy and cargo fmt and solve all the warnings.

Fi3:
The replace_field command is not documented in the README. Add docs for it

PR #594, opened 2023-07-11
This PR fixes some bugs on Message Generetor

lorbax:
The aforementioned bugs only affects the SRI compiled using serde, so no errors are displayed.
Nevertheless, as the Message Generator uses serde, they come clear.

### DIFF
--- a/protocols/v2/subprotocols/job-declaration/src/identify_transactions.rs
+++ b/protocols/v2/subprotocols/job-declaration/src/identify_transactions.rs
@@ -9,7 +9,6 @@ use core::convert::TryInto;
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 #[repr(C)]
 pub struct IdentifyTransactions {
-    #[cfg_attr(feature = "with_serde", serde(borrow))]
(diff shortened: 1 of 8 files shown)


protocols/v2/subprotocols/job-declaration/src/identify_transactions.rs
```diff
@@ -18,15 +17,16 @@ pub struct IdentifyTransactions {
...
 }
 
 #[cfg(feature = "with_serde")]
 use binary_sv2::GetSize;
 #[cfg(feature = "with_serde")]
-impl<'d> GetSize for IdentifyTransactions<'d> {
+impl GetSize for IdentifyTransactions {
     fn get_size(&self) -> usize {
-        self.user_identifier.get_size() + self.request_id.get_size()
+        self.request_id.get_size() + self.request_id.get_size()
```
4ss0, line 29: Here we should remove one of the two "self.request_id.get_size()"

4ss0 (commented):
Looks good to me for the JD part, just one little thing about a redundant code that should be removed. Thank you @lorbax

PR #614, opened 2023-09-11
Benchmark workflow fix (bencher token)

GitGab19:


GitGab19:
> It seems that by doing in this way, bencher is executed correctly. The last problem is that automatic bencher bot comments are not generated. And this is because the workflow is not triggered anymore by a `pull_request` but from a `pull_request_target`. I'm in contact with the bencher developer to try to fix this.

https://github.com/bencherdev/bencher/pull/194

PR #619, opened 2023-09-23
Coinbase output types addition 

### LINKED ISSUE #550: Improve pool coinbase outputs config
**_Originally posted by Fi3_**
Currently, the pool can only generate a coinbase with a p2pkh script as the pool's output. We should allow for custom scripts and make it configurable from the config file.

GitGab19:
This PR fixes #550.
It's still WIP since the last outuput type is still missing is the P2TR.
I will try to implement it tomorrow.


GitGab19:
This PR is finally ready to be reviewed and tested.
I had to fix some message generator tests were failing due to changes in the coinbase tx.
Fix #550

GitGab19:
Let me know if you think it's necessary to add some more explanation in the config files.
E.g `for the P2PK, P2PKH, P2WPKH and P2TR script types, the output_script_value has to be a public key` and `for the P2SH, P2WSH script types, the output_script_value has to be a valid script`.

roles/jd-server/jds-config.toml
```diff
@@ -4,16 +4,23 @@ authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
 listen_address = "0.0.0.0:34254"
-# list of compressed or uncompressed pubkeys for coinbase payout (only supports 1 item in the array at this point)
+
+# list of coinbase outputs used to build the coinbase tx
+# ! right now only one output is supported, so comment all the ones you don't need !
 coinbase_outputs = [
```
Fi3, line 10: change the invalid values in `output_script_value` and with `...`

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -43,7 +43,7 @@ impl JobDeclaratorDownstream {
         // TODO: use next variables
         let token_to_job_map = HashMap::with_hasher(BuildNoHashHasher::default());
         let tokens = Id::new();
-        crate::get_coinbase_output(config)[0]
+        crate::get_coinbase_output(config).unwrap()[0]
```
Fi3, line 46: Use expect with error message here.

roles/jd-server/src/main.rs
```diff
@@ -22,66 +27,147 @@ pub type Message = PoolMessages<'static>;
 pub type StdFrame = StandardSv2Frame<Message>;
 pub type EitherFrame = StandardEitherFrame<Message>;
 
-const BLOCK_REWARD: u64 = 5_000_000_000;
+const BLOCK_REWARD: u64 = 625_000_000;
```
Fi3, line 30: This should be automatically updated. Open an issue for it.

roles/jd-server/src/main.rs
```diff
@@ -22,66 +27,147 @@ pub type Message = PoolMessages<'static>;
...
                 value: crate::BLOCK_REWARD,
-                script_pubkey: Script::new_p2pkh(&hashed),
-            }
+                script_pubkey: output_script,
+            })
         })
-        .collect()
+        .collect::<Result<Vec<TxOut>, OutputScriptError>>();
+
+    if result.is_ok() && result.as_ref().unwrap().is_empty() {
```
Fi3, line 44:
```rust
match result.as_deref() {
   Ok([]) => Err(OutputScriptError::EmptyCoinbaseOutputs("Empty coinbase outputs".to_string(),)),
   _ => result,
}
```

roles/jd-server/src/main.rs
```diff
@@ -22,66 +27,147 @@ pub type Message = PoolMessages<'static>;
...
+    type Error = OutputScriptError;
 
-use crate::{lib::job_declarator::JobDeclarator, status::Status};
+    fn try_from(value: &CoinbaseOutput) -> Result<Self, Self::Error> {
+        match value.output_script_type.as_str() {
+            "P2PK" => {
+                if is_public_key(&value.output_script_value) {
+                    Ok({
+                        let pub_key =
+                            PublicKey::from_str(value.output_script_value.as_str()).unwrap();
```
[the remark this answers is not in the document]
Fi3: Why doing `is_public_key` if  `PublicKey::from_str` return a `Result`?

roles/pool/pool-config.toml
```diff
@@ -4,9 +4,16 @@ authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
 listen_address = "0.0.0.0:34254"
-# list of compressed or uncompressed pubkeys for coinbase payout (only supports 1 item in the array at this point)
+
+# list of coinbase outputs used to build the coinbase tx
+# ! right now only one output is supported, so comment all the ones you don't need !
 coinbase_outputs = [
```
Fi3, line 10: Remove invalid values

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -466,7 +466,7 @@ impl Pool {
             creator,
             share_per_min,
             kind,
-            pool_coinbase_outputs,
+            pool_coinbase_outputs.unwrap(),
```
Fi3, line 469: Use expect

roles/pool/src/main.rs
```diff
@@ -22,68 +27,147 @@ pub type Message = PoolMessages<'static>;
 pub type StdFrame = StandardSv2Frame<Message>;
 pub type EitherFrame = StandardEitherFrame<Message>;
 
-const BLOCK_REWARD: u64 = 5_000_000_000;
-
-//const COINBASE_ADD_SZIE: u32 = 100;
+const BLOCK_REWARD: u64 = 625_000_000;
```
Fi3, line 30: Include also this in the issue about automatically update this value

roles/pool/src/main.rs
```diff
@@ -22,68 +27,147 @@ pub type Message = PoolMessages<'static>;
...
 
-use tokio::select;
+impl TryFrom<&CoinbaseOutput> for Script {
+    type Error = OutputScriptError;
 
-use crate::status::Status;
+    fn try_from(value: &CoinbaseOutput) -> Result<Self, Self::Error> {
+        match value.output_script_type.as_str() {
+            "P2PK" => {
+                if is_public_key(&value.output_script_value) {
```
Fi3, line 59: Same as above no need to check if the value is valid cause we do that later

Fi3: Also valid for all the other branches

Fi3: Also maybe this logic and the one above about trasforming a toml value into a TxOut should stay in roles-logic-sv2 utils

roles/test-utils/sv1-mining-device/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 [dependencies]
 async-channel = "1.5.1"
 async-std = { version = "1.8.0", features = ["attributes"] }
-bitcoin = "0.27.1"
+bitcoin = "0.28.1"
```
Fi3, line 12: why here version 28 and above version 29?

test/config/interop-jd-translator/pool-config.toml
```diff
@@ -4,9 +4,15 @@ authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
 listen_address = "0.0.0.0:34254"
-# list of compressed or uncompressed pubkeys for coinbase payout (only supports 1 item in the array at this point)
+# list of coinbase outputs used to build the coinbase tx
+# ! right now only one output is supported, so comment all the ones you don't need !
 coinbase_outputs = [
-    "02e13cef1348924c49dd1f708bf38eb79ae4648c16f0301085f37547d1d25e33e0",
+    #{ output_script_type = "P2PK", output_script_value = "02e13cef1348924c49dd1f708bf38eb79ae4648c16f0301085f37547d1d25e33e0" },
```
Fi3, line 10: remove invalid values

test/config/pool-config-sri-tp-different-port.toml
```diff
@@ -5,6 +5,13 @@ authority_public_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"
 authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
+# list of coinbase outputs used to build the coinbase tx
+# ! right now only one output is supported, so comment all the ones you don't need !
 coinbase_outputs = [
-    "04466d7fcae563e5cb09a0d1870bb580344804617879a14949cf22285f1bae3f276728176c3c6431f8eeda4538dc37c865e2784f3a9e77d044f33e407797e1278a",
+    #{ output_script_type = "P2PK", output_script_value = "02e13cef1348924c49dd1f708bf38eb79ae4648c16f0301085f37547d1d25e33e0" },
```
Fi3, line 11: remove invalid values

test/config/pool-config-sri-tp.toml
```diff
@@ -5,6 +5,13 @@ authority_public_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"
 authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
+# list of coinbase outputs used to build the coinbase tx
+# ! right now only one output is supported, so comment all the ones you don't need !
 coinbase_outputs = [
-    "04466d7fcae563e5cb09a0d1870bb580344804617879a14949cf22285f1bae3f276728176c3c6431f8eeda4538dc37c865e2784f3a9e77d044f33e407797e1278a",
+    #{ output_script_type = "P2PK", output_script_value = "02e13cef1348924c49dd1f708bf38eb79ae4648c16f0301085f37547d1d25e33e0" },
```
Fi3, line 11: remove invalid values

test/config/pool-mock-tp-bad-config.toml
```diff
@@ -10,6 +10,13 @@ authority_public_key = "2di19GHYQnAZJmEpoUeP7C3Eg9TCcksHr23rZCC83dvUiZgiDL"
 authority_secret_key = "2Z1FZug7mZNyM63ggkm37r4oKQ29khLjAvEx43rGkFN47RcJ2t"
 cert_validity_sec = 3600
 test_only_listen_adress_plain =  "0.0.0.0:34250"
+# list of coinbase outputs used to build the coinbase tx
+# ! right now only one output is supported, so comment all the ones you don't need !
 coinbase_outputs = [
-    "04466d7fcae563e5cb09a0d1870bb580344804617879a14949cf22285f1bae3f276728176c3c6431f8eeda4538dc37c865e2784f3a9e77d044f33e407797e1278a",
+    #{ output_script_type = "P2PK", output_script_value = "02e13cef1348924c49dd1f708bf38eb79ae4648c16f0301085f37547d1d25e33e0" },
```
Fi3, line 16: remove invalid values

roles/jd-server/src/main.rs
Fi3: After checking spec I realized that we already get the full amount from the TP. Why we need this const? Maybe should be removed.

roles/pool/src/main.rs
Fi3: After checking spec I realized that we already get the full amount from the TP. Why we need this const? Maybe should be removed.

roles/jd-server/jds-config.toml
GitGab19: do you mean all the ones which are commented or only the P2SH and P2WSH which would require a script in the `output_script_value`?

Fi3: What I mean is that all the pair output_script_type,  output_script_value should be valid. If a script value is not valid just replace it with ...

roles/jd-server/src/main.rs
GitGab19: It's needed just to build the TxOut struct which is then modified with the `value_remaining` got from TP.
I can delete the constant and put a `0` value here if you want.

Fi3: Yep maybe is the best option. So we can check before sending the TxOut if the value is 0 and if it is it means that for whatever reason the value has not been changed so we fail. We should also add a comment where we build the TxOut that explain it.

roles/pool/src/main.rs
GitGab19: > Also maybe this logic and the one above about trasforming a CoinbaseOutput into a TxOut should stay in roles-logic-sv2 utils

I agree.
I will do this after the PR which will be opened by @vincenzopalazzo to create a stratum-common to fix the rust-bitcoin thing. Because I tried to move this logic in `roles-logic-sv2/utils.rs` but I messed up with rust-bitcoin version which doesn't support the `PublicKey::fromstr()` method anymore.

Fi3:
I'm going to merge it and fix right after in a following PR `cargo fmt` and `../../test/message-generator/test/bad-pool-config-test.json` and very minor things not addressed in the last commit.

PR #620, opened 2023-09-25
 JD tx short hash list 

4ss0:
add functionality for the DeclareMiningJob message for the values:  tx_short _hash _list and tx_hash_list_hash. These are required to let the JD server to propagate the block by checking the transactions hashes against its own TP or mempool. If they are not present IdentifyTransactions message will be sent by the JD server to retrive the missing transactions.


4ss0:
closing because wrong branch to merge to

PR #622, opened 2023-09-28
Jd short hash

4ss0:
---recreate PR because it was created on wrong branch---

Add functionality for the DeclareMiningJob message for the values: tx_short _hash _list and tx_hash_list_hash. These are required to let the JD server to propagate the block by checking the transactions hashes against its own TP or mempool. If they are not present IdentifyTransactions message will be sent by the JD server to retrive the missing transactions.

### DIFF
--- a/protocols/v2/binary-sv2/serde-sv2/src/primitives/byte_arrays/b016m.rs
+++ b/protocols/v2/binary-sv2/serde-sv2/src/primitives/byte_arrays/b016m.rs
@@ -196,4 +196,10 @@ impl<'a> B016M<'a> {
             Inner::Owned(inner) => B016M(Inner::Owned(inner)),
         }
     }
+    pub fn to_vec(self) -> Vec<u8> {
+        match self.0 {
+            Inner::Ref(v) => v.to_vec(),
+            Inner::Owned(_) => todo!(),
+        }
+    }
 }
--- a/protocols/v2/roles-logic-sv2/src/utils.rs
+++ b/protocols/v2/roles-logic-sv2/src/utils.rs
@@ -564,29 +564,31 @@ pub fn get_target(
     hash
 }
 
-pub fn short_hash_list(
+pub fn hash_lists_tuple(
     tx_data: Seq064K<'static, B016M<'static>>,
     tx_short_hash_nonce: u64,
 ) -> (Seq064K<'static, ShortTxId<'static>>, U256<'static>) {
-    let tx_short_hash_list = short_hash_list_builder(tx_short_hash_nonce, tx_data.clone());
-    let tx_hash_list_hash = tx_hash_list_hash_builder(tx_data);
-    (tx_short_hash_list, tx_hash_list_hash)
-}
-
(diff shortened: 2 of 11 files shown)


protocols/v2/binary-sv2/serde-sv2/src/primitives/byte_arrays/b016m.rs
```diff
@@ -196,4 +196,10 @@ impl<'a> B016M<'a> {
             Inner::Owned(inner) => B016M(Inner::Owned(inner)),
         }
     }
+    pub fn to_vec(self) -> Vec<u8> {
```
Fi3, line 199: Why you need this? Remove if possible.l

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -563,6 +564,47 @@ pub fn get_target(
...
+    // hash the short hash nonce
+    let nonce_hash = sha256::Hash::hash(&tx_short_hash_nonce.to_le_bytes());
+    // take first two integers from the hash
+    let k0 = u64::from_le_bytes(nonce_hash[0..8].try_into().unwrap());
+    let k1 = u64::from_le_bytes(nonce_hash[8..16].try_into().unwrap());
+    let mut tx_short_hash_list = vec![];
+    // get every transaction, hash it, remove first two bytes and push the ShortTxId in a vector
+    for tx in tx_data.to_vec() {
+        let hasher = SipHasher24::new_with_keys(k0, k1);
+        let tx_hashed = hasher.hash(&tx);
```
Fi3, line 589: you need to hash the tx id not the all tx data. I guess that the best thing here is to build a Tx using rust-bitcoin and get the tx id

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -563,6 +564,47 @@ pub fn get_target(
...
+        let tx_hashed_bytes: Vec<u8> = tx_hashed.to_le_bytes().to_vec().drain(0..2).collect();
+        let tx_hashed_bytes: ShortTxId = tx_hashed_bytes.try_into().unwrap();
+        tx_short_hash_list.push(tx_hashed_bytes);
+    }
+    Seq064K::from(tx_short_hash_list)
+}
+
+fn tx_hash_list_hash_builder(_tx_data: Seq064K<'static, B016M<'static>>) -> U256 {
+    // TODO: understand if this field is redunant and to be deleted since
```
Fi3, line 598: Yep for now is ok this solution cause we are not going to use in any way this field on our implementation. Btw we should open an issue for it.

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -563,6 +564,47 @@ pub fn get_target(
     hash
 }
 
+pub fn short_hash_list(
```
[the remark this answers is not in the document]
4ss0: will change the name

PR #638, opened 2023-10-12
critical fix for jdc config

GitGab19:
In the last PR (#637) I renamed the config file for jd-client from proxy-config.toml to jdc-config.toml.
However, the default config name was not updated, so right now if you do `cargo run` in jdc path, it doesn't work.
Here the fixes to this problem.


pavlenex:
tACK

PR #650, opened 2023-10-16
Fix jds1

Fi3:


roles/jd-server/src/lib/mempool/rpc_client.rs
```diff
@@ -47,6 +47,15 @@ impl RpcClient {
             .map(|client| RpcClient { client })
             .map_err(|e| BitcoincoreRpcError::JsonRpc(e.into()))
     }
+    pub fn submit_block(
+        &self,
+        submit_block: SubmitBlock,
+    ) -> Result<HashMap<String, SubmitBlock>, BitcoincoreRpcError> {
```
lorbax, line 53: Why we simply don't have a reply that is Result<(), Error>?  most implementations (as well as bitcoin reference) opt for this choice. Anyway, this was specified in bip22, so perhaps it is ok.
https://developer.bitcoin.org/reference/rpc/submitblock.html

roles/jd-server/jds-config.toml
```diff
@@ -16,8 +16,8 @@ coinbase_outputs = [
 listen_jd_address = "127.0.0.1:34264"
 
 core_rpc_url =  "http://127.0.0.1"
-#core_rpc_port = 18443
-core_rpc_port = 18332
+core_rpc_port = 18443
+#core_rpc_port = 18332
 core_rpc_user =  "username"
 core_rpc_pass =  "password"
 ore_rpc_pass =  "password"
```
lorbax, line 23: this should be removed

PR #654, opened 2023-10-23
Config files fixes

GitGab19:
Fixed some config files (e.g. translator proxy) in order to use the correct port of the JDC.
When merged, I will update docs accordingly


test/config/interop-jd-translator/proxy-config.toml
```diff
@@ -4,7 +4,7 @@ upstream_port = 34265
 upstream_authority_pubkey = "3VANfft6ei6jQq1At7d8nmiZzVhBFS4CiQujdgim1ign"
 
 # Local Mining Device Downstream Connection
-downstream_address = "0.0.0.0"
+downstream_address = "127.0.0.1"
```
lorbax, line 7: why this?

pavlenex: Only way it works on Windows afaik.

lorbax: no, not really important.
I think that if you are making it work with a miner in your LAN (and not just a cpu miner), than you have to set it back to 0.0.0.0. I am not sure anyway

GitGab19: > why this?

I changed it because the MG test was failing, but it seems that this is not the fix

lorbax: it shouldn't be that the problem!

lorbax: LGTM anyway

PR #657, opened 2023-10-25
Fix translator timeout connection with downstream

Fi3:


roles/translator/proxy-config.toml
GitGab19, line 1: necessary? 
can it be removed in order to be consistent with the other roles?

roles/translator/config-examples/proxy-config-local-jdc-example.toml
GitGab19, line 1: can the `downstream_address = "0.0.0.0"` be set to `127.0.0.1`?
because on mac os it seems that it doesn't work

roles/translator/config-examples/proxy-config-local-pool-example.toml
GitGab19, line 1: same here

PR #666, opened 2023-11-07
Translator proxy `clean_jobs` fix

GitGab19:
This PR fixes the issue with translator proxy, regarding the flag `clean_jobs`.
@jakubtrnka a review from you would be appreciated 🙏


jakubtrnka:
Looks good to me

PR #673, opened 2023-11-16
Difficulty management in Translator FIX

GitGab19:
This PR addresses the difficulty management in Translator. 
I left some logs in order to properly test things out with asic machines.
I will remove them after testing it.

PR #675, opened 2023-11-18
Fix dividing by zero 2

lorbax:
the functions hash_rate_from_target and hash_rate_to_target performed some divisions between integers. If the denominator is zero, the process panics. So the signatures are changed with a Result<T,E> and consequentially the proper error management is introduced. When one of these two functions go with the unhappy path, the ambient process that called them just ignore the iteration of the loop. See [here](https://github.com/lorbax/stratum/blob/fix-dividing-by-zero/roles/translator/src/status.rs#L173) for details.

A test for the function update_miner_hashrate in diff_management.rs is added.


GitGab19:
Tested with both cpuminer and asic miner, integrating the functions modified here in my PR (#673) and it seems to work properly now! 💯

Issue #166, opened 2022-06-27
Pass in `message` to `Sv2Error::PayloadTooBig` variant


rrybarczyk:
Make `Sv2Error::PayloadTooBig` -> `Sv2Error::PayloadTooBig(String)`  and pass in `message` as string in `protocols/v2/sv2-ffi/src/lib.rs`. This will help tremendously with better understanding error messages for the sv2 bindings.

Currently `message` is of type T (the first paramter passed into [`protocols/v2/framing-sv2/src/framing2.rs`](https://github.com/stratum-mining/stratum/blob/5507ae62552aca941ee89432215a5e33714b6c6c/protocols/v2/framing-sv2/src/framing2.rs#L50)). Should make the `trait Frame` inherit from `core::fmt::Display` like so:

```
pub trait Frame<'a, T: Serialize + GetSize + core::fmt::Display>: Sized {
```

This cases the following errors:

```
error[E0277]: `T` doesn't implement `Display`
  --> protocols/v2/framing-sv2/src/framing2.rs:85:64
   |
85 | impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for Sv2Frame<T, B> {
   |                                                                ^^^^^^^^^^^^ `T` cannot be formatted with the default formatter
   |
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
note: required by a bound in `Frame`
  --> protocols/v2/framing-sv2/src/framing2.rs:21:46
   |
21 | pub trait Frame<'a, T: Serialize + GetSize + core::fmt::Display>: Sized {
   |                                              ^^^^^^^^^^^^^^^^^^ required by this bound in `Frame`
help: consider further restricting this bound
   |
85 | impl<'a, T: Serialize + GetSize + core::fmt::Display, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for Sv2Frame<T, B> {
   |                                 ++++++++++++++++++++

error[E0277]: `alloc::vec::Vec<u8>` doesn't implement `Display`
   --> protocols/v2/framing-sv2/src/framing2.rs:199:10
    |
199 | impl<'a> Frame<'a, Vec<u8>> for NoiseFrame {
    |          ^^^^^^^^^^^^^^^^^^ `alloc::vec::Vec<u8>` cannot be formatted with the default formatter
    |
[...]

PR #679, opened 2023-11-24
FFI: pass in `message` to `Sv2Error::PayloadTooBig` variant

plebhash:
close https://github.com/stratum-mining/stratum/issues/166

in this PR we are making `Sv2Error::PayloadTooBig` into `Sv2Error::PayloadTooBig(*const std::ffi::c_char)` so that it can carry `message`.

as @rrybarczyk put it on the original issue description:
> This will help tremendously with better understanding error messages for the sv2 bindings.

---

Note 1: this PR is not doing `Sv2Error::PayloadTooBig(String)` like it was suggested on the original issue description.

`*const std::ffi::c_char` was chosen instead of `String` for easier `cbindgen` compatibility. If `String` is used, `sv2.h` has incomplete type ‘String’ and compilation of `examples/interop-cpp/template-provider/template-provider.cpp` breaks as a consequence.

---

Note 2: this PR is not modifying `trait Frame` like it was suggested on the original issue description.

Instead, we simply implement the `Display` trait for `Sv2Message` and the error can be created via:
```rust
let message: Sv2Message;
...
Sv2Error::PayloadTooBig(
  CString::new(format!("{}", message))
    .expect("CString::new failed")
    .into_raw(),
))
```

### DIFF
--- a/protocols/v2/sv2-ffi/src/lib.rs
+++ b/protocols/v2/sv2-ffi/src/lib.rs
@@ -144,7 +144,7 @@ pub extern "C" fn drop_sv2_error(s: Sv2Error) {
         Sv2Error::EncoderBusy => (),
         Sv2Error::InvalidSv2Frame => (),
         Sv2Error::MissingBytes => (),
-        Sv2Error::PayloadTooBig(a) => drop(a),
+        Sv2Error::PayloadTooBig(_) => (),
         Sv2Error::Unknown => (),
     }
 }
--- a/protocols/v2/sv2-ffi/src/lib.rs
+++ b/protocols/v2/sv2-ffi/src/lib.rs
@@ -1,5 +1,6 @@
 #![cfg(not(feature = "with_serde"))]
 use std::{
+    ffi::{CString, c_char},
     fmt,
     fmt::{Display, Formatter},
 };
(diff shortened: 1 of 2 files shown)


plebhash:
@Fi3 is this useful? I wrote this PR just as an exercise to help me explore `sv2-ffi`. But I can also close it if you feel it's not relevant.

Fi3:
A lot of time that I do not work on that part. Btw we should have a safer (respect to c_char) CVec. Maybe we can use it instead, not sure btw. Have you looked into it? Also remember to update the c headers with ./build-header.sh and to put them on the sv2ffi crate.

plebhash:
> Btw we should have a safer (respect to c_char) CVec. Maybe we can use it instead, not sure btw. Have you looked into it?

I saw `CVec` on some parts of the code but didn't consider it for replacing `String`. I will look into it 👍

> Also remember to update the c headers with ./build-header.sh and to put them on the sv2ffi crate.

yes, I am taking care of that 👍

Fi3:
take a look also here:

https://github.com/stratum-mining/stratum/tree/dev/examples/interop-cpp#cvec

plebhash:
@Fi3 replace `*const c_char` with `CVec` here https://github.com/stratum-mining/stratum/pull/679/commits/96878eaed7d4235f8709492fc95bcb7b8d5d973d

PR #684, opened 2023-11-27
Cross Compilation: add a `sv2-rpi.sh` script to generate tarball distributions for Raspberry Pi OS

plebhash:
Most edge network hardware that will run the SRI stack will most likely not be x86-64 cloud servers.
 
A much more likely target will be ARM SoCs and SBCs, especially when it comes to pleb mining.

Therefore it is important to make sure SRI supports the most popular off-the-shelf platforms on this category (e.g.: Raspberry Pi).

This contribution is also the first step towards SRI supporting ASIC Control Boards in the future.

stratum-rpi

This PR introduces a shell script (`sri-rpi.sh`) that automates cross compilation for tarball distributions (32 and 64 bits) of the different SRI executables.

## `sv2-rpi.sh`

Here is how to use `sv2-rpi.sh`:
```shell
Usage:
$ ./sv2-rpi.sh <role> <bits>

Available options for <role> are:
jd_client
jd_server
mining_proxy_sv2
pool_sv2
translator_sv2

Available options for <bits> are:
32
64

For example:
$ ./sv2-rpi.sh jd_server 64
```

After the package generation, you are ready to `scp` the tarball to the RPi:
```shell
$  scp tar/jd_server-aarch64-unknown-linux-gnu.tar.gz <rpiuser>@<your.rpi.ip.addr>:/home/<rpiuser>
```

## package layout

Configuration files are placed under `/etc/sri`, and executables are placed under `/bin`. For example, here is the tarball layout of `jd_server`:
```
├── bin
│   └── jd_server
└── etc
    └── sri
        ├── jds-config-hosted-example.toml
        └── jds-config-local-example.toml
```

## `cargo-cross`

Under the hood, `sv2-rpi.sh` leverages [`cargo-cross`](https://crates.io/crates/cross), so it is a pre-requisite for executing the script. It can be installed via
```
$ cargo install cross
```

Toolchain management via `cargo-cross` enables a stable cross-compilation experience. It also requires that at least one of these dependencies is available:
- [`docker`](https://www.docker.com/)
- [`podman`](https://podman.io/)


sv2-rpi.sh
```diff
@@ -0,0 +1,155 @@
...
+  echo "tar/$TARBALL created with success."
+}
+
+jd_client(){
+    build
+    pkg
+
+    install target/$TARGET/release/jd_client pkg/$PKG/bin
+    # install roles/jd-client/config-examples/jdc-config-hosted-example.toml pkg/$PKG/etc/sri
+    # install roles/jd-client/config-examples/jdc-config-local-example.toml pkg/$PKG/etc/sri
```
plebhash, line 63: @Fi3 the config files seem to have wrong names on [`roles/jd-client/config-examples`](https://github.com/stratum-mining/stratum/tree/dev/roles/jd-client/config-examples)

sv2-rpi.sh
```diff
@@ -0,0 +1,155 @@
...
+    tarball
+}
+
+translator_sv2(){
+  build
+  pkg
+
+  install target/$TARGET/release/translator_sv2 pkg/$PKG/bin
+#  install roles/translator/config-examples/translator-config-local-jdc-example.toml pkg/$PKG/etc/sri
+#  install roles/translator/config-examples/translator-config-local-pool-example.toml pkg/$PKG/etc/sri
```
plebhash, line 119: @Fi3 tsame thing on [`roles/translator/config-examples`](https://github.com/stratum-mining/stratum/tree/dev/roles/translator/config-examples)

Fi3: yep we can change them to mirror the other roles config name convention `[role type]-config-[other info].toml`. In that case we should also change the default name for the config. Right now translator looks for a file named `proxy-config.toml` in the working directory when not config arg is passed. If we go with this , in order to have a consistent behavior between roles,  we should make the translator proxy looks for `translator-config.toml` instead.

plebhash: ack 
https://github.com/stratum-mining/stratum/pull/684/commits/f9ccaae6dd93d702c40b307c053fb34644515ce9
https://github.com/stratum-mining/stratum/pull/684/commits/cfae6d01b7e3e744a68a849d3360bba123f0f181

sv2-rpi.sh
plebhash: solved via https://github.com/stratum-mining/stratum/pull/684/commits/f9ccaae6dd93d702c40b307c053fb34644515ce9

s2dd:
just some alt. variant = i can compile sv2 roles direct on a rpi and get a bin for dist.

cross compile for Antminer hosting tProxy direct is possible, a Antminer CB(Controlboard) has plenty CPU and Mem left when running its FW, a cross compile of tProxy, then can scp to device then set regular pool to 127.0.0.1:port that in turn points to next ie. pool-endpoint ... 

this for ppl with one or few machines not wanting to run role on a separate machine  ...

GitGab19:
@plebhash let us know if you think this PR is completely ready.
I just saw you pushed some changes recently, and also added `cargo cross` as suggested by @Fi3

Fi3:
@GitGab19 yesterday I suggested to add a github action that check if cross compilation works: just try to compile with cargo cross from a linux vm for the target that we care about and check that no errors or warning are emitted.

GitGab19:
> @GitGab19 yesterday I suggested to add a github action that check if cross compilation works: just try to compile with cargo cross from a linux vm for the target that we care about and check that no errors or warning are emitted.

Right, thanks for better explanation (I failed in taking notes on that yesterday 😆).
@plebhash so please have a look at @Fi3's suggestion for this.

plebhash:
> just some alt. variant = i can compile sv2 roles direct on a rpi and get a bin for dist.
> 
> cross compile for Antminer hosting tProxy direct is possible, a Antminer CB(Controlboard) has plenty CPU and Mem left when running its FW, a cross compile of tProxy, then can scp to device and set regular FW UI pool to 127.0.0.1:port that in turn points to next ie. pool-endpoint ...
> 
> this for ppl with one or few machines not wanting to run role on a separate machine ...

thanks for your input @s2dd 
indeed, this is script can be seen as redundant or overkill, depending on the angle you look

however, the vision here is to have a way to aid non-technical users to deploy SV2/SRI infrastructure with the smallest possible friction

once this script is merged and SRI has stable releases, we can simply add the `aarch64-unknown-linux-gnu` and `arm-unknown-linux-gnueabi` binaries as part of the official distribution. Then, users can simply download those binaries and get them running out-of-the-box on their RPis without the need to install rust toolchain etc.

end of the day, I'm aiming to boost SV2 adoption by combining these two factors:
- pleb-friendly hardware
- low friction UX

s2dd:
yes but to get a rpi dist it could be compiled on rpi was what i tried convey

also i find cross-compile tProxy for Antminer a interesting path forward

plebhash:
> yes but to get a rpi dist it could be compiled on rpi was what i tried convey

compiling natively on the RPi is not ideal for a release workflow for multiple reasons:
- it is much slower, since the RPi SoC has very limited computation power
- the team would need to maintain two physical RPis (32+64 bits) for this purpose

during my experience in the Embedded Software industry I learned that cross compilation workflows should always be preferred, while native compilation should only be a last resort for quick PoCs

> also i find cross-compile tProxy for Antminer a interesting path forward

that is a good idea, I wrote this issue to keep track of it https://github.com/stratum-mining/stratum/issues/711

plebhash:
> @GitGab19 yesterday I suggested to add a github action that check if cross compilation works: just try to compile with cargo cross from a linux vm for the target that we care about and check that no errors or warning are emitted.

done

- here's the workflow file: https://github.com/plebhash/stratum/blob/rpi/.github/workflows/sv2-rpi.yaml
- here's the execution of `cross_compile_64`: https://github.com/stratum-mining/stratum/actions/runs/7427061411/job/20212036182?pr=684
- here's the execution of `cross_compile_32`: https://github.com/stratum-mining/stratum/actions/runs/7427061411/job/20212035649?pr=684

the workflow calls `sv2-rpi.sh` for all roles on both 32 and 64 bits, and then checks if the binary artifacts exists and the tarballs were created correctly

this is not necessarily checking for cross compilation warnings or errors, since `cargo cross` is actually invoked by `sv2-rpi.sh`.
if cross-compilation fails, the build artifacts will not be present, and the `Check artifacts` step will fail

if we eventually decide to expand cross-compilation support to different architectures beyond only `aarch64-unknown-linux-gnu` + `arm-unknown-linux-gnueabi` (RPi 32 and 64 bits, respectively), it might make sense to add a new workflow that uses some Github Action specifically tailored for this task (e.g.: [`houseabsolute/actions-rust-cross@v0`](https://github.com/houseabsolute/actions-rust-cross))... that would allow for more detailed control over the different aspects of cross-compilation and what could be going wrong for different achitectures

however that is probably overkill at the moment, IMO

plebhash:
closing in favor of https://github.com/stratum-mining/stratum/issues/815

Issue #685, opened 2023-11-28
role executables have unclear CLI edge cases


plebhash:
CLI for `pool_sv` executable on `dev` branch is confusing:

```
$ cargo build -p pool_sv2
$ ./target/debug/pool_sv2 -h
2023-11-30T01:05:30.703148Z ERROR pool_sv2: Usage: -h/--help, -c/--config <path|default pool-config.toml> # OK
$ ./target/debug/pool_sv2 --help
2023-11-30T01:05:30.703148Z ERROR pool_sv2: Usage: -h/--help, -c/--config <path|default pool-config.toml> # OK
$ ./target/debug/pool_sv2
2023-11-30T01:04:12.789821Z ERROR pool_sv2: Failed to read config: No such file or directory (os error 2) # ???
```

if no arg is passed, the CLI should return the help message:
```
$ ./target/debug/pool_sv2
2023-11-30T01:05:30.703148Z ERROR pool_sv2: Usage: -h/--help, -c/--config <path|default pool-config.toml>
```

same applies to the following `role` executables:
- `jd_server`
- `jd_client`
- `mining_proxy_sv2`
- `translator_sv2`

GitGab19:
Yeah we decided to remove default config files (which were located at `roles/<role-name>/<role-name>-config.toml`) in order to avoid constant changes by async commits and contribution.
Every config file needed for the roles are in their relative `/config-examples` sub-directories. 
I guess you have to point the default config to one of the config file in that directory, specifying it with `-- -c path-of-the-config-file`.

PR #689, opened 2023-11-30
fix role executable CLI edge cases

plebhash:
close https://github.com/stratum-mining/stratum/issues/685

cc @pavlenex @GitGab19

### DIFF
--- a/roles/jd-client/src/args.rs
+++ b/roles/jd-client/src/args.rs
@@ -19,9 +19,14 @@ enum ArgsResult {
 
 impl Args {
     const DEFAULT_CONFIG_PATH: &'static str = "jdc-config.toml";
+    const HELP_MSG: &'static str = "Usage: -h/--help, -c/--config <path|default jdc-config.toml>";
 
     pub fn from_args() -> Result<Self, String> {
         let cli_args = std::env::args();
+
+        if cli_args.len() == 1 {
+            return Err(Self::HELP_MSG.to_string());
+        }
 
         let config_path = cli_args
             .scan(ArgsState::Next, |state, item| {
@@ -31,10 +36,7 @@ impl Args {
                             *state = ArgsState::ExpectPath;
                             Some(ArgsResult::None)
                         }
-                        "-h" | "--help" => Some(ArgsResult::Help(format!(
-                            "Usage: -h/--help, -c/--config <path|default {}>",
-                            Self::DEFAULT_CONFIG_PATH
-                        ))),
+                        "-h" | "--help" => Some(ArgsResult::Help(Self::HELP_MSG.to_string())),
                         _ => {
                             *state = ArgsState::Next;
 
--- a/roles/jd-server/src/main.rs
+++ b/roles/jd-server/src/main.rs
@@ -99,9 +99,15 @@ mod args {
 
     impl Args {
         const DEFAULT_CONFIG_PATH: &'static str = "jds-config.toml";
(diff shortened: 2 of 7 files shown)


roles/jd-client/src/args.rs
```diff
@@ -19,10 +19,15 @@ enum ArgsResult {
 
 impl Args {
     const DEFAULT_CONFIG_PATH: &'static str = "jdc-config.toml";
+    const HELP_MSG: &'static str = "Usage: -h/--help, -c/--config <path|default jdc-config.toml>";
```
GitGab19, line 22: Here it should be:
`"Usage: -h/--help, -- -c/--config <path|default jdc-config.toml>"`

The double -- is needed before indicating the config

roles/jd-server/src/main.rs
```diff
@@ -99,10 +99,16 @@ mod args {
 
     impl Args {
         const DEFAULT_CONFIG_PATH: &'static str = "jds-config.toml";
+        const HELP_MSG: &'static str =
+            "Usage: -h/--help, -c/--config <path|default jds-config.toml>";
```
GitGab19, line 103: same here

roles/mining-proxy/src/main.rs
```diff
@@ -186,10 +186,16 @@ mod args {
 
     impl Args {
         const DEFAULT_CONFIG_PATH: &'static str = "proxy-config.toml";
+        const HELP_MSG: &'static str =
+            "Usage: -h/--help, -c/--config <path|default proxy-config.toml>";
```
GitGab19, line 190: same here

GitGab19 (changes requested):
Just a minor thing to fix (indicated in comments) before it can be merged.

roles/jd-client/src/args.rs
plebhash: are you sure?

that is usually a pattern on `cargo run`, for example `cargo run -- -c jdc-config.toml`.

However this config message makes no assumptions about `cargo` vs `./target/debug/jd_client` vs `/bin/jd_client`.

f we introduce `--` for help message on any case, we mislead users of `./target/debug/jd_client` and `/bin/jd_client` approaches.

If `cargo` is used, it is assumed that the user knows he needs an extra `--` before adding the args.

by the way, this works fine:

```
$ ./target/debug/jd_client -c roles/jd-client/config-examples/proxy-config-local-example.toml
```

roles/jd-server/src/main.rs
plebhash: https://github.com/stratum-mining/stratum/pull/689#discussion_r1411197074

roles/translator/src/args.rs
```diff
@@ -19,10 +19,15 @@ enum ArgsResult {
 
 impl Args {
     const DEFAULT_CONFIG_PATH: &'static str = "proxy-config.toml";
+    const HELP_MSG: &'static str = "Usage: -h/--help, -c/--config <path|default proxy-config.toml>";
```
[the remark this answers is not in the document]
plebhash: https://github.com/stratum-mining/stratum/pull/689#discussion_r1411197074

roles/jd-client/src/args.rs
GitGab19: You're right @plebhash. I'm sorry but I did it quickly yesterday. 
I also tried it and it works as you said.
For me it can be merged 👍

PR #696, opened 2023-12-12
Parse x-only pubkey for P2TR coinbase output

Sjors:
Configuring a Taproot output for the coinbase currently expects a regular public key:

```toml
coinbase_outputs = [
    { output_script_type = "P2TR", output_script_value = "..." },
]
```

In order to extract the public key from a Bitcoin Core wallet, you'll need to do something like `getaddressinfo tr…` which will give you the _x-only_ public key. Same if you get the output descriptor.

This PR makes life easier by parsing an x-only pubkey.


Sjors:
See #697 for a more generic approach.

Sjors:
Rebased

Issue #702, opened 2023-12-19
SRI APIs for external consumption


plebhash:
## Context

As a pool operator, someone might want to write some Rust code to run their own opinionated pool while consuming to SRI APIs.

`Cargo.toml`:
```
[package]
name = "some_pool"

[dependencies]
pool_sv2 = { git = "https://github.com/stratum-mining/stratum.git", branch = "dev" }
jd_server = { git = "https://github.com/stratum-mining/stratum.git", branch = "dev" }
```

`src/main.rs`:
```
pub use pool_sv2::lib::*;
pub use jd_server::lib::*;

fn main() {
    // opinionated share accounting logic using pool_sv2::lib
    // opinionated coinbase generation logic using jd_server::lib
}
```

## Problem

Currently `some_pool` compilation breaks with:
```
error[E0433]: failed to resolve: use of undeclared crate or module `pool_sv2`
 --> src/main.rs:1:9
  |
1 | pub use pool_sv2::lib::*;
  |         ^^^^^^^^ use of undeclared crate or module `pool_sv2`

error[E0433]: failed to resolve: use of undeclared crate or module `jd_server`
 --> src/main.rs:2:9
  |
2 | pub use jd_server::lib::*;
  |         ^^^^^^^^^ use of undeclared crate or module `jd_server`

error[E0432]: unresolved imports `pool_sv2::lib::*`, `jd_server::lib::*`
 --> src/main.rs:1:9
  |
1 | pub use pool_sv2::lib::*;
  |         ^^^^^^^^^^^^^^^^
2 | pub use jd_server::lib::*;
  |         ^^^^^^^^^^^^^^^^^

```

## Feature Request

`some_pool` needs to consume the crate [`roles/pool`](https://github.com/stratum-mining/stratum/tree/dev/roles/pool) as a `lib`.

The `pool_sv2::lib` module exists, but it is not easily importable by external `bin` consumer.

---

Action items:
- [ ] split workspaces
- [ ] adapt `Cargo.toml`

plebhash:
Here is a proposed structure for the repo:
```
stratum
├── bin
│   ├── Cargo.toml (workspace)
│   ├── jd-client
│   ├── jd-server
│   ├── mining-proxy
│   ├── pool
│   ├── test-utils
│   └── translator
└── lib
    ├── Cargo.toml (workspace)
    ├── examples
    │   ├── interop-cpp
    │   ├── interop-cpp-no-cargo
[...]

PR #718, opened 2024-01-14
roles libs split

plebhash:
as discussed in https://github.com/stratum-mining/stratum/issues/702, this PR splits the `roles` crates so that they provide both an executable and a `lib` to be consumed as dependency by some external project


Fi3:
LGTM. 
Tested pool + jdc + jds + translator on regtest.

lorbax:
Tested in testnet with pool, JDS, JDC, local TP and cpuminer, everithing ok.
Tried importing in a minimal crate, stored locally at the same path as the /stratum crate, with the following toml
```
[package]
name = "consuming-stratum-v2-library"
version = "0.1.0"
edition = "2021"

[dependencies]
jd_client = { path = "../stratum/roles/jd-client" }
pool_sv2 = { path = "../stratum/roles/pool" }
mining_proxy_sv2 = { path = "../stratum/roles/mining-proxy" }
translator_sv2 = { path = "../stratum/roles/translator" }
error_handling = { path = "../stratum/utils/error-handling" }
buffer_sv2 = { path = "../stratum/utils/buffer" }
const_sv2 = { path = "../stratum/protocols/v2/const-sv2" }

```
and this main.rs

```
use const_sv2::MESSAGE_TYPE_NEW_MINING_JOB;
fn main() {
    let message_type_new_mining_job = MESSAGE_TYPE_NEW_MINING_JOB;
    println!("Message type new mining job: {:?}", message_type_new_mining_job);
}
```

It works fine.

PR #720, opened 2024-01-18
fix JDC stack overflow

plebhash:
https://github.com/stratum-mining/stratum/issues/688 has an extensive discussion around a Stack Overflow happening on JDC.

As @GitGab19 pointed out:
* JDC sends a `DeclareMiningJob` message to JDS
* JDS for some unknown reason takes some seconds to answer with a `DeclareMiningJobSuccess` message to JDC
* in the meantime, the JDC creates a new MiningJob (because a NewTemplate arrived from TP)
* JDC's call to `Self::start_templates` initiates a call to `Self::get_last_token`.
* `Self::get_last_token` checks `self.allocated_tokens.len()`, and in case it is `0`, it recursively calls itself, in order to eventually fill `self.allocated_tokens` and have something to work with. That will eventually happen, because `Self::get_last_token` always calls `Self::allocate_tokens`, which sends a `AllocateMiningJobToken ` to JDS.

The problem on the last point is that sometimes, JDS takes a bit longer than expected to reply back with a `AllocateMiningJobTokenSuccess`. So JDC's `self.allocated_tokens` remains empty for a relatively long time, and the multiple recursions on `Self::get_last_token` eventually cause the Stack to overflow.

This PR fixes this issue, because now whenever JDC is executing `Self::get_last_token`, if `self.allocated_tokens` is still empty, then it will we block/yield, and only initiate `Self::get_last_token` recursion after a message has already been received from JDS.

This way, recursion on `Self::get_last_token` is finite. There is no case where it keeps happening indefinitely and causing the Stack Overflow.

I successfully tested this solution by also applying the following modification to JDS:
```diff
diff --git a/protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs b/protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
index 40781e6c..58216e71 100644
--- a/protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
+++ b/protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
@@ -131,6 +131,7 @@ where
             Ok(JobDeclaration::DeclareMiningJob(message)) => {
                 info!("Received DeclareMiningJob with id: {}", message.request_id);
                 debug!("DeclareMiningJob: {:?}", message);
+                std::thread::sleep(std::time::Duration::from_secs(30));
                 self_
                     .safe_lock(|x| x.handle_declare_mining_job(message))
                     .map_err(|e| crate::Error::PoisonLock(e.to_string()))?
```

The JDS modification above is not committed and is not part of this PR.
@GitGab19 figured out that this is a way to force the Stack Overflow to happen deterministically.

With the changes to JDC introduced by this PR, the Stack Overflow never happens, even when we try to force it with the JDS modification technique above.

### DIFF
--- a/roles/jd-client/src/lib/job_declarator/mod.rs
+++ b/roles/jd-client/src/lib/job_declarator/mod.rs
@@ -166,7 +166,14 @@ impl JobDeclarator {
                         })
                         .unwrap();
                 }
-                tokio::task::yield_now().await;
+
+                // we block on the receiver to avoid infinite recursion
+                let receiver = self_mutex.safe_lock(|s| s.receiver.clone()).unwrap();
+                while receiver.is_empty() {
+                    tokio::time::sleep(tokio::time::Duration::from_millis(1)).await;
+                    tokio::task::yield_now().await;
+                }
+
                 Self::get_last_token(self_mutex).await
             }
             1 => {
--- a/roles/jd-client/src/lib/job_declarator/mod.rs
+++ b/roles/jd-client/src/lib/job_declarator/mod.rs
@@ -148,7 +148,7 @@ impl JobDeclarator {
     pub async fn get_last_token(
         self_mutex: &Arc<Mutex<Self>>,
     ) -> AllocateMiningJobTokenSuccess<'static> {
-        let token_len = self_mutex.safe_lock(|s| s.allocated_tokens.len()).unwrap();
+        let mut token_len = self_mutex.safe_lock(|s| s.allocated_tokens.len()).unwrap();
         match token_len {
             0 => {
                 {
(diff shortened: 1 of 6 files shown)


Fi3:
I agree with the idea of the loop so that we can avoid the recursion. Not sure why we want to check is receiver is  not empty rather then just look at:
```rust
s.allocated_tokens.len()

Fi3:
This PR should also contain a test with the MG for that specific case. One approach could be to add a JDS mock in `test/message-generator/mock/` that do the setup connection phase and when we require a new token just hang there.

plebhash:
@Fi3 I pushed some modifications based on your feedback.

I still need to familiarize myself a bit better with MG, so that might take a few days to implement.

If anyone (cc @GitGab19 @lorbax) feels they could write it faster, please feel free to send a PR towards [plebhash:fix-jdc-stack-overflow](https://github.com/plebhash/stratum/tree/fix-jdc-stack-overflow) so I can merge that into my fork and that will end up on this PR too.

GitGab19:
> @Fi3 I pushed some modifications based on your feedback.
> 
> I still need to familiarize myself a bit better with MG, so that might take a few days to implement.
> 
> If anyone (cc @GitGab19 @lorbax) feels they could write it faster, please feel free to send a PR towards [plebhash:fix-jdc-stack-overflow](https://github.com/plebhash/stratum/tree/fix-jdc-stack-overflow) so I can merge that into my fork and that will end up on this PR too.

Good job!
I also share the same feedback given by @Fi3.
Just updated the running machine with your last modifications.
I will let you know if I encounter some issues with that. 
If not, I can write the MG test and open a PR on your branch as you suggested!

lorbax:
Right now I can't work on a JDS mock. How urgently do we need it?

plebhash:
@Fi3 merged https://github.com/plebhash/stratum/pull/2

Issue #721, opened 2024-01-19
SRI uses incorrect encoding of secp256k1 public keys


jakubtrnka:
The specification says that public key should be encoded as 
concatenation of array `[1, 0]` (which is version number) and 32 bytes x-only public key.
using base58-check algorithm

Currently it does direct base58 encoding of the 32 bytes public key without checksum.

See:
https://github.com/stratum-mining/sv2-spec/blob/main/04-Protocol-Security.md#47-url-scheme-and-pool-authority-key

Sjors:
Ah, that's useful to know. I'll switch the correct encoding on the Bitcoin Core side after #722 is merged.

PR #722, opened 2024-01-19
Use correct encoding for public secp256k1 key

jakubtrnka:
This is a suggested solution for  #721

### DIFF
--- a/utils/key-utils/src/lib.rs
+++ b/utils/key-utils/src/lib.rs
@@ -2,18 +2,28 @@ use bs58::{decode, decode::Error as Bs58DecodeError};
 use core::convert::TryFrom;
 use secp256k1::{SecretKey, XOnlyPublicKey};
 use serde::{Deserialize, Serialize};
-use std::fmt::Display;
+use std::{fmt::Display, str::FromStr};
 
 #[derive(Debug)]
 pub enum Error {
     Bs58Decode(Bs58DecodeError),
     Secp256k1(secp256k1::Error),
-    Custom,
+    KeyVersion(u16),
+    KeyLength,
+    Custom(String),
 }
 
 impl Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Key Utils Error")
+        match self {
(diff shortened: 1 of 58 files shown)


utils/key-utils/src/lib.rs
```diff
@@ -58,16 +58,31 @@ impl TryFrom<String> for Secp256k1PublicKey {
...
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        let decoded = decode(value).into_vec()?;
-        let public = XOnlyPublicKey::from_slice(&decoded).expect("Invalid public key");
+        let decoded = decode(value).with_check(None).into_vec()?;
+        if decoded.len() < 34 {
+            return Err(Error::Custom);
+        }
+        if decoded[..2] != [1, 0] {
+            return Err(Error::Custom);
```
Sjors, line 66: Maybe something like `expect("Unsupported future version")`?

Sjors:
Don't forgot to update the branch to main.

utils/key-utils/src/lib.rs
```diff
@@ -29,45 +34,86 @@ impl From<secp256k1::Error> for Error {
...
 impl From<Secp256k1SecretKey> for String {
     fn from(secret: Secp256k1SecretKey) -> Self {
-        let bytes = secret.0.secret_bytes();
-        bs58::encode(bytes).into_string()
+        secret.to_string()
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+impl Display for Secp256k1SecretKey {
```
jakubtrnka, line 65: Display provides to_string method. See From impl

utils/key-utils/src/lib.rs
jakubtrnka: done

Fi3:
LGTM only cargo fmt and tests to fix, I will open a PR with the fixes against @jakubtrnka branch

Fi3:
@jakubtrnka some tests are failing with InvalidChecksum:
```
 value: Bs58Decode(InvalidChecksum { checksum: [178, 200, 220, 184], expected_checksum: [9, 43, 108, 231] })
```

Issue #723, opened 2024-01-19
Incorrect use of ECDH result for establishing shared secret


jakubtrnka:
SRI codebase uses `SharedSecret::new()` code for constructing shared secret:
```rust
        let res = SharedSecret::new(&x_public.public_key(crate::PARITY), &private);
        res.secret_bytes()
```

This is against standard, because internally this method takes x & y coordinate of the resulting EC point and calculates sha256 hash.
It is wrong because hashing the y-coordinate into the result makes the result dependent on key parity of either participants.

Correctly only the X coordinate should be taken. Without hashing:
```rust
        let ec_point_x_y = secp256k1::ecdh::shared_secret_point(
            &x_public.public_key(secp256k1::Parity::Even),
            &private,
        );
        let mut ec_point_x = [0; 32];
        ec_point_x.copy_from_slice(&ec_point_x_y[0..32]);
        ec_point_x
```

GlitchyBritches:
This bug does not surprise me, the reference implementation is generally quite poor quality - so I would not trust any security critical code they've implemented

pavlenex:
Hi @GlitchyBritches care to elaborate what's poor quality in the reference implementation? Did you try it out? Tried to implement it or something else? Would be great if you can provide some more feedback so we can learn what needs to be improved?

GlitchyBritches:
Yes I've been implementing it. And there's a lot that's low quality about it. I'll provide a full list, but starting off:

* Littering of "todos!" and "unimplemented!" all over the place
* Littering of key functions in the "main" files which should be logically organized into a struct
* Fragile parsing logic implemented in the main.rs files + reference to github repos for perfectly stable crates

I'll provide a full list with examples when I have time, but were I to do a reference implementation - I would have been embarrassed to release this publicly.

moneyball:
Which branch of code are you referring to?

pavlenex:
Thanks @GlitchyBritches. 

As @moneyball asked, it would be important to know for context which branch you've tried out, since all of our work is on `dev` and will be merged into `master` after months of work, which addresses at least a few of those you listed, though I'd love to learn what else can be improved, as it's important feedback.

Are you on [our Discord](https://discord.gg/fsEW23wFYs)? Perhaps we can continue conversation there and provide you with instructions which aren't published yet on how to run things on `dev`.

jakubtrnka:
I'd also add that the SRI code is very unstable and not very robust. There is lot of unwraps or expects.
[...]

PR #724, opened 2024-01-19
Use algorithm for ECDH that is described in the specification 

jakubtrnka:
suggestion for issue #723

### DIFF
--- a/protocols/v2/noise-sv2/src/handshake.rs
+++ b/protocols/v2/noise-sv2/src/handshake.rs
@@ -109,8 +109,13 @@ pub trait HandshakeOp<Cipher: AeadCipher>: CipherState<Cipher> {
     fn ecdh(private: &[u8], public: &[u8]) -> [u8; 32] {
         let private = SecretKey::from_slice(private).expect("Wrong key");
         let x_public = XOnlyPublicKey::from_slice(public).expect("Wrong key");
-        let res = SharedSecret::new(&x_public.public_key(crate::PARITY), &private);
-        res.secret_bytes()
+        let ec_point_x_y = secp256k1::ecdh::shared_secret_point(
+            &x_public.public_key(secp256k1::Parity::Even),
+            &private,
+        );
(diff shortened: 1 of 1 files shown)


Sjors:
Doing this on the Bitcoin Core side requires accessing lower level libsecp functions, which might raise some review eyebrows.

Sjors:
I would have to use `CKey::UnhashedECDH` that's introduced in https://github.com/bitcoin/bitcoin/pull/28122. That will probably slow down review by a lot.

Fi3:
Do you think that could make sense/is possible. To modify the spec in order to avoid it?

Sjors:
@Fi3 using EllSwift avoids this problem. Bitcoin Core has a nice high-level method for it, already used for BIP324 encryption.

Fi3:
Ok so we can wait the decision on EllSwift. If we go with it this PR get closed otherwise we merge it

jakubtrnka:
Generally I disagree that the specification should be modified according to the needs of a specific codebase (speaking of the bitcoin-core that @Sjors mentioned).
The concept suggested in this PR is either right or wrong: 
1. If it is right, it should be adopted because this was the core idea that makes the use of XOnlyPublicKey consistent.
2. If it is wrong, I'd like to hear and understand why.

But  let's wait for the Ellswift discussions

Sjors:
> I disagree that the specification should be modified according to the needs of a specific codebase

That's true, but I think there is a good reason Bitcoin Core is (probably) reluctant to add this approach. It's a non-standard way of using a cryptographic function.

lorbax:
According to what we've discussed in the [last call ](https://docs.google.com/document/d/1lkK3o-hmKeqJuHzC_GjJ778GFfQV_ObH3q6KDVgXqOE/edit#heading=h.f9ii5u3x40f4)
this solution won't be adopted. The solution with more consensus was ElligatorSwift and its ecdh procedure in libsecp256k1 and changing the specs accordingly.

PR #725, opened 2024-01-21
Clippy on all worskpaces at once

lorbax:
If we want to compile all the workspaces, we must do it for each separately. This is annoying and not practical.
This PR adds a small script that compiles all the workspaces sequentially and return an error if there are warnings or error


Fi3:
seems very helpful. Can you add also fmt and test (with prop-test feature) ?

lorbax:
yes I can

lorbax:
> seems very helpful. Can you add also fmt and test (with prop-test feature) ?

Added cargo fmt and cargo test. Does it sound good?

Fi3 (approved):

Issue #728, opened 2024-01-22
Incorrect order of arithmetic operations during handshake


jakubtrnka:
EncryptWithAd is defined in the noise-specification the following way:

```
EncryptWithAd(ad, plaintext): If k is non-empty returns ENCRYPT(k, n++, ad, plaintext). Otherwise returns plaintext.
```
That means the `n` is incremented **after the ENCRYPT operation has been performed**.

SRI, however performs this prior to the ENCRYPT (or DECRYPT) operations:
https://github.com/stratum-mining/stratum/blob/3f07a6a747ab69c1594a93fdc563f06f72d716d8/protocols/v2/noise-sv2/src/cipher_state.rs#L66

https://github.com/stratum-mining/stratum/blob/3f07a6a747ab69c1594a93fdc563f06f72d716d8/protocols/v2/noise-sv2/src/cipher_state.rs#L45

That results in handshake failure.

references:

https://noiseprotocol.org/noise.html#the-cipherstate-object

https://github.com/stratum-mining/sv2-spec/blob/main/04-Protocol-Security.md#441-cipherstate-object

Sjors:
We should also clarify this in the spec.

pavlenex:
@Sjors Should we log an issue in specs repo not to forget to clarify this a bit better in specs?

Sjors:
@pavlenex yes, you could even just make a pull request that adds the sentence from the Noise paper:

> The ++ post-increment operator applied to n means "use the current n value, then increment it"

pavlenex:
Done in #729

PR #729, opened 2024-01-22
Fix noise-handshake operations order

jakubtrnka:
This fixes issue #728

### DIFF
--- a/protocols/v2/noise-sv2/src/cipher_state.rs
+++ b/protocols/v2/noise-sv2/src/cipher_state.rs
@@ -42,8 +42,8 @@ where
         ad: &[u8],
         data: &mut T,
     ) -> Result<(), aes_gcm::Error> {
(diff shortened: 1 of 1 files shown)


jakubtrnka:
Unit tests for the correctness of each step should be added.

Fi3:
@Sjors as soon as we have it on the TP we can merge

Sjors:
A test case would indeed be very useful here, if only so that I can add it to the TP in Bitcoin Core and make sure the issue is fixed.

Sjors:
Implemented the same on the Bitcoin Core side. Untested.
https://github.com/Sjors/bitcoin/pull/33

We should coordinate the merge timing for this.

Sjors:
tACK 729cdecc0d3ae1bddeba9f118e436e5b3806da79

Tested against the Bitcoin Core branch.

Issue #730, opened 2024-01-23
Improve error message when JDS can't connect to a local TP


pavlenex:
For JDS a local Template Provider needs to run first otherwise this happens:

```2024-01-23T09:21:59.598808Z  INFO jd_server: Jds INITIALIZING with config: "jds-config.toml"
2024-01-23T09:21:59.599341Z  INFO jd_server::lib::job_declarator: JD INITIALIZED
thread 'tokio-runtime-worker' panicked at jd-server/src/lib/mempool/mod.rs:60:77:
called `Result::unwrap()` on an `Err` value: JsonRpc(Transport(SocketError(Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at jd-server/src/lib/mempool/mod.rs:75:14:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(10), ...)
```

We should ensure a better error message is presented to the user. Problem is that JDS can't connect to TP to local TP for a few reasons, and we have no way apparently to know what are these reasons  (TP not fully synced, TP down).

Suggested message could be: `Unable to connect to TP (possible reasons: not fully synced, down)`

GitGab19:
I would add that another issue is that the JDS mempool is updated every 10 seconds, which are hardcoded [here](https://github.com/stratum-mining/stratum/blob/3f07a6a747ab69c1594a93fdc563f06f72d716d8/roles/jd-server/src/main.rs#L183).
The update rate should be configurable by the user. 
I'm working on fixing both issues, I will open a PR soon

GitGab19:
> I would add that another issue is that the JDS mempool is updated every 10 seconds, which is hardcoded [here](https://github.com/stratum-mining/stratum/blob/3f07a6a747ab69c1594a93fdc563f06f72d716d8/roles/jd-server/src/main.rs#L183). The update rate should be configurable by the user. I'm working on fixing both issues, I will open a PR soon

I fixed them here: https://github.com/stratum-mining/stratum/pull/727

PR #733, opened 2024-01-25
Replaces PR #727

### LINKED ISSUE #712: Pool panics after receiving an UpdateChannel message
Pool panics after an` UpdateChannel` message is received, and takes to a `TargetError(NegativeInput)`.
It seems that something should be implemented in the code yet.
The line of interest is [roles/pool/src/lib/mining_pool/mod.rs:192:17](https://github.com/stratum-mining/stratum/blob/d5659cb8a934a0dc1e76bd7930bd02098e7e395f/roles/pool/src/lib/mining_pool/mod.rs#L192).

Here's a screenshot of the error, for a better context:
Screenshot 2024-01-08 at 11 33 46

GitGab19:
Closes #712, #730


protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -1494,6 +1494,19 @@ impl ProxyExtendedChannelFactory {
...
             .0;
+
+        if referenced_job.job_id != m.job_id {
+            let error = SubmitSharesError {
+                channel_id: m.channel_id,
+                sequence_number: m.sequence_number,
+                error_code: SubmitSharesError::invalid_job_id_error_code()
+                    .to_string()
+                    .try_into()
+                    .unwrap(),
```
Fi3, line 1505: add a comment that exaplain why the unwrap is safe

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -244,21 +243,33 @@ impl Downstream {
                     || (hashrate_delta_percentage >= 30.0) && (delta_time >= 240)
                     || (hashrate_delta_percentage >= 15.0) && (delta_time >= 300)
                 {
-                if realized_share_per_min < 0.01 {
+                if realized_share_per_min == 0.0 {
```
Fi3, line 246: this is a float are you sure that is safe to change from < to == 

If yes write a comment that explain why

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
plebhash: I agree.

Going forward, IMO we should try to keep some discipline around the usage of `unwrap`, because we are getting some criticism from the community. 

for example: https://github.com/stratum-mining/stratum/issues/723#issuecomment-1904233521

PR #742, opened 2024-02-02
Fix incorrect order of encrypt/decrypt and nonce incremental

pavlenex:
Since #729 was on `dev` rebasing it to `main`


Sjors:
LGTM, looks like the same commit.

PR #747, opened 2024-02-12
EmptyMempool-error better management and bug fix (#744) 

### LINKED ISSUE #744: Job Declarator server fails with EmptyMempool
I run a custom signet, so sometimes the mempool is actually empty.

```
     Running `/home/sjors/dev/stratum/roles/target/debug/jd_server -c /home/sjors/.stratum/signet-jd-server.toml`
2024-02-06T11:36:50.014818Z  INFO jd_server: Jds INITIALIZING with config: "/home/sjors/.stratum/signet-jd-server.toml"
2024-02-06T11:36:50.014873Z  INFO jd_server::lib::job_declarator: JD INITIALIZED
2024-02-06T11:36:50.015496Z ERROR jd_server: EmptyMempool
2024-02-06T11:36:50.015507Z ERROR jd_server: Unable to connect to Template Provider (possible reasons: not fully synced, down)
2024-02-06T11:36:50.015528Z ERROR jd_server: SHUTDOWN from Upstream: Mempool error: `EmptyMempool`
```

But when I create a transaction so that there's something in the mempool, I still the same error.

GitGab19:
Fix #744


roles/jd-server/src/main.rs
```diff
@@ -118,9 +118,20 @@ async fn main() {
...
-                    error!("{:?}", err);
-                    error!("Unable to connect to Template Provider (possible reasons: not fully synced, down)");
-                    handle_result!(sender_clone, Err(err));
+                    warn!("{:?}", err);
+                    match err {
+                        mempool::JdsMempoolError::EmptyMempool => {
+                            warn!("Template Provider is running, but its mempool is empty (possible reasons: you're testing in testnet, signet, or regtest)")
+                        }
+                        mempool::JdsMempoolError::NoClient => {
+                            error!("Unable to connect to Template Provider (possible reasons: not fully synced, down)");
```
Sjors, line 127: Maybe clarify that this is about the RPC connection, not the sv2 connection.

roles/jd-server/src/main.rs
```diff
@@ -118,9 +118,20 @@ async fn main() {
                 let updated_mempool =
                     mempool::JDsMempool::update_mempool(mempool_cloned_.clone()).await;
                 if let Err(err) = updated_mempool {
-                    error!("{:?}", err);
-                    error!("Unable to connect to Template Provider (possible reasons: not fully synced, down)");
-                    handle_result!(sender_clone, Err(err));
+                    warn!("{:?}", err);
+                    match err {
+                        mempool::JdsMempoolError::EmptyMempool => {
+                            warn!("Template Provider is running, but its mempool is empty (possible reasons: you're testing in testnet, signet, or regtest)")
```
Sjors, line 124: This works, but it's logged every second.

roles/jd-server/src/main.rs
GitGab19: > Maybe clarify that this is about the RPC connection, not the sv2 connection.

Do you have a suggestion for that?

roles/jd-server/src/main.rs
GitGab19: Rate is determined by the `[mempool_update_timeout]` parameter set in config

GitGab19: Ah, now I remember you already suggested a change in this parameter name!
Don't know where to find that suggestion btw :D

Sjors: Polling every second is fine, but the warning should just appear once, otherwise the log becomes too noisy.

Fi3: you could log it only once using something like `let mut empty_mempool_logged = false`

Sjors: Or you could make it more noisy on mainnet only.

GitGab19:
@Sjors I applied your suggestions by changing the error log (`"Unable to establish RPC connection with Template Provider (possible reasons: not fully synced, down)"`). I also added a way to log the EmptyMempool warning every 10 / mempool_update_interval.as_secs() cycles of the loop.
For example, with mempool_update_interval = 10secs in config, the warning should be displayed every 10 cycles (or 10 seconds)

roles/jd-server/src/main.rs
```diff
@@ -149,27 +107,44 @@ async fn main() {
         username,
         password,
     )));
-    let mempool_update_timeout = config.mempool_update_timeout;
+    let mempool_update_interval = config.mempool_update_interval;
     let mempool_cloned_ = mempool.clone();
     let (status_tx, status_rx) = unbounded();
     let sender = status::Sender::Downstream(status_tx.clone());
+    let empty_mempool_log_counter = 10 / mempool_update_interval.as_secs();
```
Sjors, line 114: How about (in pseudo-code):

```
let last_empty_mempool = now();
```

and then below 

```
if last_empty_mempool < 1.minute.ago() {
  log...
  last_empty_mempool = now()
}
```

GitGab19: Do you mean to log the warning every minute, right?
If so, I can push this last changes

Sjors: Not every minute: no more than once a minute. If the interval is set to more than 1 minute it would log every interval.

roles/jd-server/src/main.rs
```diff
@@ -111,11 +111,10 @@ async fn main() {
     let mempool_cloned_ = mempool.clone();
     let (status_tx, status_rx) = unbounded();
     let sender = status::Sender::Downstream(status_tx.clone());
-    let empty_mempool_log_counter = 10 / mempool_update_interval.as_secs();
+    let mut last_empty_mempool_warning = std::time::Instant::now();
```
Sjors, line 114: I think you want to set this in the past, otherwise it won't warn the first time.

Sjors:
Looks good, but I didn't re-test.

I suggest squashing (most) commits.

GitGab19:
> Looks good, but I didn't re-test.
> 
> I suggest squashing (most) commits.
> 
> (and renaming the PR to something more descriptive)

Done

Sjors:
Lgtm. Will retest later.

Fi3 (approved):
LGTM

PR #748, opened 2024-02-13
Update: translator do not accept bigger than 64K bytes lines from downstream

Fi3:
Update the translator so that it do not accept any line from downstream but only the ones shorter then 2^16 bytes. Doing that it remove a possible vector for a dos attack. It must be said that translator is supposed to be used by miner in their local network with their mining device so I'm not sure how much useful this patch can be.

Unfortunately `async_std` do not have the equivalent of `tokio_util::codec::LineCodec` (not sure if out there there is something that can be used) so i had to implement the logic manually. Should be optimized enough. If we want better performance we can open a new PR.   

EDIT:

didn't notice PR #429 it seems that we can just use `LineCodec` from `tokio_util` I will remove the manual implementation and use that one, and force push it.

### DIFF
--- a/roles/Cargo.lock
+++ b/roles/Cargo.lock
@@ -113,6 +113,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-compat"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f68a707c1feb095d8c07f8a65b9f506b117d30af431cab89374357de7c11461b"
+dependencies = [
+ "futures-core",
+ "futures-io",
+ "once_cell",
+ "pin-project-lite",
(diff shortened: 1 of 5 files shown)


Sjors:
> It must be said that translator is supposed to be used by miner in their local network with their mining device so I'm not sure how much useful this patch can be.

You never know how people use this, so seems reasonable to make it safer.

pavlenex:
@GitGab19 as agreed this one is up to you to take a quick look and then @Fi3 can merge it.

GitGab19 (approved):

GitGab19:
> @GitGab19 as agreed this one is up to you to take a quick look and then @Fi3 can merge it.

Done, I can merge it after all tests pass again

Issue #717, opened 2024-01-12
Incorrect certificate validation


Sjors:
Currently the pool role (and presumably other roles) incorrectly verifies the Template Provider authority key (and presumable other upstream keys).

https://github.com/stratum-mining/sv2-spec/blob/main/04-Protocol-Security.md#453-server-authentication

The checks happen here:

https://github.com/stratum-mining/stratum/blob/3f07a6a747ab69c1594a93fdc563f06f72d716d8/protocols/v2/noise-sv2/src/signature_message.rs#L26-L40

1. The spec requires that `server_public_key` (the upstream static key) is included in the signature hash, but `m` only covers the 10 bytes of the `SIGNATURE_NOISE_MESSAGE`. It needs to append the server static public.

2. The message should be signed by the authority key, but it's actually being checked against the static key.

The Template Provider in Bitcoin Core generated signatures with the same two issues. I fixed the bug, but then put in a workaround to re-introduce the bug. That way things keep working.

A fix for the Bitcoin Core side is here: https://github.com/Sjors/bitcoin/pull/29

I'll merge that once the SRI-side is fixed.

pavlenex:
@plebhash is picking this one 🔥

pavlenex:
It seems I can only assign you an issue if you comment back on it 😄 We have to add you to organizaiton or repo.

plebhash:
sounds good! just send me an invite and I'll gladly accept it

plebhash:
@Sjors while I'm still not at a testing stage, it seems I'll eventually need to test SRI against your Bitcoin Core fork on a specific commit.

I'm looking at https://github.com/Sjors/bitcoin/pull/29 and your last comment says:
> I included this into the main PR, with a revert commit.

Could you confirm this is the main PR? https://github.com/bitcoin/bitcoin/pull/29346

I'm not finding the revert commit, thus the confusion.

plebhash:
Another question to @Sjors:

I built `2024/01/sv2_noise` but I'm running into this while trying to start it:
```
./src/bitcoind -sv2 -sv2port=8442 -debug=sv2 -datadir=$HOME/.bitcoin/
Error: Error parsing command line arguments: Invalid parameter -sv2
```

plebhash:
question to @Fi3 :

as stated by @Sjors:

> The message should be signed by the authority key, but it's actually being checked against the static key.

So that means that I need to fix this line, correct?

[...]

Issue #746, opened 2024-02-12
Make authority public key optional?


Sjors:
It's easier to get started if the `upstream_authority_public_key` field is not mandatory. In that case we don't check the certificate.

Perhaps on mainnet it's safer to keep this mandatory though.

See also: https://github.com/lorbax/stratum/pull/3#discussion_r1485953807

Sjors:
Yes this is part of https://github.com/stratum-mining/stratum/pull/752.

PR #737, opened 2024-01-29
Elligator swift encoding

lorbax:
ElligatorSwift encoding for pubkeys in handshake operations.
This branch must be tested with thev [new TP](https://github.com/Sjors/bitcoin/tree/2024/02/sv2-poll-ellswift) by @Sjors.

Already tested in regtest. Needs to be tested in testnet.

PR #752, opened 2024-02-14
Fix certificate verification

plebhash:
replacement for https://github.com/lorbax/stratum/pull/3

close https://github.com/stratum-mining/stratum/issues/717 and https://github.com/stratum-mining/stratum/issues/746

### DIFF
--- a/protocols/v2/noise-sv2/src/initiator.rs
+++ b/protocols/v2/noise-sv2/src/initiator.rs
@@ -230,47 +230,25 @@ impl Initiator {
             .0
             .serialize();
         let rs_pk_xonly = XOnlyPublicKey::from_slice(&rs_pub_key).unwrap();
-        match &self.responder_authority_pk {
-            Some(responder_authority_pk) => {
-                if signature_message.verify(&rs_pk_xonly, responder_authority_pk) {
-                    let (temp_k1, temp_k2) = Self::hkdf_2(self.get_ck(), &[]);
-                    let c1 = ChaCha20Poly1305::new(&temp_k1.into());
-                    let c2 = ChaCha20Poly1305::new(&temp_k2.into());
-                    let c1: Cipher<ChaCha20Poly1305> = Cipher::from_key_and_cipher(temp_k1, c1);
-                    let c2: Cipher<ChaCha20Poly1305> = Cipher::from_key_and_cipher(temp_k2, c2);
-                    self.c1 = None;
-                    self.c2 = None;
-                    let mut encryptor = GenericCipher::ChaCha20Poly1305(c1);
-                    let mut decryptor = GenericCipher::ChaCha20Poly1305(c2);
-                    encryptor.erase_k();
-                    decryptor.erase_k();
-                    let codec = crate::NoiseCodec {
-                        encryptor,
-                        decryptor,
-                    };
-                    Ok(codec)
-                } else {
-                    Err(Error::InvalidCertificate(plaintext))
-                }
-            }
-            None => {
-                let (temp_k1, temp_k2) = Self::hkdf_2(self.get_ck(), &[]);
-                let c1 = ChaCha20Poly1305::new(&temp_k1.into());
-                let c2 = ChaCha20Poly1305::new(&temp_k2.into());
-                let c1: Cipher<ChaCha20Poly1305> = Cipher::from_key_and_cipher(temp_k1, c1);
-                let c2: Cipher<ChaCha20Poly1305> = Cipher::from_key_and_cipher(temp_k2, c2);
-                self.c1 = None;
-                self.c2 = None;
-                let mut encryptor = GenericCipher::ChaCha20Poly1305(c1);
-                let mut decryptor = GenericCipher::ChaCha20Poly1305(c2);
-                encryptor.erase_k();
-                decryptor.erase_k();
-                let codec = crate::NoiseCodec {
-                    encryptor,
-                    decryptor,
-                };
-                Ok(codec)
-            }
+        if signature_message.verify(&rs_pk_xonly, &self.responder_authority_pk) {
+            let (temp_k1, temp_k2) = Self::hkdf_2(self.get_ck(), &[]);
+            let c1 = ChaCha20Poly1305::new(&temp_k1.into());
+            let c2 = ChaCha20Poly1305::new(&temp_k2.into());
+            let c1: Cipher<ChaCha20Poly1305> = Cipher::from_key_and_cipher(temp_k1, c1);
+            let c2: Cipher<ChaCha20Poly1305> = Cipher::from_key_and_cipher(temp_k2, c2);
+            self.c1 = None;
+            self.c2 = None;
+            let mut encryptor = GenericCipher::ChaCha20Poly1305(c1);
+            let mut decryptor = GenericCipher::ChaCha20Poly1305(c2);
+            encryptor.erase_k();
+            decryptor.erase_k();
+            let codec = crate::NoiseCodec {
+                encryptor,
+                decryptor,
+            };
+            Ok(codec)
+        } else {
+            Err(Error::InvalidCertificate(plaintext))
         }
     }
 
--- a/protocols/v2/noise-sv2/src/initiator.rs
+++ b/protocols/v2/noise-sv2/src/initiator.rs
@@ -31,7 +31,7 @@ pub struct Initiator {
     e: Keypair,
     // upstream pub key
     #[allow(unused)]
-    responder_authority_pk: XOnlyPublicKey,
+    responder_authority_pk: Option<XOnlyPublicKey>,
     c1: Option<GenericCipher>,
     c2: Option<GenericCipher>,
 }
@@ -96,10 +96,14 @@ impl Initiator {
     pub fn from_raw_k(key: [u8; 32]) -> Result<Box<Self>, Error> {
         let pk =
             secp256k1::XOnlyPublicKey::from_slice(&key).map_err(|_| Error::InvalidRawPublicKey)?;
-        Ok(Self::new(pk))
+        Ok(Self::new(Some(pk)))
     }
 
-    pub fn new(pk: XOnlyPublicKey) -> Box<Self> {
+    pub fn without_pk() -> Result<Box<Self>, Error> {
+        Ok(Self::new(None))
+    }
+
+    pub fn new(pk: Option<XOnlyPublicKey>) -> Box<Self> {
         let mut self_ = Self {
             handshake_cipher: None,
             k: None,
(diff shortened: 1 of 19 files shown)


plebhash:
still missing:

- [ ] add the possibility to leave the auth pub key for the TP in the pool set to None (void string).
- [ ] make it mandatory on mainnet

Sjors:
> add the possibility to leave the auth pub key for the TP in the pool set to None (void string)

No need to do that in this PR.

This PR depends on #737. It can be rebased once that's merged.

Sjors (approved):
ACK 

Tested against `sjors/sv2` with commit e477666694954d6a011dcd5e3188851651a4e011 reverted.

Just ping me when this is merged so I can update `sjors/sv2` to match.

Sjors:
Just tested that 3b5cd4589088a72df2bc93110b782917bdf71499 still works against my `sv2` branch. Hopefully CI can be tamed so we can merge this...

Fi3:
> > add the possibility to leave the auth pub key for the TP in the pool set to None (void string)
> 
> No need to do that in this PR.
> 
> This PR depends on #737. It can be rebased once that's merged.

We need it in that PR otherwise after merging it will have a broken main, and it will become very hard to run the roles and test things out

Fi3:
Also note that some MG test need an hosted TP if the hosted TP go down and we need to restart it we would also need to change the config files of the tested roles to put the new key it could very easly becom a nightmare

protocols/v2/noise-sv2/src/initiator.rs
```diff
@@ -226,25 +230,47 @@ impl Initiator {
...
+                        encryptor,
+                        decryptor,
+                    };
+                    Ok(codec)
+                } else {
+                    Err(Error::InvalidCertificate(plaintext))
+                }
+            }
+            None => {
+                let (temp_k1, temp_k2) = Self::hkdf_2(self.get_ck(), &[]);
```
Sjors, line 257: This should be de-duplicated (in a followup).

Sjors (approved):
ACK 3af5a042df093aa1060ffe8873777ddb2de09be5

Don't worry about making mainnet mandatory.

I tested that it can connect to the Template Provider:
* with `tp_authority_public_key` set
* without `tp_authority_public_key` (new since my last review)

I also checked that if I used the wrong public key, it refuses.


CI no longer checks the TP certificate. That's understandable given that it needs to be manually updated, but it could cause a future regression to missed. Though as long as _someone_ uses it, we'll learn quickly enough.

protocols/v2/noise-sv2/src/initiator.rs
```diff
@@ -226,25 +230,47 @@ impl Initiator {
...
-            let codec = crate::NoiseCodec {
-                encryptor,
-                decryptor,
-            };
-            Ok(codec)
-        } else {
-            Err(Error::InvalidCertificate(plaintext))
+        match &self.responder_authority_pk {
+            Some(responder_authority_pk) => {
+                if signature_message.verify(&rs_pk_xonly, responder_authority_pk) {
```
lorbax, line 235: I think that it would be better to move the match inside the `.verify()` method, uin order to have a cleaner step2 code. Something like
`
    pub fn verify(self, pk_: &Option<XOnlyPublicKey>) -> bool {
        match pk_ {
            Some(pk) => {
                let now = SystemTime::now()
                    .duration_since(SystemTime::UNIX_EPOCH)
                    .unwrap()
                    .as_secs() as u32;
                if self.valid_from <= now && self.not_valid_after >= now {
                    let secp = Secp256k1::verification_only();
                    let (m, s) = self.split();
                    let m = Message::from_hashed_data::<sha256::Hash>(&m[0..10]);
                    let s = match Signature::from_slice(&s) {
                        Ok(s) => s,
                        _ => return false,
                    };
                    secp.verify_schnorr(&s, &m, pk).is_ok()
                } else {
                    false
                }
            },
            None => true
        }
    }
`

protocols/v2/noise-sv2/src/initiator.rs
lorbax: I wrote a comment that is equivalent to the yours, sorry

protocols/v2/noise-sv2/src/initiator.rs
plebhash: done https://github.com/stratum-mining/stratum/commit/2591a9bd2a3f47bbe27ac2ddf3fd9b13f42cfc85

roles/jd-client/src/lib/template_receiver/mod.rs
```diff
@@ -62,8 +62,11 @@ impl TemplateRx {
             .expect("Invalid coinbase output in config");
         let stream = tokio::net::TcpStream::connect(address).await.unwrap();
 
-        let pub_key: Secp256k1PublicKey = authority_public_key;
-        let initiator = Initiator::from_raw_k(pub_key.into_bytes()).unwrap();
+        let initiator = match authority_public_key {
```
lorbax, line 65: Here you can turn `authority_public_key: Option<Secp256k1PublicKey>` in a variable `authority_public_key_x_only: Option<XOnlyPublicKey>` and initialize the initiator with `Initiator::new(autority_public_key_x_only)`. I understand that you didn't write the whole code, but perhaps the methods `Initiator::from_raw_k()` and `Initiator::without_pk()` can be eliminated.

roles/pool/config-examples/pool-config-local-tp-example.toml
```diff
@@ -24,5 +24,8 @@ pool_signature = "Stratum v2 SRI Pool"
 # Template Provider config
 # Local TP (this is pointing to localhost so you must run a TP locally for this configuration to work)
 tp_address = "127.0.0.1:8442"
+# You'll need to get tp_authority_public_key from the logs of your TP, for example:
```
lorbax, line 27: should it be like this here?
`tp_authority_public_key = ""`
Also, the following comment "You'll need to get tp_authority_public_key from the logs of your TP" is against what I understood to have agreed: if you run a local TP yuu clearly trust it and there is no need of certificate authority (and no need to retrive that from TP logs)

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -43,13 +43,17 @@ impl TemplateRx {
...
         coinbase_out_len: u32,
-        authority_public_key: Secp256k1PublicKey,
+        expected_tp_authority_public_key: Option<Secp256k1PublicKey>,
     ) -> PoolResult<()> {
         let stream = TcpStream::connect(address).await?;
         info!("Connected to template distribution server at {}", address);
 
-        let pub_key: Secp256k1PublicKey = authority_public_key;
-        let initiator = Initiator::from_raw_k(pub_key.into_bytes())?;
+        let initiator = match expected_tp_authority_public_key {
```
lorbax, line 51: same thing of the [previous comment](https://github.com/stratum-mining/stratum/pull/752/files#r1492369182)

roles/pool/config-examples/pool-config-local-tp-example.toml
Sjors: In this example the TP is local, but it doesn't have to be. Maybe change "You'll need to" to "You may" and mention that it's not necessary for a TP on localhost.

Sjors (approved):
Tested that eff00b34b4e0ade1b78dd3b1c28062cdf21aa0ba still works.

roles/jd-client/src/lib/template_receiver/mod.rs
plebhash: while I agree this would be a desirable improvement to the code, this would imply refactoring multiple parts of the codebase which don't really have much to do with the context of this PR

so in order to keep PR atomicity, I created this issue so we can implement this improvement in the future:
https://github.com/stratum-mining/stratum/issues/760

roles/pool/config-examples/pool-config-local-tp-example.toml
plebhash: made some changes in https://github.com/stratum-mining/stratum/pull/752/commits/1a48fa563fe70bea4c9f3dc34b7e7ecf1d9254c3

lmk what you guys think

Fi3: Not sure if it make sense as a comment. You know the auth pub key if you see it in the log and TP is local it means that you set the TP with the auth pub key so you have to know it. I think that this comment is can confuse the reader.

protocols/v2/noise-sv2/src/signature_message.rs
```diff
@@ -24,27 +24,35 @@ impl From<[u8; 74]> for SignatureNoiseMessage {
...
-            let s = match Signature::from_slice(&s) {
-                Ok(s) => s,
-                _ => return false,
-            };
-            secp.verify_schnorr(&s, &m, pk).is_ok()
-        } else {
-            false
+    pub fn verify(self, pk: &XOnlyPublicKey, authority_pk: &Option<XOnlyPublicKey>) -> bool {
+        match authority_pk {
+            Some(authority_pk) => {
```
Fi3, line 29: not need to change it. just to let you know, in this case idiomatic rust would be to use 
```rust
if let Some(auth_pk) = auth_pk {
   ....
}
```

roles/pool/config-examples/pool-config-local-tp-example.toml
Sjors: >  it means that you set the TP with the auth pub key so you have to know it.

The TP generates its authority key automatically, so it's not true that the user set it.

Fi3: yep but that is not how it should work? we decided to have a way to communicate the key to the process via config or whatever,
Having  the TP generating it do not make any sense to me

Sjors: But then the user needs a separate tool to generate an authority key. It seems easier to generate one automatically.

I think only a mining pool needs a way to use the same authority key for both the template provider and pool. My plan is to support that use case by adding a `-sv2cert` flag where the user can pass in the certificate. The TP won't have to touch the authority private key. For that use case we also need a way to sign the certificate. This is why the TP prints its static key to the log.

lorbax: The authority key is used by some role in order to be trusted. A self generated certified authority key it like "trust me because I am certified by myself", which is a bit pointless.  Perhaps there are other use cases that I am not aware of.

Sjors: > A self generated certified authority key it like "trust me because I am certified by myself"

It's for the use case of someone else wanting to connect to my template provider.

protocols/v2/noise-sv2/src/signature_message.rs
plebhash: https://github.com/stratum-mining/stratum/pull/752/commits/b4a4387f647831f57ac8f75ce8ebdd8c8d46983a

Sjors (approved):
Untested ACK b4a4387f647831f57ac8f75ce8ebdd8c8d46983a

Let's leave "cosmetic" changes to a followup so we can merge this breaking change.

roles/pool/config-examples/pool-config-local-tp-example.toml
lorbax: If is the application itself that generates the authority key, then you have no way to find out that you are communicating with the correct endpoint. If the TP is third party, then you must ask the institution that hosts it the authority key. If it is not, than you run it yourself and you don't need any certificate.

I like the solution `-sv2cert`, with the authority key as input. If I understand it correctly, you must then communicate this key privately to every client that wants to connect to your TP, which is more or less what I wrote above.

Fi3: In the above scenario if Alice's bitcoind go down she must resend the key to Bob. I really can not imagine a scenario where one want to provide a TP service but is ok in having randomly generated auth key that change everytime that bitcoind is respwaned.

Sjors: > bitcoind go down she must resend the key to Bob.

Bitcoin Core stores the key. Only Alice deletes the data directory, and not have a backup, would she have to generate a new key.

Fi3: ok this make sense to me

lorbax: Yes but there is no need to retrieve is from the log every time. It is better to let decide the user to decide whenever to use a new authority key. I would say also that could be a little annoying to do it manually. I like the `-sv2cert` solution btw!

Fi3 (approved):

Sjors:
🎉 

I updated my `sv2` branch to reflect this change (i.e. to drop the workaround).

roles/pool/config-examples/pool-config-local-tp-example.toml
lorbax: Some messages didn't appear to me at the moment of writing my last message. The idea of generating the key and storing it seems good to me and I ack it. In order to reuse an older key (from a previous backup) we should still be able to pass it as an argument to the `bitcoind`, like the `-sv2cert` idea.

PR #754, opened 2024-02-14
RequestTransactionDataError better management in JDC

### LINKED ISSUE #709: JDC panics after receiving RequestTransactionDataError message
JDC panics after receiving a RequestTransactionDataError message.
The line of interest is [roles/jd-client/src/template_receiver/message_handler.rs:46:9](https://github.com/stratum-mining/stratum/blob/6fe31322603ae9a3917855aef17ee9e25f8aa407/roles/jd-client/src/template_receiver/message_handler.rs#L46).

For a better context, here's a screenshot of the error, we got it right after the last testnet block mined:
Screenshot 2024-01-03 at 15 20 55

GitGab19:
Fix #709


roles/jd-client/src/lib/template_receiver/message_handler.rs
```diff
@@ -43,6 +43,17 @@ impl ParseServerTemplateDistributionMessages for TemplateRx {
...
+            template_id: _m.template_id,
+            error_code: _m.error_code.into_static(),
+        };
+
+        match std::str::from_utf8(m.error_code.as_ref()).unwrap_or("unknown error code") {
+            "template-id-not-found" => Err(Error::NoValidTemplate),
+            "stale-template-id" => Ok(SendTo::None(Some(
+                TemplateDistribution::RequestTransactionDataError(m),
+            ))),
+            _ => Err(Error::NoValidTemplate),
```
Fi3, line 56: since we match string here would be useful to log which is the string that is not parsable. So i suggest to use somthing like `Error::NoValidTemplate(error-message)`

GitGab19: Done here 9693822b16924f744c31fc8d42fffed94f67776c

Fi3 (approved):
LGTM

Sjors:
Code looks reasonable to me, but I haven't tested.

PR #755, opened 2024-02-15
Fix releases

Fi3:


Fi3:
I force merge it on main cause it change only releases action and cargo version numbers. No code is touched and i need to merge it in main to check it the fixed release action do actually works

PR #761, opened 2024-02-16
new rpc-client for JD-Server mempool 

lorbax:
This is a new and rebased version of  [PR716](https://github.com/stratum-mining/stratum/pull/716)

What is in this PR.

A new Rpc client, called MiniRpcClient. This client communicates with a bitcoin node, and the requests that can do are just the following: ask for the mempool, ask for a specific transaction, submit a bitcoin block. This client is used to make the jd-server mempool communicate with the bitcoin node. The previous client was depending upon json-rpc library, which caused some issues because imported serde in a way that the SRI "doesn't like". This rpc client is made on top of the latest hyper 1.1.0 library for http connections.

The code is asynchronous and non blocking. Before was "almost" non blocking, as the function "submit_block" that concerned submitting a block to the bitcoin node, was blocking.

I also took advantage of introduce some proper error management in the jds mempool module.

This PR has been tested in regtest and needs to be tested in testnet.


Sjors (approved):
I tested this (on top of #752) against the TP on a custom Signet (using `-debug=rpc`).

I saw successful calls to `getrawmempool`.

I created a transaction to ensure that `getrawtransaction` is called.

I waited for it to find a block, which causes `submitblock` to be called (on one of the nodes, the other one got the solution via the  `SubmitSolution` sv2 message).

I _did not_ look at the code.

build-on-all-workspaces.sh
plebhash, line 1: is this meant to be used in some CI context?
or what is the motivation?

I'm not against adding this script, but I'm curious about why you wrote it.

plebhash:
IMO this PR should also add some documentation about how to interact with the new RPCs

plebhash:
why not use [`bitcoincore_rpc`](https://docs.rs/ord-bitcoincore-rpc/latest/bitcoincore_rpc/#)?

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -28,42 +32,57 @@ pub struct TransacrtionWithHash {
...
         }
     }
-    pub fn new(url: String, username: String, password: String) -> Self {
-        let auth = Auth::UserPass(username, password);
+
+    pub fn get_transaction_list(self_: Arc<Mutex<Self>>) -> Vec<Txid> {
+        let tx_list = self_.safe_lock(|x| x.mempool.clone()).unwrap();
+        let tx_list_: Vec<Txid> = tx_list.iter().map(|n| n.id).collect();
+        tx_list_
+    }
```
plebhash, line 55:
```suggestion
    }
    
```

plebhash (changes requested):
SRI has been criticized about our lack of discipline in the way we leave `TODO` comments on the code

for example: https://github.com/stratum-mining/stratum/issues/723#issuecomment-1903037506

I understand the notion of PR atomicity and I fully agree that sometimes there are things that maybe fall out of the scope of some specific PR, so I don't think writing `TODO` comments on a PR is a cardinal sin.

However, I believe we can raise the bar on how we manage these kind of practices. My suggestion is: for every `TODO` comment, open an issue with a meaningful description, and paste the link to the issue along with the `TODO` comment.

That way, future tasks are still manageable.

build-on-all-workspaces.sh
lorbax: This is just a script that runs cargo build in all the workspaces. It is very useful in a very preliminar phase of developing a feature. When the code is compiling and the feature kind of finished, I run `clippy-on-all-workspaces.sh` for running clippy.
I must say that I added `build-ol-all-workspaces.sh`n accidentally, but I thought "why not?" and I left it in the commit. It is not in the purpose of the PR, though. If it is inappropriate I will remove it.

lorbax:
> why not use [`bitcoincore_rpc`](https://docs.rs/ord-bitcoincore-rpc/latest/bitcoincore_rpc/#)?

Because it imported the json-rpc library, which in turn imported serde in a way that it is not compatible with the way it is imported ion the SRI. Moreover, import bitcoincore rpc  or just using vendored json-rpc is worse from a performance point of view

lorbax:
> SRI has been criticized about our lack of discipline in the way we leave `TODO` comments on the code
> 
> for example: [#723 (comment)](https://github.com/stratum-mining/stratum/issues/723#issuecomment-1903037506)
> 
> I understand the notion of PR atomicity and I fully agree that sometimes there are things that maybe fall out of the scope of some specific PR, so I don't think writing `TODO` comments on a PR is a cardinal sin.
> 
> However, I believe we can raise the bar on how we manage these kind of practices. My suggestion is: for every `TODO` comment, open an issue with a meaningful description, and paste the link to the issue along with the `TODO` comment.
> 
> That way, future tasks are still manageable in our roadmap and external eyes have a little bit of context on how the code is evolving.

Completely agree with that. I must say that the todo that I left are very marginal, by the way.
I will open an issue for that!

protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
```diff
@@ -158,9 +158,7 @@ where
             Ok(JobDeclaration::SubmitSolution(message)) => {
                 info!("Received SubmitSolution");
                 debug!("SubmitSolution: {:?}", message);
-                self_
-                    .safe_lock(|x| x.handle_submit_solution(message))
-                    .map_err(|e| crate::Error::PoisonLock(e.to_string()))?
+                Self::handle_submit_solution(self_, message)
```
GitGab19, line 161: Why did you treat it differently from all the other handlers?
There's no need to use safe_lock ?

build-on-all-workspaces.sh
```diff
@@ -0,0 +1,24 @@
...
+
+    echo "Running fmt on: $workspace"
+    (cd $workspace && cargo +nightly fmt)
+    if [ $? -ne 0 ]; then
+        echo "Fmt failed in: $workspace"
+        exit 1
+    fi
+done
+
+echo "build success, all tests passed!"
```
Fi3, line 22: the above script do not seem to execute any test, why you say all test passes?

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -100,6 +109,68 @@ impl JobDeclaratorDownstream {
...
+                                    txdata: tx_list.clone(),
+                                };
+
+                                block.header.merkle_root = block.compute_merkle_root().unwrap();
+
+                                let serialized_block = serialize(&block);
+                                let hexdata = hex::encode(serialized_block);
+
+                                let _ = submit_solution_sender.send(hexdata).await;
+                            }
```
GitGab19, line 173: Why did you move the SubmitSolution logic here, instead of managing it in the handlers as it is now?

roles/jd-server/Cargo.toml
```diff
@@ -35,4 +34,8 @@ key-utils = { version = "^1.0.0", path = "../../utils/key-utils" }
 secp256k1 = { version = "0.28.2", default-features = false, features =["alloc","rand","rand-std"] }
 siphasher = "1"
 hex = "0.4.3"
-
+base64 = "0.21.5"
+hyper = { version = "1.1.0", features = ["full"] }
+hyper-util = { version = "0.1", features = ["full"] }
+http-body-util = "0.1"
+bytes = "1"
```
Fi3, line 41: we usally use utils/buffer here double check if another dependency is really needed

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -28,42 +32,58 @@ pub struct TransacrtionWithHash {
...
+            let client = mini_rpc_client::MiniRpcClient::new(url.to_string(), self.auth.clone());
+            Some(client)
         } else {
             None
         }
     }
-    pub fn new(url: String, username: String, password: String) -> Self {
-        let auth = Auth::UserPass(username, password);
+
+    // This function is used only for debug purposes
```
GitGab19, line 51:
```suggestion
/// This function is used only for debug purposes and should not be used
/// in production code.
```

roles/jd-server/src/main.rs
```diff
@@ -115,43 +121,63 @@ async fn main() {
...
     if url.contains("http") {
-        let sender_clone = sender.clone();
+        let sender_update_mempool = sender.clone();
         task::spawn(async move {
             loop {
-                let updated_mempool =
+                let result: Result<(), mempool::error::JdsMempoolError> =
                     mempool::JDsMempool::update_mempool(mempool_cloned_.clone()).await;
-                if let Err(err) = updated_mempool {
+                if let Err(err) = result {
```
GitGab19, line 129: I would leave `updated_mempool` here (result is too much generic imo)

roles/jd-server/src/main.rs
```diff
@@ -115,43 +121,63 @@ async fn main() {
...
+                        _ => {
+                            mempool::error::handle_error(&err);
+                            handle_result!(sender_update_mempool, Err(err));
                         }
                     }
                 }
                 tokio::time::sleep(mempool_update_interval).await;
+                // GET TRANSACTION LIST, FOR DEBUG PURPOSES
+                //let _transactions =
+                //    mempool::JDsMempool::_get_transaction_list(mempool_cloned_.clone());
```
GitGab19, line 147: I would remove it

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -100,6 +109,68 @@ impl JobDeclaratorDownstream {
                                 Self::send(self_mutex.clone(), message).await.unwrap();
                             }
                             Ok(SendTo::None(_)) => (),
+                            Ok(SendTo::RelayNewMessage(JobDeclaration::SubmitSolution(
```
Fi3, line 112: if you move the logic from handle_submit_solution do not put it here but in a specific function. If possible add a test and prop test. Maybe in roles_logic_sv2? Or just leave it where it was and change it another time.

GitGab19 (commented):
Review done. 
I left some questions to clarify some doubts I have (especially in the `submit_solution_handler` part)

build-on-all-workspaces.sh
plebhash: I think all these scripts are very useful.

But yeah, maybe adding it on a separate PR is better for future reference.

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: because, as I remember, this would require to make the handler trait async, which is not trivial

lorbax: I try to move where it was before and see if I can fix

roles/jd-server/src/lib/mempool/mini_rpc_client.rs
```diff
@@ -0,0 +1,207 @@
+// TODO
```
Fi3, line 1: put all this mini-rpc client in a specific crate under utils

Fi3: Add some tests

build-on-all-workspaces.sh
lorbax: This is a mistake. This script is meant only for helping in an early stage of developing. I personally found it useful but I can remove it.

roles/jd-server/src/lib/mempool/mod.rs
Fi3: we have the comment and the _ in front so it should be sufficient but if possible add also a conditional compilation flag

Fi3: like keet it out of the binary if in release mode

build-on-all-workspaces.sh
lorbax: updated by the way, now it prints just  `build success!"

roles/jd-server/Cargo.toml
lorbax: I don't understand, can you explain a bit more?

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: I think this would require to make the handler trait async. Do we want to do this?

roles/jd-server/src/main.rs
```diff
@@ -102,10 +105,13 @@ async fn main() {
     let url = config.core_rpc_url.clone() + ":" + &config.core_rpc_port.clone().to_string();
     let username = config.core_rpc_user.clone();
     let password = config.core_rpc_pass.clone();
+    let (submit_solution_sender, submit_solution_receiver): (Sender<String>, Receiver<String>) =
+        unbounded();
```
Fi3, line 109: Is there a specific reason to not use a bounded channel?

roles/jd-server/src/main.rs
```diff
@@ -115,43 +121,63 @@ async fn main() {
...
-                        mempool::JdsMempoolError::NoClient => {
-                            error!("{:?}", err);
-                            error!("Unable to establish RPC connection with Template Provider (possible reasons: not fully synced, down)");
-                            handle_result!(sender_clone, Err(err));
-                        }
-                        mempool::JdsMempoolError::BitcoinCoreRpcError(_) => {
-                            error!("{:?}", err);
-                            error!("Unable to establish RPC connection with Template Provider (possible reasons: not fully synced, down)");
-                            handle_result!(sender_clone, Err(err));
+                        _ => {
```
Fi3, line 138: better have a branch for each error instead of _ if you have more error that should do the same thing you can use Error::One | Error::Two => {}

roles/jd-server/src/main.rs
lorbax: I modified please see if now it is suitable

roles/jd-server/src/lib/mempool/mod.rs
lorbax: I don't understand

roles/jd-server/src/lib/mempool/mini_rpc_client.rs
lorbax: will do

roles/jd-server/Cargo.toml
Fi3: are u sure that instead of bytes you can not use the lib crate under utils/buffer?

roles/jd-server/src/lib/job_declarator/mod.rs
Fi3: before it was not. Btw you have several option one is to have a dedicated function that you call from here not from the handler

lorbax: What do you mean with "before it was no"?

build-on-all-workspaces.sh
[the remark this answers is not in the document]
GitGab19: > we need nightly to run cargo fmt

Why? I always ran it locally with simply `cargo fmt`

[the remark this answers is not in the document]
Fi3: this is very strange in the config file we have:
```rust
imports_indent = "Block"
imports_layout = "Mixed"
imports_granularity = "Crate"
```
that use unstable features available only on nighlty

roles/jd-server/src/lib/mempool/mini_rpc_client.rs
lorbax: I moved the client within `utils` in the last force push. What tests do you suggest?

build-on-all-workspaces.sh
lorbax: > I think all these scripts are very useful.
> 
> But yeah, maybe adding it on a separate PR is better for future reference.

opened [PR765](https://github.com/stratum-mining/stratum/pull/765) for this script :-)

roles/jd-server/src/main.rs
lorbax: Ok.
Can you confirm that if create the channel like this
`let (submit_solution_sender, submit_solution_receiver): (Sender<String>, Receiver<String>) = bounded(3);`
then at most three `SubmitSolution` are sent? Sorry for the trivial question but better to ask before that getting in trouble after ^^

Fi3: no is the max number of message that can be "inside"  the channel check the documentation if is not clear

roles/jd-server/src/lib/job_declarator/mod.rs
Fi3: I mean that before your PR this logic was inside place  `handle_submit_solution` and you moved it here

roles/jd-server/src/lib/job_declarator/mod.rs
Fi3: there are sveral possible solution is moving where was before put that somewhere else but not here

protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
lorbax: Perhaps I don't understand your point. Can you be a bit more exahaustive? 
Here it must submit the block to the rpc client. We want that to be async, that would make the trait async. I opted to relaying the message to the caller JobDeclaratorDownstream::start, that is managed by a task in main.  This solution looked easier to me.

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: I moved as a method of JobDeclaratorDownastream. I could move it also to somehwere to roles-logic-sv2, I basically need the transaction list, the DeclareMiningJob and the SubmitSolutionJd message. I will do it tomorrow now I am tired. Let me know!

roles/jd-server/src/main.rs
lorbax: Do you like `update_mempool_result`? :-)

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: Now it relies on the function `roles-logic-sv2::submit_solution_to_block()`.

roles/jd-server/src/main.rs
lorbax: This have been addressed, please check if it looks fine

roles/jd-server/Cargo.toml
lorbax: I will open a PR for that later

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: added a method to JobDeclaratorDoenstream, that actually relies on a new function in `roles-logic-sv2/utils`
`
pub fn submit_solution_to_block( last_declare: DeclareMiningJob,  mut tx_list: Vec<bitcoin::Transaction>, message: SubmitSolutionJd) -> bitcoin::Block {...}
`

roles/jd-server/src/main.rs
lorbax: added `boubnded(10)` as it is done similarly to other parts of the codebase

roles/jd-server/src/main.rs
lorbax: done

protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
```diff
@@ -184,5 +182,8 @@ where
         &mut self,
         message: ProvideMissingTransactionsSuccess,
     ) -> Result<SendTo, Error>;
-    fn handle_submit_solution(&mut self, message: SubmitSolutionJd) -> Result<SendTo, Error>;
+    fn handle_submit_solution(
+        self_: Arc<Mutex<Self>>,
```
Fi3, line 186: why you change self_ from &mut self_ to Arc<Mutex<Self>> ??

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -180,62 +177,12 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
-            })
-            .unwrap();
+    fn handle_submit_solution(
+        _self: Arc<Mutex<Self>>,
+        message: SubmitSolutionJd<'_>,
+    ) -> Result<SendTo, Error> {
+        let m = JobDeclaration::SubmitSolution(message.clone().into_static());
 
-        Ok(SendTo::None(None))
+        Ok(SendTo::RelayNewMessage(m))
```
Fi3, line 186: not sure if this is semantically correct RelayNewMessage(m) meand relay a new sv2 message and the new message is the one contained inside (m) but this is not what you want to do

Fi3: This should be rather be SendTo::None(m)

Fi3: using relay new message with the same message received do not make much sense to me. It should be realy same message but this is not the case

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -67,6 +70,24 @@ impl JobDeclaratorDownstream {
...
+    fn get_block_hex(self_mutex: Arc<Mutex<Self>>, message: SubmitSolutionJd) -> String {
+        //TODO: implement logic for success or error
+        let (last_declare, tx_list, _) = match self_mutex
+            .safe_lock(|x| x.declared_mining_job.take())
+            .unwrap()
+        {
+            Some((last_declare, tx_list, _x)) => (last_declare, tx_list, _x),
+            None => {
+                warn!("Received solution but no job available");
+                todo!()
```
Fi3, line 82: here we can just close the connection with downstream

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -739,6 +741,48 @@ fn tx_hash_list_hash_builder(txid_list: Vec<bitcoin::Txid>) -> U256<'static> {
     hash.to_vec().try_into().unwrap()
 }
 
+pub fn submit_solution_to_block(
```
Fi3, line 744: can you use From<SubmitSolution> for Block here?

Fi3: also having one or two test for it wouldn't be bad

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -100,6 +125,16 @@ impl JobDeclaratorDownstream {
                                 Self::send(self_mutex.clone(), message).await.unwrap();
                             }
                             Ok(SendTo::None(_)) => (),
```
Fi3, line 127: I know that this was already like that but we are trying to make this code more stable and prod ready, so catch all should disappear if you see one from the prototype phase just remove it is few line of code

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -100,6 +125,16 @@ impl JobDeclaratorDownstream {
...
+                            Ok(SendTo::RelayNewMessage(JobDeclaration::SubmitSolution(
+                                message,
+                            ))) => {
+                                let hexdata = JobDeclaratorDownstream::get_block_hex(
+                                    self_mutex.clone(),
+                                    message,
+                                );
+
+                                let _ = submit_solution_sender.send(hexdata).await;
```
Fi3, line 136: we should at least log the the error here

roles/jd-server/src/lib/mempool/error.rs
```diff
@@ -0,0 +1,39 @@
+use rpc::mini_rpc_client::RpcError;
+use tokio::task::JoinError;
+use tracing::{error, warn};
+
+// TODO this should be includede in JdsError
```
Fi3, line 5: not clear what this comment say

roles/jd-server/src/lib/mempool/mod.rs
Fi3: why you don't use `#[cfg(debug_assertions)]`

roles/jd-server/src/lib/job_declarator/mod.rs
Fi3: In this case the catchall is dangerous is not like doing Error::ErrorKind(_) where you want to log it and fail for all the error of errorkind. In this case SendTo::None(m) is used for different thing so it is better to explicitly handle them, maybe we should rethink the SendTo type not sure, for now I would leave it like that and explicitly handle all the path

protocols/v2/roles-logic-sv2/src/handlers/job_declaration.rs
lorbax: because I made a mistake

roles/jd-server/src/lib/job_declarator/message_handler.rs
lorbax: Now the the handler returns a SendTo(None(Some(JobDeclaration::SubmitSolution())))

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: done

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: Yes but here in particular is very vague what to do. For example, I can track all the types of SendTo returned by the handler, handle them and report an error if arrives a different SendTo.

protocols/v2/roles-logic-sv2/src/utils.rs
lorbax: I don't think so, `From` trait has a non-compatible signature. Perhaps I don't understand you

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: changed, not there is a wrapper struct in roles-log::utils and From is implemented on that struct

roles/jd-server/src/lib/mempool/mod.rs
lorbax: done

roles/jd-server/src/lib/mempool/error.rs
lorbax: removed

protocols/v2/roles-logic-sv2/src/utils.rs
lorbax: done

roles/jd-server/src/lib/mempool/error.rs
lorbax: deleted

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: this function is now turned into a From for Block that takes a wrapper struct

roles/jd-server/src/lib/job_declarator/mod.rs
lorbax: done

utils/rpc/src/mini_rpc_client.rs
```diff
@@ -0,0 +1,206 @@
+// TODO
+//  - WHEN UPGRADING TO 1.0.1 Client is in hyper-utils:
+//    Struct hyper_util::client::legacy::Client
+//  - use https for security reasons
```
Sjors, line 4: The Bitcoin Core RPC does not support https. You should connect on the same machine (or via a secure tunnel). There's work in progress to support unix domain sockets: https://github.com/bitcoin/bitcoin/pull/27375

lorbax: Thank you for pointing this out!

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -739,6 +741,69 @@ fn tx_hash_list_hash_builder(txid_list: Vec<bitcoin::Txid>) -> U256<'static> {
...
+        let coinbase_suf = last_declare.coinbase_suffix.to_vec();
+        let mut path: Vec<Vec<u8>> = vec![];
+        for tx in &tx_list {
+            let id = tx.txid();
+            let id = id.as_ref().to_vec();
+            path.push(id);
+        }
+        let merkle_root =
+            merkle_root_from_path(&coinbase_pre[..], &coinbase_suf[..], &extranonce[..], &path)
+                .expect("Invalid coinbase");
```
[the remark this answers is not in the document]
lorbax: you are right but before I must make the code work. I leave this conversation open. When works I will go back and fix here!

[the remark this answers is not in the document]
lorbax: Still open issue

pavlenex (approved):

lorbax (commented):
addressed [this suggestion](https://github.com/stratum-mining/stratum/pull/761#pullrequestreview-1889290673) by @plebhash:
[https://github.com/stratum-mining/stratum/issues/782](https://github.com/stratum-mining/stratum/issues/782)


This PR we through three rounds of reviews, that are visible in the commit history

PR #766, opened 2024-02-20
docs releases

plebhash:
Refactoring of docs about releases.


pavlenex (approved):

Fi3:
we should also specify when to run the action. I would say every time that we do PR we need to bump the versions of the affected crates and then run the action in order to do a release. Keep in mine that at least for the lib crates if we try to do release without bumping the version the action will fail so is kinda safe. Of course if bump minor or major or patch is up to the dev but we have review for that.

plebhash:
> we should also specify when to run the action. I would say every time that we do PR we need to bump the versions of the affected crates and then run the action in order to do a release. Keep in mine that at least for the lib crates if we try to do release without bumping the version the action will fail so is kinda safe. Of course if bump minor or major or patch is up to the dev but we have review for that.

ack https://github.com/stratum-mining/stratum/pull/766/commits/0427297f1d2f87005ef2a3819e2337eeddbd5c87

RELEASE.md
```diff
@@ -1,66 +1,78 @@
...
-`--exclude [crate_1] --exclude [crate_2]` to the command. This is not an issue for publishing version updates since the crates are not NEW publishes.
+SRI has a global version, which uses git tags and keeps track of how the codebase evolves across time as a whole.
 
-### Current Successful Publishes
+All the other internal SRI crates also follow SemVer 2.0.0, but each crate version is independent of the global release version.
 
-Due to either github dependencies or a crate failing the build stage during publish not all crates are being published.
+Whenever a `PATCH` is introduced, it is applied to all the latest `MAJOR` releases.
+For example: imagine there's releases `v1.0.0`, `v1.1.0`, and `v2.0.0`. A bug is found, dating back all the way to `v1.0.0`.
+A patch is applied such that the following new tags are introduced: `v1.1.1` and `v2.0.1`.
```
Fi3, line 16: How do we do it? I guess that we should have a branch (or a tag?) for all the last x majior version of each libs.
This seems complicate. Another way could be to create the branch only when needed. We fix a bug for `lib1-2.0.0` we find that also `lib1-1.1.0` is affected and we want to backport the fix. We create a new branch from whatever was the state of the repo when lib1 have been go from v1 ti v2 and we apply the fix. If we do that only for last majior version, only when we have a bug that affect also the last majior version I do not think that we will end up with 'too many' branches, considering that most library in this repo will do not get a new majior version in many years. If this is a concern we can just say that we support the last majior version and that is not safe to use any version that is not the majior, or we can think about it when we will have more then one majior version (0 wont be maintained cause is unstable). Maybe we can just leave the various possibility here, for now we can say that we support only last majior so we do not make promise and when we will have more majior  we will think about it.

RELEASE.md
```diff
@@ -1,66 +1,78 @@
...
+be published are published.
 
-Versioning is handled by the `autorelease.yaml` workflow. The workflow will  will auto detect changes in crates and auto bump the patch versions 
-in the target crate, as well as any crate that uses the target crate as a dependency, if changes are detected. It will then push a new commit with the 
-versioning changes to the `bot/versioning` branch, and Lastly, it will auto create a PR (bot/versioning -> main). Since cargo-smart-release does not yet 
-support autodetection of MAJOR and MINOR changes with regards to Semver, any MAJOR/MINOR versioning will need to be manually changed in the package Cargo.toml.
+Running `cargo release` in the various workspace help to prepare the version number and everything.
 
-## Versioning Notes
+Every PR needs to increase the version of whatever crate it is touching. Otherwise, we will mess up the dependency chain of whoever is fetching from crates.io
```
Fi3, line 76: I would specify, every PR **in main**. PRs that add feature in dev are not going to increase the version number since they will be batched every [duration] into main. So lets say that when I do the release I add 5 feature to lib1 I will not increase the minor version of 5 but of 1.

So when we merge dev in main we increase the versions of the touched libs. When we merge a buge fix in main we increase the version of the touched libs. Whenever we merge something in main we run the releases actions

Fi3:
We should decide the interval between release and stick with it, 2 weeks? That means that we do not wait for feature to be ready, every 2 weeks we ship what we have on dev.

Also I would add something to automatically check the we do not mess with semver (it is very easy to think that something do not change the API but it change it). I found this linter https://crates.io/crates/cargo-semver-checks that we could add to our actions. That ofc do not remove the needs for human reviews is just an helper.

plebhash:
> We should decide the interval between release and stick with it, 2 weeks? That means that we do not wait for feature to be ready, every 2 weeks we ship what we have on dev.

ack 22e196e92c94ac9031ac34fd1556f5109e1e4a8c

RELEASE.md
plebhash: On lines [10 and 12](https://github.com/plebhash/stratum/blob/c4399a92b9e68af72d1954e97a1c40a2430d49c9/RELEASE.md?plain=1#L10-L12) of my proposal, there is a very important distiction we need to keep in mind:

```
SRI has a global version, which uses git tags and keeps track of how the codebase evolves across time as a whole.

Every internal SRI crate also follows SemVer 2.0.0, but each crate version is independent of the global release version.
```

Let's imagine this scenario, where both `crate-A` and `crate-B` are affected by a bug on the following versions:

| Global Release Version | v2.0.0 | v1.1.0 |
|------------------------|--------|--------|
| crate-A                | v1.0.0 | v0.9.0 |
| crate-B                | v2.0.0 | v2.0.0 |

Here, Global Release Versions would be indicated as git tags.
Crate versions however, would only be indicated by `Cargo.toml` and `crates.io`. Otherwise, we would end up with too many tags.

After we patch the bug, we would have the following:
| Global Release Version (new git tags) | v2.0.1 | v1.1.1 |
|------------------------|--------|--------|
| crate-A (only bump `Cargo.toml`               | v1.0.1 | v0.9.1 |
| crate-B (only bump `Cargo.toml`)               | v2.0.1 | v2.0.1 |

Fi3: Ok make sense. Only thing I wouldn't maintain version 0 and 1. I would just bump majior version of everyting to 2. And start from there 2 reasons here:
1. We didn't followed semver so if you assume semver for what we have now you risk to end up with not compatible libraries.
2. There several 0 library that now are 1 that at 0 were very very experimental and it do not make sense to to support them

An example of that is the noise crate.

Also is startum v2 so it can make sense to start from 2. Otherwise we should remove the old library from crates.io but I dont' know if is something possible or not.

RELEASE.md
```diff
@@ -1,66 +1,79 @@
...
+- `main`
+- `dev`
 
+We consider `main` to be the branch where the source code of `HEAD` always reflects a production-ready state.
 
-## Github Binary Releases
+We consider `dev` to be the branch where the source code of `HEAD` always reflects a state with the latest delivered development changes for the next release.
 
-# Versioning
+Every two weeks all the changes to `dev` are merged back into `main` and then tagged with a release number while bumping `MAJOR` and/or `MINOR`.
```
pavlenex, line 37:
```suggestion
Every two weeks all the changes to `dev` are merged back into `main` and then tagged with a release number while bumping `MAJOR` and/or `MINOR`.
```

After re-reading the PR the only thing I would be very careful is the pace of releases which we say is 2 weeks.
In startup/company structures this works, but my foss experience says we should probably be more conservative here no matter what type of release it is, I think even with 1-4 PR's it's obvious things can take quite a lot of time, especially if mainnet testing or any sort of testing is involved.

I would say we probablt shouldn't put a time-frame here, or we can put a 2-4weeks -  _until_ we do trial and error and see what kind of pace works for the structure/team we have here.

Trust me, 2-weeks is very tight in FOSS and can cause of lot of stress. As a reminder we're 8 months behind this release (ok it's major but just to give you a bit more realistic overview). What happens if we have PR's depending on each other and each of them requires testing (situation we had last 4-6 weeks).

FOSS structures are much loose and time-frames should adjust to that imo.

RELEASE.md
pavlenex: I agree with what you're saying @Fi3, the only thing that bugs me is starting with 2.0.0, afaik we didn't really have an official release yet, and in public we always referred to things as MVP and later update. 1.0.0 is really a nice clean milestone, I am worried 2.0.0 may communicate certain level of maturity and stability which I am not sure we have yet.

> Otherwise we should remove the old library from crates.io but I don't know if is something possible or not.

Doesn't sound like something impossible, I doubt there can be any consequences to this?

RELEASE.md
```diff
@@ -1,66 +1,79 @@
...
+
+- branch off from: `main`
+- merge back into: `main` + `dev`
+- naming convention: `patch-x`, where `x` describes the bug/patch
+
+# Releasing Roles Binaries
+
+The [release page of SRI repo](https://github.com/stratum-mining/stratum/releases) provides executable binaries for all SRI roles, targeting popular hardware architectures.
+
+The github binary releases of the roles are handled in the `release-bin.yaml` workflow.
```
pavlenex, line 59:
```suggestion
The GitHub binary releases of the roles are handled in the `release-bin.yaml` workflow.
```

pavlenex (commented):
Thanks for pushing this one through end ensuring there's clarity. I left 2 comments that I believe we should discuss and one minor typo suggestion.

RELEASE.md
plebhash: With regards to `2.0.0`, we should keep the **SRI vs SV2** distinction in mind.

Sure, SV2 implies v2, but only for the Specifications. 

That doesn't necessarily mean that as a Rust codebase, SRI should start at v2.

plebhash: With regards to crates.io removal, I found this:

https://github.com/rust-lang/crates.io/pull/1506#issuecomment-429102810

which is discouraging... it basically means we will not be able to remove the old libraries from crates.io, and we cannot "start from scratch".

I did understand your point on starting on `2.0.0` @Fi3 . It would be a way to put all SRI crates at the same starting point. But I still think starting on `2.0.0` would be bad, for the reasons I explained in my previous comment.

I wonder if we could do the same, but with all crates starting at `1.0.0`?
I did a quick search on all workspaces and I didn't see any crate that is above `1.0.0`.

RELEASE.md
plebhash: I'm not opinionated on this, and I do see your point.

Over Discord (and maybe during today's call) I think you expressed a slightly different idea. There you're suggesting a 4 week release cycle.

Did you change your mind after writing this comment here?

Either way is fine, just asking so I can change the wording on the PR. Also, feel free to suggest some changes direct changes to `RELEASE.md` if you want.

RELEASE.md
plebhash: from @Fi3 (over discord):

> main issue is noise crate already 1.1.0 and is not semver 1.0.0 is not compatible with 1.1.0 but jakub told me in the call that you can mark a crate on crates.io as not usable so this could be a solution we could mark all crates like this one as not usable and in that case start from 1.1.0 and mark 1.0.0 as not usable

so assuming we will set all crates to `v1.0.0` (except `noise-sv2`), I crated this PR: https://github.com/stratum-mining/stratum/pull/785

RELEASE.md
pavlenex: I would say 4 weeks was a consensus, but I still believe this is very optimistic. But for now we agreed we try with 4 weeks, and then see how it works for the team.

pavlenex:
Gave it another round of review, besides 4 week time-frame for a release, I don't see vulnerability/critical releases and how we handle those? Are those considered patch, and if so does it mean a critical vulnerability will be patched within whatever time-frame we agreed on or is it instant? I don't mean define vulnerability response process, but just define how are those handled, and step by step can be done in a separate security.md

Generally with releases, it sounds odd, but it's _storytelling_. We want to tell a story, ,motivate the users and community why certain release brings good functionality, improvements or fixes is very important for users, helps build excitement and community around releases and encourage people to try them out and report back if they find issues. With short release cycle, this is really tricky. I've checked how other projects do it, and BDK does it every 4 weeks, and LDK does it every two months. I think for us it'll be trial and error and we may decide to change things as we learn more on what works for us.

RELEASE.md
Fi3: Sounds good. All the 0s will become 1.0.0. All the one already to 1 are yanked and we start from the very next minor. And we say that 0 is not supported anymore.

RELEASE.md
```diff
@@ -1,66 +1,79 @@
...
+This workflow is manually started by navigating to the "Actions" tab in the SRI repo, then navigating to the Release workflow section, and clicking "Run Workflow".
+
+Note: in order to be able to manually trigger the "Run Workflow" button, the user needs to have "Write" permissions on the repository, otherwise the button will not show up on the UI.
+
+# Publishing Library Crates
+
+Although SRI has a global release cycle, which is attached to the binaries, each internal crate also has its own versioning history.
+
+Lib crates are published to crates.io in the `release-lib.yaml` workflow. The workflow tries to update all the library crates. 
+If a crate is not to updated, the step will fail for that each step have continue-on-error set to true.
```
GitGab19, line 70: If a create is not to BE updated (I guess it's a typo)

RELEASE.md
```diff
@@ -1,66 +1,79 @@
...
+- `main`
+- `dev`
 
+We consider `main` to be the branch where the source code of `HEAD` always reflects a production-ready state.
 
-## Github Binary Releases
+We consider `dev` to be the branch where the source code of `HEAD` always reflects a state with the latest delivered development changes for the next release.
 
-# Versioning
+Every 4 weeks all the changes to `dev` are merged back into `main` and then tagged with a release number while bumping `MAJOR` and/or `MINOR`.
```
GitGab19, line 37: I agree with all @pavlenex's considerations about the 4 weeks period. I'm very skeptical about it and I don't see a need for it to be that strictly to be honest. Why not defining a time period at all? Or put it larger (2 months)? 
Since with this release we finally implemented everything from the specs, I don't expect that many new features in a short-mid time frame. I'm sure we will need to make more patch releases than feature ones btw

RELEASE.md
plebhash: tbh the whole phrase was hard to make sense of, so I just re-wrote it on https://github.com/stratum-mining/stratum/pull/766/commits/985c438e5135334cb9449af9b2db00918fa4fbb3

RELEASE.md
plebhash: ok this point is generating a lot of confusion so I propose we leave the document like this for now:
```diff
- Every 4 weeks all the changes to `dev` are merged back into `main` and then tagged with a release number while bumping `MAJOR` and/or `MINOR`.
+ The SRI team will decide the appropriate time when the changes to `dev` are merged back into `main` and then tagged with a release number while bumping `MAJOR` and/or `MINOR`.
```

we can revisit this point in the future if necessary, but for now, we leave the date for `v1.1.0` TBD.

what do you think @Fi3 ?

plebhash: > I'm sure we will need to make more patch releases than feature ones btw

I can understand this point, and I wonder if we can do something about it. While we already have a strategy for dealing with bugs under `SemVer 2.0.0`, IMO we should keep that for exceptional situations where finding a bug is actually a **surprise**.

With that in mind, I propose that we start our release cycle with `v1.0.0-beta.0` instead of `v1.0.0`.
As we fix bugs, we increase `v1.0.0-beta.1`, `v1.0.0-beta.2`, etc.

Please let me know your thoughts @Fi3 @pavlenex @GitGab19

pavlenex: I understand your point about `v1.0.0-beta.1, v1.0.0-beta.2` but I am slightly worried on what that communicates to the public, and something as simple as a tag could drive people away from implementing and just keep them in "I'll wait a bit more stage". 

1.0.0 communicates that after 3+ years this is ready to be tested and adopted, sure there will be bugs, sure there's a LOT of work to be done to ensure stability, and probably a whole can of worms once different configurations and use-cases start being used, but... this is our opportunity to start fresh and clean, and I feel these tags would kill the momentum of excitement around something as significant as 1.0.0, so I'll really keep things as we already agreed, besides timeframe for releases seems we have a consensus. Timelines can easily be adjusted.

RELEASE.md
GitGab19: > tbh the whole phrase was hard to make sense of, so I just re-wrote it on [985c438](https://github.com/stratum-mining/stratum/commit/985c438e5135334cb9449af9b2db00918fa4fbb3)

Just saw it, now it's perfect 👍

RELEASE.md
GitGab19: Yeah now I would try to release it as `1.0.0`, without specifying a timeframe for merging `dev` into `main` and push a new release. After that, we will see how we can be more specific about timeframes, depending on external feedback and testing!

plebhash: ack https://github.com/stratum-mining/stratum/pull/766/commits/2c4cc3f71d1fc324d85086ec6e9d9396e7c9772f

plebhash:
> Gave it another round of review, besides 4 week time-frame for a release, I don't see vulnerability/critical releases and how we handle those? Are those considered a patch, and if so does it mean a critical vulnerability will be patched within whatever time-frame we agreed on (2 or 4 weeks)on or is it instant? I don't mean define vulnerability response processes, but rather when a critical vulnerabilty is found how is it handled in a release cycle.

I think that falls under the definition of bugs and patches. Since we are leaving the timelines open for releases, I guess there's no point in specifying that for patches as well?

Anyways, IMO these questions are already answered in [these lines](https://github.com/plebhash/stratum/blob/2c4cc3f71d1fc324d85086ec6e9d9396e7c9772f/RELEASE.md?plain=1#L15-L17):
```
Whenever a `PATCH` is introduced, it is applied to all the latest `MAJOR` releases.
For example: imagine there's releases `v1.0.0`, `v1.1.0`, and `v2.0.0`. A bug is found, dating back all the way to `v1.0.0`.
A patch is applied such that the following new tags are introduced: `v1.1.1` and `v2.0.1`.
```

@pavlenex let me know if you feel we should elaborate more on this.

pavlenex (approved):
LGTM

Issue #750, opened 2024-02-14
README(s) in /roles/ directory are not updated 


GitGab19:
We need to check validity of our current README files present under every [role dir](https://github.com/stratum-mining/stratum/tree/main/roles), and update them with latest changes

ikeogu:
Good day, please I wish to contribute to getting this resolved

GitGab19:
> Good day, please I wish to contribute to getting this resolved

Hey, I'm sorry but we've been very busy during past days. 
Looks good to me, go on! I mentioned this issue in a comment [here](https://github.com/stratum-mining/stratum/pull/772#issuecomment-1961809392), I can eventually push these details to your work

ikeogu:
Good day, I will be waiting for more details I wish to complete it within the shortes possiblbe time

PR #772, opened 2024-02-23
JDS mempool management enhancement

### LINKED ISSUE #740: JDS randomly answers to JDC with delay
As pointed out under issue #688 discussion, the root cause of the JDC stack overflow was JDS delay (seconds) in answering to JDC with a `AllocateMiningJob.Success` message. 
The stack overflow on JDC has been solved in #720, but we still don't know what can be the cause of JDS delay.

I opened this issue just to better track this strange behaviour in case something similar will be spotted in the future.

GitGab19:
This PR is an enhancement of the mempool instance inside Job Declarator Server.
As discussed during last dev call (here's [dev notes](https://discord.com/channels/950687892169195530/958814900770205739/1209570137905299598)), now JDS mempool is updated every X seconds (set in config file) by retrieving all data for every tx contained in the node's mempool. While on the testnet this is not a big issues (since there are few txs), on the mainnet this method is not sustainable at all. 
As pointed out during mainnet testing, this operation is too slow, and so after a DeclareMiningJob is received by JDC, JDS answers with a `ProvideMissingTransactions` message (asking basically all the txs proposed by JDC, because its mempool is not updated yet). The answers by JDC is huge, so lot of time is taken to process all data, leading to delays, which lead to crashes (this is exactly a reference to issue #740)

My proposal in this PR is a big change in how JDS manages its local mempool:
1) every X seconds (set in config) JDS updates it local mempool by inserting ONLY current tx_ids into a hashtable
2) as soon as a `DeclareMiningJob` is received, JDS retrieves txs data (using RPC call `getrawtransaction`) from Bitcoin node, filing them into the same hashtable
3) if some txs contained in `DeclareMiningJob` are not present into its local mempool (hashtable), JDS asks for them with `ProvideMissingTransactions` message (as usual)
4) JDC sends txs data trough the answer (`ProvideMissingTransactions.Success`) and JDS adds those txs data to its local mempool
5) txs which are mined in the meantime are automatically deleted from JDS mempool since the update is done by calling `RPC getrawmempool` so there's no risk that hashmap blows up

Already tested on both testnet and mainnet and it seems to work very well.
FIx #740


roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -20,14 +20,14 @@ pub struct Amount(f64);
 pub struct BlockHash(Hash);
 
 #[derive(Clone, Debug)]
-pub struct TransacrtionWithHash {
-    id: Txid,
-    tx: Transaction,
+pub struct TransactionWithHash {
+    pub id: Txid,
+    pub tx: Option<Transaction>,
```
Sjors, line 25: What is the purpose of storing a txid without its corresponding full transaction?

Sjors (commented):
If the JDS calls `getrawmempool` and then deletes absent transactions from its local pseudo-mempool, it will delete transactions that have just been mined. But it will also delete transactions that have been (RBF) replaced.

If miners are still working on templates that point to the replaced transaction, then you can't construct the block. The clients themselves can still send SubmitSolution to the Template Provider and it can still relay the block.

But it would be safer to hold on to transactions in templates that are still pending. Perhaps you could hold on to each template, and the template hold a reference to the pseudo-mempool entries it needs. And then you have a background job that throws out old templates some time after a block is found, which lets go of the reference. The mempool pruning process then just needs to make sure no references are held to an entry before deleting it.


I find it difficult to review the code, because I'm not a Rust dev and there's no comments explaining what it's trying to achieve.

GitGab19:
> If the JDS calls `getrawmempool` and then deletes absent transactions from its local pseudo-mempool, it will delete transactions that have just been mined. But it will also delete transactions that have been (RBF) replaced.
> 
> If miners are still working on templates that point to the replaced transaction, then you can't construct the block. The clients themselves can still send SubmitSolution to the Template Provider and it can still relay the block.
> 
> But it would be safer to hold on to transactions in templates that are still pending. Perhaps you could hold on to each template, and the template hold a reference to the pseudo-mempool entries it needs. And then you have a background job that throws out old templates some time after a block is found, which lets go of the reference. The mempool pruning process then just needs to make sure no references are held to an entry before deleting it.
> 
> I find it difficult to review the code, because I'm not a Rust dev and there's no comments explaining what it's trying to achieve.

Good point, thanks for reviewing!
I double-checked and every time a DeclareMiningJob is received from JDC, JDS stores a `declared_mining_job` which contains every tx data needed to build the block before propagating it. So there shouldn't be any problem with the case you mentioned here. If a miner works and finds a solution for a template which contains a RBF-replaced tx, JDS is able to construct it because it stores all data necessary, even if its local mempool has just been refreshed/updated and does not contain anymore that specific tx data.

roles/jd-server/src/lib/mempool/mod.rs
GitGab19: When you run the JDS, it starts creating its local mempool (through the use of a hashmap) by just loading the txids (which are the key of the hashmap).
As soon as `DeclareMiningJob` messages are received, JDS fills those entries by retrieving tx data.
In the meantime, it keeps on updating its local mempool, and the txids are used to retrieve (since are the hashmap keys) and pass tx data present in the current mempool to the updated one (so that it's no needed a second fetch from the node).
Here's the specific lines I'm mentioning: https://github.com/GitGab19/stratum/blob/93901fb4c525ccbedfe0dce7714ba11001d78ad1/roles/jd-server/src/lib/mempool/mod.rs#L86

Sjors:
> JDS stores a declared_mining_job which contains every tx data needed to build the block before propagating it

Ok, but then what's the point of the local mempool?

Sjors:
I see, so the local mempool feeds the `declared_mining_job` data structure?

GitGab19:
> I see, so the local mempool feeds the `declared_mining_job` data structure?
> 
> (maybe someone can draw a diagram of the data structures and information flow of the JDS)

Yeah exactly.
Regarding diagram, I can do it while tackling issue #750, adding it to JDS readme

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -85,10 +87,27 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
+                    Some(tx_data) => match &tx_data.tx {
+                        Some(tx) => txs_in_job.push(tx.clone()),
+                        None => {
+                            let new_tx_data = self
+                                .mempool
+                                .safe_lock(|x| x.get_client())
+                                .unwrap()
+                                .unwrap()
+                                .get_raw_transaction(&tx_data.id.to_string(), None);
+                            if let Ok(new_tx) = new_tx_data {
```
lorbax, line 100: in the case that `new_tx_data` is `Err(...)` should we add the index to the list of missing transactions? like below `missing_txs.push(i as u16),`

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -43,31 +43,55 @@ impl JDsMempool {
...
+                    let key_id = Txid::from_str(&id).unwrap();
+                    let tx = self_.safe_lock(|x| {
+                        match x.mempool.get(&key_id) {
+                            Some(entry) => {
+                                entry.clone()
+                            }
+                            None => None,
+                        }
+                    });
+                    let id = Txid::from_str(&id).unwrap();
```
lorbax, line 89: After rebasing remember to manage the unhappy path!

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -85,10 +87,32 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
+                                    );
+                                    txs_in_job.push(new_tx);
+                                }
+                                Err(_) => {
+                                    missing_txs.push(i as u16);
+                                }
+                            }
+                        }
+                    },
+                    None => missing_txs.push(i as u16),
```
Fi3, line 115: Why you do not try to get the tx data from the node, before adding it to missing_txs?

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -43,31 +43,44 @@ impl JDsMempool {
...
     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
-        let mut mempool_ordered: Vec<TransacrtionWithHash> = Vec::new();
+        let mut mempool_ordered: HashMap<Txid, Option<Transaction>> = HashMap::new();
         let client = self_
             .safe_lock(|x| x.get_client())
             .unwrap()
             .ok_or(JdsMempoolError::NoClient)?;
-        let new_mempool: Result<Vec<TransacrtionWithHash>, JdsMempoolError> =
+        let self_clone = self_.clone();
+        let new_mempool: Result<HashMap<Txid, Option<Transaction>>, JdsMempoolError> =
```
Fi3, line 71: you can do:
```rust
let new_mempool: Result<HashMap<Txid, Option<Transaction>>, JdsMempoolError> = {
    let self_ = self_.clone();
    ...
}
 ```
 instead of `let self_clone = self_.clone();`

File: roles/jd-server/src/lib/job_declarator/message_handler.rs (hunk 4 above)
Comment (GitGab19) [reply roles/jd-server/src/lib/job_declarator/message_handler.rs:115]: Becuase this path is the one taken when the txid is not present into the current mempool, so it means that tx is not present at all.

File: roles/jd-server/src/lib/mempool/mod.rs (hunk 5 above)
Comment (GitGab19) [reply roles/jd-server/src/lib/mempool/mod.rs]: Done 4be66e0e70c7f392b70fbb0f3df8eb18dcc4fed8

**Review by Fi3 — APPROVED** [review]

File: roles/jd-server/src/lib/job_declarator/message_handler.rs [hunk 6]
```diff
@@ -153,7 +168,15 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
                                     ),
                                 ),
                             )))? as usize;
-                    transactions.insert(index, tx);
+                    if index >= transactions.len() {
```
lorbax, line 171: From doc.rs
```
let mut vec = vec!["hello"];
vec.resize(3, "world");
assert_eq!(vec, ["hello", "world", "world"]);
```
This is the vector for the transactions that will be mined, in this order. I think that this is used to reconstruct the block. If some transaction is duplicated, this may lead propagation of invalid block. Be careful!
I suggest not resizing, as, in fact, it is already performed by `insert()` method few lines below.

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -93,6 +104,10 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
                 missing_txs.clone(),
             ));
 
+            if !txs_to_retrieve.is_empty() {
```
lorbax, line 107: Here you are asking the transactions to the rpc client. You are doing this _before_ sending `ProvideMissingTransactions` (PMT). My fear is that in this way the JD-server must wait the rpc client response in order to send `PMT` message. Please be careful!
If so, I suggest to save the variable `txs_to_retrieve` in the `declared_mining_job` field of `JobDeclaratorDownstream` and modify `add_tx_data_to_job` in order to be called in `main`.

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -182,3 +205,33 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
+
+fn add_tx_data_to_job(tx_id_list: Vec<(String, usize)>, jdd: &mut JobDeclaratorDownstream) {
+    let mempool = jdd.mempool.clone();
+    let mut declared_mining_job = jdd.declared_mining_job.clone();
+    tokio::task::spawn(async move {
+        for tx in tx_id_list.iter().enumerate() {
+            let index = tx.1 .1;
+            let new_tx_data: Result<Transaction, JdsMempoolError> = mempool
+                .safe_lock(|x| x.get_client())
+                .unwrap()
```
lorbax, line 217: Is it a big deal to propagate errors instead panicking?

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -114,14 +128,18 @@ impl JDsMempool {
         Ok(())
     }
 
-    pub fn to_short_ids(&self, nonce: u64) -> Option<HashMap<[u8; 6], Transaction>> {
+    pub fn to_short_ids(&self, nonce: u64) -> Option<HashMap<[u8; 6], TransactionWithHash>> {
```
lorbax, line 131: The struct `TransactionWithHash` was introduced on to represent transactions in the mempool, but you changed this representation. Did you consider to get rid of it?
You can return `Option<HashMap<[u8; 6], (Txid, Transaction)>>`.

lorbax (commented):
This is a very good PR!
I left some comments.

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -77,14 +78,24 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
             for (i, sid) in short_hash_list.iter().enumerate() {
                 let sid_: [u8; 6] = sid.to_vec().try_into().unwrap();
-                if let Some(tx_data) = short_id_mempool.get(&sid_) {
-                    txs_in_job.push(tx_data.clone());
-                } else {
-                    missing_txs.push(i as u16);
+                match short_id_mempool.get(&sid_) {
+                    Some(tx_data) => match &tx_data.tx {
+                        Some(tx) => {
+                            if i >= txs_in_job.len() {
```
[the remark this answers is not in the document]
GitGab19: I need to do it because we need to maintain the same order defined in `DeclaredMiningJob` message in the `txs_in_job`. Using `push()` method, if a tx is not present in current JDS mempool (and so it's added to the `missing_txs` vec), another one would be inserted into that specific index (causing a wrong tx order in `declared_mining_job`)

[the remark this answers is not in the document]
lorbax: I don't think that it is as you are describing. Here they will be repositioned correctly. If you have (0, tx0), (1, tx1), (2, tx2), (3, _), (4, tx4), (5, _) and (6, tx6) then it will be
`
tx_in_job = vec![tx0, tx1, tx2, tx4, tx6];
missing_ts = vec![3, 5]
`
then when you receive the message PMTS
tx3 is inserted in tx_in_job[3], which becomes
```
assert_eq[tx_in_job, vec![tx0, tx1, tx3, tx2, tx4, tx6]]
```
and then tx5 is inserted in tx_in_job[5]
```
assert_eq[tx_in_job, vec![tx0, tx1, tx3, tx2, tx4, tx5, tx6]]
```
It is tricky and we should find a way to manage it better, possibly to store each transaction with its index in the block payload. Like it is now works by it is very error prone

roles/jd-server/src/lib/job_declarator/message_handler.rs
GitGab19: I don't think it's happening, since add_tx_data_to_job is not async and the handler doesn't await anything. Also analyzing logs it seems that PMT message is created and sent to JDC without waiting nothing

lorbax: Yes the point is is `add_tx_data_to_job` is not async. If it was, there wouldn't be any problem I think.

Fi3:
I think that the overall architecture can be improved. From my understanding we clone txdata into too
many places and that can lead to memory issues at scale.

Each `JobDeclaratorDownstream` have a full copy of the transaction in the last declared block that
means that you need at least `mempool_size + average_block_size * number_of_clients` of memory. Is very
easy too see that this scale very badly.

What I instead suggest is to have txdata only in the `JDsMempool` and get them when we construct the
block then we get the data and we cretae the hex blok. I don't see any reason to have the data for each 
`JobDeclaratorDownstream` instead of just having the ids.

Another issue that I spotted is that `get_block_hex` when called assume that all txs data have been
fetched from the node. I don't think that this is the case but if it is maybe we should explain why
it is in a comment.

GitGab19:
> I think that the overall architecture can be improved. From my understanding we clone txdata into too many places and that can lead to memory issues at scale.
> 
> Each `JobDeclaratorDownstream` have a full copy of the transaction in the last declared block that means that you need at least `mempool_size + average_block_size * number_of_clients` of memory. Is very easy too see that this scale very badly.
> 
> What I instead suggest is to have txdata only in the `JDsMempool` and get them when we construct the block then we get the data and we cretae the hex blok. I don't see any reason to have the data for each `JobDeclaratorDownstream` instead of just having the ids.
> 
> Another issue that I spotted is that `get_block_hex` when called assume that all txs data have been fetched from the node. I don't think that this is the case but if it is maybe we should explain why it is in a comment.

You're right.
Done here 4cc2624df081a45c6e6b7e43d0772c45604bf5d9

GitGab19:
Just mined a block (https://mempool.space/it/testnet/block/000000000000001313d29fcd6627788c457cc4d8a1e5747abe34db0ea2117dae), and everything went right: JDS correctly fetched txs from its local mempool, reconstructed the block and propagated it through the node.

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -137,43 +149,45 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
+                TransactionState::ToBeRetrievedFromNodeMempool(_) => continue,
+                TransactionState::Missing => return Err(Error::JDSMissingTransactions),
             }
-            None => Err(Error::LogicErrorMessage(Box::new(
-                AllMessages::JobDeclaration(JobDeclaration::ProvideMissingTransactionsSuccess(
-                    message.clone().into_static(),
-                )),
-            ))),
         }
+        // TODO check it
```
Fi3, line 179: can we remove this TODO?

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -65,23 +88,115 @@ impl JobDeclaratorDownstream {
...
+        for txid in transactions_to_be_retrieved {
+            let transaction = client
+                .get_raw_transaction(&txid.to_string(), None)
+                .await
+                .map_err(|e| JdsError::MempoolError(mempool::error::JdsMempoolError::Rpc(e)))?;
+            let txid = transaction.txid();
+            mempool::JDsMempool::add_tx_data_to_mempool(
+                mempool_.clone(),
+                txid,
+                Some(transaction.clone()),
```
Fi3, line 137: no need to clone the tx data here, just save the id before call add_tx_data_to_mempool and pass the tx data witgout cloning it

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -65,23 +88,115 @@ impl JobDeclaratorDownstream {
...
+                    "Missing transactions".to_string(),
+                ));
+                if let Ok(Some(tx)) = tx {
+                    transactions_list.push(tx);
+                } else {
+                    return Err(Box::new(JdsError::ImpossibleToReconstructBlock(
+                        "Missing transactions".to_string(),
+                    )));
+                }
+            } else {
```
Fi3, line 190: you are assuming that when you receive the solution the task that fetch tx data from the bitcoin node has finished. This assumption is not always true. You can put an ref count with an atomic bool set to false in that tuple:
```rust
....
    declared_mining_job: (
        Option<DeclareMiningJob<'static>>,
        Vec<TransactionState>,
        Vec<u16>,
    ),
```
will be:
```rust
    declared_mining_job: (
        Option<DeclareMiningJob<'static>>,
        Vec<TransactionState>,
        Vec<u16>,
        Arc<AtomicBool>,
    ),
```
And you set it to true when retrieve_transactions_via_rpc finish. If for now you don't want to handle it cause is very unlikely I'm not sure that the right thing to do is to close the connection with downstream.

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -193,6 +312,16 @@ impl JobDeclaratorDownstream {
                         break;
                     }
                 }
+                let retrieve_transactions =
+                    JobDeclaratorDownstream::retrieve_transactions_via_rpc(self_mutex.clone())
+                        .await;
```
Fi3, line 317: why you are awaiting it? the all point of this PR is to get tx data without blocking here you just block the main loop. This should run in a new task and not be awaited.

Fi3:
I dont' see any way to remove tx from mempool, didn't look for it actually, I leave this comment just to be sure that we have a method to remove tx from mempool (eg when they go in a block) and that we don't let it growth to infinity. If not we should open an issue for it.

roles/jd-server/src/lib/job_declarator/mod.rs
Fi3: The assumption is correct but just cause we don't create a task and we await the future before continue with the main loop (that invalidate the all point of this PR) so we should have a parallel task that fetch txs, and when we have it the assumption is not anymore true, so my comment above still need to be addressed.  

https://github.com/stratum-mining/stratum/pull/772#discussion_r1527129448

GitGab19:
> I dont' see any way to remove tx from mempool, didn't look for it actually, I leave this comment just to be sure that we have a method to remove tx from mempool (eg when they go in a block) and that we don't let it growth to infinity. If not we should open an issue for it.

It's indirectly done at https://github.com/GitGab19/stratum/blob/b0caab442b834625660f626fac0c86a845aaf020/roles/jd-server/src/lib/mempool/mod.rs#L77.
Every time the mempool is updated, since `getrawmempool` RPC is called, the new mempool is built on its result. So it means that old txs are automatically cleaned up!

roles/jd-server/src/lib/job_declarator/mod.rs
GitGab19: @lorbax is working on this, he will push a commit on the PR he opened against my branch in order to check it and properly manage this case

roles/jd-server/src/lib/job_declarator/mod.rs
GitGab19: This has been fixed by @lorbax in https://github.com/GitGab19/stratum/pull/15/commits/b63587037769b5cc12dfad6108cafe2db4b88b14

PR #780, opened 2024-03-04
JDS block construction and propagation fix

GitGab19:
This PR fixes the way JDS manages and uses `version` field in the block header, taking it from the `successful share` instead of taking it from the `DeclareMiningJob` message.

Right now, since version-rolling is used, this field was changed many times by the miner and so JDS was not able to correctly build the right block header, as pointed out in the Template Provider logs as well:
`2024-02-29T17:45:55Z ERROR: ProcessNewBlock: AcceptBlock FAILED (high-hash, proof of work failed)`.
In the past we didn't figure out the exact reason of this message, because in the meantime block were propagated also from JDC. While testing PR #761 we forced JDC to not propagate it and the result was that we couldn't propagate valid blocks found because of the aforementioned reason.

Since I applied this patch to that PR, we finally mined and properly propagated many blocks on the testnet.

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/utils.rs
+++ b/protocols/v2/roles-logic-sv2/src/utils.rs
@@ -785,7 +785,7 @@ impl<'a> From<BlockCreator<'a>> for bitcoin::Block {
 
         let prev_blockhash = u256_to_block_hash(message.prev_hash.into_static());
         let header = stratum_common::bitcoin::blockdata::block::BlockHeader {
-            version: last_declare.version as i32,
+            version: message.version as i32,
             prev_blockhash,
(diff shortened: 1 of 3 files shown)


plebhash (approved):
LGTM

pavlenex:
@GitGab19 There's a conflict that needs to be resolved.

GitGab19:
> > @GitGab19 There's a conflict that needs to be resolved.
> 
> Yeah I will solve it right after #772 is merged

Since async functions were introduced in PR #761, it will take a little bit to correctly rebase this on `main`.
I would proceed by merging this one first, so that we can keep on with testing in the meantime!

PR #790, opened 2024-03-12
`release-libs.yaml` automatically triggered on PRs to `main`

plebhash:
Every time we merge some PR from `dev` to `main`, we need to make sure every modified crate version is properly increased. Otherwise, we will mess up the dependency chain of whoever is fetching from crates.io

This PR:
- changes `release-libs.yaml` so that the `Release Libs` workflow is triggered every time a PR is created against `main`.
- introduces a shell script (`check-versioning-lib-release.sh`) that checks the diff between `main` and `dev` branches... if some lib crate has been changed but its version wasn't bumped, the CI will trigger an error and stop (here's an [example](https://github.com/plebhash/stratum/actions/runs/8530384769/job/23368058479?pr=47#step:3:15) on my fork)
- removes `roles` from being published to crates.io (it doesn't make sense until https://github.com/stratum-mining/stratum/issues/702 is completed, which will take a long time)

### DIFF
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -11,7 +11,7 @@
 name: Release Libs
 
 on:
-  pull_request:
+  push:
     branches:
       - main
 
@@ -21,6 +21,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          ref: 'main'
 
       - name: Run check-versioning-lib-release.sh
         run: |
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -10,9 +10,10 @@
 
 name: Release Libs
 
-on: 
-  # Manually run by going to "Actions/Release" in Github and running the workflow
-  workflow_dispatch:
+on:
+  push:
+    branches:
+      - main
 
 jobs:
   libs_publish:
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -11,7 +11,7 @@
 name: Release Libs
 
 on:
-  push:
+  pull_request:
     branches:
       - main
 
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -21,8 +21,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-        with:
-          ref: 'main'
 
       - name: Run check-versioning-lib-release.sh
         run: |
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -136,23 +136,3 @@ jobs:
         run: |
           cd roles/roles-utils/rpc
           cargo publish
-      - name: Publish crate jd_client
-        continue-on-error: true
-        run: |
-          cargo publish --manifest-path=roles/jd-client/Cargo.toml
-      - name: Publish crate jd_server
-        continue-on-error: true
-        run: |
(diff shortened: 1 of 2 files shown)


Fi3:
It would be nice to have a script that look at diff with main and for the crates that do diff, it run cargo release --dry run. So that we can run the script as an action before merging to main and spot if we forgot to change a version of a changed crate. (less thing to think about when we do release that is not bad)

plebhash:
> It would be nice to have a script that look at diff with main and for the crates that do diff, it run cargo release --dry run. So that we can run the script as an action before merging to main and spot if we forgot to change a version of a changed crate. (less thing to think about when we review release that is not bad)

that's a good idea, I'll write a new script and add to this PR

plebhash:
> It would be nice to have a script that look at diff with main and for the crates that do diff, it run cargo release --dry run. So that we can run the script as an action before merging to main and spot if we forgot to change a version of a changed crate. (less thing to think about when we review release that is not bad)

I wrote a shell script on this commit https://github.com/stratum-mining/stratum/pull/790/commits/83733a86b4b85eab202c8447a8debfc3dbd13481

however, it is purely shell scripting + git, it does not really use `cargo release` in any way

GitGab19 (approved):
Great idea to have a script to check that before merging `dev` into `main` 👏

.github/workflows/release-libs.yaml
```diff
@@ -134,24 +146,4 @@ jobs:
...
-        run: |
-          cargo publish --manifest-path=roles/mining-proxy/Cargo.toml
-      - name: Publish crate pool_sv2
-        continue-on-error: true
-        run: |
-          cargo publish --manifest-path=roles/pool/Cargo.toml
-      - name: Publish crate translator_sv2
-        continue-on-error: true
-        run: |
-          cargo publish --manifest-path=roles/translator/Cargo.toml
```
plebhash, line 157: these are being removed because it doesn't make sense to publish roles to `crates.io` before https://github.com/stratum-mining/stratum/issues/702 is solved

and that will probably take a long time

check-versioning-lib-release.sh
```diff
@@ -0,0 +1,51 @@
...
+      version_dev=$(git show origin/dev:./Cargo.toml | awk -F' = ' '$1 == "version" {gsub(/[ "]+/, "", $2); print $2}')
+      version_main=$(git show origin/main:./Cargo.toml | awk -F' = ' '$1 == "version" {gsub(/[ "]+/, "", $2); print $2}')
+      if [ "$version_dev" = "$version_main" ]; then
+         echo "Changes detected in crate $crate between dev and main branches! Versions on dev and main branches are identical ($version_dev), so you should bump the crate version on dev before merging into main."
+         exit 1
+      fi
+  fi
+
+  cd - >/dev/null
+done
```
plebhash, line 51: here's an example of how this shell script will prevent us from merging to `main` if we forget to bump some crate's version:

https://github.com/plebhash/stratum/actions/runs/8530384769/job/23368058479?pr=47#step:3:15

plebhash:
`release-libs.yaml` is failing on CI with this error message:
```
please paste the token found on https://crates.io/me below
error: credential provider `cargo:token` failed action `login`

Caused by:
  please provide a non-empty token
Error: Process completed with exit code 101.
```

I suspect this is because this workflow is being triggered from my fork, which doesn't have the login info to crates.io configured.

Also, this failed workflow is not blocking the PR from being merged.

IMO I would ignore this and merge (if reviews ok). After merging, I can do a dummy PR from `dev` to `main`  to see if this error remains (I suspect it won't).

GitGab19 (approved):

PR #799, opened 2024-03-19
Translator should trigger downstream mining job clean

### LINKED ISSUE #797: Translator should trigger downstream mining job clean-up for Extended Mining Jobs
Currently [handle_new_exteded_mining_job_](https://github.com/stratum-mining/stratum/blob/361f12ba72053422d64bfcb7ab010d5d45e1f0a5/roles/translator/src/lib/proxy/bridge.rs#L402-L452) is setting the `clean_job` parameter of [clean_job](https://github.com/stratum-mining/stratum/blob/361f12ba72053422d64bfcb7ab010d5d45e1f0a5/roles/translator/src/lib/proxy/next_mining_notify.rs#L15-L58) to `false` always.

This is to rigid and should only be done, if the downstream client does signal support for Extended Mining Jobs. Perhaps a default value of `true` is more sensible.

For additional context please find see the [Discord message](https://discord.com/channels/950687892169195530/958814900770205739/1219201652322730094)  message below:

--- 

First of all, please take everything I write with a grain of salt. I'm a complete stratum noob :wink:

Basically you are right, the "delay" refers to the fact that upstream is sending a `mining.notify` for job-id 7 while downstream is sending a `mining.submit` for job-id 4 *after* the notification for job-id 7 has been sent out.This will result in upstream reject the job-id 4 share submission. #791 is about handling this error condition properly.

The interesting question is, why both, `CPUminer` and `bitaxe` are submitting stale shares? SV1 `mining.notify` has a `clean_jobs` parameter which should prevent the above scenario or at least result in only rather infrequent invalid share submissions. Looking at the logs provided with [#788](https://github.com/stratum-mining/stratum/issues/788) it looks like the invalid share submission happens in conjunction with so called Extended Mining Jobs.

Looking at the translator code reveals that [create_notify()](https://github.com/stratum-mining/stratum/blob/361f12ba72053422d64bfcb7ab010d5d45e1f0a5/roles/translator/src/lib/proxy/next_mining_notify.rs#L15-L58) is responsible to create SV1 `mining.notify` requests. In the signature of create_notify() we will find the `clean_jobs` parameter talked about earlier.

[handle_new_prev_hash_()](https://github.com/stratum-mining/stratum/blob/361f12ba72053422d64bfcb7ab010d5d45e1f0a5/roles/translator/src/lib/proxy/bridge.rs#L304-L361) is calling create_notify() with `clean_jobs` set to `true`. So this should trigger downstream to clean up stale job-ids, which at least the `bitaxe` FW is doing.

However, [handle_new_extended_mining_job_()](https://github.com/stratum-mining/stratum/blob/361f12ba72053422d64bfcb7ab010d5d45e1f0a5/roles/translator/src/lib/proxy/bridge.rs#L402-L452) is *always* setting `clean_jobs` to `false` and I think. this is wrong. If a downstream client in a group channel doesn't signal support for extended mining jobs, `clean_jobs` should be `true` IMO.

---

adoerr:
Translator should trigger downstream mining job clean up for Extended Mining Jobs.

Please see #797 for more context.

Fixes #797

### DIFF
--- a/roles/translator/src/lib/proxy/bridge.rs
+++ b/roles/translator/src/lib/proxy/bridge.rs
@@ -433,11 +433,12 @@ impl Bridge {
 
             let j_id = sv2_new_extended_mining_job.job_id;
             // Create the mining.notify to be sent to the Downstream.
-            // clean_jobs must be false because it's not a NewPrevHash template
+            // `clean_jobs` must be `true` otherwise downstream clients will
+            // never clean up invalid jobs and submit stale shares.
             let notify = crate::proxy::next_mining_notify::create_notify(
                 last_p_hash,
                 sv2_new_extended_mining_job.clone(),
-                false,
+                true,
             );
             // Get the sender to send the mining.notify to the Downstream
             tx_sv1_notify.send(notify.clone())?;


adoerr:
Closing the PR for now because it needs a more detailed discussion and perhaps some infrastructure support.

Discussion will continue as part of issue #797

PR #804, opened 2024-03-20
Roles dir readme

ikeogu:


plebhash:
as already stated, this file is not necessary

Issue #456, opened 2023-02-27
Update all `toml` dependencies to a published version


coleFD:
We cannot publish crates that have github dependencies and because we need `no_std` support we are currently depending on a fork of the original toml repo since the published version will not compile. toml is used in all of the roles crates to deserialize the config files so we need to fix each of them

pavlenex:
@plebhash any updates if https://github.com/stratum-mining/stratum/pull/812#issuecomment-2021581672 affected this and if we can freely close it?

plebhash:
#456 is still an unsolved issue

keeping track of this on the roles CLI standardization discussion: https://github.com/stratum-mining/stratum/discussions/857#discussioncomment-9143815

Shourya742:
Removing the TOML dependency is a critical requirement for the integration test. The TOML dependency obstructs the addition of the Bitcoind crate because of its `no_std` constraint. 
cc: @plebhash @jbesraa

PR #812, opened 2024-03-22
rename lib crates

### LINKED ISSUE #811: naming issues with lib crates
We are publishing our lib crates to crates.io, but there's some issues with naming.

Multiple crates have generic names with no indication to the SV2 or SRI context. The following crates indeed are non-SV2 specific, which I guess is fine:
- `bip32_derivation`
- `error_handling`
- `key-utils`

However, crates that could only be used as dependencies under a SV2 context should indicate that on their names.

- `rpc`
- `network-helpers`

In fact, [`rpc` is causing confusion on our `Release Libs` Github Action](https://github.com/stratum-mining/stratum/actions/runs/8392074993/job/22983876271#step:27:12):
```
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `rpc = "^1.0.0"`
  candidate versions found which didn't match: 0.0.1-dev.1
  location searched: crates.io index
  required by package `jd_server v0.1.0 (/home/runner/work/stratum/stratum/roles/jd-server)`
  if you are looking for the prerelease package it needs to be specified explicitly
      rpc = { version = "0.0.1-dev.1" }
  perhaps a crate was updated and forgotten to be re-vendored?
```

That is because there's already another crate with the same name: https://crates.io/crates/rpc

IMO these crates should be renamed to:
- `network_helpers_sv2`
- `rpc_sv2`

Another comment about these two crates is that they are only being used by `roles`, so IMO makes sense to move them to `roles/roles-utils`. That way, every crate inside `utils` is generic and we could eventually move them to separate repositories in the future.

plebhash:
close #811 

renames:
- `network_helpers` crate into `network_helpers_sv2`
- `rpc` crate into `rpc_sv2`

also moves both crates to `roles/roles-utils`


plebhash:
I suspect we will still run into https://github.com/stratum-mining/stratum/issues/456 after we merge this PR, but let's see

if that is the case, we will have to live with the fact that `roles` crates are not published to crates.io until we split them into `lib` + `bin`, which would allow us to leave `toml` as a dependency only for `bin` but not for `lib`

roles/jd-server/Cargo.toml
```diff
@@ -16,8 +16,8 @@ binary_sv2 = { version = "^1.0.0", path = "../../protocols/v2/binary-sv2/binary-
 buffer_sv2 = { version = "^1.0.0", path = "../../utils/buffer" }
 codec_sv2 = { version = "1.0.0", path = "../../protocols/v2/codec-sv2", features = ["noise_sv2"] }
 const_sv2 = { version = "^1.0.0", path = "../../protocols/v2/const-sv2" }
-network_helpers = { version = "1.0.0", path = "../../utils/network-helpers", features = ["with_tokio"] }
-noise_sv2 = { version = "1.0.0", path = "../../protocols/v2/noise-sv2" }
+network_helpers_sv2 = { version = "1.0.0", path = "../roles-utils/network-helpers", features = ["with_tokio"] }
+noise_sv2 = { version = "1.1.0", path = "../../protocols/v2/noise-sv2" }
```
GitGab19, line 20: Why you bumped this?

GitGab19 (approved):

roles/jd-server/Cargo.toml
plebhash: `noise_sv2` is on `v1.1.0`

https://github.com/stratum-mining/stratum/blob/23227812bad2973e3169c3e664853bba782d021a/protocols/v2/noise-sv2/Cargo.toml#L3

I'm not sure why `jd-server` even compiled before, but should be fixed.

PR #814, opened 2024-03-22
standardize roles CLI

### LINKED ISSUE #456: Update all `toml` dependencies to a published version

### LINKED ISSUE #813: standardize `roles` CLI
# context

`Stratum Reference Implementation` implies that the community will look at our roles to understand how to implement one.

therefore, code readability on the application level of the SRI stack (`roles`) implementations is a requirement

# problem

all crates under `roles` follow a similar pattern for CLI args parsing, but each with slightly different approaches, making them clumsy to navigate and reason about

# proposal

standardizing coding patterns across all roles CLI is essential for code readability

plebhash:
close #813
close #456 

- renames `*Result` and `*Error` structs to follow a standardized naming convention for all roles
- replaces `toml` dependency with `config` (see #456)
- introduces `clap` for managing CLI
- introduces a separate `args.rs` module into all roles for managing CLI args
- introduces a separate `*_config.rs` module into all roles for managing role `*Config` structs


plebhash:
we agreed to close this PR and replace it with a conversation over Github Discussions so we can agree on the conceptual direction for this refactor

PR #819, opened 2024-03-28
Improve sv2 cpuminer

Fi3:
Add a nicer user interface to the sv2 cpuminer with help messages. Add noise to the cpuminer.

This is the output of  `cargo run -- --help`

```console
Usage: mining-device [OPTIONS] --address-pool <ADDRESS_POOL>

Options:
  -p, --pubkey-pool <PUBKEY_POOL>    Pool pub key, when left empty the pool certificate is not checked
  -i, --id-device <ID_DEVICE>        Sometimes used by the pool to identify the device
  -a, --address-pool <ADDRESS_POOL>  Address of the pool in this format ip:port or domain:port
      --handicap <HANDICAP>          This value is used to slow down the cpu miner, it rapresents the number of micro-seconds that are awaited between one hashes, default is 0 [default: 0]
  -h, --help                         Print help
  -V, --version                      Print version
```

### DIFF
--- a/roles/test-utils/mining-device/src/main.rs
+++ b/roles/test-utils/mining-device/src/main.rs
@@ -236,7 +236,7 @@ pub struct Device {
 fn open_channel(device_id: Option<String>) -> OpenStandardMiningChannel<'static> {
     let user_identity = device_id.unwrap_or_default().try_into().unwrap();
     let id: u32 = 10;
-    info!("Misuring pc hashrate");
+    info!("Measuring pc hashrate");
     let nominal_hash_rate = measure_hashrate(5) as f32;
     info!("Pc hashrate is {}", nominal_hash_rate);
     info!("MINING DEVICE: send open channel with request id {}", id);
--- a/roles/test-utils/mining-device/src/main.rs
+++ b/roles/test-utils/mining-device/src/main.rs
@@ -1,35 +1,107 @@
 use async_std::net::TcpStream;
-use network_helpers_sv2::PlainConnection;
+use key_utils::Secp256k1PublicKey;
+use network_helpers_sv2::Connection;
 use roles_logic_sv2::utils::Id;
-use std::{
-    net::{IpAddr, Ipv4Addr, SocketAddr},
-    sync::Arc,
-};
+use std::{net::SocketAddr, sync::Arc, thread::sleep, time::Duration};
 
+use async_std::net::ToSocketAddrs;
+use clap::Parser;
 use stratum_common::bitcoin::{
     blockdata::block::BlockHeader, hash_types::BlockHash, hashes::Hash, util::uint::Uint256,
 };
+use tracing::{error, info};
 
-async fn connect(address: SocketAddr, handicap: u32) {
-    let stream = TcpStream::connect(address).await.unwrap();
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+struct Args {
+    #[arg(
+        short,
+        long,
+        help = "Pool pub key, when left empty the pool certificate is not checked"
+    )]
+    pubkey_pool: Option<Secp256k1PublicKey>,
+    #[arg(
+        short,
+        long,
+        help = "Sometimes used by the pool to identify the device"
+    )]
+    id_device: Option<String>,
+    #[arg(
+        short,
+        long,
+        help = "Address of the pool in this format ip:port or domain:port"
+    )]
+    address_pool: String,
+    #[arg(
+        long,
+        help = "This value is used to slow down the cpu miner, it rapresents the number of micro-seconds that are awaited between one hashes, default is 0",
+        default_value = "0"
+    )]
+    handicap: u32,
+}
+
+async fn connect(
+    address: String,
+    pub_key: Option<Secp256k1PublicKey>,
+    device_id: Option<String>,
+    handicap: u32,
+) {
+    let address = address
+        .clone()
+        .to_socket_addrs()
+        .await
+        .expect("Invalid pool address, use one of this formats: ip:port, domain:port")
+        .next()
+        .expect("Invalid pool address, use one of this formats: ip:port, domain:port");
+    info!("Connecting to pool at {}", address);
+    let socket = loop {
+        let pool =
+            async_std::future::timeout(Duration::from_secs(5), TcpStream::connect(address)).await;
+        match pool {
+            Ok(result) => match result {
+                Ok(socket) => break socket,
+                Err(e) => {
+                    error!(
+                        "Failed to connect to Upstream role at {}, retrying in 5s: {}",
+                        address, e
+                    );
+                    sleep(Duration::from_secs(5));
+                }
+            },
+            Err(_) => {
+                error!("Pool is unresponsive, terminating");
+                std::process::exit(1);
+            }
+        }
+    };
+    info!("Pool tcp connection established at {}", address);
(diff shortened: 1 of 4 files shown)


roles/test-utils/mining-device/src/main.rs
```diff
@@ -1,35 +1,107 @@
...
+    id_device: Option<String>,
+    #[arg(
+        short,
+        long,
+        help = "Address of the pool in this format ip:port or domain:port"
+    )]
+    address_pool: String,
+    #[arg(
+        long,
+        help = "This value is used to slow down the cpu miner, it rapresents the number of micro-seconds that are awaited between one hashes, default is 0",
```
plebhash, line 37:
```suggestion
        help = "This value is used to slow down the cpu miner, it represents the number of micro-seconds that are awaited between one hashes, default is 0",
```

plebhash:
I see this is using `clap` for CLI, while all other roles crates do not.
Was there any specific reason not to use `clap` there? Would it be desirable to use `clap` on roles implementations?

For context, see https://github.com/stratum-mining/stratum/issues/813 and https://github.com/stratum-mining/stratum/pull/814

roles/test-utils/mining-device/src/main.rs
```diff
@@ -143,7 +226,7 @@ pub struct Device {
 fn open_channel() -> OpenStandardMiningChannel<'static> {
     let user_identity = "ABC".to_string().try_into().unwrap();
```
plebhash, line 227: `ABC` could be made into a CLI arg?

roles/test-utils/mining-device/src/main.rs
plebhash, line 1: for better modularity and consistency with other role implementation, IMO the `main.rs` file could be broken down into:
- `args.rs`
- `setup_connection.rs`
- `device.rs`
- `miner.rs`

roles/test-utils/mining-device/src/main.rs
Fi3: yep I will do

roles/test-utils/mining-device/src/main.rs
Fi3: is 500 line of code file it seem an overkill for now ti split it in 4 modules. I will do when grow.

plebhash: IMO the number of lines of code shouldn't be the main criteria for this kind of decision

we should aim to keep consistent code structure across the project and think about readability and maintainability for future contributors

Fi3: having 4 files instead of one impact readability and maintainability in a bad way

Fi3: in a complex application is mandtory split everything in modules but this is almost a script. We risk to have one module per function not very pleasent to mantain :)

Fi3: Btw if you have an **objective** criteria that can be applied everywhere in the SRI repo on how and when to split files, we can discuss about that (maybe in another place).

plebhash: I am starting to work on something along those lines together with @lorbax 

we will present a comprehensive proposal soon

lorbax: I am reviewing it and for me the code would be a bit easier to understand with a light module subdivision. Just my opinion, no changes requested.

roles/test-utils/mining-device/src/main.rs
```diff
@@ -209,7 +308,7 @@ impl Device {
 
         loop {
             let mut incoming: StdFrame = receiver.recv().await.unwrap().try_into().unwrap();
-            let message_type = incoming.get_header().unwrap().msg_type();
+            let message_type = dbg!(incoming.get_header().unwrap().msg_type());
```
lorbax, line 311: this dbg! is supposed to be here?

lorbax:
I reviewed the code and LGTM.
This PR basically introduces two things: noise encrypted connections and command line arguments (with the aid of `clap` library). The first is easy to test, because it just works.
For testing the second, I launched the pool with local TP and I launched the mining_device (where the pool-pubkey is retrieved from pool's config)
`cargo run -- --pubkey-pool 9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72 --address-pool 127.0.0.2:34254`
I also tested the --address option, by launching the pool listening on a different address
`
listen_address = "127.0.0.2:34254"
`
 and the mining device
 `
 cargo run -- --pubkey-pool 9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72 --address-pool 127.0.0.2:34254
 `
 which works as expected.
 Similarly, all the other options works (-`V` and `--help`), in particupar `--id-user`
 `
 2024-03-30T18:04:59.209022Z  INFO roles_logic_sv2::handlers::common: Received SetupConnection: version=2, flags=1
2024-03-30T18:04:59.209254Z  INFO roles_logic_sv2::handlers::mining: Received OpenStandardMiningChannel from: pippo with id: 10
`

roles/test-utils/mining-device/src/main.rs
```diff
@@ -1,35 +1,114 @@
...
+    id_device: Option<String>,
+    #[arg(
+        short,
+        long,
+        help = "Address of the pool in this format ip:port or domain:port"
+    )]
+    address_pool: String,
+    #[arg(
+        long,
+        help = "This value is used to slow down the cpu miner, it rapresents the number of micro-seconds that are awaited between hashes",
```
plebhash, line 37:
```suggestion
        help = "This value is used to slow down the cpu miner, it represents the number of micro-seconds that are awaited between hashes",
```

plebhash (approved):

Sjors:
Thanks for opening this! Added to my test list...

Sjors:
I was able to mine a block!

Followup suggestions (feel free to merge this first, if it's holding back other things):

Use all CPU threads (currently only uses 1).

Rather than `--handicap` I would prefer being able to set a difficulty target, like the `--nbits` argument for Bitcoin Core's signet miner: https://github.com/bitcoin/bitcoin/blob/master/contrib/signet/README.md (this even messes with the timestamps to achieve the desired difficulty adjustments)

Perhaps simpler:
* a way to set the max hash rate
* optionally have it wait for x (default 10) minutes before mining the next block

However that can wait for a followup.

Sjors:
For the second block (on a custom signet) I'm getting errors:

```
2024-04-05T11:02:30.047785Z  INFO roles_logic_sv2::handlers::template_distribution: Received NewTemplate with id: 2, is future: true
2024-04-05T11:02:30.047829Z  INFO roles_logic_sv2::handlers::template_distribution: Received SetNewPrevHash for template: 2
2024-04-05T11:05:21.072403Z ERROR roles_logic_sv2::channel_logic::channel_factory: Share does not meet any target: Standard((SubmitSharesStandard { channel_id: 1, sequence_number: 2, job_id: 2, nonce: 360646192, ntime: 1712315010, version: 536870912 }, 0))
2024-04-05T11:08:48.166951Z ERROR roles_logic_sv2::channel_logic::channel_factory: Share does not meet any target: Standard((SubmitSharesStandard { channel_id: 1, sequence_number: 3, job_id: 2, nonce: 789285428, ntime: 1712315010, version: 536870912 }, 0))
```

PR #821, opened 2024-03-28
split roles bin lib

### LINKED ISSUE #820: split roles crates into `lib` + `bin`
as stated in https://github.com/stratum-mining/stratum/issues/702, it is desirable to have some clear distinction between:

- reference implementation for roles (bin)
- reusable high-level APIs for roles implementations (lib)

so we should split the roles workspace in two distinct directories, one for lib(APIs) and one for bin(executable applications).

plebhash:
close #820

as stated in https://github.com/stratum-mining/stratum/issues/702, it is desirable to have some clear distinction between:
- reference implementation for roles (`bin`)
- reusable high-level APIs for roles implementations (`lib`)

so we should split the roles workspace in two distinct directories, one for `lib` (APIs) and one for `bin` (executable applications).

a `*_bin` suffix was added to the name of binary crates (to avoid name clash), which now have the following names:
- `jd_client_sv2_bin`  (also added `_sv2` for consistency with other roles)
- `jd_server_sv2_bin` (also added `_sv2` for consistency with other roles)
- `proxy_sv2_bin` (removed `mining_` from the beginning to avoid redundancy)
- `pool_sv2_bin`
- `translator_proxy_sv2_bin` (added `_proxy` to make it clear this is another kind of proxy)

library crates now have the following names:
- `jd_client_sv2`  (added `_sv2` for consistency with other roles)
- `jd_server_sv2`  (added `_sv2` for consistency with other roles)
- `proxy_sv2` (removed `mining_` from the beginning to avoid redundancy)
- `network_helpers_sv2`
- `pool_sv2`
- `rpc_sv2`
- `translator_proxy_sv2`  (added `_proxy` to make it clear this is another kind of proxy)

`config-examples` only exist for binary crates
`main.rs` and `args.rs` only exist for binary crates
`lib.rs` (+ all other internal logic) only exist for library crates


lorbax:
Is this PR supposed to be against main?

plebhash:
> Is this PR supposed to be against main?

no, just fixed for `dev`

Issue #828, opened 2024-03-30
eliminate `handle_result!` macro


plebhash:
@lorbax suggested that we eliminate `handle_result!` macro:
> The `handle_error` function is called by the `handle_result!` macro, which performs a task with is similar to `?` operator, returning `val` if encounters an `Ok(val)` and calls `handle_error(sender, error)` if encounters an `Err(error)`.

in order to keep PR atomicity and make it easy to review, we should achieve that in separate PRs, one for each role:
- [ ] jd-client
- [ ] jd-server
- [ ] pool https://github.com/stratum-mining/stratum/pull/829
- [ ] proxy
- [ ] translator

plebhash:
I misunderstood things here. I no longer think this is possible, since it has implications on how we are spawning many `async move {}` blocks, which require propagation of errors via channels

simply replacing this macro with `?` means that many errors would go unnoticed, which would be undesirable

and simply expanding the macros (like I did in #829) doesn't achieve anything useful, aside from bloating the code

PR #829, opened 2024-03-30
remove `handle_result!` macro from pool

plebhash:
as described in #828 , this PR is eliminating the usage of `handle_result!` macro on the pool role implementation

it is the first step to allow us to evaluate if we really want to go in this direction

if we get consensus, new PRs will be opened making similar changes to other roles, and ultimately, the `utils/error-handling` can be eliminated

### DIFF
--- a/roles/Cargo.lock
+++ b/roles/Cargo.lock
@@ -320,41 +320,19 @@ checksum = "d86b93f97252c47b41663388e6d155714a9d0c398b99f1005cbc5f978b29f445"
 
 [[package]]
 name = "binary_codec_sv2"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Fi3:
I dont' get what is the direction. This PR just substitute the macro with the generated code.Which is goal? How do we want to handle errors? Only changing the macro with the respective generated code do not make much sens IMO. We have all the same issues as before with more code.

plebhash:
> I dont' get what is the direction. This PR just substitute the macro with the generated code.Which is goal? How do we want to handle errors? Only changing the macro with the respective generated code do not make much sens IMO. We have all the same issues as before with more code.

I agree this is making code more complex. Does @lorbax have some ideas of how we could mitigate this? Am I missing something on the direction you suggested?

Issue #375, opened 2022-12-11
Rm 3rd-party `hex` crate dependency


rrybarczyk:
SRI cannot take on any 3rd-party crates.

Remove 3rd-party `hex` crate dependency from `protocols/v1` and replace with custom serialize/deserialize logic.

keonglee1978:
If pay trible price show up havie any efectivr probllem or not

suyashjoshi404:
I could try and tackle this issue if there is no problem.

Fi3:
Hi @Kixunil we use rust-bitcoin in roles-logic-sv2 that is used by all the roles (pool, proxy, translator) so yes it would good to be able to reexport the hex crate from rust-bitcoin

Kixunil:
I don't expect this to be done soon but thanks for feedback!

plebhash:
@Kixunil what's the status on this? Would we be able to use `rust-bitcoin` as our re-export dependency for `hex`?

Kixunil:
@plebhash we made an independent `hex-conservative` crate, you can use that one as it's maintained by the same people as `bitcoin`.

plebhash:
@Fi3 should we replace `hex` with `hex-conservative` here?

https://github.com/stratum-mining/stratum/blob/7c96c4ecf48c490aa0494aca328f7ae2dfce233c/protocols/v1/Cargo.toml#L17

Fi3:
seem a good idea

star-gazer111:
@plebhash @Fi3 I think [this](https://github.com/stratum-mining/stratum/pull/830) does removes the 3rd-party hex crate dependency from protocols/v1 .

rrybarczyk:
@Kixunil sorry for the late response, but basically we just don't want to take on external dependencies (like `hex`) as it is a security risk. Looks like you have it handled now tho!

PR #830, opened 2024-03-31
fix: Rm 3rd party hex crate dependency 

star-gazer111:
fixes [375](https://github.com/stratum-mining/stratum/issues/375)

### DIFF
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -10,14 +10,26 @@
 
 name: Release Libs
 
-on: 
-  # Manually run by going to "Actions/Release" in Github and running the workflow
-  workflow_dispatch:
+on:
+  pull_request:
+    branches:
+      - main
 
 jobs:
   libs_publish:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Run check-versioning-lib-release.sh
+        run: |
+          ./check-versioning-lib-release.sh
+          if [ $? -eq 1 ]; then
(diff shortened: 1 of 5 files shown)


Fi3:
this code do not compile

star-gazer111:
Oh I see, am I making some basic mistake here?

Fi3:
have you tried to build the code before pushing it? what is the output of cargo build?

star-gazer111:
It is throwing me the following error : 
`error: failed to select a version for the requirement `hex-conservative = "^0.4.3"`
candidate versions found which didn't match: 0.2.0, 0.1.1, 0.1.0, ...
location searched: crates.io index
required by package `sv1_api v1.0.0 (/home/vasuk/stratum/protocols/v1)`
perhaps a crate was updated and forgotten to be re-vendored? `

If I change the version of hex-conservative in cargo.toml from 0.4.3 to 0.2.0 it gives me a very obvious error as follows:
`error[E0433]: failed to resolve: use of undeclared crate or module 'hex',  ^^^ use of undeclared crate or module 'hex'`

plebhash:
@star-gazer111 this task is not about simply replacing the dependency on the `Cargo.toml`

you also need to adapt the internal logic inside the code so that it uses the new dependency

that implies understanding how `hex` was used previously and how `hex-conservative` can be used as a replacement

star-gazer111:
got it, on to it!

star-gazer111:
just deleted the fork because of merge conflicts in #837 so the pr is closed for now, I'm working on it and almost done with the part of shifting the logic from `hex` to `hex-conservative` and will open a new pr in the next couple of days.

PR #833, opened 2024-04-02
Refractor update_mempool module in JDS mempool module for simplified async flow

### LINKED ISSUE #784: small bugs in the JDS mempool
Here
[https://github.com/stratum-mining/stratum/blob/main/roles/jd-server/src/lib/mempool/mod.rs#L85](https://github.com/stratum-mining/stratum/blob/main/roles/jd-server/src/lib/mempool/mod.rs#L85)
there is a tokio task that is awaited, which does not make any sense. Since the function `update_mempool` is already an async method, I think that the generation of a new task can be definitely can be avoided without any substantial work.

Also the rpc method `get_raw_mempool_verbose` should be renamed with `get_raw_mempool`, as at the end the rpc call is `getrawmempool` without verbose.

good for first issue!

NonsoAmadi10:
### What does this PR do?

Closes #784 

- Removed the unnecessary use of tokio::task::spawn in update_mempool, leveraging the existing async function context for all operations.
- Removed the TokioJoin(JoinError) implementation since it is not been used

### DIFF
--- a/roles/jd-server/src/lib/mempool/mod.rs
+++ b/roles/jd-server/src/lib/mempool/mod.rs
@@ -69,8 +69,7 @@ impl JDsMempool {
         let txids = add_txs_to_mempool_inner.known_transactions;
         let transactions = add_txs_to_mempool_inner.unknown_transactions;
         let client = self_
-            .safe_lock(|a| a.get_client())
-            .map_err(|e| JdsMempoolError::PoisonLock(e.to_string()))?
+            .safe_lock(|a| a.get_client())?
             .ok_or(JdsMempoolError::NoClient)?;
         // fill in the mempool the transactions id in the mempool with the full transactions
         // retrieved from the jd client
@@ -99,26 +98,18 @@ impl JDsMempool {
         let mut mempool_ordered: HashMap<Txid, Option<Transaction>> = HashMap::new();
 
         let client = self_
-            .safe_lock(|x| x.get_client())
-            .map_err(|e| JdsMempoolError::PoisonLock(e.to_string()))?
+            .safe_lock(|x| x.get_client())?
             .ok_or(JdsMempoolError::NoClient)?;
 
-        let mempool: Vec<String> = client.get_raw_mempool().await.map_err(|err| {
-            let err_msg = format!("Error occurred during deserialization: {:?}", err);
-            JdsMempoolError::Rpc(RpcError::Deserialization(err_msg))
-        })?;
+        let mempool: Vec<String> = client.get_raw_mempool().await?;
         for id in &mempool {
-            let key_id = Txid::from_str(id).map_err(|err| {
-                let err_msg = format!("Error occurred during deserialization: {:?}", err);
-                JdsMempoolError::Rpc(RpcError::Deserialization(err_msg))
-            })?;
+            let key_id = Txid::from_str(id)
+                .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
 
-            let tx = self_
-                .safe_lock(|x| match x.mempool.get(&key_id) {
-                    Some(entry) => entry.clone(),
-                    None => None,
-                })
-                .map_err(|e| JdsMempoolError::PoisonLock(e.to_string()))?;
+            let tx = self_.safe_lock(|x| match x.mempool.get(&key_id) {
+                Some(entry) => entry.clone(),
+                None => None,
+            })?;
 
             mempool_ordered.insert(key_id, tx);
         }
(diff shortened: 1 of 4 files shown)


NonsoAmadi10:
@GitGab19 I have re-reviewed and made some changes here, please I am requesting a review here and also to point out that the MG test is still failing

GitGab19:
> @GitGab19 I have re-reviewed and made some changes here, please I am requesting a review here and also to point out that the MG test is still failing

I would ask to @lorbax to have a look at this, since he opened issue #784 and he worked on the async management during last release 🙏

lorbax:
I will do it by the next week!

lorbax (commented):
This is a good PR
I would be good to have the minor changes that I left as comments

NonsoAmadi10:
@lorbax i have been able to manage the unhappy paths in the the update_mempool function, please I request your feedback on this before I pick up the next issue. Thanks

plebhash:
@lorbax once you do the final review here, are you able to approve?

Fi3:

GitGab19 (approved):
LGTM.
@lorbax could you double check when you get a chance pls?

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -97,43 +97,37 @@ impl JDsMempool {
...
-                let _ = self_.safe_lock(|x| {
-                    x.mempool = new_mempool_;
-                });
-                Ok(())
-            }
-            Err(a) => Err(a),
+
+        let mempool: Vec<String> = client.get_raw_mempool().await.map_err(|err| {
+            let err_msg = format!("Error occurred during deserialization: {:?}", err);
+            JdsMempoolError::Rpc(RpcError::Deserialization(err_msg))
```
Fi3, line 108: mapping the error to a de-serialization error do not seems right to me. We can have network error for example.

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -97,43 +97,37 @@ impl JDsMempool {
...
+        })?;
+        for id in &mempool {
+            let key_id = Txid::from_str(id).map_err(|err| {
+                let err_msg = format!("Error occurred during deserialization: {:?}", err);
```
Fi3, line 112: Also this mapping seems problematic. Is true that here you can have only de-serialization error. But my question is what this mapping add to the error? Why not just immediately return the error without mapping it?

NonsoAmadi10: I spoke to @lorbax ftom our previous conversation on this PR and he said I should return a deserialization error

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -97,43 +97,37 @@ impl JDsMempool {
...
+                JdsMempoolError::Rpc(RpcError::Deserialization(err_msg))
+            })?;
+
+            let tx = self_
+                .safe_lock(|x| match x.mempool.get(&key_id) {
+                    Some(entry) => entry.clone(),
+                    None => None,
+                })
+                .map_err(|e| JdsMempoolError::PoisonLock(e.to_string()))?;
```
Fi3, line 121: This is superfluous, as above, when there is a `TryFom ErrorA to ErrorB` as in this case for JdsMempoolError  and PoinsonLock you can just use the question mark to immidatlyb return the error (you are implicity calling into())

roles/jd-server/src/lib/mempool/mod.rs
Fi3: If there is not TryFrom for it would be better to add it but is ok also like that, for the other 2 map you need to remove the mapping

roles/jd-server/src/lib/mempool/mod.rs
lorbax: @NonsoAmadi10 perhaps you can just propagate the error
```
        let mempool: Vec<String> = client
            .get_raw_mempool()
            .await
            .map_err(JdsMempoolError::Rpc)?;
```

lorbax:
@NonsoAmadi10 you must  also remove the TokioJoin variant of JdsMempoolError

roles/jd-server/src/lib/mempool/mod.rs
lorbax: Are suggesting to add an implementation of try_from that converts a PoisonLock to the relative variant in JdsMempolError? otherwise I cannot see how you can "use the question mark to immidatly"

Fi3: we already have the try from you can just use ?

Fi3: sorry the From not TryFrom

Fi3: a ok is another kind of error you are not returning JdsError but a JdsMempoolError that do not have the from. Yes I suiggest to implement it   in  that case as we have for the JdsError

roles/jd-server/src/lib/mempool/mod.rs
lorbax: BTW for sure the most elegant solution is to implement from for error conversion as @Fi3 already suggested

lorbax (approved):
Run local TP, pool, jds, jdc, tproxy and cpuminer without issues.
The code looks good to me.

Issue #845, opened 2024-04-12
Rust Docs + refactor: `protocols` crates


plebhash:
We recently published our `protocols` crates to `crates.io`.
We need to make sure the APIs provided by those crates are properly documented under the [Rust Best Practices](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html).

This is also the perfect opportunity to map and address technical debt on those crates. As we go through the code and try to understand it, we take the chance to write down a list of ways the code could be improved. Things such as:
- unintuitive, redundant, broken or suboptimal APIs
- unsafe error handling (`unwrap`, `panic!`, `expect`)
- typos, bad variable names
- `// todo` comments, `todo!` macros, edge cases with `unimplemented!` or `panic!`

---

# Task list

## Rust Docs

- [x] `v1` (already documented but worth a look before proceding)
  - INPUT: Q&A Discussion:: https://github.com/stratum-mining/stratum/discussions/897
  - OUTPUT: not planned
- [ ] `v2/binary-sv2/binary-sv2`
  - [x] INPUT: Q&A Discussion: todo
  - [x] INPUT: Rust Docs Issue #1008
  - [x] OUTPUT: Rust Docs PR: #1231 
  - [x] OUTPUT: Technical Debt issue: #1131
  - [ ] OUTPUT: Refactoring PRs: todo
- [ ] `v2/binary-sv2/no-serde-sv2/codec`
  - [x] INPUT: Q&A Discussion: todo 
  - [x] INPUT: Rust Docs Issue: #1009
  - [x] OUTPUT: Rust Docs PRs: #1046 (small doc additions) #1231
  - [x] OUTPUT: Technical Debt issue: #1132 
  - [ ] OUTPUT: Refactoring PRs: todo
- [ ] `v2/binary-sv2/no-serde-sv2/derive_codec`
  - [x] INPUT: Q&A Discussion: todo 
  - [x] INPUT: Rust Docs Issue: #1010
  - [x] OUTPUT: Rust Docs PR: #1231 
  - [x] OUTPUT: Technical Debt issue: #1133 
  - [ ] OUTPUT: Refactoring PRs: todo
- [x] `v2/binary-sv2/serde-sv2`
  - [x] INPUT: Q&A Discussion: not planned 
  - [x] INPUT: Rust Docs Issue: not planned #1011
  - [x] OUTPUT: Rust Docs PR: not planned
  - [x] OUTPUT: Technical Debt issue: not planned #1134
[...]

PR #848, opened 2024-04-13
docs `framing_sv2`

plebhash:
adds Rust Docs for `framing_sv2`, as described in https://github.com/stratum-mining/stratum/issues/845

also renames `NoiseFrame` into `HandShakeFrame`

the documentation effort was guided by discussion with @Fi3 on https://github.com/stratum-mining/stratum/discussions/858

### DIFF
--- a/protocols/v2/framing-sv2/src/framing2.rs
+++ b/protocols/v2/framing-sv2/src/framing2.rs
@@ -113,6 +113,8 @@ impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for 
     type Deserialized = B;
 
     /// Write the serialized `Sv2Frame` into `dst`.
+    /// This operation when called on an already serialized frame is very cheap.
+    /// When called on a non serialized frame, it is not so cheap (because it serializes it).
     #[inline]
     fn serialize(self, dst: &mut [u8]) -> Result<(), Error> {
         if let Some(mut serialized) = self.serialized {
--- a/protocols/v2/framing-sv2/src/framing2.rs
+++ b/protocols/v2/framing-sv2/src/framing2.rs
@@ -205,7 +205,7 @@ impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for 
     }
 
     /// If `Sv2Frame` is serialized, returns the length of `self.serialized`,
-    /// otherwise, returns the length of `self.pauload`.
+    /// otherwise, returns the length of `self.payload`.
     #[inline]
     fn encoded_length(&self) -> usize {
         if let Some(serialized) = self.serialized.as_ref() {
--- a/protocols/v2/framing-sv2/src/framing2.rs
+++ b/protocols/v2/framing-sv2/src/framing2.rs
@@ -98,14 +98,6 @@ pub struct NoiseFrame {
 /// todo
 pub type HandShakeFrame = NoiseFrame;
 
-/// todo (why unreachable?)
-#[cfg(feature = "with_buffer_pool")]
-impl<A> From<EitherFrame<A, Vec<u8>>> for Sv2Frame<A, buffer_sv2::Slice> {
-    fn from(_: EitherFrame<A, Vec<u8>>) -> Self {
-        unreachable!()
-    }
-}
-
 impl NoiseFrame {
     /// Returns payload of `NoiseFrame` as a `Vec<u8>`
     pub fn get_payload_when_handshaking(&self) -> Vec<u8> {
@@ -240,12 +232,6 @@ impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for 
             serialized: None,
         })
     }
-}
-
-#[inline]
-pub fn build_noise_frame_header(frame: &mut [u8], len: u16) {
-    frame[0] = len.to_le_bytes()[0];
-    frame[1] = len.to_le_bytes()[1];
 }
 
 impl<'a> Frame<'a, Slice> for NoiseFrame {
--- a/protocols/v2/framing-sv2/src/framing2.rs
+++ b/protocols/v2/framing-sv2/src/framing2.rs
@@ -15,6 +15,8 @@ type Slice = Vec<u8>;
 type Slice = buffer_sv2::Slice;
 
 impl<A, B> Sv2Frame<A, B> {
+    /// Maps a `Sv2Frame<A, B>` to `Sv2Frame<C, B>` by applying `fun`,
+    /// which is assumed to be a closure that converts `A` to `C`
     pub fn map<C>(self, fun: fn(A) -> C) -> Sv2Frame<C, B> {
         let serialized = self.serialized;
         let header = self.header;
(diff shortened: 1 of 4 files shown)


protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -31,31 +33,35 @@ pub trait Frame<'a, T: Serialize + GetSize>: Sized {
...
+    /// Write the serialized `Frame` into `dst`.
     fn serialize(self, dst: &mut [u8]) -> Result<(), Error>;
 
-    //fn deserialize(&'a mut self) -> Result<Self::Deserialized, serde_sv2::Error>;
-
+    /// Get the payload
     fn payload(&'a mut self) -> &'a mut [u8];
 
-    /// If is an Sv2 frame return the Some(header) if it is a noise frame return None
+    /// `Sv2Frame` returns `Some(self.header)`, `NoiseFrame` return `None`.
```
Fi3, line 42: Since NoiseFrame is no more a thing I would at least change the name everywhere and use HandShakeFrame instead. Here we can say return a frame when the frame have an header like Sv2Frame, when it do not (like HandShakeFrame) return None.

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -31,31 +33,35 @@ pub trait Frame<'a, T: Serialize + GetSize>: Sized {
...
     fn get_header(&self) -> Option<crate::header::Header>;
 
-    /// Try to build an Frame frame from raw bytes.
-    /// It return the frame or the number of the bytes needed to complete the frame
-    /// The resulting frame is just a header plus a payload with the right number of bytes nothing
-    /// is said about the correctness of the payload
+    /// Try to build a `Frame` from raw bytes.
+    /// Returns `Self` on success, or the number of the bytes needed to complete the frame
+    /// as an error. Nothing is assumed or checked about the correctness of the payload.
```
Fi3, line 47: We should say that this function check that the payload have the right byte len (the one stated in the header) where from_bytes_unchecked does not

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -31,31 +33,35 @@ pub trait Frame<'a, T: Serialize + GetSize>: Sized {
...
     fn from_bytes(bytes: Self::Buffer) -> Result<Self, isize>;
 
     fn from_bytes_unchecked(bytes: Self::Buffer) -> Self;
 
+    /// Helps to determine if the frame size encoded in a byte array correctly representing the size of the frame.
+    /// - Returns `0` if the byte slice is of the expected size according to the header.
+    /// - Returns a negative value if the byte slice is smaller than a Noise Frame header; this value
+    ///   represents how many bytes are missing.
+    /// - Returns a positive value if the byte slice is longer than expected; this value
+    ///   indicates the surplus of bytes beyond the expected size.
```
Fi3, line 57: Note for the future, it could make sense to return a type that encode the above infos: enum Ok, Bigger, Smaller. This would be an API change so for now I would live it like it is

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -82,20 +89,19 @@ impl<T, B> Default for Sv2Frame<T, B> {
     }
 }
 
+/// Abstraction for a Noise Frame
+/// In practice, it is only used as the alias `HandShakeFrame`
+/// Contains only a `Slice` payload
```
Fi3, line 94: I would call it HandShakeFrame and use the alias to rename it NoiseFrame, so we don't break the API but is more explicit what this is

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -64,11 +70,12 @@ pub trait Frame<'a, T: Serialize + GetSize>: Sized {
     ) -> Option<Self>;
 }
 
+/// Abstraction for a SV2 Frame.
```
Fi3, line 73: I would add some info about the Sv2Frame in general.

protocols/v2/framing-sv2/src/framing2.rs
Fi3: I would also add some info about the HandShakeFrame in general. (is not a frame is just an handshake message and have fixed len)

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -82,20 +89,19 @@ impl<T, B> Default for Sv2Frame<T, B> {
...
 pub type HandShakeFrame = NoiseFrame;
 
-#[cfg(feature = "with_buffer_pool")]
-impl<A> From<EitherFrame<A, Vec<u8>>> for Sv2Frame<A, buffer_sv2::Slice> {
-    fn from(_: EitherFrame<A, Vec<u8>>) -> Self {
-        unreachable!()
-    }
-}
 impl NoiseFrame {
+    /// Returns payload of `NoiseFrame` as a `Vec<u8>`
```
Fi3, line 104: As already said above I would always call it HandShakeFrame for now (then we will remove it entirely).

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -105,8 +111,7 @@ impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for
     type Buffer = B;
     type Deserialized = B;
 
-    /// Serialize the frame into dst if the frame is already serialized it just swap dst with
-    /// itself
+    /// Write the serialized `Sv2Frame` into `dst`.
```
Fi3, line 114: I would say (also if it seems obvious but can be helpful when you write code not having to check the implementation for things like that) that this operation when called on an already serialized frame is very cheap (consume the Sv2Frame) but when called on a non serialized frame is not so cheap cause it have to serialize it.

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -175,23 +178,34 @@ impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for
...
                 } else {
+                    // Returns how many excess bytes are beyond the expected frame size
                     (bytes.len() - Header::SIZE) as isize + header.len() as isize
                 }
             }
         }
     }
 
+    /// If `Sv2Frame` is serialized, returns the length of `self.serialized`,
+    /// otherwise, returns the length of `self.pauload`.
```
Fi3, line 208: typo

protocols/v2/framing-sv2/src/header.rs
```diff
@@ -52,13 +54,15 @@ impl Header {
         })
     }
 
+    /// Get the message length
```
Fi3, line 57: I would use payload rather then message here

protocols/v2/framing-sv2/src/header.rs
```diff
@@ -52,13 +54,15 @@ impl Header {
...
     #[allow(clippy::len_without_is_empty)]
     #[inline]
     pub fn len(&self) -> usize {
         let inner: u32 = self.msg_length.into();
         inner as usize
     }
 
+    /// Construct a `Header` from message length, type and extension type.
```
Fi3, line 65: same as above

protocols/v2/framing-sv2/src/framing2.rs
Fi3: Ignore the  general info about the Sv2Frame in general I didn't see that you have a module level comment for that.

protocols/v2/framing-sv2/src/lib.rs
```diff
@@ -1,25 +1,31 @@
...
+//! | `msg_type` | `U8` | Unique identifier of the extension describing this protocol message. |
+//! | `msg_length` | `U24` | Length of the protocol message, not including this header. |
+//! | `payload` | `BYTES` | Message-specific payload of length `msg_length`. If the MSB in `extension_type` (the `channel_msg` bit) is set the first four bytes are defined as a `U32` `"channel_id"`, though this definition is repeated in the message definitions below and these 4 bytes are included in `msg_length`. |
+//!
+//! # Features
+//! This crate can be built with the following features:
+//! - `with_serde`: builds `binary_sv2` and `buffer_sv2` crates with `serde`-based encoding and decoding.
```
Fi3, line 17: I would also add that we need it only for the message generator. And that is deprecated for everything else. (Maybe in the future we will remove it entirely)

protocols/v2/framing-sv2/src/framing2.rs
```diff
@@ -129,30 +134,28 @@ impl<'a, T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Frame<'a, T> for
...
-    // deserialized it contain the rust type that represant the Sv2 message. If the type is
-    // deserialized self.paylos.is_some() is true. To get the serialized payload the inner type
-    // should be serialized and this function should never be used, cause is intended as a fast
-    // function that return a reference to an already serialized payload. For that the function
-    // panic.
+    /// `self` can be either serialized (`self.serialized` is `Some()`) or
+    /// deserialized (`self.serialized` is `None`, `self.payload` is `Some()`).
+    /// This function is only intended as a fast way to get a reference to an
+    /// already serialized payload. If the frame has not yet been
+    /// serialized, this function should never be used (it will panic).
```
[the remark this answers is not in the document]
plebhash: sorry @Fi3 I don't fully understand the suggestion

I did some research but couldn't find an easy way to print a warning during compilation.
I only found [this pre-RFC](https://internals.rust-lang.org/t/pre-rfc-add-compile-warning-macro/9370) which proposes a `compile_warning!` macro but it will probably take a long time before this is available.

Do you have some examples on how to achieve this?

protocols/v2/framing-sv2/src/framing2.rs
plebhash: @Fi3 can you elaborate on the comments here? I'm confused about the suggestion

protocols/v2/framing-sv2/src/framing2.rs
plebhash: https://github.com/stratum-mining/stratum/pull/848/commits/cbafbeae58a160d1a9daec4eb013546f58c261e8

protocols/v2/framing-sv2/src/framing2.rs
plebhash: 1fe9e1af00ef2f09a30862b8f4aa60cfc4af6819

protocols/v2/framing-sv2/src/framing2.rs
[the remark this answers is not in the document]
Fi3: No I was just supposing that it was already doable on stable, maybe we can leave a comment for our future self.

protocols/v2/framing-sv2/src/framing2.rs
Fi3: I was suggesting to add thing that you already added so just ignore it

protocols/v2/framing-sv2/src/framing2.rs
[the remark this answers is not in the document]
plebhash: well, as we already discussed briefly, I believe the `framing_sv2` crate should be refactored in the future.

the `Frame` trait is forcing `Sv2Frame` and `HandShakeFrame` to share APIs while they don't really share a lot of common features, and as a consequence we end up with APIs that are inuntuitive and dirty (e.g.: panic on undersired scenarios)

so if we're taking a note for the future, I think we should simply refactor this crate (removing the `Frame` trait) so that the APIs don't need to be emitting compilation warnings in order to avoid footguns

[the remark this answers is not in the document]
plebhash: I wrote a summary of all things to be refactored on `framing_sv2` in the future here:
https://github.com/stratum-mining/stratum/issues/903

Fi3 (approved):

PR #851, opened 2024-04-15
fix(translator): fix HexError(OddLength) in mining.subscribe handler

nikicat:
This PR fixes 
`ERROR translator_sv2::lib::status: Error: V1Protocol(Method(ParsingMethodError((HexError(OddLength), StandardRequest(StandardRequest { id: 150166, method: "mining.subscribe", params: Array [String("bmminer/2.0.0"), String("265800001")] })))))`
errors.

I'm new to Rust and made this PR with the help of ChatGPT, so I'm not sure it's fully correct. Nevertheless, it fixes the actual error in my case.

This error happens when the miner sends an odd-sized extranonce.

### DIFF
--- a/benches/benches/src/sv1/lib/client.rs
+++ b/benches/benches/src/sv1/lib/client.rs
@@ -199,8 +199,7 @@ impl IsClient<'static> for Client {
     }
 }
 pub fn extranonce_from_hex(hex: &str) -> Extranonce<'static> {
-    let data = utils::decode_hex(hex).unwrap();
-    Extranonce::try_from(data).expect("Failed to convert hex to U256")
+    Extranonce::try_from(hex).expect("Failed to convert hex to U256")
 }
 pub fn prevhash_from_hex<'a>(hex: &str) -> PrevHash<'a> {
     let data = utils::decode_hex(hex).unwrap();
--- a/protocols/v1/src/lib.rs
+++ b/protocols/v1/src/lib.rs
@@ -333,10 +333,7 @@ pub trait IsClient<'a> {
                 // so it doesnt really matter what the server sets the extranonce to in the mining.configure handler
                 debug!("NOTICE: Subscribe extranonce is hardcoded by server");
                 let subscribe = self
-                    .subscribe(
-                        configure.id,
-                        Some(Extranonce::try_from(hex::decode("08000002")?)?),
-                    )
+                    .subscribe(configure.id, Some(Extranonce::try_from("08000002")?))
                     .ok();
                 Ok(subscribe)
             }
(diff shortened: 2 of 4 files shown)


GitGab19:
@nikicat some tests are failing, please fix them.
As a reference, you can have a look at this paragraph: https://github.com/stratum-mining/stratum/blob/main/CONTRIBUTING.md#contribution-workflow

nikicat:
@GitGab19 Hi, I've fixed `rustfmt` errors.
The test `downstream_sv1::diff_management::test::test_diff_management` is just flapping - passes or fails randomly. I suppose it's out of the scope of this PR to fix it.

plebhash:
@nikicat as pointed out in [`RELEASE.md`](https://github.com/stratum-mining/stratum/blob/main/RELEASE.md#git-branching-strategy), we ask that all external contributions are made as PRs against `dev` branch.

Please change the base branch of this PR, but be mindful that a rebase might be required.

GitGab19:
@nikicat haven't you added some tests?
I'm not sure if they were messed up during your last pushes. But since yesterday this PR seems different.
Can you double check it?

nikicat:
@GitGab19 This is because my initial fix turned out to be wrong - it was just a coincidence that a miner with an odd-sized extranonce stopped subscribing.
After that, I made a fix that indeed works, but I struggled to find the correct place for a test - it should request the translator proxy API. I've tried to add odd-sized extranonce [here](https://github.com/stratum-mining/stratum/blob/dev/test/message-generator/test/sv1-test.json#L23) but it doesn't lead to the test failure even without this patch. Where to add a test?

GitGab19:
> @GitGab19 This is because my initial fix turned out to be wrong - it was just a coincidence that a miner with an odd-sized extranonce stopped subscribing. After that, I made a fix that indeed works, but I struggled to find the correct place for a test - it should request the translator proxy API. I've tried to add odd-sized extranonce [here](https://github.com/stratum-mining/stratum/blob/dev/test/message-generator/test/sv1-test.json#L23) but it doesn't lead to the test failure even without this patch. Where to add a test?

I'm not following you. Are you saying there's no issue at all?
From the changes you made, it doesn't seem to be a fix to something. 
If you could better explain in the PR description what is the issue and what is your fix it would be appreciated 🙏

nikicat:
There is an issue: when a miner sends `mining.subscribe` RPC request with an odd-sized second parameter (hex string like `abcde` or `112233445` ) then the default hex-decoder (`hex::decode`) fails to decode it because it expects an even-sized hex string (e.g. `aabbcc` or `123456`).
The code that patches incoming strings before passing them to the hex decoder to make them even-sized was already in the codebase for a long time, but it was not used in all the cases of parsing `extranonce`.
So, this patch modifies all the places where `extranonce` was passed to the `hex::decode` so that the even-sizer modification for the `extranonce` is applied.

Fi3:
Hi @nikicat ty for the patch. Can you also add a test that fail without the patch?

Fi3:
@nikicat you can add an unit test in the v1 crate  writing it from scratch is ok

nikicat:
@Fi3 done

Fi3:
Also remember to bump the patch version number in the translator and in the v1 lib

Fi3:

plebhash (approved):

PR #862, opened 2024-04-19
tproxy: panic when failed to load config

nikicat:
This patch adds a panic with an error description when tproxy fails to load the config file for any reason. It helps greatly to spot errors in the configuration file.

### DIFF
--- a/roles/translator/src/main.rs
+++ b/roles/translator/src/main.rs
@@ -41,7 +41,7 @@ async fn main() {
 
     let proxy_config = match process_cli_args() {
         Ok(p) => p,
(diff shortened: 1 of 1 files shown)


plebhash (approved):

plebhash:
hey @nikicat thanks for the multiple contributions over the past few days!

please join us on Discord, where we can keep in touch more closely: https://discord.gg/fsEW23wFYs

we also have a weekly dev call where your presence would be much appreciated!

https://discord.gg/FH8wJdcS?event=1232265523987742741

PR #866, opened 2024-04-22
Fix sv2 cpu miner hash rate calculation

Fi3:
In order to calculte the hash rate of the machine in which is running the sv2 cpu miner was using a fster function that the one that use to actually mine. This commit change the function that the cpu miner use to calculte the hash rate using the same one that is used when it mine.

### DIFF
--- a/roles/test-utils/mining-device/src/main.rs
+++ b/roles/test-utils/mining-device/src/main.rs
@@ -7,7 +7,6 @@ use std::{net::SocketAddr, sync::Arc, thread::sleep, time::Duration};
 use async_std::net::ToSocketAddrs;
 use clap::Parser;
 use rand::{thread_rng, Rng};
-use sha2::{Digest, Sha256};
 use std::time::Instant;
 use stratum_common::bitcoin::{
     blockdata::block::BlockHeader, hash_types::BlockHash, hashes::Hash, util::uint::Uint256,
@@ -110,7 +109,7 @@ async fn main() {
 }
 
 use async_channel::{Receiver, Sender};
-use binary_sv2::{u256_from_int, U256};
+use binary_sv2::u256_from_int;
 use codec_sv2::{Frame, Initiator, StandardEitherFrame, StandardSv2Frame};
 use roles_logic_sv2::{
     common_messages_sv2::{Protocol, SetupConnection, SetupConnectionSuccess},
@@ -236,7 +235,7 @@ pub struct Device {
 fn open_channel(device_id: Option<String>) -> OpenStandardMiningChannel<'static> {
     let user_identity = device_id.unwrap_or_default().try_into().unwrap();
     let id: u32 = 10;
-    info!("Measuring pc hashrate");
+    info!("Measuring CPU hashrate");
     let nominal_hash_rate = measure_hashrate(5) as f32;
     info!("Pc hashrate is {}", nominal_hash_rate);
     info!("MINING DEVICE: send open channel with request id {}", id);
@@ -290,7 +289,7 @@ impl Device {
         let handicap = miner.safe_lock(|m| m.handicap).unwrap();
         std::thread::spawn(move || loop {
             std::thread::sleep(std::time::Duration::from_micros(handicap.into()));
-            if miner.safe_lock(|m| m.next_share()).unwrap().is_ok() {
+            if miner.safe_lock(|m| m.next_share()).unwrap().is_valid() {
                 let nonce = miner.safe_lock(|m| m.header.unwrap().nonce).unwrap();
                 let time = miner.safe_lock(|m| m.header.unwrap().time).unwrap();
                 let job_id = miner.safe_lock(|m| m.job_id).unwrap();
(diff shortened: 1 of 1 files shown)


Fi3:
@Sjors can you be more specific about the helper function? Where do you thing that should be added?

Sjors:
I don't understand the code organisation well enough.

> using the same one that is used when it mine

At first glance it looks to me that we now have two functions that are copy-pasted, rather than one function that's called from two places. But maybe I missed something.

Fi3:
In `measure_hashrate` (the function modified by this code). We have:
* Initialize an arbitrary block header (most of the body)
* Call `miner.next()` for a `duration` (this incr the nonce and do the dsha)
* return `number of hashes calculated / duration`

Each of the above 3 part are very specific of this function, one could thing that the first part can be de-duplicated,  but in the function we create an header from some random vectors, in other parts of the code we compose them from SetPrevhash + NewJob messages. I don't think that could make much sense to have a function that do both things. Also would be very strange to create 2 messages SetPrevhash + NewJob in the `measure_hashrate` function only to derive an arbitrary block header when we could just create it right away and put it in the miner.

Not sure if this clarify your doubts? Or if I even answered your questions? Please let me know if your comment was about it or not,

Sjors:
Alright, let's keep it this way then. Is there something that needs testing, or does it just need code review? Maybe @plebhash & @lorbax can take a look?

Fi3:
Code review will be enough

Fi3:

roles/test-utils/mining-device/src/main.rs
```diff
@@ -614,33 +613,40 @@ impl Miner {
 
 // returns hashrate based on how fast the device hashes over the given duration
 fn measure_hashrate(duration_secs: u64) -> f64 {
-    let mut share = generate_random_80_byte_array();
+    let mut rng = thread_rng();
+    let prev_hash: [u8; 32] = generate_random_32_byte_array().to_vec().try_into().unwrap();
```
plebhash, line 617: so we are using a random number for the prevhash?
why?

I assume prevhash can be anything on this context, since all shares will be discarded anyways.

but maybe we should add a comment for making this explicitly clear?

roles/test-utils/mining-device/src/main.rs
```diff
@@ -614,33 +613,40 @@ impl Miner {
...
+            .unwrap()
+            .as_secs() as u32,
+        bits: rng.gen(),
+        nonce: 0,
+    };
     let start_time = Instant::now();
     let mut hashes: u64 = 0;
     let duration = Duration::from_secs(duration_secs);
+    let mut miner = Miner::new(0);
+    miner.new_target(vec![0_u8; 32]);
```
plebhash, line 636: why are we setting the target as 0?

doesn't this make is basically impossible to find a share?

roles/test-utils/mining-device/src/main.rs
```diff
@@ -140,14 +232,17 @@ pub struct Device {
     sequence_numbers: Id,
 }
 
-fn open_channel() -> OpenStandardMiningChannel<'static> {
-    let user_identity = "ABC".to_string().try_into().unwrap();
+fn open_channel(device_id: Option<String>) -> OpenStandardMiningChannel<'static> {
+    let user_identity = device_id.unwrap_or_default().try_into().unwrap();
     let id: u32 = 10;
-    println!("MINING DEVICE: send open channel with request id {}", id);
+    info!("Measuring pc hashrate");
```
plebhash, line 238:
```suggestion
    info!("Measuring CPU hashrate");
```

plebhash:
@Fi3 it seems this PR needs to be rebased after the `dev` branch cleanup?

roles/test-utils/mining-device/src/main.rs
```diff
@@ -140,14 +232,17 @@ pub struct Device {
...
+    let nominal_hash_rate = measure_hashrate(5) as f32;
+    info!("Pc hashrate is {}", nominal_hash_rate);
```
plebhash, line 240:
```suggestion
    info!("CPU hashrate is {} H/s", nominal_hash_rate);
```

roles/test-utils/mining-device/src/main.rs
```diff
@@ -499,3 +610,43 @@ impl Miner {
...
+    let mut hashes: u64 = 0;
+    let duration = Duration::from_secs(duration_secs);
+    let mut miner = Miner::new(0);
+    miner.new_target(vec![0_u8; 32]);
+    miner.header = Some(header);
+
+    while start_time.elapsed() < duration {
+        let _ = miner.next_share();
+        hashes += 1;
+    }
```
plebhash, line 642: why are we setting the target as 0?

doesn't this make it basically impossible to find a share?
```
difficulty = genesis_target / target = genesis_target / 0 = ∞
```

we are bypassing this restriction because we are ignoring the error returned by `miner.next_share()`, but that makes the code more confusing and counter-intuitive than it needs to be

but in order to improve readability and make a more clear usage of the API provided by `miner.next_share()`, I would refactor this snippet into something like:

```suggestion
    // set the max possible target (genesis hash), so that every share is valid
    let max_target = "00000000ffff0000000000000000000000000000000000000000000000000000";
    let target = hex::decode(max_target).expect("Decoding failed");
    miner.new_target(target);
    miner.header = Some(header);

    while start_time.elapsed() < duration {
        miner.next_share().expect("every share is valid under max_target");
        hashes += 1;
    }
```

but when I test these changes, I get a panic:
```
2024-05-03T20:48:04.540935Z  INFO mining_device: Set target to 0000000000000000000000000000000000000000000000000000ffff00000000
thread 'main' panicked at test-utils/mining-device/src/main.rs:644:28:
every share is valid under max_target: ()
```

which means the implementation of `fn next_share` is probably broken, probably on [this line](https://github.com/Fi3/stratum-1/blob/e4ad646c344d7ea124fbf651c575e4b956940bc3/roles/test-utils/mining-device/src/main.rs#L602):
```
if hash < *self.target.as_ref().ok_or(())?
```

plebhash (dismissed):
I thought I found a bug on `fn next_share`, but it was a false alarm

roles/test-utils/mining-device/src/main.rs
plebhash: this log line is also concerning:
```
2024-05-03T20:48:04.540935Z  INFO mining_device: Set target to 0000000000000000000000000000000000000000000000000000ffff00000000
```

why is the target hash being displayed backwards? maybe an indication of the root cause of the bug?

plebhash, line 642: why are we setting the target as 0?

doesn't this make it basically impossible to find a share?
```
difficulty = genesis_target / target = genesis_target / 0 = ∞
```

we are bypassing this restriction because we are ignoring the error returned by `miner.next_share()`, but that makes the code more confusing and counter-intuitive than it needs to be

but in order to improve readability and make a more clear usage of the API provided by `miner.next_share()`, I would refactor this snippet into something like:

```suggestion
    // set the max possible target, so that every share is valid
    let max_target = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
    let target = hex::decode(max_target).expect("Decoding failed");
    miner.new_target(target);
    miner.header = Some(header);

    while start_time.elapsed() < duration {
        miner.next_share().expect("every share is valid under max_target");
        hashes += 1;
    }
```

the downside for this approach is that now `next_share` will spam the console:
```
2024-05-03T20:59:11.630062Z  INFO mining_device: Found share with nonce: 0, for target: Some(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff), with hash: 0x03986db2375a8106218561993e202c09b8ed969157417664ccba54c1ef8c5e6e
2024-05-03T20:59:11.630065Z  INFO mining_device: Found share with nonce: 0, for target: Some(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff), with hash: 0x03986db2375a8106218561993e202c09b8ed969157417664ccba54c1ef8c5e6e
2024-05-03T20:59:11.630082Z  INFO mining_device: Found share with nonce: 0, for target: Some(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff), with hash: 0x03986db2375a8106218561993e202c09b8ed969157417664ccba54c1ef8c5e6e
...
```

so `next_share` could probably have a filter for [the line that prints this](https://github.com/Fi3/stratum-1/blob/e4ad646c344d7ea124fbf651c575e4b956940bc3/roles/test-utils/mining-device/src/main.rs#L603):
```diff
-        if hash < *self.target.as_ref().ok_or(())? {
-            info!(
-                "Found share with nonce: {}, for target: {:?}, with hash: {:?}",
-                header.nonce, self.target, hash,
-            );
+        let target = *self.target.as_ref().ok_or(())?;
+        let max_target_hashrate_measurement =
+            Uint256::from_be_bytes(vec![255_u8; 32].try_into().unwrap());
+        if hash < target {
+            if target < max_target_hashrate_measurement {
+                info!(
+                    "Found share with nonce: {}, for target: {:?}, with hash: {:?}",
+                    header.nonce, self.target, hash,
+                );
+            }

```

roles/test-utils/mining-device/src/main.rs
Fi3: better to use random so you are sure that there wont be optimization.

roles/test-utils/mining-device/src/main.rs
Fi3: we don't car if share are valid or not we just want to execute the dsha function in the time interval and count how many time we are doing that so we know the hashrate of the machine that is running this cpuminer.

Fi3: the majior issue with the above code is that next_share return an Result (for sake of simplicity) instead of a custom type like enum ValidShare | InvalidShare

Fi3: i'm going do push something next week

plebhash: ok

it took me some time but I ended up understanding everything is going on

the suggested change is just a small cosmetic change to improve code readability, but not essential

feel free to ignore this and merge if you wish

plebhash (approved):
I made some suggestions for minor improvements, but they're not essential. I'll approve and leave it up to @Fi3  to decide if he wants to merge it right away or implement those improvements before doing so.

Fi3:
> I made some suggestions for minor improvements, but they're not essential. I'll approve and leave it up to @Fi3 to decide if he wants to merge it right away or implement those improvements before doing so.

your suggestion should be already addressed

PR #867, opened 2024-04-23
Make mg GitHub actions more stable moving compilation before running tests

lorbax:
In the message-generator.sh script the binaries are pre-compiled so that the MG-tests do not take forever for run, reducing probability of timeouts

### DIFF
--- a/message-generator-tests.sh
+++ b/message-generator-tests.sh
@@ -6,17 +6,33 @@ else
     RUST_LOG="debug"
 fi
 
-search_dir="../../test/message-generator/test/"
 message_generator_dir="./utils/message-generator/"
-
 cd $message_generator_dir
 cargo llvm-cov clean
+
+cd ../../roles
+cargo build -p mining-device
+cargo llvm-cov -p pool_sv2
+cargo llvm-cov -p jd_server
+cargo llvm-cov -p jd_client
+cargo llvm-cov -p translator_sv2 
+cargo llvm-cov -p mining_proxy_sv2 
+cd ./roles-utils
+cargo llvm-cov 
+cd ../../
+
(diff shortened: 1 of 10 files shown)


Fi3:
I noticed that you added llvm-cov everywhere can you explain what it do and why il called with --no-report? Also having it documented somewhere in the project would be nice. Ty

lorbax:
> > Hi you are actually right. It is called with --no-report everywhere except [here](https://github.com/stratum-mining/stratum/blob/dev/test/message-generator/test/pool-sri-test-reject-auth.json#L101). I am not practical with llvm-cov and I don't know if and when we need it. I used llvm-cov everywhere in order to use the same binaries, avoiding recompiling whenever possible.
> 
> @lorbax ok since no one is practical with it I think that we should add some documentation about it in this PR and justify why we use it.

I will spend some time with it and understand what should we do with it.

Fi3:

message-generator-tests.sh
```diff
@@ -6,18 +6,34 @@ else
...
+cd $message_generator_dir
+cargo build
+
 for entry in `ls $search_dir`; do
     if [ "$entry" = "interop-jdc-change-upstream.json" ]; then
         echo "Skipping $entry"
         continue
     fi
 
+    sleep 10
```
plebhash, line 35: maybe add a comment justifying this sleep?

lorbax: in order to give time to every process to close successfully

plebhash: my suggestion was to add a comment to the shell script

lorbax:
replaced by https://github.com/stratum-mining/stratum/pull/905

PR #868, opened 2024-04-23
Custom transactions from JDC removed from mempool during update

### LINKED ISSUE #840: Custom transactions from JDC removed from mempool during update
The declared transaction in declared mining job are stored as a whole in the JDS mempool. The transactions in in declared jobs are recognized because they have this form (like `txid: Txid <-- Some(Transaction)` in the hashmap). In `JDsMemppol::update_mempool()` we create a new hashmap `mempool_ordered`. Inside the task spawned, the node's mempool is stored in in the variable `mempool: Vec<Txid>`. For every entry `txid` of this vector, if there was a transaction in the mutex mempool of the form `txid: Txid <-- Some(Transaction)`, then the whole transaction is stored in `mempool_ordered`. The other transactions in `mempool` are stored in `mempool_ordered` as `txid: Txid <-- None`. Then JDS mempool is updated with `mempool_ordered`
[https://github.com/stratum-mining/stratum/pull/833](https://github.com/stratum-mining/stratum/pull/833)
As a consequence of that, if the downstream declares a transaction which is unknown to the node, then it gets lost during the update of the JDS mempool. If the mempool is updated frequently this may prevent to the block propagation for the JDS, which is not good.

If this is fixed, we must also remove the transactions of old declared jobs, otherwise the there is (very remote) risk of overflow.

Fix this issue after merging of [PR833.](https://github.com/stratum-mining/stratum/pull/833)

NonsoAmadi10:
### What does this PR do?
- Closes #840 
- It reads transactions from the jds mempool and the node's mempool, compares them and adds them to the new mempool that is updated. In cases where the txids are unknown it adds the fat transactions to the updated mempool as well.
- To prevent an overflow, old fat transactions are cleared before a new minning job is declared

### DIFF
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -42,49 +42,54 @@ impl JobDeclaratorDownstream {
 }
 
 pub fn clear_declared_mining_job(
-    mining_job: DeclareMiningJob,
+    old_mining_job: DeclareMiningJob,
+    new_mining_job: &DeclareMiningJob,
     mempool: Arc<Mutex<JDsMempool>>,
 ) -> Result<(), Error> {
-    // If there is an old declared mining job, remove its transactions from the mempool
-    // Retrieve necessary data from the old job
-    let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
-    if transactions_to_remove.is_empty() {
+    let old_transactions = old_mining_job.tx_short_hash_list.inner_as_ref();
+    let new_transactions = new_mining_job.tx_short_hash_list.inner_as_ref();
+
+    if old_transactions.is_empty() {
         info!("No transactions to remove from mempool");
         return Ok(());
     }
 
-    let nonce = mining_job.tx_short_hash_nonce;
+    let nonce = old_mining_job.tx_short_hash_nonce;
 
-    for short_id in transactions_to_remove {
-        let result = mempool.safe_lock(|mempool_| -> Result<(), Error> {
-            // Try to manage this unwrap, we use .ok_or() method to return the proper error
-            let short_ids_map = mempool_
-                .to_short_ids(nonce)
-                .ok_or(Error::JDSMissingTransactions)?;
-            let transaction_with_hash = short_ids_map
-                .get(short_id);
+    let result = mempool.safe_lock(|mempool_| -> Result<(), Error> {
+        let short_ids_map = mempool_
+            .to_short_ids(nonce)
+            .ok_or(Error::JDSMissingTransactions)?;
 
-            match transaction_with_hash {
-                Some(transaction_with_hash) => {
+        for short_id in old_transactions {
+            if !new_transactions.contains(&short_id) {
+                if let Some(transaction_with_hash) = short_ids_map.get(short_id) {
                     let txid = transaction_with_hash.id;
-                    match mempool_.mempool.remove(&txid) {
-                        Some(transaction) => {
-                            debug!("Fat transaction {:?} in job with request id {:?} removed from mempool", transaction, mining_job.request_id);
-                            info!("Fat transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id);
+                    match mempool_.mempool.get_mut(&txid) {
+                        Some(Some((transaction, counter))) => {
+                            if *counter > 1 {
+                                *counter -= 1;
+                                info!("Fat transaction {:?} decreased mempool counter because job id {:?} was dropped", txid, old_mining_job.request_id);
+                            } else {
+                                mempool_.mempool.remove(&txid);
+                                info!("Fat transaction {:?} in job with request id {:?} removed from mempool", txid, old_mining_job.request_id);
+                            }
                         },
-                        None => info!("Thin transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id),
+                        Some(None) => info!("Thin transaction {:?} in job with request id {:?} removed from mempool", txid, old_mining_job.request_id),
+                        None => {},
                     }
-                },
-                None => debug!("Transaction with short id {:?} not found in mempool while clearing old jobs", short_id),
+                } else {
+                    debug!("Transaction with short id {:?} not found in mempool while clearing old jobs", short_id);
+                }
             }
-            Ok(())  // Explicitly return Ok(()) inside the closure for proper flow control
-        });
-
-        // Propagate any error from the closure
-        if let Err(err) = result {
-            return Err(Error::PoisonLock(err.to_string()));
         }
+        Ok(())
+    });
+
+    if let Err(err) = result {
+        return Err(Error::PoisonLock(err.to_string()));
     }
+
     Ok(())
 }
 
@@ -110,10 +115,12 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
         Ok(SendTo::Respond(message_enum))
     }
 
-    fn handle_declare_mining_job(&mut self, message: DeclareMiningJob) -> Result<SendTo, Error> {
-        // Clone the old declared mining job to retain its data
-        if let Some(old_declare_mining_job_) = self.declared_mining_job.0.clone() {
-            clear_declared_mining_job(old_declare_mining_job_, self.mempool.clone())?;
+    fn handle_declare_mining_job(
+        &mut self,
+        new_mining_job: DeclareMiningJob,
+    ) -> Result<SendTo, Error> {
+        if let Some(old_mining_job) = self.declared_mining_job.0.take() {
+            clear_declared_mining_job(old_mining_job, &new_mining_job, self.mempool.clone())?;
         }
 
         // the transactions that are present in the mempool are stored here, that is sent to the
@@ -121,15 +128,15 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
         // The unknown transactions is a vector that contains the transactions that are not in the
         // jds mempool, and will be non-empty in the ProvideMissingTransactionsSuccess message
         let mut known_transactions: Vec<Txid> = vec![];
-        self.tx_hash_list_hash = Some(message.tx_hash_list_hash.clone().into_static());
-        if self.verify_job(&message) {
-            let short_hash_list: Vec<ShortTxId> = message
+        self.tx_hash_list_hash = Some(new_mining_job.tx_hash_list_hash.clone().into_static());
+        if self.verify_job(&new_mining_job) {
+            let short_hash_list: Vec<ShortTxId> = new_mining_job
                 .tx_short_hash_list
                 .inner_as_ref()
                 .iter()
                 .map(|x| x.to_vec().try_into().unwrap())
                 .collect();
-            let nonce = message.tx_short_hash_nonce;
+            let nonce = new_mining_job.tx_short_hash_nonce;
             // TODO return None when we have a collision handle that case as weel
             let short_id_mempool = self
                 .mempool
@@ -154,7 +161,7 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
                 }
             }
             self.declared_mining_job = (
-                Some(message.clone().into_static()),
+                Some(new_mining_job.clone().into_static()),
                 transactions_with_state,
                 missing_txs.clone(),
             );
@@ -167,9 +174,9 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
 
             if missing_txs.is_empty() {
                 let message_success = DeclareMiningJobSuccess {
-                    request_id: message.request_id,
+                    request_id: new_mining_job.request_id,
                     new_mining_job_token: signed_token(
-                        message.tx_hash_list_hash.clone(),
+                        new_mining_job.tx_hash_list_hash.clone(),
                         &self.public_key.clone(),
                         &self.private_key.clone(),
                     ),
@@ -178,7 +185,7 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
                 Ok(SendTo::Respond(message_enum_success))
             } else {
                 let message_provide_missing_transactions = ProvideMissingTransactions {
-                    request_id: message.request_id,
+                    request_id: new_mining_job.request_id,
                     unknown_tx_position_list: missing_txs.into(),
                 };
                 let message_enum_provide_missing_transactions =
@@ -189,7 +196,7 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
             }
         } else {
             let message_error = DeclareMiningJobError {
-                request_id: message.request_id,
+                request_id: new_mining_job.request_id,
                 error_code: Vec::new().try_into().unwrap(),
                 error_details: Vec::new().try_into().unwrap(),
             };
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -17,7 +17,7 @@ use crate::mempool::JDsMempool;
 use super::{signed_token, TransactionState};
 use roles_logic_sv2::{errors::Error, parsers::PoolMessages as AllMessages};
 use stratum_common::bitcoin::consensus::Decodable;
-use tracing::{info, debug};
+use tracing::{debug, info};
 
 use super::JobDeclaratorDownstream;
 
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -52,18 +52,40 @@ pub fn clear_declared_mining_job(
         info!("No transactions to remove from mempool");
         return Ok(());
     }
-    let clear_transactions = |jds_mempool: &mut JDsMempool| {
-        for txid in transactions_to_remove {
-            match jds_mempool.mempool.remove(txid) {
-                Some(_) => info!("Transaction {:?} removed from mempool", txid),
-                None => debug!("Transaction {:?} not found in mempool", txid),
-            };
+
+    let nonce = mining_job.tx_short_hash_nonce;
+
+    for short_id in transactions_to_remove {
+        let result = mempool.safe_lock(|mempool_| -> Result<(), Error> {
+            // Try to manage this unwrap, we use .ok_or() method to return the proper error
+            let short_ids_map = mempool_
+                .to_short_ids(nonce)
+                .ok_or(Error::JDSMissingTransactions)?;
+            let transaction_with_hash = short_ids_map
+                .get(short_id);
+
+            match transaction_with_hash {
+                Some(transaction_with_hash) => {
+                    let txid = transaction_with_hash.id;
+                    match mempool_.mempool.remove(&txid) {
+                        Some(transaction) => {
+                            debug!("Fat transaction {:?} in job with request id {:?} removed from mempool", transaction, mining_job.request_id);
+                            info!("Fat transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id);
+                        },
+                        None => info!("Thin transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id),
+                    }
+                },
+                None => debug!("Transaction with short id {:?} not found in mempool while clearing old jobs", short_id),
+            }
+            Ok(())  // Explicitly return Ok(()) inside the closure for proper flow control
+        });
+
+        // Propagate any error from the closure
+        if let Err(err) = result {
+            return Err(Error::PoisonLock(err.to_string()));
         }
-    };
-    match mempool.safe_lock(clear_transactions) {
-        Ok(_) => Ok(()),
-        Err(e) => Err(Error::PoisonLock(e.to_string())),
     }
+    Ok(())
 }
 
 impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -111,11 +111,9 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
     }
 
     fn handle_declare_mining_job(&mut self, message: DeclareMiningJob) -> Result<SendTo, Error> {
-        {
-            // Clone the old declared mining job to retain its data
-            if let Some(old_declare_mining_job_) = self.declared_mining_job.0.clone() {
-                clear_declared_mining_job(old_declare_mining_job_, self.mempool.clone())?;
-            }
+        // Clone the old declared mining job to retain its data
+        if let Some(old_declare_mining_job_) = self.declared_mining_job.0.clone() {
+            clear_declared_mining_job(old_declare_mining_job_, self.mempool.clone())?;
         }
 
         // the transactions that are present in the mempool are stored here, that is sent to the
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -7,14 +7,17 @@ use roles_logic_sv2::{
         ProvideMissingTransactions, ProvideMissingTransactionsSuccess, SubmitSolutionJd,
     },
     parsers::JobDeclaration,
+    utils::Mutex,
 };
-use std::{convert::TryInto, io::Cursor};
+use std::{convert::TryInto, io::Cursor, sync::Arc};
 use stratum_common::bitcoin::{Transaction, Txid};
 pub type SendTo = SendTo_<JobDeclaration<'static>, ()>;
+use crate::mempool::JDsMempool;
+
 use super::{signed_token, TransactionState};
 use roles_logic_sv2::{errors::Error, parsers::PoolMessages as AllMessages};
 use stratum_common::bitcoin::consensus::Decodable;
-use tracing::info;
+use tracing::{info, debug};
 
 use super::JobDeclaratorDownstream;
 
@@ -35,6 +38,31 @@ impl JobDeclaratorDownstream {
         // 3. right prev-hash
         // 4. right nbits
         self.token_to_job_map.contains_key(&(token_u32))
+    }
+}
+
+pub fn clear_declared_mining_job(
+    mining_job: DeclareMiningJob,
+    mempool: Arc<Mutex<JDsMempool>>,
+) -> Result<(), Error> {
+    // If there is an old declared mining job, remove its transactions from the mempool
+    // Retrieve necessary data from the old job
+    let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
+    if transactions_to_remove.is_empty() {
+        info!("No transactions to remove from mempool");
+        return Ok(());
+    }
+    let clear_transactions = |jds_mempool: &mut JDsMempool| {
+        for txid in transactions_to_remove {
+            match jds_mempool.mempool.remove(txid) {
+                Some(_) => info!("Transaction {:?} removed from mempool", txid),
+                None => debug!("Transaction {:?} not found in mempool", txid),
+            };
+        }
+    };
+    match mempool.safe_lock(clear_transactions) {
+        Ok(_) => Ok(()),
+        Err(e) => Err(Error::PoisonLock(e.to_string())),
     }
 }
 
@@ -61,6 +89,13 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
     }
 
     fn handle_declare_mining_job(&mut self, message: DeclareMiningJob) -> Result<SendTo, Error> {
+        {
+            // Clone the old declared mining job to retain its data
+            if let Some(old_declare_mining_job_) = self.declared_mining_job.0.clone() {
+                clear_declared_mining_job(old_declare_mining_job_, self.mempool.clone())?;
+            }
+        }
+
         // the transactions that are present in the mempool are stored here, that is sent to the
         // mempool which use the rpc client to retrieve the whole data for each transaction.
         // The unknown transactions is a vector that contains the transactions that are not in the
--- a/roles/jd-server/src/lib/mempool/mod.rs
+++ b/roles/jd-server/src/lib/mempool/mod.rs
@@ -12,12 +12,12 @@ use stratum_common::bitcoin::{self, hash_types::Txid};
 #[derive(Clone, Debug)]
 pub struct TransactionWithHash {
     pub id: Txid,
-    pub tx: Option<Transaction>,
+    pub tx: Option<(Transaction, u64)>,
 }
 
 #[derive(Clone, Debug)]
 pub struct JDsMempool {
-    pub mempool: HashMap<Txid, Option<Transaction>>,
+    pub mempool: HashMap<Txid, Option<(Transaction, u64)>>,
     auth: mini_rpc_client::Auth,
     url: String,
     new_block_receiver: Receiver<String>,
@@ -50,7 +50,7 @@ impl JDsMempool {
         new_block_receiver: Receiver<String>,
     ) -> Self {
         let auth = mini_rpc_client::Auth::new(username, password);
-        let empty_mempool: HashMap<Txid, Option<Transaction>> = HashMap::new();
+        let empty_mempool: HashMap<Txid, Option<(Transaction, u64)>> = HashMap::new();
         JDsMempool {
             mempool: empty_mempool,
             auth,
@@ -82,14 +82,15 @@ impl JDsMempool {
                     .get_raw_transaction(&txid.to_string(), None)
                     .await
                     .map_err(JdsMempoolError::Rpc)?;
-                let _ =
-                    self_.safe_lock(|a| a.mempool.insert(transaction.txid(), Some(transaction)));
+                let _ = self_
+                    .safe_lock(|a| a.mempool.insert(transaction.txid(), Some((transaction, 1))));
             }
         }
 
         // fill in the mempool the transactions given in input
         for transaction in transactions {
-            let _ = self_.safe_lock(|a| a.mempool.insert(transaction.txid(), Some(transaction)));
+            let _ =
+                self_.safe_lock(|a| a.mempool.insert(transaction.txid(), Some((transaction, 1))));
         }
         Ok(())
     }
--- a/roles/jd-server/src/lib/mempool/mod.rs
+++ b/roles/jd-server/src/lib/mempool/mod.rs
@@ -99,12 +99,13 @@ impl JDsMempool {
             .safe_lock(|x| x.get_client())?
             .ok_or(JdsMempoolError::NoClient)?;
         let node_mempool: Vec<String> = client.get_raw_mempool().await?;
-        let node_mempool_deserialized: Result<Vec<Txid>, JdsMempoolError> = node_mempool
-            .iter()
-            .map(|id| Txid::from_str(id).map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string()))))
-            .collect();
-        let node_mempool_deserialized = node_mempool_deserialized?;
-        
+        let mut node_mempool_deserialized: Vec<Txid> = vec![];
+        for id in &node_mempool {
+            let key_id = Txid::from_str(id)
+                .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
+            node_mempool_deserialized.push(key_id);
+        }
+
         self_.safe_lock(|x| {
             let jds_mempool = &mut x.mempool;
             // the fat transactions in the jds-mempool are those declared by some downstream and we
@@ -113,7 +114,7 @@ impl JDsMempool {
             // here we add all the new transactions
             for id in &node_mempool_deserialized {
                 jds_mempool.entry(*id).or_insert(None);
-            };
+            }
             if jds_mempool.is_empty() {
                 Err(JdsMempoolError::EmptyMempool)
             } else {
--- a/roles/jd-server/src/lib/mempool/mod.rs
+++ b/roles/jd-server/src/lib/mempool/mod.rs
@@ -95,12 +95,11 @@ impl JDsMempool {
     }
 
     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
-
-        let mut new_jds_mempool: HashMap<Txid, Option<Transaction>> = self_.safe_lock(|x| x.mempool.clone())?;
+        let mut new_jds_mempool: HashMap<Txid, Option<Transaction>> =
+            self_.safe_lock(|x| x.mempool.clone())?;
         // the fat transactions in the jds-mempool are those declared by some downstream and we
         // don't want to remove them, but we can get rid of the others
         new_jds_mempool.retain(|_, val| val.is_some());
-
 
         let client = self_
             .safe_lock(|x| x.get_client())?
--- a/roles/jd-server/src/lib/mempool/mod.rs
+++ b/roles/jd-server/src/lib/mempool/mod.rs
@@ -7,7 +7,7 @@ use hashbrown::HashMap;
 use roles_logic_sv2::utils::Mutex;
 use rpc_sv2::{mini_rpc_client, mini_rpc_client::RpcError};
 use std::{convert::TryInto, str::FromStr, sync::Arc};
-use stratum_common::{bitcoin, bitcoin::hash_types::Txid};
+use stratum_common::bitcoin::{self, hash_types::Txid};
 
 #[derive(Clone, Debug)]
 pub struct TransactionWithHash {
@@ -95,29 +95,29 @@ impl JDsMempool {
     }
 
     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
-        let mut mempool_ordered: HashMap<Txid, Option<Transaction>> = HashMap::new();
+
+        let mut new_jds_mempool: HashMap<Txid, Option<Transaction>> = self_.safe_lock(|x| x.mempool.clone())?;
+        // the fat transactions in the jds-mempool are those declared by some downstream and we
+        // don't want to remove them, but we can get rid of the others
+        new_jds_mempool.retain(|_, val| val.is_some());
+
 
         let client = self_
             .safe_lock(|x| x.get_client())?
             .ok_or(JdsMempoolError::NoClient)?;
-
-        let mempool: Vec<String> = client.get_raw_mempool().await?;
-        for id in &mempool {
+        let node_mempool: Vec<String> = client.get_raw_mempool().await?;
+        // here we add all the new transactions
+        for id in &node_mempool {
             let key_id = Txid::from_str(id)
                 .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
-
-            let tx = self_.safe_lock(|x| match x.mempool.get(&key_id) {
-                Some(entry) => entry.clone(),
-                None => None,
-            })?;
-
-            mempool_ordered.insert(key_id, tx);
+            // not sure if correct, check it!
+            new_jds_mempool.entry(key_id).or_insert(None);
         }
 
-        if mempool_ordered.is_empty() {
+        if new_jds_mempool.is_empty() {
             Err(JdsMempoolError::EmptyMempool)
         } else {
-            let _ = self_.safe_lock(|x| x.mempool = mempool_ordered);
+            let _ = self_.safe_lock(|x| x.mempool = new_jds_mempool);
             Ok(())
         }
     }
--- a/roles/jd-server/src/lib/mempool/mod.rs
+++ b/roles/jd-server/src/lib/mempool/mod.rs
@@ -95,30 +95,31 @@ impl JDsMempool {
     }
 
     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
-        let mut new_jds_mempool: HashMap<Txid, Option<Transaction>> =
-            self_.safe_lock(|x| x.mempool.clone())?;
-        // the fat transactions in the jds-mempool are those declared by some downstream and we
-        // don't want to remove them, but we can get rid of the others
-        new_jds_mempool.retain(|_, val| val.is_some());
-
         let client = self_
             .safe_lock(|x| x.get_client())?
             .ok_or(JdsMempoolError::NoClient)?;
         let node_mempool: Vec<String> = client.get_raw_mempool().await?;
-        // here we add all the new transactions
-        for id in &node_mempool {
-            let key_id = Txid::from_str(id)
-                .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
-            // not sure if correct, check it!
-            new_jds_mempool.entry(key_id).or_insert(None);
-        }
-
-        if new_jds_mempool.is_empty() {
-            Err(JdsMempoolError::EmptyMempool)
-        } else {
-            let _ = self_.safe_lock(|x| x.mempool = new_jds_mempool);
-            Ok(())
-        }
+        let node_mempool_deserialized: Result<Vec<Txid>, JdsMempoolError> = node_mempool
+            .iter()
+            .map(|id| Txid::from_str(id).map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string()))))
+            .collect();
+        let node_mempool_deserialized = node_mempool_deserialized?;
+        
+        self_.safe_lock(|x| {
+            let jds_mempool = &mut x.mempool;
+            // the fat transactions in the jds-mempool are those declared by some downstream and we
+            // don't want to remove them, but we can get rid of the others
+            jds_mempool.retain(|_, val| val.is_some());
+            // here we add all the new transactions
+            for id in &node_mempool_deserialized {
+                jds_mempool.entry(*id).or_insert(None);
+            };
+            if jds_mempool.is_empty() {
+                Err(JdsMempoolError::EmptyMempool)
+            } else {
+                Ok(())
+            }
+        })?
     }
 
     pub async fn on_submit(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
--- a/roles/jd-server/src/lib/job_declarator/mod.rs
+++ b/roles/jd-server/src/lib/job_declarator/mod.rs
@@ -136,7 +136,7 @@ impl JobDeclaratorDownstream {
         let mut transactions_list: Vec<Transaction> = Vec::new();
         for tx_with_state in transactions_with_state.iter().enumerate() {
             if let TransactionState::PresentInMempool(txid) = tx_with_state.1 {
-                let tx = mempool
+                let (tx, _) = mempool
                     .safe_lock(|x| x.mempool.get(txid).cloned())
                     .map_err(|e| JdsError::PoisonLock(e.to_string()))?
                     .ok_or(Box::new(JdsError::ImpossibleToReconstructBlock(


Fi3:

NonsoAmadi10:
@lorbax I have made the required changes, I welcome your thoughts

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -102,6 +102,18 @@ impl JDsMempool {
             .ok_or(JdsMempoolError::NoClient)?;
 
         let mempool: Vec<String> = client.get_raw_mempool().await?;
+
```
lorbax, line 105: you may want to change this function with something like this
```
    pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {

        let mut new_jds_mempool: HashMap<Txid, Option<Transaction>> = self_.safe_lock(|x| x.mempool.clone())?;
        // the fat transactions in the jds-mempool are those declared by some downstream and we
        // don't want to remove them, but we can get rid of the others
        // TODO remove this clone!!!!
        for (key, val) in new_jds_mempool.clone().iter() {
            if let None = val {
                new_jds_mempool.remove(key);
            }
        }

        let client = self_
            .safe_lock(|x| x.get_client())?
            .ok_or(JdsMempoolError::NoClient)?;
        let node_mempool: Vec<String> = client.get_raw_mempool().await?;
        // here we add all the new transactions
        for id in &node_mempool {
            let key_id = Txid::from_str(id)
                .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
            // not sure if correct, check it!
            new_jds_mempool.entry(key_id).or_insert(None);
        }

        if new_jds_mempool.is_empty() {
            Err(JdsMempoolError::EmptyMempool)
        } else {
            let _ = self_.safe_lock(|x| x.mempool = new_jds_mempool);
            Ok(())
        }
    }

```
this because
1. keep the fat txs which are those that appear in some DeclareMiningJob
2. get rid of all other thin transactions
3. add the new transactions retrieved from the bitcoin node

lorbax (changes requested):
Need rebase and change the code logic

plebhash:
@NonsoAmadi10 would you be willing to work on https://github.com/stratum-mining/stratum/issues/955 as well?

we are trying to implement a practice in the SV2 development community where every bug coming from protocol edge cases needs to be mapped into MG tests.

we have some video tutorials teaching how to write MG tests:
- https://youtu.be/gKCCq5EVxOc?si=Wlq1Fi-n9L9y4ftc
- https://youtu.be/rNdm9Q4KTWE?si=uYayax6EIY3UScqd

NonsoAmadi10:
> @NonsoAmadi10 would you be willing to work on #955 as well?
> 
> we are trying to implement a practice in the SV2 development community where every bug coming from protocol edge cases needs to be mapped into MG tests
> 
> that is strategically important for the robustness of SRI
> 
> here's a link for a Github Discussion on this topic: #944
> 
> we have some video tutorials teaching how to write MG tests:
> 
> * https://www.youtube.com/playlist?list=PL4okm1s0ZpT9YveJBiU51gaaOmNJviugi
> * https://www.youtube.com/playlist?list=PL4okm1s0ZpT8gWyM0UNKNbFoy5-djIheg

Sure I would love to take a go at it. Sign me up!! Lol

plebhash (changes requested):
CI seems broken on rustfmt

please do the following

run:
```
cargo +nightly fmt
```

on all crates included in the CI errors, commit and push again

NonsoAmadi10:
I have formatted now @plebhash

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -61,6 +62,61 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
     }
 
     fn handle_declare_mining_job(&mut self, message: DeclareMiningJob) -> Result<SendTo, Error> {
+        {
```
lorbax, line 65: put this into a function that takes input a the OldDeclareMiningJob and the Arc<Mutex<JdsMempool>> and removes the old transactions from there

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -61,6 +62,61 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
...
+                let old_nonce = old_job.tx_short_hash_nonce;
+
+                // Get the mempool of the old transaction IDs
+                let old_short_id_mempool =
+                    match self.mempool.safe_lock(|x| x.to_short_ids(old_nonce)) {
+                        // If successfully retrieved, store it
+                        Ok(Some(mempool)) => mempool,
+                        // If the mempool is not found, return an error
+                        Ok(None) => {
+                            return Err(roles_logic_sv2::Error::PoisonLock(
```
lorbax, line 82: Do not return a poison lock here

lorbax (commented):
Needs some minor fixes

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: What error class should i return please? A regular error string?

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: Alright

roles/jd-server/src/lib/job_declarator/message_handler.rs
lorbax: I thought about.
[Few lines](https://github.com/stratum-mining/stratum/blob/main/roles/jd-server/src/lib/job_declarator/message_handler.rs#L82) below there already a double unwrap. I think that fix that is out of the scope of this PR, so imo you can double unwrap, as long as you open an issue for that.
So
```
                // Get the mempool of the old transaction IDs
                let old_short_id_mempool = self
                    .mempool
                    .safe_lock(|x| x.to_short_ids(old_nonce))
                    .unwrap()
                    .unwrap();
```

lorbax (approved):

GitGab19:
@NonsoAmadi10 @lorbax haven't you decided to add a MG test for this?
I'm trying to understand it from discussion here https://github.com/stratum-mining/stratum/issues/955

NonsoAmadi10:
> @NonsoAmadi10 @lorbax haven't you decided to add a MG test for this? I'm trying to understand it from discussion here #955

Hi @GitGab19 I promised to open an issue later that writes the MG test as well as fix issue #956

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -95,29 +95,30 @@ impl JDsMempool {
...
             let key_id = Txid::from_str(id)
                 .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
-
-            let tx = self_.safe_lock(|x| match x.mempool.get(&key_id) {
-                Some(entry) => entry.clone(),
-                None => None,
-            })?;
-
-            mempool_ordered.insert(key_id, tx);
+            new_jds_mempool.entry(key_id).or_insert(None);
```
GitGab19, line 115: Where is the tx data inserted?
Maybe I'm missing something

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -95,29 +95,30 @@ impl JDsMempool {
     }
 
     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
-        let mut mempool_ordered: HashMap<Txid, Option<Transaction>> = HashMap::new();
+        let mut new_jds_mempool: HashMap<Txid, Option<Transaction>> =
+            self_.safe_lock(|x| x.mempool.clone())?;
+
+        // the fat transactions in the jds-mempool are those declared by some downstream and we
+        // don't want to remove them, but we can get rid of transactions that are none
+        new_jds_mempool.retain(|_, val| val.is_some());
```
GitGab19, line 103: What if the txs declared in the job are still not inserted in the mempool, but this function is executed? Are they gonna by deleted or not?

roles/jd-server/src/lib/mempool/mod.rs
NonsoAmadi10: Oh I am going to make modifications to it now, I don't know how I missed it. The idea was not adding the data if the key_id exist, but I see that was not what was pushed. I apologise

roles/jd-server/src/lib/mempool/mod.rs
NonsoAmadi10: If a transaction has not yet been inserted and its value is None, this line would remove it from the mempool

NonsoAmadi10: would you suggest we refrain from removing transactions that are none? I am doing this so that there is not an overflow in the jds mempool of old transactions

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,41 @@ impl JobDeclaratorDownstream {
     }
 }
 
+pub fn clear_old_mempool_transactions(
+    old_declare_mining_job: Option<DeclareMiningJob>,
```
jbesraa, line 45: I don't think you need the `Option` here

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,41 @@ impl JobDeclaratorDownstream {
...
+        // Get the mempool of the old transaction IDs
+        let old_short_id_mempool = mempool
+            .safe_lock(|x| x.to_short_ids(old_nonce))
+            .unwrap()
+            .unwrap();
+
+        // Convert old transaction IDs to short IDs
+        let old_short_hash_list: Vec<[u8; 6]> = old_tx_hash_list
+            .iter()
+            .map(|x| x.to_vec().try_into().unwrap())
```
jbesraa, line 63: Lets avoid this `unwrap`

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,41 @@ impl JobDeclaratorDownstream {
     }
 }
 
+pub fn clear_old_mempool_transactions(
```
jbesraa, line 44: I am a bit on the fence with this name. The struct is `DeclareMiningJob`  so maybe it would make more sense to change this to `fn clear_old_mining_jobs`. Thinking about it, why not adding this function in a 
```rust
impl DeclareMiningJob { 
  pub(crate) fn clear_from_mempool(&self, mempool: Arc<Mutex<..>>) {} 
}
```

jbesraa, line 44: Could we compare old tx list and new tx list and only remove whats out then? that could be more efficient if there is enough similarity between the mining jobs

jbesraa (commented):
Looks good. Added a few comments.

roles/jd-server/src/lib/mempool/mod.rs
GitGab19: Aren't you managing the old transactions removal with the clear_old_mempool_transactions? Is it not enough in your opinion?

NonsoAmadi10: The clear_old_mempool_transactions handles the removal of old transactions, ensuring the mempool is kept clean and up-to-date. While this retain function would  ensures that only transactions that have been fully fetched and are valid remain in the mempool. Isn't this what we want to achieve? Except maybe there is some edge case I am missing or maybe I might need clarity because the later part of the update_mempool function updates the mempool and ensures transaction IDs that are not in the JDS mempool are added along with their transaction data

NonsoAmadi10: I feel this should be enough

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: Okay, thanks I would fix this

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: alright I will do just that

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: I am not clearing old mining jobs in this case just emptying the mempool from the old mining jobs

NonsoAmadi10: @lorbax  @GitGab19 what do you think? This looks like a good idea, it would mean there would be no need for the retain function in the update_mempool function

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,41 @@ impl JobDeclaratorDownstream {
...
+    if let Some(old_job) = old_declare_mining_job {
+        // Retrieve necessary data from the old job
+        let old_tx_hash_list = old_job.tx_short_hash_list.inner_as_ref();
+        let old_nonce = old_job.tx_short_hash_nonce;
+
+        // Get the mempool of the old transaction IDs
+        let old_short_id_mempool = mempool
+            .safe_lock(|x| x.to_short_ids(old_nonce))
+            .unwrap()
+            .unwrap();
```
GitGab19, line 58: Why do you convert mempool into short_ids? Isn't it already made by Txid?

roles/jd-server/src/lib/job_declarator/message_handler.rs
GitGab19: What new tx list are you talking about? And which old tx list? Are you referring to node mempool or jds mempool?

GitGab19: If you're referring to mempool management (it's done in the other part of this PR), I think that @NonsoAmadi10 is already trying to do what you are mentioning here (maybe I'm missing something)

roles/jd-server/src/lib/mempool/mod.rs
lorbax: IMO the two functions have two purposes.
`clear_old_mempool_transactions` is called when a new job arrived and removes the transactions of the old job
`update_mempool` keep in sync the jds light mempool with the node mempool. You want to keep the transactions that are declared by some downstream, because there is the possibility that the node is not aware of some of them. BTW in the future I'll write an rpc  method to push txs to the node when they arrive to the jds, this improves a lot

roles/jd-server/src/lib/job_declarator/message_handler.rs
lorbax: > I am not clearing old mining jobs in this case just emptying the mempool from the old mining jobs

you are clearing the txs of old mining jobs from the jds light mempool.
What @jbesraa  wrote make sense to me

lorbax: > Could we compare old tx list and new tx list and only remove whats out then? that could be more efficient if there is enough similarity between the mining jobs

yes would make a lot of sense actually, but this is supposed to be a kind of first issue for the author. Perhaps to keep it like it is and open an issue? What do you think?

roles/jd-server/src/lib/mempool/mod.rs
NonsoAmadi10: This means there would be no need for the filtration I am doing on the JDS mempool? @lorbax

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: Alright, will adjust that now

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: I have modified that in the new commit i will push now. Thanks @GitGab19

roles/jd-server/src/lib/job_declarator/message_handler.rs
jbesraa: we should probably move this function to live under `DeclareMiningJob` implementation and the signature can be `pub fn clear_mining_jobs(&self, mempol: ..) -> Result<(), Error> { }`. 
Then you can use `old_declare_mining_job_.as_ref()...` in L91

jbesraa: Another option is just to replace  `Option<..>` with `DeclareMiningJob` and keep it simple for now

jbesraa: @NonsoAmadi10 Look here please https://github.com/jbesraa/stratum/commit/4b520cd3322f7508d011cad5d22bdbb26ec4919f
you can `git cherry-pick commit_hash` this  commit and it should resolve this

jbesraa (commented):
Whats the best way to test this pull request? I am happy to help but not sure where to start

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,31 @@ impl JobDeclaratorDownstream {
     }
 }
 
+pub fn clear_declared_mining_job(
+    mining_job: DeclareMiningJob,
+    mempool: Arc<Mutex<JDsMempool>>,
+) -> Result<(), Error> {
+    // If there is an old declared mining job, remove its transactions from the mempool
+        // Retrieve necessary data from the old job
+        let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
```
GitGab19: Why do you use tx_short_hash_list?
Afaik it doesn't contain TxIds

NonsoAmadi10: It was the feedback I got from @jbesraa. I used the short hash list before and used the mempool to get the transaction ID from it But the earlier feedback you get suggested I can use it directly

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: @GitGab19 Here

Shourya742 (changes requested):
I tried running the branch to test a few of the edge cases I encountered, as mentioned here: [GitHub issue #1005](https://github.com/stratum-mining/stratum/issues/1005). Interestingly, even though my testnet blockchain is fully synced, I am still receiving an error indicating that my chain is not in sync.

Error screenshot: 
Screenshot from 2024-07-02 03-31-13

Blockchain info snapshot: 
Screenshot from 2024-07-02 03-29-42

As you can see, the `initialblockdownload` is set to false, indicating that the blockchain is in sync.

NonsoAmadi10:
> I tried running the branch to test a few of the edge cases I encountered, as mentioned here: [GitHub issue #1005](https://github.com/stratum-mining/stratum/issues/1005). Interestingly, even though my testnet blockchain is fully synced, I am still receiving an error indicating that my chain is not in sync.
> 
> Error screenshot: Screenshot from 2024-07-02 03-31-13
> 
> Blockchain info snapshot: Screenshot from 2024-07-02 03-29-42
> 
> As you can see, the `initialblockdownload` is set to false, indicating that the blockchain is in sync.

Thanks for pointing this out, I will look to recreate and see what went wrong

GitGab19:
> I tried running the branch to test a few of the edge cases I encountered, as mentioned here: [GitHub issue #1005](https://github.com/stratum-mining/stratum/issues/1005). Interestingly, even though my testnet blockchain is fully synced, I am still receiving an error indicating that my chain is not in sync.
> 
> Error screenshot: 
> Screenshot from 2024-07-02 03-31-13
> 
> Blockchain info snapshot: 
> Screenshot from 2024-07-02 03-29-42
> 
> As you can see, the `initialblockdownload` is set to false, indicating that the blockchain is in sync.

From your screenshot you are running testnet3, right?
Check that RPC port is correct in JDS config file you're using

Shourya742:
> > I tried running the branch to test a few of the edge cases I encountered, as mentioned here: [GitHub issue #1005](https://github.com/stratum-mining/stratum/issues/1005). Interestingly, even though my testnet blockchain is fully synced, I am still receiving an error indicating that my chain is not in sync.
> > Error screenshot:
> > Screenshot from 2024-07-02 03-31-13
> > Blockchain info snapshot:
> > Screenshot from 2024-07-02 03-29-42
> > As you can see, the `initialblockdownload` is set to false, indicating that the blockchain is in sync.
> 
> From your screenshot you are running testnet3, right? Check that RPC port is correct in JDS config file you're using

Have we switched to testnet4? The last time I checked, I was encountering some errors.

Shourya742 (changes requested):
ACK. One minor nit.

This PR also resolves this issue: https://github.com/stratum-mining/stratum/issues/1005.

roles/jd-server/src/lib/mempool/mod.rs
Shourya742: Hey @NonsoAmadi10, have you implemented this suggestion from @lorbax?

lorbax (commented):
function update_mempool does not do what it is supposed to do

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -95,29 +95,35 @@ impl JDsMempool {
...
+        // Add new transactions from the node's mempool
+        for id in &node_mempool {
             let key_id = Txid::from_str(id)
                 .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
 
-            let tx = self_.safe_lock(|x| match x.mempool.get(&key_id) {
-                Some(entry) => entry.clone(),
-                None => None,
-            })?;
+            if !new_jds_mempool.contains_key(&key_id) {
```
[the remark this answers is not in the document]
lorbax: I already wrote a suggestion for this function in this comment
https://github.com/stratum-mining/stratum/pull/868/files#r1617572236

roles/jd-server/src/lib/mempool/mod.rs
NonsoAmadi10: In implementing the TODO for removing the clone, is it not the same thing as using the retain method, keeping only values in the new_jds_mempool that are some which removes the need for cloning again as this way:

`  new_jds_mempool.retain(|_, val| val.is_some());`

roles/jd-server/src/lib/mempool/mod.rs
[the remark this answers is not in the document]
NonsoAmadi10: Thanks @lorbax.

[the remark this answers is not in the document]
NonsoAmadi10: Done. Please review again @lorbax

Shourya742 (approved):
ACK 15452b479a0a751c37f7158bdc49ffd1e1e9219f

lorbax:
> ACK [15452b4](https://github.com/stratum-mining/stratum/commit/15452b479a0a751c37f7158bdc49ffd1e1e9219f)

just rebased the branch

NonsoAmadi10:
> I point out that there is a better implementation for the `update_mempool` method, that avoids to clone the jds light mempool. Firstly, the the bitcoin node mempool is retrieved and transformed in a vector of `Txid`s. Then the jds mempool is updated right inside the lock of the mutex.
> 
> ```rust
>     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
>         let client = self_
>             .safe_lock(|x| x.get_client())?
>             .ok_or(JdsMempoolError::NoClient)?;
>         let node_mempool: Vec<String> = client.get_raw_mempool().await?;
>         let mut node_mempool_deserialized: Vec<Txid> = vec![];
>         for id in &node_mempool {
>             let key_id = Txid::from_str(id)
>                 .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
>             node_mempool_deserialized.push(key_id);
>         };
>         
>         self_.safe_lock(|x| {
>             let jds_mempool = &mut x.mempool;
>             // the fat transactions in the jds-mempool are those declared by some downstream and we
>             // don't want to remove them, but we can get rid of the others
>             jds_mempool.retain(|_, val| val.is_some());
>             // here we add all the new transactions
>             for id in &node_mempool_deserialized {
>                 jds_mempool.entry(*id).or_insert(None);
>             };
>             if jds_mempool.is_empty() {
>                 Err(JdsMempoolError::EmptyMempool)
>             } else {
>                 Ok(())
>             }
>         })?
>     }
> ```

@lorbax I agree. I was thinking, to minimse lock duration, do you think this could work, while also implementing your feedback: 
```rust
 pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
    // First, retrieve the client outside of the lock.
    let client = {
        let self_guard = self_.lock().unwrap();
        self_guard.get_client().ok_or(JdsMempoolError::NoClient)?
    };

    // Fetch the node mempool and deserialize it.
    let node_mempool: Vec<String> = client.get_raw_mempool().await?;
    let node_mempool_deserialized: Result<Vec<Txid>, JdsMempoolError> = node_mempool
        .iter()
        .map(|id| Txid::from_str(id).map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string()))))
        .collect();
    let node_mempool_deserialized = node_mempool_deserialized?;

    // Lock the mempool only once and for a short duration.
    let mut self_guard = self_.lock().unwrap();
    let jds_mempool = &mut self_guard.mempool;

    // Retain fat transactions and add new transactions.
    jds_mempool.retain(|_, val| val.is_some());
    for id in node_mempool_deserialized {
        jds_mempool.entry(id).or_insert(None);
    }

    // Ensure the mempool is not empty and return the appropriate result.
    if jds_mempool.is_empty() {
        Err(JdsMempoolError::EmptyMempool)
    } else {
        Ok(())
    }
 }

```

lorbax:
Second thing, I am testing the PR with testnet4, jds, jdc, tproxy and cpuminer, and seems that the old job are not cleared from mempool, which grows only and never shrink.
I can see that by using the following modified version of the suggested [update_mempool](https://github.com/stratum-mining/stratum/pull/868#issuecomment-2225833093) above, which prints the jds-mempool transactions marked with `None` and the number of those that are `Some(...)`.
```rust
    pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
        let client = self_
            .safe_lock(|x| x.get_client())?
            .ok_or(JdsMempoolError::NoClient)?;
        let node_mempool: Vec<String> = client.get_raw_mempool().await?;
        let mut node_mempool_deserialized: Vec<Txid> = vec![];
        for id in &node_mempool {
            let key_id = Txid::from_str(id)
                .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
            node_mempool_deserialized.push(key_id);
        };
        
        self_.safe_lock(|x| {
            println!("AAAAAAAAAAAA");
            let jds_mempool = &mut x.mempool;
            println!("BEFORE RETAINING");
            println!("is_none() transactions");
            let mut jds_mempool_ = jds_mempool.clone();
            jds_mempool_.retain(|_, val| val.is_none());
            dbg!(jds_mempool_);
            println!("is_some() transactions");
            let mut jds_mempool_ = jds_mempool.clone();
            jds_mempool_.retain(|_, val| val.is_some());
            dbg!(jds_mempool_.len());
            // the fat transactions in the jds-mempool are those declared by some downstream and we
            // don't want to remove them, but we can get rid of the others
            jds_mempool.retain(|_, val| val.is_some());
            println!("AFTER RETAINING");
            println!("is_none() transactions");
            let mut jds_mempool_ = jds_mempool.clone();
            jds_mempool_.retain(|_, val| val.is_none());
            dbg!(jds_mempool_);
            println!("is_some() transactions");
            let mut jds_mempool_ = jds_mempool.clone();
            jds_mempool_.retain(|_, val| val.is_some());
            dbg!(jds_mempool_.len());
            // here we add all the new transactions
            for id in &node_mempool_deserialized {
                jds_mempool.entry(*id).or_insert(None);
            };
            println!("AFTER UPDATING");
            println!("is_none() transactions");
            let mut jds_mempool_ = jds_mempool.clone();
            jds_mempool_.retain(|_, val| val.is_none());
            dbg!(jds_mempool_);
            println!("is_some() transactions");
            let mut jds_mempool_ = jds_mempool.clone();
            jds_mempool_.retain(|_, val| val.is_some());
            dbg!(jds_mempool_.len());
            if jds_mempool.is_empty() {
                Err(JdsMempoolError::EmptyMempool)
            } else {
                Ok(())
            }
        })?
    }
```
I also used a modified version of `clear_declare_mining_job` function, for debuggin purpose
```rust
pub fn clear_declared_mining_job(
    mining_job: DeclareMiningJob,
    mempool: Arc<Mutex<JDsMempool>>,
) -> Result<(), Error> {
    info!("BBBBBBBBBBB");
    info!("BEFORE CLEARING OLD JOB");
    info!("length of jds mempool");
    dbg!(mempool.safe_lock(|x| x.mempool.len()).unwrap());
    // If there is an old declared mining job, remove its transactions from the mempool
    // Retrieve necessary data from the old job
    let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
    if transactions_to_remove.is_empty() {
        info!("No transactions to remove from mempool");
        return Ok(());
    }
    let clear_transactions = |jds_mempool: &mut JDsMempool| {
        for txid in transactions_to_remove {
            match jds_mempool.mempool.remove(txid) {
                Some(_) => info!("Transaction {:?} removed from mempool", txid),
                None => debug!("Transaction {:?} not found in mempool", txid),
            };
        }
    };
    match mempool.safe_lock(clear_transactions) {
        Ok(_) => {
            info!("AFTER CLEAR OLD JOB");
            info!("length of jds mempool");
            dbg!(mempool.safe_lock(|x| x.mempool.len()).unwrap());
            Ok(())
        },
        Err(e) => Err(Error::PoisonLock(e.to_string())),
    }
}
```
with this two modification I get this log (from jds launched RUST_LOG=info) most of the times
```
AAAAAAAAAAAA
BEFORE RETAINING
is_none() transactions
[jd-server/src/lib/mempool/mod.rs:116] jds_mempool_ = {
    2cd40b599c84999cc0aea1b9b68c87b98e30b2d56ea101faeb52571fe7e90cba: None,
    74b5e06ca418cf3757bef90cf8f1c68bdfd7fca7309f9a312ea6366ed1dd557a: None,
    02e253138faba485aef61e7c899d8cd1c9c926006df6a79c1c1758c1f6222f08: None,
    5505d5c8dad2fc03740f432b6cd60d9b817c4532088bdafad11276e3fb1655a7: None,
}
is_some() transactions
[jd-server/src/lib/mempool/mod.rs:120] jds_mempool_.len() = 24
AFTER RETAINING
is_none() transactions
[jd-server/src/lib/mempool/mod.rs:128] jds_mempool_ = {}
is_some() transactions
[jd-server/src/lib/mempool/mod.rs:132] jds_mempool_.len() = 24
AFTER UPDATING
is_none() transactions
[jd-server/src/lib/mempool/mod.rs:141] jds_mempool_ = {
    2cd40b599c84999cc0aea1b9b68c87b98e30b2d56ea101faeb52571fe7e90cba: None,
    74b5e06ca418cf3757bef90cf8f1c68bdfd7fca7309f9a312ea6366ed1dd557a: None,
    02e253138faba485aef61e7c899d8cd1c9c926006df6a79c1c1758c1f6222f08: None,
    5505d5c8dad2fc03740f432b6cd60d9b817c4532088bdafad11276e3fb1655a7: None,
}
is_some() transactions
[jd-server/src/lib/mempool/mod.rs:145] jds_mempool_.len() = 24
2024-07-12T15:41:08.221937Z  INFO roles_logic_sv2::handlers::job_declaration: Received DeclareMiningJob with id: 11
2024-07-12T15:41:08.221962Z  INFO jd_server::lib::job_declarator::message_handler: BBBBBBBBBBB
2024-07-12T15:41:08.221971Z  INFO jd_server::lib::job_declarator::message_handler: BEFORE CLEARING OLD JOB
2024-07-12T15:41:08.221981Z  INFO jd_server::lib::job_declarator::message_handler: length of jds mempool
[jd-server/src/lib/job_declarator/message_handler.rs:51] mempool.safe_lock(|x| x.mempool.len()).unwrap() = 28
2024-07-12T15:41:08.222008Z  INFO jd_server::lib::job_declarator::message_handler: AFTER CLEAR OLD JOB
2024-07-12T15:41:08.222016Z  INFO jd_server::lib::job_declarator::message_handler: length of jds mempool
[jd-server/src/lib/job_declarator/message_handler.rs:71] mempool.safe_lock(|x| x.mempool.len()).unwrap() = 28
```
which shows that no transaction is removed
apparently whenever a new block is mined, I get this output, telling me again that no transaction is removed from jds mempool

```

BEFORE RETAINING
is_none() transactions
[jd-server/src/lib/mempool/mod.rs:116] jds_mempool_ = {}
is_some() transactions
[jd-server/src/lib/mempool/mod.rs:120] jds_mempool_.len() = 30
AFTER RETAINING
is_none() transactions
[jd-server/src/lib/mempool/mod.rs:128] jds_mempool_ = {}
is_some() transactions
[jd-server/src/lib/mempool/mod.rs:132] jds_mempool_.len() = 30
AFTER UPDATING
is_none() transactions
[jd-server/src/lib/mempool/mod.rs:141] jds_mempool_ = {
    eff860506c7cf48f7f557cde40a8e90ee669d1725d9719174bf22fca445c6d9a: None,
    5961066cd2e4871c99c580c7e6936355518a01df081ecdf7080c85fffae75faf: None,
}
is_some() transactions
[jd-server/src/lib/mempool/mod.rs:145] jds_mempool_.len() = 30
2024-07-12T15:58:08.227340Z  INFO roles_logic_sv2::handlers::job_declaration: Received DeclareMiningJob with id: 14
2024-07-12T15:58:08.227364Z  INFO jd_server::lib::job_declarator::message_handler: BBBBBBBBBBB
2024-07-12T15:58:08.227373Z  INFO jd_server::lib::job_declarator::message_handler: BEFORE CLEARING OLD JOB
2024-07-12T15:58:08.227381Z  INFO jd_server::lib::job_declarator::message_handler: length of jds mempool
[jd-server/src/lib/job_declarator/message_handler.rs:51] mempool.safe_lock(|x| x.mempool.len()).unwrap() = 32
2024-07-12T15:58:08.227399Z  INFO jd_server::lib::job_declarator::message_handler: No transactions to remove from mempool
```

lorbax:
> > I point out that there is a better implementation for the `update_mempool` method, that avoids to clone the jds light mempool. Firstly, the the bitcoin node mempool is retrieved and transformed in a vector of `Txid`s. Then the jds mempool is updated right inside the lock of the mutex.
> > ```rust
> >     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
> >         let client = self_
> >             .safe_lock(|x| x.get_client())?
> >             .ok_or(JdsMempoolError::NoClient)?;
> >         let node_mempool: Vec<String> = client.get_raw_mempool().await?;
> >         let mut node_mempool_deserialized: Vec<Txid> = vec![];
> >         for id in &node_mempool {
> >             let key_id = Txid::from_str(id)
> >                 .map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string())))?;
> >             node_mempool_deserialized.push(key_id);
> >         };
> >         
> >         self_.safe_lock(|x| {
> >             let jds_mempool = &mut x.mempool;
> >             // the fat transactions in the jds-mempool are those declared by some downstream and we
> >             // don't want to remove them, but we can get rid of the others
> >             jds_mempool.retain(|_, val| val.is_some());
> >             // here we add all the new transactions
> >             for id in &node_mempool_deserialized {
> >                 jds_mempool.entry(*id).or_insert(None);
> >             };
> >             if jds_mempool.is_empty() {
> >                 Err(JdsMempoolError::EmptyMempool)
> >             } else {
> >                 Ok(())
> >             }
> >         })?
> >     }
> > ```
> 
> @lorbax I agree. I was thinking, to minimse lock duration, do you think this could work, while also implementing your feedback:
> 
> ```rust
>  pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
>     // First, retrieve the client outside of the lock.
>     let client = {
>         let self_guard = self_.lock().unwrap();
>         self_guard.get_client().ok_or(JdsMempoolError::NoClient)?
>     };
> 
>     // Fetch the node mempool and deserialize it.
>     let node_mempool: Vec<String> = client.get_raw_mempool().await?;
>     let node_mempool_deserialized: Result<Vec<Txid>, JdsMempoolError> = node_mempool
>         .iter()
>         .map(|id| Txid::from_str(id).map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string()))))
>         .collect();
>     let node_mempool_deserialized = node_mempool_deserialized?;
> 
>     // Lock the mempool only once and for a short duration.
>     let mut self_guard = self_.lock().unwrap();
>     let jds_mempool = &mut self_guard.mempool;
> 
>     // Retain fat transactions and add new transactions.
>     jds_mempool.retain(|_, val| val.is_some());
>     for id in node_mempool_deserialized {
>         jds_mempool.entry(id).or_insert(None);
>     }
> 
>     // Ensure the mempool is not empty and return the appropriate result.
>     if jds_mempool.is_empty() {
>         Err(JdsMempoolError::EmptyMempool)
>     } else {
>         Ok(())
>     }
>  }
> ```

it is basically the same code of mine, with the exception of `.lock()`, which shouldn't be used in favor of `.safe_lock()`.
ack for this part in your suggestion
```
let node_mempool: Vec<String> = client.get_raw_mempool().await?;
    let node_mempool_deserialized: Result<Vec<Txid>, JdsMempoolError> = node_mempool
        .iter()
        .map(|id| Txid::from_str(id).map_err(|err| JdsMempoolError::Rpc(RpcError::Deserialization(err.to_string()))))
        .collect();
    let node_mempool_deserialized = node_mempool_deserialized?;
    ```
    I should read it more carefully but seems to be a bit more elegant. Mine solution is raw lol

**Comment by lorbax** [comment]
I think I found the bug.
```rust
pub fn clear_declared_mining_job(
    mining_job: DeclareMiningJob,
    mempool: Arc<Mutex<JDsMempool>>,
) -> Result<(), Error> {
    // If there is an old declared mining job, remove its transactions from the mempool
    // Retrieve necessary data from the old job
    let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
    if transactions_to_remove.is_empty() {
        info!("No transactions to remove from mempool");
        return Ok(());
    }
    let clear_transactions = |jds_mempool: &mut JDsMempool| {
        for txid in transactions_to_remove {
            match jds_mempool.mempool.remove(txid) {
                Some(_) => info!("Transaction {:?} removed from mempool", txid),
                None => debug!("Transaction {:?} not found in mempool", txid),
            };
        }
    };
    match mempool.safe_lock(clear_transactions) {
        Ok(_) => Ok(()),
        Err(e) => Err(Error::PoisonLock(e.to_string())),
    }
}
```
There is no way this is gonna work because the transaction id's in the DeclareMiningJob are 6bytes short hash, and there is not way to reconstruct the orginal 32bytes `Txid` from them.  At the same time, the in the jds-mempool the txs are stored with `Txid` key. as a result here
```rust
        for txid in transactions_to_remove {
            match jds_mempool.mempool.remove(txid) {
                Some(_) => info!("Transaction {:?} removed from mempool", txid),
                None => debug!("Transaction {:?} not found in mempool", txid),
            };
        }
```
cannot find any declared transaction. Don't have time to check it deeper.

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,31 @@ impl JobDeclaratorDownstream {
     }
 }
 
+pub fn clear_declared_mining_job(
```
lorbax, line 44: as suggested in this [comment](https://github.com/stratum-mining/stratum/pull/868#issuecomment-2226012537), this function cannot clear from jds mempool transactions declared in old jobs.
I tested the following function and seems to work fine, even though requires more testing for other devs. Also, there are some suggestions left as a comment.
```rust
pub fn clear_declared_mining_job(
    mining_job: DeclareMiningJob,
    mempool: Arc<Mutex<JDsMempool>>,
) -> Result<(), Error> {
    // If there is an old declared mining job, remove its transactions from the mempool
    // Retrieve necessary data from the old job
    let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
    if transactions_to_remove.is_empty() {
        debug!("No transactions to remove from mempool");
        return Ok(()); };
    let nonce = mining_job.tx_short_hash_nonce;

    for short_id in transactions_to_remove {
        // try to return en error if the following lock results in an error 
        let _ = mempool.safe_lock(|mempool_| {
            // try to manage this unwrap, use .ok_or() method to return the proper error
            match mempool_.to_short_ids(nonce).unwrap().get(short_id){
               Some(transaction_with_hash) => {
                   let txid = transaction_with_hash.id;
                   match mempool_.mempool.remove(&txid) {
                       Some(transaction) => {
                           debug!("Fat transaction {:?} in job with request id {:?} removed from mempool", transaction, mining_job.request_id);
                           info!("Fat transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id);
                       },
                       None => info!("Thin transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id)
                   }
               },
               None => debug!("Transaction with short id {:?} not found in mempool while clearing old jobs", short_id),
            }
        });
    };
    Ok(())
}
```

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -95,31 +95,34 @@ impl JDsMempool {
     }
 
     pub async fn update_mempool(self_: Arc<Mutex<Self>>) -> Result<(), JdsMempoolError> {
```
lorbax, line 97: I suggested a better implementation of this function in this [comment](https://github.com/stratum-mining/stratum/pull/868#issuecomment-2225833093)

lorbax (changes requested):

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,42 @@ impl JobDeclaratorDownstream {
...
+        info!("No transactions to remove from mempool");
+        return Ok(());
+    }
+
+    let nonce = mining_job.tx_short_hash_nonce;
+
+    for short_id in transactions_to_remove {
+        // try to return en error if the following lock results in an error
+        let _ = mempool.safe_lock(|mempool_| {
+            // try to manage this unwrap, use .ok_or() method to return the proper error
```
lorbax, line 61: address this comment

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,42 @@ impl JobDeclaratorDownstream {
...
+    let transactions_to_remove = mining_job.tx_short_hash_list.inner_as_ref();
+    if transactions_to_remove.is_empty() {
+        info!("No transactions to remove from mempool");
+        return Ok(());
+    }
+
+    let nonce = mining_job.tx_short_hash_nonce;
+
+    for short_id in transactions_to_remove {
+        // try to return en error if the following lock results in an error
```
lorbax, line 59: address this comment

lorbax (changes requested):

rrybarczyk:
Please update title to follow the commit guidelines.

NonsoAmadi10:
> Please update title to follow the commit guidelines.

@rrybarczyk i don't quite understand.

GitGab19:
I think she was referring to this document which is linked in our `CONTRIBUTING.md`: https://cbea.ms/git-commit/ 

In this specific case, the PR title should be named in a way it tries to wrap the fix you did, instead of just using the same name of the issue.

NonsoAmadi10:
@rrybarczyk @GitGab19 I have done that and even squashed my commits. Please review

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,53 @@ impl JobDeclaratorDownstream {
     }
 }
 
+pub fn clear_declared_mining_job(
+    mining_job: DeclareMiningJob,
+    mempool: Arc<Mutex<JDsMempool>>,
+) -> Result<(), Error> {
+    // If there is an old declared mining job, remove its transactions from the mempool
+    // Retrieve necessary data from the old job
```
jbesraa, line 49: No need for these comments. Also, the`pub` is redundant.

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,53 @@ impl JobDeclaratorDownstream {
...
+                    }
+                },
+                None => debug!("Transaction with short id {:?} not found in mempool while clearing old jobs", short_id),
+            }
+            Ok(())  // Explicitly return Ok(()) inside the closure for proper flow control
+        });
+
+        // Propagate any error from the closure
+        if let Err(err) = result {
+            return Err(Error::PoisonLock(err.to_string()));
```
jbesraa, line 85: Is this necessarily a `PoisonLock` error?

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,53 @@ impl JobDeclaratorDownstream {
...
+                .ok_or(Error::JDSMissingTransactions)?;
+            let transaction_with_hash = short_ids_map
+                .get(short_id);
+
+            match transaction_with_hash {
+                Some(transaction_with_hash) => {
+                    let txid = transaction_with_hash.id;
+                    match mempool_.mempool.remove(&txid) {
+                        Some(transaction) => {
+                            debug!("Fat transaction {:?} in job with request id {:?} removed from mempool", transaction, mining_job.request_id);
```
jbesraa, line 72: we can remove this one and the `info!` beneath it

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,53 @@ impl JobDeclaratorDownstream {
...
+        info!("No transactions to remove from mempool");
+        return Ok(());
+    }
+
+    let nonce = mining_job.tx_short_hash_nonce;
+
+    for short_id in transactions_to_remove {
+        let result = mempool.safe_lock(|mempool_| -> Result<(), Error> {
+            // Try to manage this unwrap, we use .ok_or() method to return the proper error
+            let short_ids_map = mempool_
```
jbesraa, line 61: do we need to call `mempool.safe_lock` in each iteration? couldnt we retrieve `short_ids_map` before the `for` and just use it?

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -61,6 +111,13 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
     }
 
     fn handle_declare_mining_job(&mut self, message: DeclareMiningJob) -> Result<SendTo, Error> {
+        {
```
jbesraa, line 114: redundant brackets

roles/jd-server/src/lib/mempool/mod.rs
```diff
@@ -7,7 +7,7 @@ use hashbrown::HashMap;
 use roles_logic_sv2::utils::Mutex;
 use rpc_sv2::{mini_rpc_client, mini_rpc_client::RpcError};
 use std::{convert::TryInto, str::FromStr, sync::Arc};
-use stratum_common::{bitcoin, bitcoin::hash_types::Txid};
+use stratum_common::bitcoin::{self, hash_types::Txid};
```
jbesraa, line 10: why this change?

jbesraa (changes requested):

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: what do you suggest? Because that is the error class being returned from the safe_lock `PoisonError<MutexGuard<JDsMempool>>`

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: This was a feedback from lorban I implemented and modified. If you have a better approach I can compare that

roles/jd-server/src/lib/mempool/mod.rs
NonsoAmadi10: I rebased with dev and it shows you made those commit last month @jbesraa

jbesraa: sorry can you point to the commit?

roles/jd-server/src/lib/job_declarator/message_handler.rs
jbesraa: In the mutex closure we have this error, will this be propagated to LockPoison?
```rust
           let short_ids_map = mempool_
                .to_short_ids(nonce)
                .ok_or(Error::JDSMissingTransactions)?;
```

roles/jd-server/src/lib/mempool/mod.rs
NonsoAmadi10: this is the commit hash: a76d1022eaca42c9578b5d2bec511acfc2a43e10

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: Yes please! it would be wrapped and propagated as a poison lock

jbesraa:
@NonsoAmadi10 could you please rebase and update the branch? Ill help with the outstanding issues

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: That means I would still need to handle the double unwraps well? @lorbax

roles/jd-server/src/lib/job_declarator/message_handler.rs
NonsoAmadi10: @jbesraa do you suggest i write my own error class maybe something like a mutex poisoned class similar to the one the JDS error has?

roles/jd-server/src/lib/job_declarator/message_handler.rs
lorbax: yes, but I suggest first to write a the code as easy as possible, thest it, then add error handling and then test it deeply. You should check that the code you introduced does what it is supposed to do. In particular, the old job must be removed from jds mempool and while updating the jds mempool the fat transactions are not discarded.
You can do that using testnet4, or a custom signet

Fi3: I'm worried that to_short_ids is very expensive not sure if is really better to lock the mutex for all that time. Btw we should have a benchmark for cases where there is contention for the lock.

Fi3: considering that this is JDC and not JDS it shouldn't be an issue as we expect to have only one connection

Fi3: this should look into the new received DeclareMiningJob and remove only the txs that are in the old one but not in the new one. Also better would be to have a rc for each txs and decrement it everytime that the tx is in the old one and not in the new one (this cause we have multiple downstreams). And when we reach 0 we remove the tx

Fi3: somthing like that:  `    pub mempool: HashMap<Txid, Option<(Transaction,u8)>>`

plebhash:
hey @NonsoAmadi10 what's the status here?

do you need any input or help unblocking anything?

NonsoAmadi10:
> hey @NonsoAmadi10 what's the status here?
> 
> do you need any input or help unblocking anything?

@plebhash , @jbesraa told me to rebase and update the branch that he would help with outstanding issues. I think that was why I halted. Other than that I intended on closing the opened conversation here. I have time this month to work here

roles/jd-server/src/lib/job_declarator/message_handler.rs
plebhash: @NonsoAmadi10 the redundant brackets remain, I don't see why this comment should be resolved just yet

NonsoAmadi10: @plebhash i have removed the redundants brackets

roles/jd-server/src/lib/job_declarator/message_handler.rs
Shourya742: @NonsoAmadi10, The comment is resolved, but I don't see the requested changes.

Shourya742 (changes requested):
Hey @NonsoAmadi10 , I see the comments have been marked as resolved, but the requested changes don't seem to be in place. Could you please check?

roles/jd-server/src/lib/job_declarator/message_handler.rs
plebhash: > this should look into the new received DeclareMiningJob and remove only the txs that are in the old one but not in the new one. Also better would be to have a rc for each txs and decrement it everytime that the tx is in the old one and not in the new one (this cause we have multiple downstreams). And when we reach 0 we remove the tx

I implemented this suggestion here https://github.com/plebhash/stratum/commit/24b952a296b2949729e85ea4b29e5e58b356824b

roles/jd-server/src/lib/job_declarator/message_handler.rs
```diff
@@ -38,6 +41,53 @@ impl JobDeclaratorDownstream {
...
+            let transaction_with_hash = short_ids_map
+                .get(short_id);
+
+            match transaction_with_hash {
+                Some(transaction_with_hash) => {
+                    let txid = transaction_with_hash.id;
+                    match mempool_.mempool.remove(&txid) {
+                        Some(transaction) => {
+                            debug!("Fat transaction {:?} in job with request id {:?} removed from mempool", transaction, mining_job.request_id);
+                            info!("Fat transaction {:?} in job with request id {:?} removed from mempool", txid, mining_job.request_id);
```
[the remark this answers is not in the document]
NonsoAmadi10: okay, I will

NonsoAmadi10:
@plebhash @lorbax @jbesraa @Fi3 @Shourya742. I have taken the feedback from previous unresolved conversations and a huge help from @plebhash commit. Here is what I did:

Optimize clear_declared_mining_job function

- Move mempool lock outside the loop to reduce contention
- Compare old and new mining jobs to avoid unnecessary removals( as @Fi suggested)
- Modify transaction counters in-place for better efficiency(thank you @plebhash)
- Remove unnecessary cloning and HashMap operations
- Improve error handling and logging

I don't know if what I did addressed it or if it has introduced added complexity. I would welcome your feedback

roles/jd-server/src/lib/job_declarator/message_handler.rs
jbesraa: > somthing like that: ` pub mempool: HashMap<Txid, Option<(Transaction,u64)>>`

Could we do take advantage of the `token` or other property in `JobDeclaratorDownstream` in order to do that?

rrybarczyk:
Thank you so much for the work here, but right now our plan is to refactor the `roles` completely and we want to avoid large `roles` PR's as it will eventually be refactored again. We will work on this once we have the proper integration test framework to test it in #1138.

Issue #870, opened 2024-04-25
Dev branch is not on top of Main branch


lorbax:
Seems that dev branch is not on top of main. This is bad because if someone wants contribute and fork the project from main, we must ask them to rewrite their PRs.
The most recent commit that both dev and main share is [this one](https://github.com/stratum-mining/stratum/commit/7c96c4ecf48c490aa0494aca328f7ae2dfce233c).

Fi3:
this is bad cause when we want to merge dev on main wont be automatic. If someone fork from main he will always need to rebase on dev and it just shouldn't create the new branch from main

plebhash:
I created #871 merging `main` into `dev`, which should make the branches back to sync again

plebhash:
closed via https://github.com/stratum-mining/stratum/pull/871 +  https://github.com/stratum-mining/stratum/pull/872

PR #871, opened 2024-04-25
merge dev main

plebhash:
fix https://github.com/stratum-mining/stratum/issues/870

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,109 @@
+<!-- omit in toc -->
+# Contributing to SRI (Stratum V2 Reference Implementation)
+
+First off, thanks for taking the time to contribute! ❤️
+
+All types of contributions are encouraged and valued. See the [Table of Contents](#table-of-contents) for different ways to help and details about how this project handles them. Please make sure to read the relevant section before making your contribution. It will make it a lot easier for us maintainers and smooth out the experience for all involved. The community looks forward to your contributions. 🎉
+
+> And if you like the project, but just don't have time to contribute, that's fine. There are other easy ways to support the project and show your appreciation, which we would also be very happy about:
+> - Star the project
+> - Tweet about it
+> - Refer this project in your project's readme
+> - Mention the project at local meetups and tell your friends/colleagues
+
+<!-- omit in toc -->
+## Table of Contents
+
+- [I Have a Question](#i-have-a-question)
+- [What Should I Know Before I Get Started](#what-should-i-know-before-i-get-started)
+  - [Important Resources About SRI](#important-resources-about-sri)
+  - [Project Communications](#project-communications)
+- [I Want To Contribute](#i-want-to-contribute)
+  - [Project Structure](#project-structure)
+  - [Contribution workflow](#contribution-workflow)
+  - [Your First Code Contribution](#your-first-code-contribution)
+  
+
+## I Have a Question
+
+> If you want to ask a question, we assume that you have read the documentation available at [stratumprotocol.org/docs](https://stratumprotocol.org).
+
+Best way to ask a question is to hop onto our community [Discord](https://discord.com/invite/fsEW23wFYs). Two most suitable places to post a question are:
+- #newbies-qs and
+- #dev, for technical questions, suitable for developers building on top of SRI, or contributing to it.
+
+If you then still feel the need to ask a question and need clarification, we recommend the following:
+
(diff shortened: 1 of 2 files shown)


pavlenex (approved):

lorbax:
The least commit shared between this branch and main is [this one](https://github.com/stratum-mining/stratum/commit/1611cdb84951d8cc8b27c655f3cd4530ab734161) which is not on top of main. Seems that this does not solve the issue

plebhash:
> The least commit shared between this branch and main is [this one](https://github.com/stratum-mining/stratum/commit/1611cdb84951d8cc8b27c655f3cd4530ab734161) which is not on top of main. Seems that this does not solve the issue

@lorbax can you elaborate on what you see as a problem here?

these are the steps that led me to this PR:
```
git checkout dev
git checkout -b dev-merge-main
git merge origin main
# there was a confict on roles/Cargo.lock, which I solved by picking the state from `main`
git merge --continue
git push plebhash dev-merge-main
# opened PR
```

> tried to rebase manually commit by commit dev on main
this is the result
https://github.com/lorbax/stratum/tree/dev-rebased

I think manually rebasing every commit is too error prone. We should be able to fix this with a few commands.

Fi3:
> # there was a confict on roles/Cargo.lock, which I solved by picking the state from `main`

If you are going to merge on dev you should take the state from dev not main

lorbax:
I was wrong, the least common ancestor between this branch and main branch is the HEAD of main branch, so this PR solves the issue. It would have been better to rebase dev on main, so that we do not mess up with history of commits, which is what I tried to do in my branch (before you publish this PR)
[https://github.com/lorbax/stratum/tree/dev-rebased](https://github.com/lorbax/stratum/tree/dev-rebased)
Next time we should declare the issues we are working in!

plebhash:
> > there was a confict on roles/Cargo.lock, which I solved by picking the state from `main`
> 
> If you are going to merge on dev you should take the state from dev not main
> 
> Considering that this PR do not change any code
> 
> This ofc assuming that lock on dev do not have any issue

`roles/Cargo.lock` is outdated on `dev`.

I noticed that because I use CLion and it automatically updates the lockfiles. So I got many conflicts on git because of that.

https://github.com/stratum-mining/stratum/commit/3ec6b50875e49dfed7415b0d99060302f437e467 updated this lockfile on `main`, and that was the motivation for picking the state of `main` on this merge conflict.

Fi3:
> > > there was a confict on roles/Cargo.lock, which I solved by picking the state from `main`
> > 
> > 
> > If you are going to merge on dev you should take the state from dev not main
> > Considering that this PR do not change any code
> > This ofc assuming that lock on dev do not have any issue
> 
> `roles/Cargo.lock` is outdated on `dev`.
> 
> I noticed that because I use CLion and it automatically updates the lockfiles. So I got many conflicts on git because of that.
> 
> [3ec6b50](https://github.com/stratum-mining/stratum/commit/3ec6b50875e49dfed7415b0d99060302f437e467) updated this lockfile on `main`, and that was the motivation for picking the state of `main` on this merge conflict.

ok make sense

plebhash:
> It would have been better to rebase dev on main, so that we do not mess up with history of commits, which is what I tried to do in my branch (before you publish this PR) https://github.com/lorbax/stratum/tree/dev-rebased 

https://github.com/stratum-mining/stratum/compare/dev...lorbax:stratum:dev-rebased

This diff shows 23 files being changed, where most of those changes were already introduced in PRs recently merged into `dev`. Is that expected? Git rebase is a bit confusing to me sometimes, that is why I went for a merge.

But if you prefer this approach we can replace this PR with another one from your branch, that would be no problem for me.

lorbax:
> > It would have been better to rebase dev on main, so that we do not mess up with history of commits, which is what I tried to do in my branch (before you publish this PR) https://github.com/lorbax/stratum/tree/dev-rebased
> 
> [dev...lorbax:stratum:dev-rebased](https://github.com/stratum-mining/stratum/compare/dev...lorbax:stratum:dev-rebased)
> 
> This diff shows 23 files being changed, where most of those changes were already introduced in PRs recently merged into `dev`. Is that expected? Git rebase is a bit confusing to me sometimes, that is why I went for a merge.
> 
> But if you prefer this approach we can replace this PR with another one from your branch, that would be no problem for me.

For me it is fine whatever does work. There is not huge history that can possibly be messed up btw

Fi3:
no you shouldn't get 23 diff files. IMO merging main on dev is ok.

PR #888, opened 2024-05-03
DO NOT MERGE (diff dev)

Fi3:


xyephy:
Hi @Fi3 could the deletion of : ```sha2 = "0.10.6" ``` in [roles/test-utils/mining-device/Cargo.toml](https://github.com/stratum-mining/stratum/pull/888#diff-257f60c1fb859b4368447d41cf196b8a0b9c62787439be88d27f7c84cc0b4bc7) be bringing this error: 
```
error[E0432]: unresolved import `sha2`
  --> test-utils/mining-device/src/main.rs:10:5
   |
10 | use sha2::{Digest, Sha256};
   |     ^^^^ use of undeclared crate or module `sha2`
```

Issue #749, opened 2024-02-13
better import for libsecp256k1


lorbax:
As suggested by @Fi3 in the review of https://github.com/stratum-mining/stratum/pull/737, we should
 1. move the dependency of secp256k1 in stratum_common and import this library from there
 2. avoid to import secp256k1 in roles (it is used by the JDS for signing tokens and by the pool for verifying tokens' signature). Rather, some sort of wrapper structure in roles-logic for making and verifying signatures would be more appropriate.

rockcoolsaint:
I think this is something I can handle. Can I go ahead with it? @lorbax

lorbax:
Don't know if it is really a good "first issue" but if feel comfortable you can DM on discord and discuss how to proceed

Fi3:
now you can import libs from the workspace Cargo.toml and refer to it in the crate Cargo.toml when you have common dep between crates in the same workspace.

rockcoolsaint:
> Don't know if it is really a good "first issue" but if feel comfortable you can DM on discord and discuss how to proceed

What's your discord handle?

plebhash:
closed via #894

PR #894, opened 2024-05-03
better import for libsecp256k1

rockcoolsaint:
fixes issue raised in #749 

This was essentially a refactoring issue

### DIFF
--- a/benches/Cargo.lock
+++ b/benches/Cargo.lock
@@ -1070,7 +1070,7 @@ dependencies = [
 
 [[package]]
 name = "network_helpers_sv2"
-version = "1.0.1"
+version = "2.0.0"
 dependencies = [
  "async-channel 1.9.0",
  "async-std",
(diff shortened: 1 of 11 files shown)


plebhash:
I think this PR is diverging from the description of https://github.com/stratum-mining/stratum/issues/749

this PR should be removing these lines:

https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/roles/pool/Cargo.toml#L33

https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/roles/jd-server/Cargo.toml#L33

and adding it to [`common/Cargo.toml`](https://github.com/stratum-mining/stratum/blob/main/common/Cargo.toml)

rockcoolsaint:
@plebhash I don't get error on `cargo fmt` and `cargo clippy` in development
I don't understand why they both fails here

plebhash:
> @plebhash I don't get error on `cargo fmt` and `cargo clippy` in development I don't understand why they're both failing here

do `cargo +nightly fmt`

plebhash (approved):

PR #899, opened 2024-05-07
Fix network helper panic if peer close connection

Fi3:
Fix network helper panic if peer close connection

    When the remote peer try to open an encrypted sv2 connection and then
    immediatly close the tcp connection, the helper for async std and tokio
    it panic, mostly of the time we don't want to panic in this situation.
    This commit add an error for this case and return it.


lorbax (approved):

lorbax:
This PR just introduce an new variant to an error type and returns this variant in a part of the code where there was an unwrap. There is no need to do further test more than those on actions. 
Looks good to me.

PR #904, opened 2024-05-10
Better management of LastDeclareJobs - no more wrong fallback-system activations

### LINKED ISSUE #901: JDC executes fallback system too frequently
While digging into issue #844, I discovered that JDC enters in the fallback mechanism too frequently.
I discovered the reason for that, and it's related to the `prev_hash` contained in shares. 
- Sometime (I still don't know exactly why) a valid share is sent to the pool (in this case through JDC) with the correct `prev_hash`.
- When the valid share is checked by the pool, the pool uses a wrong `prev_hash` in the `check_target` function, and so the share is not a valid one for the pool.
- Pool sends a `SubmitShareError` message to JDC 
- JDC thinks that Pool is wrongly refusing the valid share and pool-fallback is started

JDC logs:
Screenshot 2024-05-08 at 18 25 43

Pool logs:
Screenshot 2024-05-08 at 18 26 48

GitGab19:
This PR addresses the way `last_declare_mining_jobs_sent` are managed by JDC. 
Before this PR, we could end up in a situation in which we were not able to exit from the loop inside `on_set_new_prev_hash` function because of this line: https://github.com/stratum-mining/stratum/blob/0b131916f8bbc739a57b739c059fe887bc538931/roles/jd-client/src/lib/job_declarator/mod.rs#L362 
The situation occured every time a new job was created and declared right after the future job linked to the new prev_hash. Because of [this line](https://github.com/stratum-mining/stratum/blob/0b131916f8bbc739a57b739c059fe887bc538931/roles/jd-client/src/lib/job_declarator/mod.rs#L275) the second job (not future) was extracted in this scenario (instead of the right previous one, the future one). In this way the future_job was not able to be inserted into `future_jobs` through [this line](https://github.com/stratum-mining/stratum/blob/0b131916f8bbc739a57b739c059fe887bc538931/roles/jd-client/src/lib/job_declarator/mod.rs#L289) and so we never exited from the loop.

This implied that that specific prev_hash was constantly set by the loop, causing the behaviour described under issue #901.

Fixes #901

### DIFF
--- a/roles/jd-client/src/lib/job_declarator/mod.rs
+++ b/roles/jd-client/src/lib/job_declarator/mod.rs
@@ -377,14 +377,12 @@ impl JobDeclarator {
                 s.last_set_new_prev_hash = Some(set_new_prev_hash.clone());
                 s.set_new_prev_hash_counter += 1;
             });
-
-            let mut future_job_tuple_opt = None;
-
-            loop {
+            let (job, up, merkle_path, template, mut pool_outs) = loop {
                 match self_mutex
                     .safe_lock(|s| {
                         if s.set_new_prev_hash_counter > 1
                             && s.last_set_new_prev_hash != Some(set_new_prev_hash.clone())
+                        //it means that a new prev_hash is arrived while the previous hasn't exited the loop yet
                         {
                             s.set_new_prev_hash_counter -= 1;
                             Some(None)
@@ -401,39 +399,31 @@ impl JobDeclarator {
                     })
                     .unwrap()
                 {
-                    Some(Some(future_job_tuple)) => {
-                        future_job_tuple_opt = Some(future_job_tuple);
-                        break;
-                    }
-                    Some(None) => {
-                        break;
-                    }
+                    Some(Some(future_job_tuple)) => break future_job_tuple,
+                    Some(None) => return,
                     None => {}
                 };
                 tokio::task::yield_now().await;
-            }
-
-            if let Some((job, up, merkle_path, template, mut pool_outs)) = future_job_tuple_opt {
-                let signed_token = job.mining_job_token.clone();
-                let mut template_outs = template.coinbase_tx_outputs.to_vec();
-                pool_outs.append(&mut template_outs);
-                Upstream::set_custom_jobs(
-                    &up,
-                    job,
-                    set_new_prev_hash,
-                    merkle_path,
-                    signed_token,
-                    template.coinbase_tx_version,
-                    template.coinbase_prefix,
-                    template.coinbase_tx_input_sequence,
-                    template.coinbase_tx_value_remaining,
-                    pool_outs,
-                    template.coinbase_tx_locktime,
-                    template.template_id,
-                )
-                .await
-                .unwrap();
-            }
+            };
+            let signed_token = job.mining_job_token.clone();
+            let mut template_outs = template.coinbase_tx_outputs.to_vec();
+            pool_outs.append(&mut template_outs);
+            Upstream::set_custom_jobs(
+                &up,
+                job,
+                set_new_prev_hash,
+                merkle_path,
+                signed_token,
+                template.coinbase_tx_version,
+                template.coinbase_prefix,
+                template.coinbase_tx_input_sequence,
+                template.coinbase_tx_value_remaining,
+                pool_outs,
+                template.coinbase_tx_locktime,
+                template.template_id,
+            )
+            .await
+            .unwrap();
         });
     }
 
--- a/roles/jd-client/src/lib/job_declarator/mod.rs
+++ b/roles/jd-client/src/lib/job_declarator/mod.rs
@@ -57,6 +57,7 @@ pub struct JobDeclarator {
     // (Sent DeclareMiningJob, is future, template id, merkle path)
     last_declare_mining_jobs_sent: HashMap<u32, Option<LastDeclareJob>>,
     last_set_new_prev_hash: Option<SetNewPrevHash<'static>>,
+    set_new_prev_hash_counter: u8,
     #[allow(clippy::type_complexity)]
     future_jobs: HashMap<
         u64,
@@ -121,6 +122,7 @@ impl JobDeclarator {
             task_collector,
             coinbase_tx_prefix: vec![].try_into().unwrap(),
             coinbase_tx_suffix: vec![].try_into().unwrap(),
+            set_new_prev_hash_counter: 0,
         }));
 
         Self::allocate_tokens(&self_, 2).await;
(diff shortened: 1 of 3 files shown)


Fi3:
Can you add an MG test for this situation?

GitGab19:
> Can you add an MG test for this situation?

I never wrote a MG test, but I can try to dig into it and add it

Fi3:
This is a perefect candidate for an MG test, whenever we do a PR that fix an issue that arise from message ordering we should add an MG test. This will add a lot of value since these kind of issue are hard to find.

GitGab19:
> This is a perefect candidate for an MG test, whenever we do a PR that fix an issue that arise from message ordering we should add an MG test. This will add a lot of value since these kind of issue are hard to find.
> 
> Spec are important, but a test suite that handle all the corner cases, that for obviuous reasons are not discussed in the spec, is also important

Agree, will do it after some more tests.
I just turned this PR into draft since I noticed there's still a possible case not handled correctly, leading to the issue related to this PR.

roles/jd-client/src/lib/job_declarator/mod.rs
```diff
@@ -114,7 +114,7 @@ impl JobDeclarator {
             allocated_tokens: vec![],
             req_ids: Id::new(),
             min_extranonce_size,
-            last_declare_mining_job_sent: None,
+            last_declare_mining_jobs_sent: HashMap::with_capacity(10),
```
plebhash, line 117: why initialize via `with_capacity(10)`? seems quite arbitrary?
why not simply `new()`?

roles/jd-client/src/lib/job_declarator/mod.rs
```diff
@@ -55,7 +55,7 @@ pub struct JobDeclarator {
     req_ids: Id,
     min_extranonce_size: u16,
     // (Sent DeclareMiningJob, is future, template id, merkle path)
-    last_declare_mining_job_sent: Option<LastDeclareJob>,
+    last_declare_mining_jobs_sent: HashMap<u32, Option<LastDeclareJob>>,
```
plebhash, line 58: now that we are introducing a `HashMap`, I guess it makes sense to get rid of the `Option`?

I mean, there will never be a situation where we have a `(u32, None)` entry on this `HashMap`, right?
there's either a `(key, value)` entry, or there is not.

plebhash:
the original description on #901 says:

> - Sometime (I still don't know exactly why) a valid share is sent to the pool (in this case through JDC) with the correct `prev_hash`.
> - When the valid share is checked by the pool, the pool uses a wrong `prev_hash` in the `check_target` function, and so the share is not a valid one for the pool.

Do we also have a bug on the pool? Should we create an issue for that?
Or is that a false alarm and this PR solves everything related to this problem?

roles/jd-client/src/lib/job_declarator/mod.rs
Fi3: make sure to not leak memory through this hash map.

roles/jd-client/src/lib/job_declarator/mod.rs
GitGab19: I put 10 in order to not leak memory

roles/jd-client/src/lib/job_declarator/mod.rs
GitGab19: > now that we are introducing a `HashMap`, I guess it makes sense to get rid of the `Option`?
> 
> I mean, there will never be a situation where we have a `(u32, None)` entry on this `HashMap`, right? there's either a `(key, value)` entry, or there is nothing.

Yeah, but I preferred to not change too many things, because I think that the Option was put from someone else in the past to better manage some cases that I did not explore in this PR.

Fi3: this is not enough you have to remove things from it. When you initialize with capacity it just means that it allocate memory to get 10 elements. When you add the 11 you just reallocate.

GitGab19: > this is not enough you have to remove things from it. When you initialize with capacity it just means that it allocate memory to get 10 elements. When you add the 11 you just reallocate.

Ok, once stabilized the fix I'm working on, I will push this changes, adding checks and reallocation of this hashmap

GitGab19: @Fi3 done in 47dc7aa514740f6ac7c2c8065834a3082c75d714

plebhash (approved):

GitGab19:
I reverted last commit since it introduced no meaningful changes.
And as discussed with @Fi3, it's better (and cleaner) as it was before.

GitGab19:
I also identified a new bug (described in issue #920).
I pushed a basic (but not elegant) fix for that.
But issue #920 will be still to be solved properly, removing the timeout I just introduced.

roles/jd-client/src/lib/job_declarator/mod.rs
Fi3: see my comment above, initializing it with 10 do not do anything in order to not leak memory

roles/jd-client/src/lib/job_declarator/mod.rs
```diff
@@ -114,34 +115,50 @@ impl JobDeclarator {
...
     }
 
-    fn get_last_declare_job_sent(self_mutex: &Arc<Mutex<Self>>) -> LastDeclareJob {
+    fn get_last_declare_job_sent(self_mutex: &Arc<Mutex<Self>>, request_id: u32) -> LastDeclareJob {
         self_mutex
             .safe_lock(|s| {
-                s.last_declare_mining_job_sent
+                s.last_declare_mining_jobs_sent
+                    .get(&request_id)
+                    .expect("LastDeclareJob not found")
```
Fi3, line 138: we should never panic while we are keeping a lock otherwise the shared data get unusable for everyone. If there is something that is obviously unreachable, is ok to put an expect but is not very clear why the code can never panic you should write an explanation. Otherwise, will be a lot better to take the Result or the Option out of the safe_lock and if needed unwrap it out

Fi3: here you should remove the job btw

roles/jd-client/src/lib/job_declarator/mod.rs
```diff
@@ -114,34 +115,50 @@ impl JobDeclarator {
...
+        self_mutex: &Arc<Mutex<Self>>,
+        request_id: u32,
+        j: LastDeclareJob,
+    ) {
         self_mutex
-            .safe_lock(|s| s.last_declare_mining_job_sent = Some(j))
-            .unwrap()
+            .safe_lock(|s| {
+                //check hashmap size in order to not let it grow indefinetely
+                if s.last_declare_mining_jobs_sent.len() < 10 {
```
Fi3, line 153: this number "10" really require a comment. Why is 10 and not 2 or 3? How many job we expect to have in the same time? I guess max is 2? If not why? And is ok having more than 3 jobs, if not we should just close the process cause something unexpected is happening, and we don't want people keep mining (paying electricity bill) when we are not sure that they are producing shares that will get payed.

Fi3: maybe would be better to just have an array with [job-1, job0]. Are there cases where we would need to access an older job? If downstream send share for a job that have is 7 job old, I would say that something is off either here or in the downstream so the safest thing to do would be close the connection. Maybe I'm  missing something but I would like to have it addressed in a comment.

roles/jd-client/src/lib/job_declarator/mod.rs
Fi3: Somthing like that:
```rust
    fn get_last_declare_job_sent(self_mutex: &Arc<Mutex<Self>>, request_id: u32) -> LastDeclareJob {
        let id = self_mutex
            .safe_lock(|s| {
                s.last_declare_mining_jobs_sent
                    .remove(&request_id)
                    .clone()
            })
            .unwrap();
        id.expect("Impossible to get last declare job sent").clone().expect("This is ok")
    }
```
Some consideration:
1. having and HashMap of Option do no make much sense, I would remove the inner Option so we do not have last exepct.
2. not having the required id, is a possibility, the function should return an Option and the caller should handle the case of downstream sending shares with wrong id (closing the connection, or the entire process)
3. as I said below could be enough having an array of 2 elements rather then an hashmap here, (1) and (2) still apply

GitGab19:
@Fi3 Thanks for review.
I opened an issue to better track the changes you suggested and to add MG test for this specific scenario

PR #905, opened 2024-05-10
Every MG test in a separate action

lorbax:
The AIM of this PR is to split every Message Generator test in a separate GH action
the test `interop-jdc-change-upstream` is excluded from MG action jobs. It refuses to pass even with all the needed applications precompiled. This requires more investigation, possibly spitting this test in smaller tests. There is the possiblity that GH has not enough resources for making it run successfully. Worth noting that this test was excluded from GH actions also before this PR.
Now each test is launched by a script which precompiles the binaries that are going to be used. This reduces the risk of test fail due to timeouts.


test/message-generator/test/bad-pool-config-test/bad-pool-config-test.sh
```diff
@@ -0,0 +1,9 @@
+cd roles
+cargo llvm-cov --no-report -p pool_sv2
+
+cd ../utils/message-generator/
+cargo build
```
plebhash, line 5: do we really need to do `cargo build` here?
it feels redundant, since we are doing `cargo run` on the next step

test/message-generator/test/interop-jd-translator/interop-jd-translator.sh
```diff
@@ -0,0 +1,13 @@
+cd roles
+cargo llvm-cov --no-report -p pool_sv2
+cargo llvm-cov --no-report -p jd_server
+cargo llvm-cov --no-report -p jd_client
+cargo llvm-cov --no-report -p translator_sv2
+cargo build -p sv1-mining-device
+
+cd ../utils/message-generator/
+cargo build
```
plebhash, line 9: [ditto](https://github.com/stratum-mining/stratum/pull/905/files#r1603974488)

test/message-generator/test/interop-proxy-with-multi-ups-extended/interop-proxy-with-multi-ups-extended.sh
```diff
@@ -0,0 +1,11 @@
+cd roles
+cargo llvm-cov --no-report -p pool_sv2
+cargo llvm-cov --no-report -p mining_proxy_sv2
+cargo build --no-report -p mining-device
+
+cd ../utils/message-generator/
+cargo build
```
plebhash, line 7: [ditto](https://github.com/stratum-mining/stratum/pull/905/files#r1603974488)

test/message-generator/test/interop-proxy-with-multi-ups/interop-proxy-with-multi-ups.sh
```diff
@@ -0,0 +1,11 @@
+cd roles
+cargo llvm-cov --no-report -p pool_sv2
+cargo llvm-cov --no-report -p mining_proxy_sv2
+cargo build -p mining-device
+
+cd ../utils/message-generator/
+cargo build
```
plebhash, line 7: [ditto](https://github.com/stratum-mining/stratum/pull/905/files#r1603974488)

I could do this for every file, but it would be too repetitive.
I think you get the point

test/message-generator/test/standard-coverage-test/standard-coverage-test.json
```diff
@@ -181,27 +181,6 @@
...
-                            "output_location": "StdErr",
-                            "late_condition": false,
-                            "condition": true
-                        }
-                    ],
-                    "timer_secs": 320,
-                    "warn_no_panic": false
-                }
-            }
-        },
```
plebhash, line 204: why is this being removed?

test/message-generator/test/bad-pool-config-test/bad-pool-config-test.sh
lorbax: yes because we want to minimize  the risk that a test does not pass because of a timeout

test/message-generator/test/standard-coverage-test/standard-coverage-test.json
lorbax: because
1. the TP mock is launched in the same command above
2. there is no file at that address

.github/workflows/coverage.yaml
```diff
@@ -70,16 +70,59 @@ jobs:
...
-      - name: Run message generator tests
-        run: sh ./message-generator-tests.sh
+      - name: Run bad-pool-config-test
+        run: sh ./test/message-generator/test/bad-pool-config-test/bad-pool-config-test.sh
+
+      - name: Run interop-jd-translator
+        run: sh ./test/message-generator/test/interop-jd-translator/interop-jd-translator.sh
+
+          #- name: Run interop-jdc-change-upstream
+          #  run: sh ./test/message-generator/test/interop-jdc-change-upstream/interop-jdc-change-upstream.sh
```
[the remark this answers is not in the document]
lorbax: you are right, but this test wasn't performed even before this pr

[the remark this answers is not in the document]
plebhash: well sure but eventually we need to do something about it, right?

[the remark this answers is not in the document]
plebhash: ok, but eventually we need to do something about it, right?

see my comment https://github.com/stratum-mining/stratum/pull/905#issuecomment-2116078933

Fi3:
target directory should be cached (and shared) between action, so that you can reuse most of the already built artifact.  
I don't see the need to have a shell script for each test, I guess that having only one and passing to it the name of the test will be enough.

PR #908, opened 2024-05-14
Fix cpu miner panic when not header

Fi3:


roles/test-utils/mining-device/src/main.rs
```diff
@@ -595,17 +595,21 @@ impl Miner {
...
+                info!(
+                    "Found share with nonce: {}, for target: {:?}, with hash: {:?}",
+                    header.nonce, self.target, hash,
+                );
+                NextShareOutcome::ValidShare
+            } else {
+                NextShareOutcome::InvalidShare
+            }
         } else {
+            std::thread::yield_now();
```
plebhash, line 612: what is the purpose of this `yield_now`?

Fi3: avoid entering a spin loop, this i better than the hint as we can just wait here as it can happen only when the application is starting, the there will always be an header

plebhash: what do you mean by hint?

Fi3: https://doc.rust-lang.org/std/hint/index.html in that case a spin_loop hint

plebhash (approved):

Issue #791, opened 2024-03-12
`invalid-job-id` errors on `tProxy` don't result in rejection of shares


plebhash:
I was trying a BitAxe Ultra (BM1366) and I noticed multiple errors like this:
```
ERROR translator_sv2::lib::proxy::bridge: Submit share error Ok("invalid-job-id")
```

It also happens with `CPUminer`.

It was reported here: https://github.com/stratum-mining/stratum/issues/788

as [stated](https://github.com/stratum-mining/stratum/issues/788#issuecomment-1987870478) by @Fi3:
> This happens because the miner keep sending shares the for the old job and the translator mark them as invalid 

That explains the root cause of the error. Let's call these shares "delayed shares".

However, delayed shares are still accounted as "accepted" on the BitAxe side.

![image](https://github.com/stratum-mining/stratum/assets/147345153/5ba63f03-b2eb-4dce-9fa4-0b1865c0bcc1)

That is likely because the delayed shares only trigger an error event, but never really trigger some share rejection.

plebhash:
I'm labelling this issue as `good first issue`. As instructions on how to dive into this:
1. You would need an ASIC. CPUmining will not provide enough hashpower to reproduce this issue reliably.
2. Before starting, follow SRI [`Getting Started` tutorial](https://stratumprotocol.org/getting-started/) and do some CPU mining anyways. That will help you get familiar with the process of deploying a local SRI setup.
3. Try to reproduce it with your ASIC.

Fi3:
yep if I remember correctly the translator do not bother to send a response downstream for share error. I guess that this is mainly cause is supposed to be a proxy used by miners, so it is where metrics are collected and you use it as observable point. Said that returning errors for invalid shares would be a very nice thing to add. Also I don't know if the downstream need that message for stop sending shares and start working on the new job (don't think so but maybe is the case)

GitGab19:
Translator should be able to communicate to the miners the refused share whenever it's refused from upstream or it's not valid. What translator should do is to send downstream the following string: {"id": job_id, "result": `false`, "error": null} in those cases.
Right now it always sends it with `"result": true`, I just checked.

@plebhash we can reinsert the label good-first-issue imo, what do you think?

plebhash:
Cool, I'm putting the `good-first-issue` label back again.

We believe this is a good first issue because the fix is a relatively simple change, so it should provide some smooth intro for new contributors.

[...]

PR #910, opened 2024-05-16
[patch + MG test] tproxy receiving old share

plebhash:
This PR adds a MG test against the bug identified on https://github.com/stratum-mining/stratum/issues/791

Changelog:
- added the actual fix to the [#791](#791) bug
- adds MG test

Todo:
- it seems the bugfix is breaking `message-generator/test/translation-proxy/translation-proxy.json`


test/message-generator/test/translation-proxy-old-share.json
```diff
@@ -0,0 +1,105 @@
...
+                        [
+                            [
+                                "result",
+                                true
+                            ]
+                        ]
+                    ]
+                }
+            ],
+            "actiondoc": "Checks that the mining.submit request (with wrong job_id) generates a response with false, indicating that the share was rejected"
```
lorbax, line 58: this is not what the test does

lorbax:
look at how are the other tests. There is a directory with the name of the test, that contains the json tests and a script (that in your case will be `translation-proxy-old-share.sh`) which precompiles the binaries used by the test and the test itself. This is done in order to launch the test from the root directory of the project. Recall also to add the 
```
      - name: Run translation-proxy
        run: sh ./test/message-generator/test/translation-proxy/translation-proxy-old-share.sh
```
here
[https://github.com/plebhash/stratum/blob/mg-test-tproxy-old-share/.github/workflows/coverage.yaml#L123](https://github.com/plebhash/stratum/blob/mg-test-tproxy-old-share/.github/workflows/coverage.yaml#L123)

I would personally specify in the test (or wherever it is clear to see) that the test passes because there is a bug, and have to be changed whenever the bug is fixed.

lorbax (changes requested):
It must be added the machinery needed to run this new test among the others during the workflow of github actions.
Other two minor fixes should be addressed.

Apart from this, looks good to me.

test/message-generator/test/translation-proxy-old-share.json
plebhash: 78749a06cc81d8f43bb1b2e510bd544920e12343

plebhash:
@lorbax based on your feedback, I changed the following:
- added the actual fix to the bug, changing the following files:
  - `roles/translator/src/lib/downstream_sv1/downstream.rs`
  - `protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs`
  - `roles/translator/src/lib/proxy/bridge.rs`
- followed the new MG test template
- added a call to `test/message-generator/test/translation-proxy-old-share/translation-proxy-old-share.sh` on `.github/workflows/coverage.yaml`

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
xyephy, line 1: new error:
```
 v2/roles-logic-sv2/src/channel_logic/channel_factory.rs:52:10
     |
52   | pub enum OnNewShare {
     |          ^^^^^^^^^^
...
74   |     InvalidShare(SubmitSharesError<'static>),
     |     ------------ not covered
     = note: the matched value is of type `channel_factory::OnNewShare`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
     |
1982 ~             OnNewShare::ShareMeetDownstreamTarget => panic!(),
1983 ~             channel_factory::OnNewShare::InvalidShare(_) => todo!(),
```

plebhash: I removed `InvalidShare` from the PR, should be good now

xyephy (approved):

plebhash:
@lorbax can you approve the PR if you feel it's ok?

lorbax (approved):

plebhash:
it seems the bugfix is breaking `message-generator/test/translation-proxy/translation-proxy.json`

reverting to draft, since this needs to be fixed before the PR is merged

lorbax:
> it seems the bugfix is breaking `message-generator/test/translation-proxy/translation-proxy.json`
> 
> reverting to draft, since this needs to be fixed before the PR is merged

actually I think that your test is correct. Every time we launched `translation_proxy.json` MG test was kind of a false positive. This is due basically to the fact that we are using the `sv1-mining-device` test utils. This mining device is partially developed, and since the TProxy accepted the shares with invalid ID, presumably this mining device is sending only invalid shares.
I see the output of the error
image
which tells us that expected a SubmitSharesExtended and instead received an UpdateChannel. Clearly, since the TProxy is not receiving any valid share from the miner, SubmitSharesExtended will never be sent. On the other hand, it can send UpdateChannel to the upstream, and this behaviour is perfectly fine.
A list of things that we can do:

1. (the worst) change TM test action to await the UpdateChannel message, regardless of the invalid shares produced by the miner to the proxy. This is a rapid fix, but doing so we are actually giving up some interoperability. It may be viable for the purpose of this PR, given that we open an issue to fix the miner and solve it with a bit of prio. 
2. include the cpu miner binary in the repo and use that instead. It is a solution that works, but in general I am against the practice of including binaries in a repo.
3. fix the sv1-mining-device within this PR, but I actually don't know how much time is needed (I should look at the mining-device code)

Give me your impressions about it, I can help with it.

plebhash:
> > it seems the bugfix is breaking `message-generator/test/translation-proxy/translation-proxy.json`
> > reverting to draft, since this needs to be fixed before the PR is merged
> 
> actually I think that your test is correct. Every time we launched `translation_proxy.json` MG test was kind of a false positive. This is due basically to the fact that we are using the `sv1-mining-device` test utils. This mining device is partially developed, and since the TProxy accepted the shares with invalid ID, presumably this mining device is sending only invalid shares. I see the output of the error image which tells us that expected a SubmitSharesExtended and instead received an UpdateChannel. Clearly, since the TProxy is not receiving any valid share from the miner, SubmitSharesExtended will never be sent. On the other hand, it can send UpdateChannel to the upstream, and this behaviour is perfectly fine. A list of things that we can do:
> 
> 1. (the worst) change TM test action to await the UpdateChannel message, regardless of the invalid shares produced by the miner to the proxy. This is a rapid fix, but doing so we are actually giving up some interoperability. It may be viable for the purpose of this PR, given that we open an issue to fix the miner and solve it with a bit of prio.
> 2. include the cpu miner binary in the repo and use that instead. It is a solution that works, but in general I am against the practice of including binaries in a repo.
> 3. fix the sv1-mining-device within this PR, but I actually don't know how much time is needed (I should look at the mining-device code)
> 
> Give me your impressions about it, I can help with it.

that is a very good and comprehensive outline, thanks for the info!

I think the best approach is to fix `sv1-mining-device`. I found the bug:
https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/roles/test-utils/sv1-mining-device/src/job.rs#L25-L31

I will fix it on this same PR.

cc @rrybarczyk just so you're aware as seems you have some git history on this area of the codebase

plebhash:
@lorbax I fixed the bug described on my previous message via https://github.com/stratum-mining/stratum/pull/910/commits/61c572673b3f199c9a057f9c454a7b30aa8ea5b5

however, the problem persists when I run the following:
```
$ ./test/message-generator/test/translation-proxy/translation-proxy.sh
...
STD OUT: 2024-06-19T11:21:38.689859Z DEBUG translator_sv2::lib::downstream_sv1::downstream: To DOWN: OkResponse(Response { id: 0, error: None, result: Bool(false) })
STD OUT: 2024-06-19T11:21:38.689879Z DEBUG translator_sv2::lib::downstream_sv1::downstream: Sending to Mining Device: 127.0.0.1:35852 - "{\"id\":0,\"error\":null,\"result\":false}\n"
RECV Sv2Frame {
    header: Header {
        extension_type: 32768,
        msg_type: 22,
        msg_length: U24(
            40,
        ),
    },
    payload: None,
    serialized: Some(
        Slice {
            offset: 0x00007f007d68a010,
            len: 46,
            index: 1,
            shared_state: SharedState(
                128,
            ),
            owned: None,
        },
    ),
}
WRONG MESSAGE TYPE expected: 27 received: 22
STD OUT: 2024-06-19T11:21:38.744611Z ERROR translator_sv2::lib::status: Error: Io(Custom { kind: ConnectionAborted, error: "Connection closed by client" })
STD OUT: 2024-06-19T11:21:38.744645Z  WARN translator_sv2::lib::downstream_sv1::downstream: Downstream: Shutting down sv1 downstream reader
STD OUT: 2024-06-19T11:21:38.744654Z  INFO translator_sv2: HEALTHY message: I/O error: `Custom { kind: ConnectionAborted, error: "Connection closed by client" }
STD OUT: 2024-06-19T11:21:38.744666Z  WARN translator_sv2::lib::downstream_sv1::downstream: Downstream: Shutting down sv1 downstream job notifier for 127.0.0.1:35852
STD OUT: 2024-06-19T11:21:38.744676Z  WARN translator_sv2::lib::downstream_sv1::downstream: Downstream: Shutting down sv1 downstream writer: 127.0.0.1:35852
STD OUT: 2024-06-19T11:21:38.759169Z  INFO translator_sv2: Interrupt received
Disconnected from client while reading : early eof - 127.0.0.1:41024
PanicInfo {
    payload: Any { .. },
    message: Some(
        test failed!!!,
    ),
    location: Location {
        file: "src/executor.rs",
        line: 786,
        col: 13,
    },
    can_unwind: true,
    force_no_backtrace: false,
}
Disconnecting from client due to error receiving: receiving from an empty and closed channel - 127.0.0.1:41024
thread 'main' panicked at src/main.rs:438:13:
TEST FAILED
```

I suspect that `translator-proxy.json` also needs some adjusments

plebhash:
@lorbax I empirically that if I change `0x1b` for `0x16`, the test passes ok

https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/test/message-generator/test/translation-proxy/translation-proxy.json#L144-L154

lorbax:
> @lorbax I fixed the bug described on my previous message via [61c5726](https://github.com/stratum-mining/stratum/commit/61c572673b3f199c9a057f9c454a7b30aa8ea5b5)
> 
> however, `translation-proxy.sh` kept breaking at `WRONG MESSAGE TYPE expected: 27 received: 22`
> 
> That seems to come from the fact that shares are being rejected by tProxy, as we can verify on these logs:
> 
> ```
> STD OUT: 2024-06-19T11:55:52.361642Z DEBUG translator_sv2::lib::downstream_sv1::downstream: Down: Handling mining.submit: Submit { user_name: "user", job_id: "1", extra_nonce2: Extranonce(Owned([0, 0, 0, 0, 0, 0, 0, 0])), time: HexU32Be(1718798212), nonce: HexU32Be(3089), version_bits: None, id: 0 }
> STD OUT: 2024-06-19T11:55:52.361647Z DEBUG translator_sv2::lib::downstream_sv1::downstream: To DOWN: OkResponse(Response { id: 0, error: None, result: Bool(false) })
> STD OUT: 2024-06-19T11:55:52.361665Z DEBUG translator_sv2::lib::downstream_sv1::downstream: Sending to Mining Device: 127.0.0.1:44818 - "{\"id\":0,\"error\":null,\"result\":false}\n"
> ```
> 
> a `false` result on the context of `mining.submit` means the share was rejected, and the patch introduced by this PR is changing the logic on share rejection.
> 
> I tweaked `test/config/tproxy-config-no-jd-sv1-cpu-md.toml` to set a smaller `min_individual_miner_hashrate` via [aadd202](https://github.com/stratum-mining/stratum/commit/aadd2028ff7dc470943d5c830cf9a78c90de8dc6)
> 
> The goal was to avoid tProxy rejecting shares. The test seems stable now.

I think that what you've done is to make the t-proxy make a lot more shares that it is supposed to do and consequentially provide a share to the upstream before UpdateChannel is sent. Makes sense.

jbesraa:
@lorbax @plebhash Could we please make the commit history nicer by not pushing "fmt" or other similar commits into dev/main branches?

jbesraa:
I would revert this PR TBH, clear the git history and also get two approvals. First approval was not reconfirmed after additional code pushes.

plebhash:
> @lorbax @plebhash Could we please make the commit history nicer by not pushing "fmt" or other similar commits into dev/main branches?

sure, I think this is a good suggestion

maybe we could make a requirement of squashing PRs when are merged to `dev` branch?

@Fi3 what do you think?

> I would revert this PR TBH, clear the git history and also get two approvals. First approval was not reconfirmed after additional code changes.

that wouldn't be simple, because we would need to force-push to `dev`. I don't have rights for that.

but I think the most important action would be to establish this as a development policy and make sure all core devs are aligned

lorbax:
> I would revert this PR TBH, clear the git history and also get two approvals. First approval was not reconfirmed after additional code changes.

Your points are good. We can discuss it, I was agree with the author in merging it, but we should have found another reviewer!

PR #915, opened 2024-05-18
add `taplo` CI + format all TOML files

### LINKED ISSUE #914: formatting on TOML files
our TOML files have random formatting standards

we should have some CI to make sure all TOML files are properly formatted under the same standard, just like we do with `rustfmt`

plebhash:
close #914 by:
- adding a CI step based on [`taplo`](https://taplo.tamasfe.dev/)
- applying to all TOML files

### DIFF
--- a/protocols/v1/Cargo.toml
+++ b/protocols/v1/Cargo.toml
@@ -15,10 +15,15 @@ byteorder = "1.2.7"
 pretty_env_logger = "0.4.0"
 log = "0.4.0"
 hex = "0.4.3"
-serde = { version = "1.0.89", default-features = false, features = ["derive", "alloc"] }
-serde_json = { version = "1.0.64", default-features = false, features = ["alloc"] }
-tracing = {version = "0.1"}
-binary_sv2 = { version = "^1.0.0", path = "../v2/binary-sv2/binary-sv2"}
+serde = { version = "1.0.89", default-features = false, features = [
+  "derive",
+  "alloc",
+] }
+serde_json = { version = "1.0.64", default-features = false, features = [
+  "alloc",
+] }
+tracing = { version = "0.1" }
+binary_sv2 = { version = "^1.0.0", path = "../v2/binary-sv2/binary-sv2" }
 
 [dev-dependencies]
 quickcheck = "1"
--- a/.github/workflows/fmt.yaml
+++ b/.github/workflows/fmt.yaml
@@ -8,7 +8,7 @@ on:
       - main
       - dev
 
-name: Rustfmt
+name: Rustfmt + Taplo
 
 jobs:
   fmt:
@@ -32,6 +32,7 @@ jobs:
           toolchain: nightly
           override: true
           components: rustfmt
+
       - name: Run fmt in different workspaces and crates
         run: |
           cargo fmt --all --manifest-path=benches/Cargo.toml -- --check
@@ -40,3 +41,18 @@ jobs:
           cargo fmt --all --manifest-path=roles/Cargo.toml -- --check
           cargo fmt --all --manifest-path=utils/Cargo.toml -- --check
           cargo fmt --all --manifest-path=utils/message-generator/Cargo.toml -- --check
+
+      - name: Install Taplo
+        run: cargo install taplo-cli
+
+      - name: Run Taplo to check formatting on Cargo.toml files
+        id: taplo-check
+        run: taplo fmt --check -c .taplo-cargo.toml
+
+      - name: Run Taplo to check formatting on user-facing TOML files
+        id: taplo-check
+        run: taplo fmt --check -c .taplo-user-config.toml
+
+      - name: Run Taplo to check formatting on misc TOML files (not user-facing configs nor Cargo.toml)
+        id: taplo-check
+        run: taplo fmt --check -c .taplo-misc.toml
--- a/.taplo-cargo.toml
+++ b/.taplo-cargo.toml
@@ -1,0 +1,2 @@
+# these formatting rules only apply to Cargo.toml files
+include = ["*/**/Cargo.toml"]
--- a/.taplo-misc.toml
+++ b/.taplo-misc.toml
@@ -1,0 +1,7 @@
+# these formatting rules only apply to misc TOML files (not user-facing configs nor Cargo.toml)
+exclude = [
+  "roles/**/*config*.toml",
+  "test/config/**/*.toml",
+  "examples/**/*config*.toml",
+  "*/**/Cargo.toml",
+]
--- a/.taplo-user-config.toml
+++ b/.taplo-user-config.toml
@@ -1,0 +1,6 @@
+# these formatting rules only apply to user-facing TOML files
+include = [
+  "roles/**/*config*.toml",
+  "test/config/**/*.toml",
+  "examples/**/*config*.toml",
+]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,8 @@ The SRI project follows an open contributor model, where anyone is welcome to co
 
     `cargo fmt`: this command formats your code according to the project's style guidelines. Make sure to run this command to ensure consistency in code formatting.
 
+    `taplo fmt`: this command formats TOML files according to the project's style guidelines. Make sure to run this command to ensure consistency in code formatting.
+
 5. **Submit a Pull Request:** once you're satisfied with your changes, submit a pull request to the original SRI repository. Provide a clear and concise description of the changes you've made. If your pull request addresses an existing issue, reference the issue number in the description. In order to contribute to the protocol implementation, every PR must be opened against `dev` branch. To better understand which is the SRI branches structure, please have a look [here](https://github.com/stratum-mining/stratum/blob/main/RELEASE.md#principal-branches).
 
 6. **Review and Iterate** 
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -4,32 +4,39 @@ version = "1.0.0"
 edition = "2021"
 
 [dependencies]
-async-std={version = "1.10.0", features = ["attributes"]}
+async-std = { version = "1.10.0", features = ["attributes"] }
 criterion = "0.5.1"
 async-channel = "1.4.0"
-v1 = { path="../protocols/v1", package="sv1_api", version = "^1.0.0" }
-serde_json = { version = "1.0.64", default-features = false, features = ["alloc"] }
-iai="0.1"
+v1 = { path = "../protocols/v1", package = "sv1_api", version = "^1.0.0" }
+serde_json = { version = "1.0.64", default-features = false, features = [
+  "alloc",
+] }
+iai = "0.1"
 mining_sv2 = { path = "../protocols/v2/subprotocols/mining", version = "^1.0.0" }
 roles_logic_sv2 = { path = "../protocols/v2/roles-logic-sv2", version = "^1.0.0" }
 framing_sv2 = { version = "1.1.0", path = "../protocols/v2/framing-sv2" }
-serde = { version = "1.0.89", default-features = false, features = ["derive", "alloc"] }
+serde = { version = "1.0.89", default-features = false, features = [
+  "derive",
+  "alloc",
+] }
 num-bigint = "0.4.3"
 num-traits = "0.2.15"
-bitcoin="0.28.1"
-codec_sv2 = { path = "../protocols/v2/codec-sv2", features=["noise_sv2"] }
+bitcoin = "0.28.1"
+codec_sv2 = { path = "../protocols/v2/codec-sv2", features = ["noise_sv2"] }
 binary_sv2 = { path = "../protocols/v2/binary-sv2/binary-sv2" }
-network_helpers_sv2 = { path = "../roles/roles-utils/network-helpers", features=["async_std"] }
+network_helpers_sv2 = { path = "../roles/roles-utils/network-helpers", features = [
+  "async_std",
+] }
 rand = "0.8.4"
 
 [[bench]]
 name = "criterion_sv1_benchmark"
-path="benches/src/sv1/criterion_sv1_benchmark.rs"
+path = "benches/src/sv1/criterion_sv1_benchmark.rs"
 harness = false
 
 [[bench]]
 name = "iai_sv1_benchmark"
-path="benches/src/sv1/iai_sv1_benchmark.rs"
+path = "benches/src/sv1/iai_sv1_benchmark.rs"
 harness = false
 
 [[bench]]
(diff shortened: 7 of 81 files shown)


jbesraa:
Thanks @plebhash. Could you please split the second commit(the actual formatting) into commit-for-each-crate?

plebhash:
> Thanks @plebhash. Could you please split the second commit(the actual formatting) into commit-for-each-crate?

done

taplo.toml
```diff
@@ -0,0 +1,26 @@
...
+# Align consecutive entries vertically.
+align_entries = false
+# Append trailing commas for multi-line arrays.
+array_trailing_comma = true
+# Expand arrays to multiple lines that exceed the maximum column width.
+array_auto_expand = true
+# Collapse arrays that don't exceed the maximum column width and don't contain comments.
+array_auto_collapse = false
+# Omit white space padding from single-line arrays
+compact_arrays = true
```
rrybarczyk, line 11: Personally, I like the space between the `[` and `]` in arrays because it is easier on my eyes, if you don't have a preference my suggestion is:

```suggestion
# Include white space padding in single-line arrays for better readability.
compact_arrays = false
```

taplo.toml
```diff
@@ -0,0 +1,26 @@
+[formatting]
+# Align consecutive entries vertically.
+align_entries = false
+# Append trailing commas for multi-line arrays.
+array_trailing_comma = true
+# Expand arrays to multiple lines that exceed the maximum column width.
+array_auto_expand = true
+# Collapse arrays that don't exceed the maximum column width and don't contain comments.
```
rrybarczyk, line 8:
```suggestion
# Collapse arrays that do not exceed the maximum column width and do not contain comments.
```

taplo.toml
```diff
@@ -0,0 +1,26 @@
...
+array_auto_collapse = false
+# Omit white space padding from single-line arrays
+compact_arrays = true
+# Omit white space padding from the start and end of inline tables.
+compact_inline_tables = false
```
rrybarczyk, line 13: Just like with `compact_arrays`, I like to have spacing between `{` and `}`. If you agree in setting `compact_arrays = true`, we should change this to true as well:

```suggestion
# Include white space padding in inline tables for better readability.
compact_inline_tables = true
```

taplo.toml
```diff
@@ -0,0 +1,26 @@
...
+# Maximum column width in characters, affects array expagit nsion and collapse, this doesn't take whitespace into account.
+# Note that this is not set in stone, and works on a best-effort basis.
+column_width = 160
```
rrybarczyk, line 16: If you disagree with the above suggestion of setting this to 100 instead of 160, let's still update the comment (there is a minor typo there as well, "expagit nsion" -> "expansion". So if you do not merge any of my suggestions on this, be sure to fix that).

```suggestion
# Maximum column width in characters.
column_width = 160
```

taplo.toml
```diff
@@ -0,0 +1,26 @@
...
+# Indent based on tables and arrays of tables and their subtables, subtables out of order are not indented.
+indent_tables = false
+# The substring that is used for indentation, should be tabs or spaces (but technically can be anything).
+indent_string = '	'
+# Add trailing newline at the end of the file if not present.
+trailing_newline = true
+# Alphabetically reorder keys that are not separated by empty lines.
```
rrybarczyk, line 23:
```suggestion
# Do not reorder keys within tables; keep the original order.
```

taplo.toml
```diff
@@ -0,0 +1,26 @@
...
+reorder_keys = false
+# Maximum amount of allowed consecutive blank lines. This does not affect the whitespace at the end of the document, as it is always stripped.
```
rrybarczyk, line 25: Update comment for clarity of what setting is doing and ensure comments do not exceed 100 (if you keep 160 for max width, follow that instead).

```suggestion
# Limit the number of consecutive blank lines to 1 to reduce unnecessary whitespace. This does not
# affect the whitespace at the end of the document, as it is always stripped.
```

rrybarczyk (changes requested):
Love the addition of `taplo`.

* Should we add instructions somewhere to install `talpo` with `cargo install taplo-cli`?
* Should we add instructions on running `taplo` to apply the format with `taplo fmt`?

taplo.toml
plebhash: that would have been the opposite outcome

`compact_inline_tables = false` results in:
```
async-std           = { version = "1.10.0", features = [ "attributes" ] }
```

`compact_inline_tables = true` results in:
```
async-std           = {version = "1.10.0", features = [ "attributes" ]}
```

we don't need to change anything here

taplo.toml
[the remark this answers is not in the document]
plebhash: from these suggestions, the only one I would prefer to diverge from is `align_entries = true`

plebhash:
> * Should we add instructions somewhere to install `talpo` with `cargo install taplo-cli`?
> * Should we add instructions on running `taplo` to apply the format with `taplo fmt`?

I added some instructions https://github.com/stratum-mining/stratum/pull/915/commits/dcd231f7505219096617e2f73892954165afe603

rrybarczyk (approved):
I noticed that the `align = true` and the `sort = true` does not apply to all `Cargo.toml`s `[dependencies]` table (but appears to for all other tables), for example `roles/jd-server/Cargo.toml`. I have cloned this branch and ran `taplo` and tinkered with some settings, but I cannot figure out why these are not being universally applied. 

I think this is fine as is and we should merge. Just good to get this in the books and can always modify later.

plebhash:
> I noticed that the `align = true` and the `sort = true` does not apply to all `Cargo.toml`s `[dependencies]` table (but appears to for all other tables), for example `roles/jd-server/Cargo.toml`. I have cloned this branch and ran `taplo` and tinkered with some settings, but I cannot figure out why these are not being universally applied.
> 
> I think this is fine as is and we should merge. Just good to get this in the books and can always modify later.

Here's the taplo docs on the options available: https://taplo.tamasfe.dev/configuration/formatter-options.html

The `align_entries` is described as:
> Align entries vertically. Entries that have table headers, comments, or blank lines between them are not aligned.

So it does seem to come with some caveats and edge cases.

I fed `roles/jd-server/Cargo.toml` to chatGPT and asked why `align_entries` wasn't being applied to `[dependencies]`, and here's the answer:

> The alignment is not being applied to the dependencies section due to the presence of entries with multiline values (lists and tables) that are not aligned by the align_entries option in Taplo. This option only aligns single-line key-value pairs. When multiline values are present, alignment is not performed to avoid disrupting the readability and structure of these values.

> Here’s your dependencies section with comments indicating where multiline values are causing the alignment to be skipped:

```
[dependencies]
stratum-common      = { version = "1.0.0", path = "../../common" }
async-channel       = "1.5.1"
async-recursion     = "0.3.2"
binary_sv2          = { version = "^1.0.0", path = "../../protocols/v2/binary-sv2/binary-sv2" }
buffer_sv2          = { version = "^1.0.0", path = "../../utils/buffer" }
codec_sv2           = { version = "^1.0.1", path = "../../protocols/v2/codec-sv2", features = [
    "noise_sv2",
    "with_buffer_pool",
] }  # Multiline value
framing_sv2         = { version = "^1.1.0", path = "../../protocols/v2/framing-sv2" }
network_helpers_sv2 = { version = "2.0.0", path = "../roles-utils/network-helpers", features = [
    "with_tokio",
    "with_buffer_pool",
] }  # Multiline value
roles_logic_sv2     = { version = "^1.0.0", path = "../../protocols/v2/roles-logic-sv2" }
serde               = { version = "1.0.89", default-features = false, features = [ "derive", "alloc" ] }  # Multiline value
futures             = "0.3.25"
tokio               = { version = "1", features = [ "full" ] }  # Multiline value
toml                = { version = "0.5.6", git = "https://github.com/diondokter/toml-rs", default-features = false, rev = "c4161aa" }  # Multiline value
tracing             = { version = "0.1" }
tracing-subscriber  = { version = "0.3" }
error_handling      = { version = "1.0.0", path = "../../utils/error-handling" }
nohash-hasher       = "0.2.0"
key-utils           = { version = "^1.0.0", path = "../../utils/key-utils" }
```

> In this case, entries like codec_sv2, network_helpers_sv2, serde, tokio, and toml contain multiline values (e.g., arrays or tables), preventing the align_entries option from aligning the entire dependencies section.

---

with regards to the `sort` option, I couldn't find it anywhere in the docs so I'm actually wondering if that is really a feature that `taplo` provides?

@rrybarczyk I see you suggested its introduction [here](https://github.com/stratum-mining/stratum/pull/915#discussion_r1650148061), do you recall what you took as reference for that suggestion?

also note that `taplo` will not complain if you feed it an unexistant option... for example, having `ssssort = true` inside `.taplo.toml` doesn't elicit any errors when I do `taplo fmt`, so it might be the case that `sort = true` is simply being ignored because it's not a real option

in case we confirm that `sort` is not a real option for `taplo`, we should probably remove this line since it's not really doing anything

jbesraa (commented):
I am not really convinced about formatting `Cargo.toml` files. As long as they are not huge, the normal formatting is just fine. whats the motivation for including them here?

plebhash:
> I am not really convinced about formatting `Cargo.toml` files. As long as they are not huge, the normal formatting is just fine. whats the motivation for including them here?

original motivation was just to bring some uniformity to them... there's random patterns accross multiple files

jbesraa:
> > I am not really convinced about formatting `Cargo.toml` files. As long as they are not huge, the normal formatting is just fine. whats the motivation for including them here?
> 
> original motivation was just to bring some uniformity to them... there's random patterns accross multiple files
> 
> but also, if we include taplo on CI, we need all TOML files to conform to those rules, otherwise CI will break for every PR

I think if you are eager to format `Cargo.toml` files, the formatting should at least follow the original formatting. i.e.,  similar to what `cargo` produces when creating new crate. 

Cant `taplo` exclude files?

plebhash:
> > > I am not really convinced about formatting `Cargo.toml` files. As long as they are not huge, the normal formatting is just fine. whats the motivation for including them here?
> > 
> > 
> > original motivation was just to bring some uniformity to them... there's random patterns accross multiple files
> > but also, if we include taplo on CI, we need all TOML files to conform to those rules, otherwise CI will break for every PR
> 
> I think if you are eager to format `Cargo.toml` files, the formatting should at least follow the original formatting. i.e., similar to what `cargo` produces when creating new crate.

When I first started this PR, I introduced a `.taplo.toml` file without giving much thought.
Later on the first round of reviews @rrybarczyk made some suggestions on this config file.

I assume that using `taplo fmt` without any config file would resort to whatever `cargo` defaults are, but I haven't validated this in practice. 

tbh I'm fine with this simplified approach as well, I don't really care for one particular flavor of formatting... my overall aim is to just have **some** standard rather than none, which leads to random formatting and gives room for sloppy PRs.

@rrybarczyk how do you feel about this?

> Cant `taplo` exclude files?

from my research on CLI args, no... but it is able to selectively format files if you feed it via stdin.

for example, if you only do `taplo fmt`, it will recursively format all `.toml` files on that directory/subdirectories.

but if you do `taplo fmt config.toml`, it will ignore all other files on that directory/subdirectories and focus specifically on `config.toml`

rrybarczyk:
@plebhash my bad, I don't know where I got `sort` from, so yes let's remove since it DNE!

rrybarczyk:
> > > > I am not really convinced about formatting `Cargo.toml` files. As long as they are not huge, the normal formatting is just fine. whats the motivation for including them here?
> > > 
> > > 
> > > original motivation was just to bring some uniformity to them... there's random patterns accross multiple files
> > > but also, if we include taplo on CI, we need all TOML files to conform to those rules, otherwise CI will break for every PR
> > 
> > 
> > I think if you are eager to format `Cargo.toml` files, the formatting should at least follow the original formatting. i.e., similar to what `cargo` produces when creating new crate.
> 
> When I first started this PR, I introduced a `.taplo.toml` file without giving much thought. Later on the first round of reviews @rrybarczyk made some suggestions on this config file.
> 
> I assume that using `taplo fmt` without any config file would resort to whatever `cargo` defaults are, but I haven't validated this in practice.
> 
> tbh I'm fine with this simplified approach as well, I don't really care for one particular flavor of formatting... my overall aim is to just have **some** standard rather than none, which leads to random formatting and gives room for sloppy PRs.
> 
> @rrybarczyk how do you feel about this?
> 
> > Cant `taplo` exclude files?
> 
> from my research on CLI args, no... but it is able to selectively format files if you feed it via stdin.
> 
> for example, if you only do `taplo fmt`, it will recursively format all `.toml` files on that directory/subdirectories.
> 
> but if you do `taplo fmt config.toml`, it will ignore all other files on that directory/subdirectories and focus specifically on `config.toml`

Yes I am good with the simplified approach!

plebhash:
@jbesraa I removed `.taplo.toml` with the opinionated formatting from the PR

now, we are just doing the default formatting

jbesraa:
> @jbesraa I removed `.taplo.toml` with the opinionated formatting from the PR
> 
> now, we are just doing the default formatting

You can exclude `Cargo.toml` from `taplo`, see this https://taplo.tamasfe.dev/configuration/file.html#exclude. 
There example is exactly what we want: you specify which folder you want to include in the `include` prop and in `exclude` you ignore files from the `included` stuff.

plebhash:
> My main concern here is that we are using the same CI process + same formatting setup for a config intended for users and a config(Cargo.toml) we use, requirements between those two could diverge easily

I rebased. Now we have three separate `taplo` config files, each applying to specific categories of TOML files.

They don't have any specific formatting policies however, only `inlcude`/`exclude` filters.

@jbesraa do you have any specific ideas on formatting policies, or did you just want to make this PR "future-proof" in that sense?

protocols/v1/Cargo.toml
```diff
@@ -15,10 +15,15 @@ byteorder = "1.2.7"
 pretty_env_logger = "0.4.0"
 log = "0.4.0"
 hex = "0.4.3"
-serde = { version = "1.0.89", default-features = false, features = ["derive", "alloc"] }
-serde_json = { version = "1.0.64", default-features = false, features = ["alloc"] }
-tracing = {version = "0.1"}
-binary_sv2 = { version = "^1.0.0", path = "../v2/binary-sv2/binary-sv2"}
+serde = { version = "1.0.89", default-features = false, features = [
```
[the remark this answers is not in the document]
plebhash: I'm not sure, the `.taplo*.toml` files are all empty with regards to actual formatting policies, they only do filtering for which file they will format

so I assume this is coming from the default formatting policy that `taplo` brings

plebhash:
reverting this back to Draft status

merging this will have an undesired impact on our semver checking on CI, see this comment for more details: https://github.com/stratum-mining/stratum/pull/985#issuecomment-2181652506

we should get https://github.com/stratum-mining/stratum/pull/985 sorted out before merging this

@pavlenex we can remove this from milestone 1.0.2, since #985 could take a while and we don't want the next release blocked

rrybarczyk:
Maybe someday.

PR #925, opened 2024-05-23
Release 1.0.1

plebhash:
Global Release `v1.0.1`.

# PR Changelog
- #933
- #929
- #930
- #927
- #924
- #926
- #904
- #905
- #913
- #900
- #908
- #875
- #894
- #848
- #898
- #866
- #879
- #800
- #802
- #805
- #887

# Crate Version changes

## `protocols`
- `framing-sv2`: v1.0.0 → v1.1.0
- `codec-sv2`: v1.0.1 → v1.1.0
- `roles-logic-sv2`: v1.0.0 → v1.1.0

## `roles`
- `jd-client`: v0.1.0 → 0.1.1
- `jd-server`:  v0.1.0 → v0.1.1
- `mining-proxy`: v0.1.0 → v0.1.1
- `pool`: v0.1.0 → v0.1.1
- `translator`: v0.1.0 → v0.1.1
- `roles-utils/network-helpers`: v0.1.0 → v2.0.0

## `utils`
- `key-utils`: v1.0.0 → v1.1.0

PR #927, opened 2024-05-23
bump codec sv2 version

plebhash:
while fixing `check-versioning-lib-release.sh` (https://github.com/stratum-mining/stratum/pull/926) and preparing for Release `v1.0.1` (https://github.com/stratum-mining/stratum/pull/925), I noticed we were forgetting to bump `codec_sv2` crate version, as a result of a patch applied to one of it's dependencies (`framing_sv2`) via https://github.com/stratum-mining/stratum/pull/913

`framing_sv2` (which is a dependency of `codec_sv2`) had it's `MINOR` version bumped, which means `codec_sv2` also needs to bump it's `MINOR` version

### DIFF
--- a/protocols/v2/codec-sv2/Cargo.toml
+++ b/protocols/v2/codec-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codec_sv2"
-version = "1.0.1"
+version = "1.1.0"
 authors = ["fi3 <email@email.org>"]
 edition = "2018"
 description = "Sv2 data format"


GitGab19:
Since it was a patch, why `framing_sv2` had its MINOR bumped, instead of PATCH?

plebhash:
> Since it was a patch, why `framing_sv2` had its MINOR bumped, instead of PATCH?

https://github.com/stratum-mining/stratum/pull/848 was not a patch, it added docs and renamed/cleaned up some code

GitGab19 (approved):

PR #932, opened 2024-05-25
make `no_std` into an optional feature on `protocols` crates

plebhash:
Many of the low level libs under `protocols` are `no_std`. That is great, since it allows for potential firmware integrations using `rust-embedded` and targets without MMUs (e.g.: Xtensa ESP32, ARM Cortex-M, etc).

Some projects in the community even have some potential to leverage SRI `protocols` crates for firmware integration, e.g.: https://github.com/bitaxeorg/esp-miner-rs

However, for most common deployment scenarios (e.g.: x86-64), high level application code (e.g.: `roles`) has no hard requirement for `no_std`. Also, having it as a mandatory feature becomes a blocker to many desirable changes.

### DIFF
--- a/protocols/v2/binary-sv2/serde-sv2/Cargo.toml
+++ b/protocols/v2/binary-sv2/serde-sv2/Cargo.toml
@@ -13,3 +13,6 @@ repository = "https://github.com/stratum-mining/stratum"
 [dependencies]
 serde = { version = "1.0.89", features = ["derive", "alloc"], default-features = false }
 buffer_sv2 = {version = "^1.0.0",  path = "../../../../utils/buffer"}
+
+[features]
+no_std = []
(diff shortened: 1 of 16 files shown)


plebhash:
> I am not fully sure adding `anyhow` is a great solution for our error handling. We could maintain different errors internally and expose one single error that maps to the rest. Our minimum for adding a new crate should be more than "it makes this thing slightly easier", in this case its preferable to write our own code. Disadvantages of adding anyhow could be: slower compiling time, we could be imposing `anyhow` for downstream users, its more appropriate for app layer..

that's actually a good point, and @rrybarczyk just made a similar argument just a while ago.

rrybarczyk:
A little late to the party, but as @plebhash pointed out, I also think we should handle our own errors in the roles rather than using `anyhow`.

plebhash:
I'm trying to unblock https://github.com/stratum-mining/stratum/pull/985#issuecomment-2234328927 and it seems that this (amongst other reasons) is a blocker there

rrybarczyk (approved):

PR #933, opened 2024-05-25
Update `Cargo.lock` files with version updates

plebhash:
As we prepare for Release v1.0.1 (https://github.com/stratum-mining/stratum/pull/925), we recently bumped the versions for a few modified crates:

- #930 
- #927 
- #913 

I made a mistake on those PRs: I forgot to run `cargo build` to update all `Cargo.lock` files that list these crates as dependencies.

This also happened on Release v1.0.0 and `Cargo.lock` became PITA as a result.

This PR should avoid repeating this mistake on Release v1.0.1.

---

Also, as I built the `examples` and `utils` crates, I noticed git was trying to commit some `target` directories and `Cargo.lock` files. So this PR also fixes `.gitignore` such that those files are now ignored by git.


examples/interop-cpp/Cargo.lock
GitGab19, line 1: Do we really need to commit Cargo.lock for this?
Looking at our current repo state, we are not committing them for inner crates

.gitignore
```diff
@@ -1,9 +1,10 @@
 .idea
-*/target
```
GitGab19, line 2: I would leave also */target, why did you remove that?

plebhash: `*/**/target` also covers `*/target`

GitGab19: Sure, you're right 👍

GitGab19 (approved):

PR #935, opened 2024-05-26
Standardize `roles` CLI

plebhash:
close https://github.com/stratum-mining/stratum/issues/456 (indirectly)
close https://github.com/stratum-mining/stratum/issues/813 (directly)

---

as discussed on https://github.com/stratum-mining/stratum/discussions/857, this PR is proposing some standardization changes to the CLI of all `roles` crates.

I tried to split this PR into multiple on a per-role basis, but unfortunately that is not possible.
That is because we are removing `toml` crate, which is a custom fork.

If we remove it from one crate in `roles`, but leave it in others, that breaks workspace compilation. Many CI actions compile at the workspace level, which means that breaking this PR into separate roles is not a viable option. All roles need to get rid of `toml` dependency on the same PR, so that the entire workspace is free from this dependency.

Also, all the proposed changes below are closely correlated, so it's not practical to submit separate PRs for each.

changelog:
- [x] [`toml` dependency replacement](https://github.com/stratum-mining/stratum/discussions/857#discussioncomment-9143815)
- [x] [`clap` for managing CLI args](https://github.com/stratum-mining/stratum/discussions/857#discussioncomment-9143819)
- [x] [`args` module split](https://github.com/stratum-mining/stratum/discussions/857#discussioncomment-9143824)
- [x] [`*Result` and `*Error` structs](https://github.com/stratum-mining/stratum/discussions/857#discussioncomment-9143827)
- [x] [`*Config` structs](https://github.com/stratum-mining/stratum/discussions/857#discussioncomment-9143831)

All of these modifications was applied to each role implementation, namely:
- [x] `jd-server`
- [x] `jd-client`
- [x] `mining-proxy`
- [x] `pool`
- [x] `translator` 

So unfortunately, this PR is very similar to https://github.com/stratum-mining/stratum/pull/814, in the sense that they are both quite big.

Although it might seem like this will be a time-consuming review (too many files), I kindly ask the reviewers to pay attention to the patterns that are already described on the changelog list above. That's a very good description of all the changes this PR is introducing.

Also, in order to make the PR review more practical, the reviewer can resort to automated [Message Generator](https://github.com/stratum-mining/stratum/tree/main/utils/message-generator) tests to make sure that nothing is broken.

---

Note: this is a low-prio PR (thus draft). I will explicitly ask for reviews when the timing is appropriate.

### DIFF
--- a/roles/Cargo.lock
+++ b/roles/Cargo.lock
@@ -54,12 +54,6 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.3.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8fd72866655d1904d6b0997d0b07ba561047d070fbe29de039031c641b61217"
-
-[[package]]
-name = "ahash"
 version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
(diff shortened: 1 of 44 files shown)


plebhash:
closing in favor of https://github.com/stratum-mining/stratum/pull/937

PR #939, opened 2024-05-27
Specify version on filenames inside `Release Bin` CI

plebhash:
I was trying to release the binary executables and I noticed there was a naming clash with the previous release.

https://github.com/stratum-mining/stratum/actions/runs/9258045890/job/25467603017#step:13:46

This PR is adjusting the `Release Bin` CI Job so that each output executable has the Global Release version as a substring.

That way, there will be no name clashing in the future.

It starts from the assumption that the `Release Bin` Job was triggered from a branch named `vMAJOR.MINOR.PATCH`, which is highlighted on the changes to `RELEASE.md`

### DIFF
--- a/.github/workflows/release-bin.yaml
+++ b/.github/workflows/release-bin.yaml
@@ -49,7 +49,7 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: roles/target/release/pool_sv2
-          asset_name: pool_sv2-x86_64-linux-gnu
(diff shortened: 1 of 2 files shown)


.github/workflows/release-bin.yaml
GitGab19, line 1: I would leave `sv2` instead of `sri`, so that it's more clear and in accordance to what we already uploaded

plebhash: https://github.com/stratum-mining/stratum/pull/939/commits/013131fd5d11e7a05a254cfa3cda99940d722b6b

GitGab19 (approved):

PR #949, opened 2024-06-01
Message Generator: fix primitives serialization, fix broken unit tests, added unit test for (de)serialization of NMJ message

lorbax:
Unity tests in main.rs of MG are fixed. One is added. This required to fix the serialization the primitives in the serde case (/protocols/v2/binary_sv2/serde_sv2)
test.json fixed in src directory of MG
mining-proxy config file fixed


protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/mod.rs
```diff
@@ -22,6 +23,23 @@ struct Seq<'s, T: Clone + Serialize + TryFromBSlice<'s>> {
...
 
+impl<'s, T: Clone + Serialize + TryFromBSlice<'s>> Seq<'s, T> {
+    pub fn parse(self) -> Result<Vec<T>, ()> {
+        let mut ret = Vec::with_capacity(self.size as usize);
+        let elem_len = self.size as usize;
+        let mut first = 0;
+        let mut last = elem_len;
+        while last <= self.data.len() {
+            let slice = &self.data[first..last];
+            let elem = T::try_from_slice(slice).map_err(|_| ())?;
```
jbesraa, line 34: Should we add an error variant here?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -65,11 +68,20 @@ impl<'s, T: Clone + Serialize + TryFromBSlice<'s>> Serialize for Sv2Option<'s, T
...
+                    seq.end()
+                } else {
+                    let tuple = (data.len() as u8, &data[..]);
+                    let mut seq = serializer.serialize_tuple(2)?;
+                    seq.serialize_element(&tuple.0)?;
+                    seq.serialize_element(tuple.1)?;
+                    seq.end()
+                }
             }
             _ => panic!(),
```
jbesraa, line 86: Could you handle this panic please?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -227,7 +239,15 @@ impl<'s> Sv2Option<'s, U256<'s>> {
                 data: Some(data),
             }
         } else {
-            panic!()
+            // this is an already valid seq should be safe to call the unwraps.
```
jbesraa, line 242: why is this unwrap safe?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -242,7 +262,14 @@ impl<'s> Sv2Option<'s, u32> {
                 data: Some(inner),
             }
         } else {
-            panic!()
+            // this is an already valid seq should be safe to call the unwraps.
```
jbesraa, line 265: why is this unwrap safe?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
```diff
@@ -68,11 +72,20 @@ impl<'s, T: Clone + Serialize + TryFromBSlice<'s>> Serialize for Seq0255<'s, T>
                 seq.end()
             }
             (None, Some(data)) => {
-                let tuple = (data.len() as u8, &data[..]);
-                let mut seq = serializer.serialize_tuple(2)?;
-                seq.serialize_element(&tuple.0)?;
-                seq.serialize_element(tuple.1)?;
-                seq.end()
+                if serializer.is_human_readable() {
+                    let data_ = self.data.clone().unwrap();
```
jbesraa, line 76: same, why not use `data` from `match` result?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
```diff
@@ -230,7 +243,15 @@ impl<'s> Seq0255<'s, U256<'s>> {
                 data: Some(data),
             }
         } else {
-            panic!()
+            // this is an already valid seq should be safe to call the unwraps.
+            // also this library shouldn't be used for priduction envs so is ok do thigs like this
+            // one
+            let data = self.seq.unwrap().parse().unwrap();
```
jbesraa, line 249: why is this unwrap safe?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
```diff
@@ -245,7 +266,14 @@ impl<'s> Seq0255<'s, u32> {
                 data: Some(inner),
             }
         } else {
-            panic!()
+            // this is an already valid seq should be safe to call the unwraps.
+            // also this library shouldn't be used for priduction envs so is ok do thigs like this
+            // one
+            let data = self.seq.unwrap().parse().unwrap();
```
jbesraa, line 272: why is this unwrap safe?

jbesraa (commented):
@lorbax did first initial pass. Also,  could you please tide up the commits so you have a couple of detailed commits explaining the problem and the solution you are proposing?

protocols/v2/binary-sv2/serde-sv2/src/primitives/byte_arrays/b064k.rs
```diff
@@ -4,7 +4,7 @@ use core::convert::{TryFrom, TryInto};
 use serde::{de::Visitor, ser, ser::SerializeTuple, Deserialize, Deserializer, Serialize};
 
 #[derive(Debug, Clone, Eq)]
-enum Inner<'a> {
+pub enum Inner<'a> {
```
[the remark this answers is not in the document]
lorbax: makes sense

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -65,11 +68,20 @@ impl<'s, T: Clone + Serialize + TryFromBSlice<'s>> Serialize for Sv2Option<'s, T
                 seq.end()
             }
             (None, Some(data)) => {
-                let tuple = (data.len() as u8, &data[..]);
-                let mut seq = serializer.serialize_tuple(2)?;
-                seq.serialize_element(&tuple.0)?;
-                seq.serialize_element(tuple.1)?;
-                seq.end()
+                if serializer.is_human_readable() {
+                    let data_ = self.data.clone().unwrap();
```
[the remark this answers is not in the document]
lorbax: Because I had to do this modification several times and I did past and copy, but you are right I should use the `data` in the match branch!
Thank you!

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/mod.rs
lorbax: An `crate::error::Error::PrimitiveConversionError` should be the right solution, what do you think? 
BTW this part of code is used only in the message generator and not for production.

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
lorbax: Yes because if `self.data` is `None` then `self.seq` is `Some(Inner)` (it is the same as above!)

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
lorbax: my mistake!

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
lorbax: Same as above, if `self.data` is `None` then `self.seq` is `Some()`

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
lorbax: same as above

lorbax:
> @lorbax did first initial pass. Also, could you please tide up the commits so you have a couple of detailed commits explaining the problem and the solution you are proposing?

Do you suggest to have just 2 commits? I don't think that it would be better as each commit describe a precise small change (except for fmt)

lorbax:
> @lorbax did first initial pass. Also, could you please tide up the commits so you have a couple of detailed commits explaining the problem and the solution you are proposing?

Tomorrow I will remove the fmt btw. 
What do you suggest for apply the review? to just apply changes on older commits and force pushing or apply one review commit with a comment pointing to this PR? I would like to do it properly...

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
jbesraa: maybe add `assert_debug!(false, "EXPLAIN_WHY_THIS_SHOULD_NOT_HAPPEN")` instead

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/mod.rs
jbesraa: Ah, we can leave as is then

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
jbesraa: Hm, that sounds risky. Here we also have two unwraps. 
maybe something like this:
```rust
impl<'s> Seq0255<'s, u32> {
    pub fn into_static(self) -> Seq0255<'static, u32> {
        match (self.data, self.seq) {
            (Some(data),_) => {
            },
            (_, Some(seq)) => {
            },
            _ => {
                debug_assert!(false, "This is unreachable because `req` most be defined if `data` is `None`");
                unreachable!();
            }

        }
    }
}
```

jbesraa:
Thanks @lorbax for the answers. You can add as many commits as you like as long as they are reasonable(not fmt or typo fixes) and informative. 

I prefer that you add `f [INFO_ABOUT_FIX]` commits and after review they can be squashed, but I am happy with either.

lorbax:
> Thanks @lorbax for the answers. You can add as many commits as you like as long as they are reasonable(not fmt or typo fixes) and informative.
> 
> I prefer that you add `f [INFO_ABOUT_FIX]` commits and after review they can be squashed, but I am happy with either.

Thank you! 
Tomorrow I'll read in detail your replies. Thanks for your review!!!

protocols/v2/binary-sv2/serde-sv2/src/primitives/byte_arrays/b064k.rs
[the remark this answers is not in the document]
lorbax: actually it is not even necessary to expose Inner, I removed `pup` and seems to be compiling without issues ^^

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
lorbax: done

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
lorbax: I believe because this library is used only for the MessageGenerator and the seq variant is never used in that case, but we should ask @fi3 for that

jbesraa:
@lorbax feel free to ping me when this is ready for another review

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
Fi3: we need data "variant" for it to be static we can't have a static seq

roles/mining-proxy/config-examples/proxy-config-example.toml
```diff
@@ -1,5 +1,5 @@
 upstreams = [
-    { channel_kind = "Extended", address = "0.0.0.0", port = 34265, pub_key = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"}
+    { channel_kind = "Extended", address = "0.0.0.0", port = 34254, pub_key = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"}
```
plebhash, line 2: why is this being changed here? seems totally unrelated to this PR

lorbax:
@jbesraa took me a while, but now I have a better understanding of `sequences` datatypes in `serde_sv2`. They are a struct with two fields, one of type `Option<Vec<T>>` and the other `Option<Seq<T>>`. This struct behaves as an enum with `data: Vec<T>` and `seq: Seq<T>` variants. At the time of writing the code, it was written like this because was more practical for some reasons (that I don't know, haven't tried to replace it with a true enum).
At the beginning, `serde_sv2` was equivalent to `no_serde` library for serialization and deserialization of datatypes in binary format. They had the same functionality and it was possible to use the application with serde by just importing some libraries with `with_serde` features (for having an idea of which of them imported like this, look at the Message Generator toml). Now I strongly suggest to do not use roles that import libraries with "with_serde" feature, they are not working anymore with that. Nevertheless, it is possible to use this libraries with "with_serde", and indeed it is the correct way to do this, because the MG it is meant to work with this feature.
Now I try to explain what I understood.
In MG test `pool-sri-test-extended-1` there is an action that awaits the message `NewExtendedMiningJob` . This message contains the fields `min_ntime` and `merkle_path` which are resp a Sv2Option (which is a Seq0255<U32> under the hood) and Seq0255<U256> . Before deserialization these are represented with the `seq: Some(T)` and `data: None`. This is the tipical backtrace if we add a panic [here](https://github.com/stratum-mining/stratum/blob/dev/protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs#L174)
the call of the function `parsers::TryFrom` [here](https://github.com/stratum-mining/stratum/blob/dev/utils/message-generator/src/executor.rs#L249) will eventually panic.
Worth noting that deserialization is performed using the custom deserializer in `binary_sv2::serde_sv2`. I still have to understand why `is_human_readable` is used to distinguish when `seq` field or `data` field have to be used. I suspect that for serialization `is_human_readable` is true, and for deserialization is set to false (in `serde_sv2::ser` and `serde_sv2::de`). In this case, perhaps it may be possible to use a light version of serde_sv2 just for deserialization, and standard serde (with few modifications) for serialization.

jbesraa:
@lorbax thanks for digging into this! Maybe its worth breaking this pr to smaller ones if possible so you can get things merged while you are looking into the serde stuff

lorbax:
> @lorbax thanks for digging into this! Maybe its worth breaking this pr to smaller ones if possible so you can get things merged while you are looking into the serde stuff

the PR is ready as it is, I won't dig any longer for this serde stuff if not needed.
I am a bit reluctant to split it in smaller pieces, but if it gets not reviewed I will eventually do it.

jbesraa:
It would be nice to change the commits to describe how/what/why its fixed rather than `fix X`, also I am not sure all comments are addressed? for example the `pub Inner`..

lorbax:
> It would be nice to change the commits to describe how/what/why its fixed rather than `fix X`, also I am not sure all comments are addressed? for example the `pub Inner`..

Now I changed the commit history, let me see it does look fine to you

jbesraa:
sorry for the delay here @lorbax, Ill review this in a bit

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/mod.rs
jbesraa: Having second thoughts about this. As it is exposed through the public API, it might make more sense to handle errors properly here.
Also, I have a small question:
If `self.size` is bigger than `self.data.len`, is returning empty array with `self.size` cap(the current behavior) makes sense or an error should be returned?

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -65,13 +69,28 @@ impl<'s, T: Clone + Serialize + TryFromBSlice<'s>> Serialize for Sv2Option<'s, T
                 seq.end()
             }
             (None, Some(data)) => {
-                let tuple = (data.len() as u8, &data[..]);
-                let mut seq = serializer.serialize_tuple(2)?;
-                seq.serialize_element(&tuple.0)?;
-                seq.serialize_element(tuple.1)?;
-                seq.end()
+                if serializer.is_human_readable() {
+                    let data_ = data.clone();
```
jbesraa, line 73: u can use masking and do `let data = data.clone();`. I think its more clear

jbesraa: you might also be able to remove this `clone` as `serialize_element` takes a reference

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -65,13 +69,28 @@ impl<'s, T: Clone + Serialize + TryFromBSlice<'s>> Serialize for Sv2Option<'s, T
...
+                    let mut seq = serializer.serialize_tuple(2)?;
+                    seq.serialize_element(&tuple.0)?;
+                    seq.serialize_element(tuple.1)?;
+                    seq.end()
+                }
+            }
+            _ => {
+                debug_assert!(
+                    false,
+                    "sv2option can never have boh fields of the same type"
```
jbesraa, line 90: Is there an action expected from a user who hit this? regarding how to debug it? If is, its worth mentioning it IMO

protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/option.rs
```diff
@@ -219,15 +238,30 @@ impl<'a, T: Clone + FixedSize + Serialize + TryFromBSlice<'a>> GetSize for Sv2Op
...
     pub fn into_static(self) -> Sv2Option<'static, U256<'static>> {
-        if let Some(inner) = self.data {
-            let inner = inner.clone();
-            let data = inner.into_iter().map(|i| i.into_static()).collect();
-            Sv2Option {
-                seq: None,
-                data: Some(data),
+        match (self.data, self.seq) {
+            (None, Some(seq)) => {
+                let data = seq.parse().unwrap();
```
jbesraa, line 243: It would be useful to use `.expect` and explain why this should be `Some`

jbesraa (changes requested):
@lorbax Thanks for breaking this down. 
I would appreciate if @plebhash can validate the MG part.

I wonder also if we could avoid some of the repeated code, for example this code (and a couple of others) is repeated across different files
```rust
      if serializer.is_human_readable() {
                    let data_ = self.data.clone().unwrap();
                    let mut seq = serializer.serialize_seq(Some(data_.len()))?;
                    for item in data_ {
                        seq.serialize_element(&item)?;
                    }
                    seq.end()
                } else {
                    let tuple = (data.len() as u16, &data[..]);
                    let mut seq = serializer.serialize_tuple(2)?;
                    seq.serialize_element(&tuple.0)?;
                    seq.serialize_element(tuple.1)?;
                    seq.end()
                }
```

I understand that those changes intended for MG but they live in `protocols` as public APIs, so I think we should probably handle all edge cases. Let me know if you think otherwise.

utils/message-generator/src/main.rs
```diff
@@ -710,111 +733,123 @@ mod test {
...
+                "--datadir=./test/appdata/bitcoin_data/",
+                "generatetoaddress",
+                "16",
+                "bcrt1qttuwhmpa7a0ls5kr3ye6pjc24ng685jvdrksxx",
+            ],
+            ExternalCommandConditions::None,
+        )
+        .await;
+        child.unwrap().wait().await.unwrap();
+        */
```
plebhash, line 762: why are we keeping these comments?

`//dont clean the following commented code` doesn't give any context

utils/message-generator/src/main.rs
```diff
@@ -710,111 +733,123 @@ mod test {
...
+        // dont clean the following commented code
+        /*
+        let mut child = os_command(
+            "rm",
+            vec!["-rf", "./test/appdata/bitcoin_data/regtest"],
+            ExternalCommandConditions::None,
+        )
+        .await;
+        child.unwrap().wait().await.unwrap();
+        */
```
plebhash, line 828: why are we keeping these comments?

`//dont clean the following commented code` doesn't give any context

utils/message-generator/src/main.rs
```diff
@@ -710,111 +733,123 @@ mod test {
...
+        assert!(true)
+    }
 
+    //// the previous version of this test used mining-proxy, which is currently broken. Possibly
+    //// this is due to the lack of elligator swift supporto for the mining-proxy, and in general the
+    //// handshake needs maintainance
     //#[tokio::test]
     //async fn it_test_against_remote_endpoint() {
-    //    let proxy = match os_command(
+    //    let _proxy = match os_command(
```
plebhash, line 841: why is this being changed if it's still disabled?

utils/message-generator/src/main.rs
lorbax: because can still be useful in future

utils/message-generator/src/main.rs
lorbax: because `it_test_against_remote_endpoint` is a test that I want to make it pass and not just ignore it. It is also true that there is nothing against to uncomment it and have it as "non-passing test"

utils/message-generator/src/main.rs
Fi3: maybe you want to change the:
        //dont clean the following commented code
with: TODO fix the below tests

the first comment don't say anything about why you want to keep the below comment and people get confused.

utils/message-generator/src/main.rs
Fi3: same as above. comments are not memo for you.

Fi3:
for me is ok also one PR

jbesraa:
> for me is ok also one PR

Where is the second PR? 
 I feel we are mixing MG stuff with protocol level stuff and iam not a big fan of that..

lorbax:
I recently saw that we have decided to close [PR1068](https://github.com/stratum-mining/stratum/pulls/1068) because we want to replace the MG with something else for the CI. I don't see why this does not apply to this PR too, which is needed only for the MG. The library `serde_sv2`  works only with the MG (anything else is likely to be broken)

pavlenex:
> I recently saw that we have decided to close [PR1068](https://github.com/stratum-mining/stratum/pulls/1068) because we want to replace the MG with something else for the CI. I don't see why this does not apply to this PR too, which is needed only for the MG. The library `serde_sv2` works only with the MG (anything else is likely to be broken)

I wanted to ask this question as well, cc @jbesraa @plebhash Could we  add it to the list https://github.com/stratum-mining/stratum/issues/1077 and just close for now?

plebhash:
> > I recently saw that we have decided to close [PR1068](https://github.com/stratum-mining/stratum/pulls/1068) because we want to replace the MG with something else for the CI. I don't see why this does not apply to this PR too, which is needed only for the MG. The library `serde_sv2` works only with the MG (anything else is likely to be broken)
> 
> I wanted to ask this question as well, cc @jbesraa @plebhash Could we add it to the list #1077 and just close for now?

#1041 is already keeping track of this

#1077 is tracking issues of a different nature

but yeah we can close for now

PR #975, opened 2024-06-18
Add missing messages errors  constructors

pythcoiner:
This PR implement some missing messages errors 'constructor' in order to have all errors types from specs implemented.

### DIFF
--- a/protocols/v2/subprotocols/job-declaration/src/declare_mining_job.rs
+++ b/protocols/v2/subprotocols/job-declaration/src/declare_mining_job.rs
@@ -1,10 +1,8 @@
-#[cfg(not(feature = "with_serde"))]
-use alloc::vec::Vec;
+use alloc::{format, string::ToString, vec::Vec};
 #[cfg(not(feature = "with_serde"))]
 use binary_sv2::binary_codec_sv2;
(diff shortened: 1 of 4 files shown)


pythcoiner:
Btw, i don't really like the way i populate the `error_details` field, as this field is optionnal, shouldn't it have an Option<B064K> type? Or B064K should have `None` variant if length == 0? should `error_details` passed as argument?

jbesraa (commented):
I see new functions added but are not used anywhere. what iam missing? 
could you please provide a link to the mentioned error types?

pythcoiner:
> > I see new functions added but are not used anywhere. what iam missing? could you please provide a link to the mentioned error types?
> 
> I hold a similar view. Could you provide more context on the motivations behind this PR?

a part of the errors messages defined in the specs had 'constructors' implemented but not all, this PR try to add more consistency by implementing missing ones.

jbesraa:
Yea but there is no point in adding functions that are not used. If we are not handling errors probably we should that, otherwise I dont see a point in pushing this code to dev

rrybarczyk:
Closing, see [this comment](https://github.com/stratum-mining/stratum/pull/975#issuecomment-2200292376).

PR #969, opened 2024-06-12
Refactor `Framing` crate

jbesraa:
relates to #903 

This commit aims to refactor the `framing-sv2` crate. List of changes:
0. Cleand up `header.rs`
1. Removed `Frame` trait
2. Changed `EitherFrame` enum to `Frame`
3. Removed unused functions from `HandShakeFrame` and `Sv2Frame`
4. Added documentation where missing
5. Made sure there is consistency in function naming in `Frame`, i.e, in
   order to get the header you call `Frame::header`, for payload
   `Frame::payload` and so on.
6. Made sure we dont return mut data from `Frame` get functions(leave it
   to the caller to decide).
7. Fixed all the modules in `protocols` based on the above changes
8. Fixed all the modules in `roles` based on the above changes

For 7 & 8 its mainly removing the `Frame` trait import plus naming fixes
where I renamed `StandardEitherFrame` to `StandardFrame` and other name
fixes and also fixing the `header` and `payload` calls to align with the
new API.

Issue #928, opened 2024-05-23
MG CI issues


plebhash:
even though we merged https://github.com/stratum-mining/stratum/pull/905, we still haven't fixed all issues on MG CI

@Fi3 has some suggestions here: https://github.com/stratum-mining/stratum/pull/905#issuecomment-2122146609

cc @lorbax

plebhash:
another one while running [`translation-proxy`](https://github.com/stratum-mining/stratum/actions/runs/9213712745/job/25348434528?pr=929#step:21:2070) on #929:

```
EXECUTING ../../test/message-generator/test/translation-proxy/translation-proxy.json

...
STD OUT: 2024-05-23T19:44:16.766948Z DEBUG roles_logic_sv2::channel_logic::channel_factory: Upstream target: "000002cbd3e84cb8c95196d545017a21b956ff20c06c583a8aaebeee82979711"
STD OUT: 2024-05-23T19:44:16.766954Z DEBUG roles_logic_sv2::channel_logic::channel_factory: Hash: "000062117926ff023b81b40081470d60629b7122e89a344de62e863e97f56a17"
STD OUT: 2024-05-23T19:44:16.766959Z DEBUG translator_sv2::lib::proxy::bridge: SHARE MEETS DOWNSTREAM TARGET
STD OUT: 2024-05-23T19:44:16.766965Z DEBUG translator_sv2::lib::downstream_sv1::downstream: Sending to Mining Device: 127.0.0.1:52420 - "{\"id\":0,\"error\":null,\"result\":true}\n"
RECV Sv2Frame {
    header: Header {
        extension_type: 32768,
        msg_type: 22,
        msg_length: U24(
            40,
        ),
    },
    payload: None,
    serialized: Some(
        Slice {
            offset: 0x00007fbc2a97a010,
            len: 46,
            index: 1,
            shared_state: SharedState(
                128,
            ),
            owned: None,
        },
    ),
}
WRONG MESSAGE TYPE expected: 27 received: 22
STD OUT: 2024-05-23T19:44:16.819583Z ERROR translator_sv2::lib::status: Error: Io(Custom { kind: ConnectionAborted, error: "Connection closed by client" })
STD OUT: 2024-05-23T19:44:16.819617Z  WARN translator_sv2::lib::downstream_sv1::downstream: Downstream: Shutting down sv1 downstream reader
[...]

Issue #807, opened 2024-03-21
Improve flaky MG tests


pavlenex:
Placeholder,until someone more knowledgable document sspecifics, but life shouldn't be waiting for MG tests to ✅ , so we should enure MG tests aren't flaky.

![Waiting-Memes-52918](https://github.com/stratum-mining/stratum/assets/36959754/17c0a65b-b6e4-4f01-b92a-f2e76a5efa32)

plebhash:
this is a similar issue from a while ago: https://github.com/stratum-mining/stratum/issues/377

we should check if it is still relevant, and if not, close it

pavlenex:
https://blog.btcpayserver.org/bitcoin-bonus-scam/

plebhash:
I'm starting some investigation on this. I notice MG is quite non-deterministic, so I think a good first step is to start logging there the different error messages we get with it. I'll do that in the comments below.

plebhash:
```
EXECUTING ../../test/message-generator/test/bad-pool-config-test.json

STD ERR:     Finished dev [unoptimized + debuginfo] target(s) in 0.06s
STD ERR:      Running `target/debug/message_generator_sv2 ../../test/message-generator/mock/template-provider-mock0.json`
STD OUT: 
STD OUT: EXECUTING ../../test/message-generator/mock/template-provider-mock0.json
STD OUT: 
STD ERR: info: cargo-llvm-cov currently setting cfg(coverage); you can opt-out it by passing --no-cfg-coverage
STD ERR:    Compiling libc v0.2.153
STD ERR:    Compiling cfg-if v1.0.0
STD ERR:    Compiling version_check v0.9.4
STD ERR:    Compiling typenum v1.17.0
STD ERR:    Compiling proc-macro2 v1.0.79
STD ERR:    Compiling generic-array v0.14.7
STD ERR:    Compiling unicode-ident v1.0.12
STD ERR:    Compiling cpufeatures v0.2.12
STD ERR:    Compiling cc v1.0.90
STD ERR:    Compiling getrandom v0.2.12
STD ERR:    Compiling rand_core v0.6.4
STD ERR:    Compiling quote v1.0.35
STD ERR:    Compiling syn v2.0.55
STD ERR:    Compiling opaque-debug v0.3.1
STD ERR:    Compiling zeroize v1.7.0
STD ERR:    Compiling crypto-common v0.1.6
STD ERR:    Compiling inout v0.1.3
[...]

PR #976, opened 2024-06-18
`Framing` crate refactor: cleanup `header.rs` and remove `Frame` trait

### LINKED ISSUE #903: refactor: `framing_sv2` crate

## rename `framing2.rs`

## Frame trait removal

## code organization

## `sv2_frame::get_header`

jbesraa:
I am breaking #969 into smaller PRs because its touching too many parts of the code and doing it all in a single PR could result in endless conflicts and make it also hard to review. 

In this part the `header.rs` file is tied up with a few things removed and replaced by constants and adding docs. the second commit removes the `Frame` trait and adjust the code accordingly. 

Partially resolves #903

### DIFF
--- a/benches/benches/src/sv2/criterion_sv2_benchmark.rs
+++ b/benches/benches/src/sv2/criterion_sv2_benchmark.rs
@@ -1,4 +1,4 @@
-use codec_sv2::{Frame, StandardEitherFrame, StandardSv2Frame};
+use codec_sv2::{StandardEitherFrame, StandardSv2Frame};
 use criterion::{black_box, Criterion};
 use roles_logic_sv2::{
     handlers::{common::ParseUpstreamCommonMessages, mining::ParseUpstreamMiningMessages},
--- a/benches/benches/src/sv2/iai_sv2_benchmark.rs
+++ b/benches/benches/src/sv2/iai_sv2_benchmark.rs
@@ -1,4 +1,4 @@
-use codec_sv2::{Frame, StandardEitherFrame, StandardSv2Frame};
+use codec_sv2::{StandardEitherFrame, StandardSv2Frame};
 use iai::{black_box, main};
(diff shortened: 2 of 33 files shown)


plebhash (approved):

jbesraa:
@plebhash did you change anything here?

plebhash:
> @plebhash did you change anything here? I see you committed but cant see diff

I rebased to keep it up with `dev` branch

image

jbesraa:
rebased without further changes

rrybarczyk (approved):
This looks good to me. Any idea why the MG `bad-pool-config-test` is failing in the CI?

Screen Shot 2024-07-01 at 10 47 36 AM

jbesraa:
> This looks good to me. Any idea why the MG `bad-pool-config-test` is failing in the CI?
> Screen Shot 2024-07-01 at 10 47 36 AM

I think its a false alarm

plebhash:
> > This looks good to me. Any idea why the MG `bad-pool-config-test` is failing in the CI?
> > Screen Shot 2024-07-01 at 10 47 36 AM
> 
> I think its a false alarm

correct, this is a false alarm

unfortunately the state of our MG CI is that we get many of those, and the solution is to keep re-running them with this button until they're all green:

image

only if they keep failing in a deterministic way we know that there's actually something wrong with it

very suboptimal, I know! something should be done about this... MG CI already takes a really long time to run, and by having to re-run it several times (due to false alarms or rebasing) is something that really slows down development

we have some issues for that:
- https://github.com/stratum-mining/stratum/issues/928
- https://github.com/stratum-mining/stratum/issues/807

Fi3:
Not a solution but running them locally should be more stable.

plebhash:
I just rebased this PR to get it ready for merging

jbesraa:
@plebhash I think this is good to go

PR #977, opened 2024-06-18
Refactor ping pong example

plebhash:
This PR is refactoring the `examples/ping-pong-without-noise` example into `examples/ping-pong`.

I wrote it from scratch as an exercise to get familiar with `protocols` crates.

The overall idea is quite similar to the previous implementation, but with some simplifications to make the example easier to understand.

---

`ping-pong` is an example of how to encode and decode SV2 binary frames (without any encryption layer) while leveraging the following crates:
- [`binary_sv2`](http://docs.rs/binary_sv2)
- [`codec_sv2`](http://docs.rs/codec_sv2)
- [`framing_sv2`](http://docs.rs/framing_sv2) (which is actually just re-exported by `codec_sv2`)

We establish a simple `Ping`-`Pong` protocol with a server and a client communicating over a TCP socket.

The server expects to receive a `Ping` message encoded as a SV2 binary frame.
The `Ping` message contains a `nonce`, which is a `u8` generated randomly by the client.

The client expects to get a `Pong` message in response, also encoded as a SV2 binary frame, with the same `nonce`.

The messages are assigned arbitrary values for binary encoding:
```rust
pub const PING_MSG_TYPE: u8 = 0xfe;
pub const PONG_MSG_TYPE: u8 = 0xff;
```


plebhash:
@jbesraa @Shourya742 @pythcoiner since you're all interested in learning more about the low-level crates, how about we do a "PR Review Club"-style call? We could follow this agenda:
- explain PR
- code walkthrough
- Q&A

We could record it and share on [SRI YouTube](https://www.youtube.com/@Stratumv2) to spread the knowledge to the community.

Also, this could be a good opportunity to brainstorm design decisions on the `framing_sv2` API refactoring that's going on in https://github.com/stratum-mining/stratum/pull/969

pythcoiner:
> @jbesraa @Shourya742 @pythcoiner since you're all interested in learning more about the low-level crates, how about we do a "PR Review Club"-style call? We could follow this agenda:
> 
> * explain PR
> * code walkthrough
> * Q&A
> 
> We could record it and share on [SRI YouTube](https://www.youtube.com/@Stratumv2) to spread the knowledge to the community.
> 
> Also, this could be a good opportunity to brainstorm design decisions on the `framing_sv2` API refactoring that's going on in #969 that @jbesraa and @Shourya742 are working on

sure!

examples/ping-pong/src/client.rs
```diff
@@ -0,0 +1,82 @@
...
+    let ping_encoded = encoder.encode(ping_frame)?;
+
+    println!("CLIENT: Sending Ping to server with nonce: {}", ping_nonce);
+    stream.write_all(ping_encoded)?;
+
+    // ok, we have successfully sent the ping message
+    // now it's time to receive and verify the pong response
+
+    // initialize decoder
+    let mut decoder = StandardDecoder::<Ping>::new();
```
Shourya742, line 35:
```suggestion
    let mut decoder = StandardDecoder::<Pong>::new();
```
I believe this should actually be Pong.

Shourya742 (approved):
tACK

examples/ping-pong/src/client.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    // right now, the decoder buffer can only read a frame header
```
jbesraa, line 37: sorry, I am not getting why the first read fails, could you pleas elaborate?

jbesraa, line 37: sorry, I am not getting why the first read fails, could you pleas elaborate?

jbesraa (commented):
Looks good. Added one small question.
Could you please add a gh workflow to run this example instead of the removed one?

examples/ping-pong/src/client.rs
plebhash: > It seems like we're only reading the first 6 bytes of this specific message type and not reaching the total byte length, which causes an error signaling that more bytes are needed to match the payload size. 

correct

> But I'm wondering, why are we taking this approach? Couldn't we just read all the bytes in one go and then decipher from there?

this is technical debt we need to clean up

https://github.com/stratum-mining/stratum/issues/903 is the first step of that journey

plebhash:
> @jbesraa @Shourya742 @pythcoiner since you're all interested in learning more about the low-level crates, how about we do a "PR Review Club"-style call? We could follow this agenda:
> 
> * explain PR
> * code walkthrough
> * Q&A
> 
> We could record it and share on [SRI YouTube](https://www.youtube.com/@Stratumv2) to spread the knowledge to the community.
> 
> Also, this could be a good opportunity to brainstorm design decisions on the `framing_sv2` API refactoring that's going on in #969 that @jbesraa and @Shourya742 are working on

[recording](https://drive.google.com/file/d/1tNqUeTo75KVUXbUTOXrtMy77RDU2e2xE/view?usp=sharing)

jbesraa:
@plebhash Thanks alot for the call and the thorough explanation about this.  
I think this would be good to go after you add it to github actions and remove the old one, and squash all the commits into a single one

plebhash:
> @plebhash Thanks alot for the call and the thorough explanation about this. I think this would be good to go after you add it to github actions and remove the old one, and squash all the commits into a single one

done

jbesraa (approved):

examples/ping-pong/src/client.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    // initialize decoder
+    let mut decoder = StandardDecoder::<Pong>::new();
+
+    // right now, the decoder buffer can only read a frame header
+    // because decoder.missing_b is initialized with a header size
+    let mut decoder_buf = decoder.writable();
+
+    // read frame header into decoder_buf
+    stream.read_exact(&mut decoder_buf)?;
```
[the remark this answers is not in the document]
pythcoiner: @plebhash, i think you forgot this one :smile:

examples/ping-pong/src/server.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    // initialize decoder
+    let mut decoder = StandardDecoder::<Ping>::new();
+
+    // right now, the decoder buffer can only read a frame header
+    // because decoder.missing_b is initialized with a header size
+    let mut decoder_buf = decoder.writable();
+
+    // read frame header into decoder_buf
+    stream.read_exact(&mut decoder_buf)?;
```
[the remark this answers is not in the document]
plebhash: done, thanks!

pythcoiner:
utACK [60b722a](https://github.com/stratum-mining/stratum/pull/977/commits/60b722a542ec1d336654a1075980b886c52a1542)

pythcoiner:
utACK 6e3334b

plebhash:
reverting back to draft, since this could be affected by #903 (and other similar issues that are output from #845

rrybarczyk:
Note: Wait to merge until the following is merged:
- #1049 
- #1033

plebhash:
> reverting back to draft, since this could be affected by #903 (and other similar issues that are output from #845 )



> Note: Wait to merge until the following is merged:
> 
> * [`Framing` refactor: test `Sv2Frame` and `HandShakeFrame` #1049](https://github.com/stratum-mining/stratum/pull/1049)
> * [`Framing` refactor: `framing.rs` api #1033](https://github.com/stratum-mining/stratum/pull/1033)

the tasks mentioned above have been de-prioritized

@Shourya742 is working to remove `async_std` support from SRI, and this PR helps achieve that goal

therefore, I'm removing the draft status so we can merge this soon

GitGab19 (approved):
tACK

PR #981, opened 2024-06-19
Add MSRV 1.75 workflow check

### LINKED ISSUE #980: Add MSRV check in CI
currently the crates MSRV is 1.75 and included in the different `.toml` files and others.
this would allow us to remove all the version pinning.

jbesraa:
resolves #980

### DIFF
--- a/.github/workflows/rust-msrv.yaml
+++ b/.github/workflows/rust-msrv.yaml
@@ -1,0 +1,37 @@
+on:
+  push:
+    branches:
+      - main
+      - dev
+  pull_request:
+    branches:
+      - main
+      - dev
+
+name: MSRV 1.75 Check
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        rust:
+          - 1.75.0 # MSRV
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1.2.0
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - name: Build Benches
+        run: cargo build --manifest-path=benches/Cargo.toml
+      - name: Build Protocols
+        run: cargo build --manifest-path=protocols/Cargo.toml
+      - name: Build Roles
+        run: cargo build --manifest-path=roles/Cargo.toml
+      - name: Build Utils
+        run: cargo build --manifest-path=utils/Cargo.toml
--- a/protocols/Cargo.toml
+++ b/protocols/Cargo.toml
@@ -2,7 +2,6 @@ name = "stratum_v2_protocols"
 version = "1.0.0"
 authors = ["The Stratum v2 Developers"]
 edition = "2021"
-rust-version = "1.75.0"
 description = "The Stratum protocol defines how miners, proxies, and pools communicate to contribute hashrate to the Bitcoin network. Stratum v2 is a robust set of primitives which anyone can use to expand the protocol or implement a role."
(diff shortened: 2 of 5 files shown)


protocols/Cargo.toml
```diff
@@ -2,7 +2,6 @@ name = "stratum_v2_protocols"
 version = "1.0.0"
 authors = ["The Stratum v2 Developers"]
 edition = "2021"
-rust-version = "1.75.0"
```
plebhash, line 5: I get the point of CI, but why is it desirable to remove it from the different `Cargo.toml` files?

.github/workflows/rust-msrv.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+      # IF NEEDED
+      # - name: Fixes for MSRV
+      #   if: matrix.rust == '1.75.0'
+      #   run: |
+          # cargo update
+          # cargo update -p [CRATE_NAME] --precise [VERSION]
```
plebhash, line 35: what is the context for this?

jbesraa: Sometimes other crate introduce MSRV changes and we will want to pin those crates into the latest version that supports our MSRV

jbesraa: here is an example https://github.com/payjoin/rust-payjoin/blob/master/.github/workflows/rust.yml#L26

plebhash: under which circumstances would that be needed?

The codebase already has a lot of littering in terms of comments saying "todo" and things of that nature. So we should aim for a mindset where we are always aiming for clean code.

I think it's good to be strategize about potential solutions that we might apply for future problems, but maybe we should document these kind of things on different places.

plebhash:
what is the implication about MSRV on the project as a whole?

we have a section touching on MSRV on [README](https://github.com/stratum-mining/stratum/blame/7841021e3f20280c62af119de5de58f421c6e428/README.md#L125), should we modify it somehow?

cc @pavlenex @Fi3

.github/workflows/rust-msrv.yaml
jbesraa: Yea good point. deleted.

protocols/Cargo.toml
Shourya742: Yes, that has happened to me as well; my code editor continuously updates the `Cargo.lock` file because of this pinning. It would be nice to remove this pinning.

jbesraa:
@plebhash I dont think we have much to update. If we had any `crate` we needed to pin in order to run on 1.75, we would mention it in the README.md, but we are good for now

plebhash (approved):

jbesraa:
Thanks @rrybarczyk 

protocols/Cargo.toml
jbesraa: the `README.md` already mention the MSRV, not sure about the hooks tbh as it seems its not widely used

rrybarczyk (approved):
Ltgm. Do we want to add a bash script in `scipts/` to call this? I can also do this in the #1039 PR I have open if you want to add that but don't want to do it in this PR.

rrybarczyk:
@jbesraa can you rebase and we will merge?

jbesraa:
> Ltgm. Do we want to add a bash script in `scipts/` to call this? I can also do this in the #1039 PR I have open if you want to add that but don't want to do it in this PR.

lets do it in #1039 please

jbesraa:
@plebhash I think this is good to go

PR #982, opened 2024-06-19
`Framing` crate refactor: cleanup `framing.rs` 

### LINKED ISSUE #903: refactor: `framing_sv2` crate

jbesraa:
`framing-sv2` Crate refactor p2, first part here #976 

Partially resolves #903 

This is part two of the refactor of `framing-sv2` crate. 

Code changes:
- Rename `framing2.rs` to `framing.rs`
- Reorder `framing.rs` file
- Remove unused functions from `HandShakeFrame`
- Rename `framing::EitherFrame` to `framing::Frame`


rrybarczyk (changes requested):
Looks great. Very happy to see the renaming. Looks like you need to update this branch with `main`, then we can merge.

jbesraa:
> Looks great. Very happy to see the renaming. Looks like you need to update this branch with `main`, then we can merge.

Thank you!

This PR is part two of the refactor, this PR https://github.com/stratum-mining/stratum/pull/976 should reviewed and merged first, so I would appreciate your review there as well. thanks!

plebhash:
> Looks great. Very happy to see the renaming. Looks like you need to update this branch with `main`, then we can merge.

do you mean `dev`?

jbesraa:
rebased without further changes

rrybarczyk (approved):

plebhash (approved):

PR #983, opened 2024-06-19
Add parallelism to cpu miner

Fi3:
Make the cpu miner multithread as requested by @Sjors


roles/test-utils/mining-device/Cargo.toml
```diff
@@ -24,3 +23,4 @@ clap = { version = "^4.5.4", features = ["derive"] }
 tracing = { version = "0.1" }
 tracing-subscriber = "0.3"
 sha2 = "0.10.6"
+tokio = "^1.36.0"
```
jbesraa, line 26: Could we use latest `1.38.0` version?

roles/test-utils/mining-device/src/main.rs
```diff
@@ -1,10 +1,14 @@
-use async_std::net::TcpStream;
 use key_utils::Secp256k1PublicKey;
-use network_helpers_sv2::Connection;
+use network_helpers_sv2::noise_connection_tokio::Connection;
 use roles_logic_sv2::utils::Id;
-use std::{net::SocketAddr, sync::Arc, thread::sleep, time::Duration};
+use std::{
+    net::{SocketAddr, ToSocketAddrs},
+    sync::Arc,
+    thread::sleep,
```
jbesraa, line 11: `thread::sleep` is used in the `connect` method which uses `tokio` as well, so I think it would be better to use `tokio::time::sleep` here

roles/test-utils/mining-device/src/main.rs
```diff
@@ -276,6 +285,7 @@ impl Device {
             prev_hash: None,
             channel_id: None,
             sequence_numbers: Id::new(),
+            notify_changes_to_mining_thread: (true, notify_changes_to_mining_thread),
```
jbesraa, line 288: could we move `(true, notify_changes_to_mining_thread)` to a type to improve readability? will also eliminate the `.0` \ `.1` ambiguity

roles/test-utils/mining-device/src/main.rs
```diff
@@ -286,24 +296,8 @@ impl Device {
...
-                    .try_send((nonce, job_id.unwrap(), version.unwrap(), time))
-                    .unwrap();
-            }
-            miner
-                .safe_lock(|m| m.header.as_mut().map(|h| h.nonce += 1))
-                .unwrap();
-        });
-
-        async_std::task::spawn(async move {
+        start_mining_threads(update_miners, miner, share_send);
```
jbesraa, line 299: I think it would be possible to combine this call with the call in L300 and the `loop` afterward(all async) in a `tokio::select!` macro, but that could be a different task

roles/test-utils/mining-device/src/main.rs
```diff
@@ -669,3 +674,70 @@ fn generate_random_32_byte_array() -> [u8; 32] {
...
 }
+
+fn start_mining_threads(
+    have_new_job: Receiver<()>,
+    miner: Arc<Mutex<Miner>>,
+    share_send: Sender<(u32, u32, u32, u32)>,
+) {
+    tokio::task::spawn(async move {
+        let mut killers: Vec<Arc<AtomicBool>> = vec![];
+        loop {
```
jbesraa, line 685: nit: maybe we could use only `tokio` in this function with https://docs.rs/tokio/latest/tokio/runtime/#multi-thread-scheduler

jbesraa (commented):
Looks good, added a few small comments

roles/test-utils/mining-device/src/main.rs
Fi3: my plan was to dedicate only one thread for the async runtime (I forgot to do it) and have one dedicated thread for each miner. I think is better like that cause miners thread have no blocking operation. I would stick with my plan. If you have any particular reason in mind to no do it pls let me know.

roles/test-utils/mining-device/src/main.rs
Fi3: yea I didn't see it ty

roles/test-utils/mining-device/src/main.rs
```diff
@@ -669,3 +674,70 @@ fn generate_random_32_byte_array() -> [u8; 32] {
...
+            let p = std::thread::available_parallelism().unwrap().get() as u32 - 1;
```
Shourya742, line 686: Noob Questions:
1. In systems that have only a single core without hyper-threading, wouldn't `p`  reduce to 0, potentially leading to undefined behavior? Should we add a check to prevent this?
2. Regarding nonce distribution, we are currently not utilizing the entire nonce space, as `u32::MAX % p ` values are not considered, although the number missed is relatively small, at most `p-1`.

Shourya742 (commented):
Concept ACK.
nit: https://github.com/stratum-mining/stratum/blob/3f2a191bba40dd18bbe403fcd53247cc0f8fda24/roles/test-utils/mining-device/src/main.rs#L47. `rapresents` -> `represents`

Sjors:
While you're at it, can you add a README file to `roles/test-utils/mining-device`?

Mined a block! https://mempool.space/testnet4/block/00000000f6b1355bcf0077bce98b21566728ad10bd7e0856dbc38e0bb2c9f74a

roles/test-utils/mining-device/src/main.rs
Fi3: I mark as solved for now let me know if you still want to use select

roles/test-utils/mining-device/src/main.rs
Fi3: I'm going to dedicate only one thread to tokio runtime and all the others to mining. I don't see any advantage in using work stealing since mining thread are constantly doing something. Mark as resolved for now, let me know if you are ok with that

roles/test-utils/mining-device/src/main.rs
```diff
@@ -669,3 +674,70 @@ fn generate_random_32_byte_array() -> [u8; 32] {
...
+                let job_id = miner.job_id.unwrap();
+                let version = miner.version;
+                share_send
+                    .try_send((nonce, job_id, version.unwrap(), time))
+                    .unwrap();
+            }
+            miner.header.as_mut().map(|h| h.nonce += 1);
+        }
+    }
+}
```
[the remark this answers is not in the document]
Fi3: I dont' like it. I don't want branches in this code cause is very hot. I know that cpu branch predictor should handle  this case very well, but I still prefer to keep them separated.

roles/test-utils/mining-device/src/main.rs
Fi3: 1. definitively.
2. I don't see it as an issue.

Fi3:
@GitGab19 there are issues with benches no way that this PR impact the performance of the SRI libraries

GitGab19:
> @GitGab19 there are issues with benches no way that this PR impact the performance of the SRI libraries 

Yeah I definitely need to figure out what's going on there. I will try to fix it by eof

roles/test-utils/mining-device/README.md
```diff
@@ -0,0 +1,21 @@
...
+  -a, --address-pool <ADDRESS_POOL>  Address of the pool in this format ip:port or domain:port
+      --handicap <HANDICAP>          This value is used to slow down the cpu miner, it represents the number of micro-seconds that are awaited between hashes [default: 0]
+      --id-user <ID_USER>            User id, used when a new channel is opened, it can be used by the pool to identify the miner
+  -h, --help                         Print help
+  -V, --version                      Print version
+```
+
+Usage example:
+```
+cargo run --release -- --address-pool 127.0.0.1:20000 --id-device device_id::SOLO::bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh
```
Sjors, line 20: I guess `--address-upstream` is more accurate, but that's unrelated to this PR

roles/test-utils/mining-device/src/main.rs
```diff
@@ -669,3 +686,74 @@ fn generate_random_32_byte_array() -> [u8; 32] {
...
+            if miner.next_share().is_valid() {
+                let nonce = miner.header.unwrap().nonce;
+                let time = miner.header.unwrap().time;
+                let job_id = miner.job_id.unwrap();
+                let version = miner.version;
+                share_send
+                    .try_send((nonce, job_id, version.unwrap(), time))
+                    .unwrap();
+            }
+            miner.header.as_mut().map(|h| h.nonce += 1);
```
johnnyasantoss, line 740: Aren't all the threads searching the *same* nonce space?

Fi3: look at line 713

roles/test-utils/mining-device/src/main.rs
```diff
@@ -322,6 +324,17 @@ impl Device {
...
+                .safe_lock(|s| s.notify_changes_to_mining_thread.clone())
+                .unwrap();
+            if notify_changes_to_mining_thread.should_send {
+                notify_changes_to_mining_thread
+                    .sender
+                    .send(())
+                    .await
+                    .unwrap();
+                notify_changes_to_mining_thread.should_send = false;
+            };
```
plebhash, line 337: currently `Device` is notifying changes to mining threads on **any** message received from upstream, including  `SubmitShareSuccess`

so every time `Device` submits a new share, it restarts all mining threads and start from scratch, which results in shares being submitted repeatedly:
```
2024-07-17T19:52:30.201538Z  INFO mining_device: Found share with nonce: 1185614795, for target: Some(0x00000005b5246dc512185093364f5f8fb78345dd01ec6397d5cd30e1232d5039), with hash: 0x00000004ac66a0b64a11bca5ad95832c12f28f5e3ce88468b750a538a28f51e8
2024-07-17T19:52:30.203086Z  INFO mining_device: SUCCESS SubmitSharesSuccess { channel_id: 6, last_sequence_number: 3, new_submits_accepted_count: 1, new_shares_sum: 0 }
2024-07-17T19:52:42.398146Z  INFO mining_device: Found share with nonce: 1185614795, for target: Some(0x00000005b5246dc512185093364f5f8fb78345dd01ec6397d5cd30e1232d5039), with hash: 0x00000004ac66a0b64a11bca5ad95832c12f28f5e3ce88468b750a538a28f51e8
2024-07-17T19:52:42.399670Z  INFO mining_device: SUCCESS SubmitSharesSuccess { channel_id: 6, last_sequence_number: 4, new_submits_accepted_count: 1, new_shares_sum: 0 }
```

the suggested changes below make sure that notification is only sent when `Device` receives a `NewMiningJob`, `SetNewPrevHash` or `SetTarget` (did I miss any message that could be relevant on this context? @Fi3 please double check)
```suggestion
            if notify_changes_to_mining_thread.should_send
                && (message_type == const_sv2::MESSAGE_TYPE_NEW_MINING_JOB
                    || message_type == const_sv2::MESSAGE_TYPE_SET_NEW_PREV_HASH
                    || message_type == const_sv2::MESSAGE_TYPE_SET_TARGET)
            {
                notify_changes_to_mining_thread
                    .sender
                    .send(())
                    .await
                    .unwrap();
                notify_changes_to_mining_thread.should_send = false;
            };
```

roles/test-utils/mining-device/src/main.rs
```diff
@@ -669,3 +686,74 @@ fn generate_random_32_byte_array() -> [u8; 32] {
...
+                let job_id = miner.job_id.unwrap();
+                let version = miner.version;
+                share_send
+                    .try_send((nonce, job_id, version.unwrap(), time))
+                    .unwrap();
+            }
+            miner.header.as_mut().map(|h| h.nonce += 1);
+        }
+    }
+}
```
plebhash, line 759: we can remove some redundancies in the logic of this function, there's no need for two separate loops:
```suggestion
fn mine(mut miner: Miner, share_send: Sender<(u32, u32, u32, u32)>, kill: Arc<AtomicBool>) {
    loop {
        if kill.load(Ordering::Relaxed) {
            break;
        }
        if miner.handicap > 0 {
            std::thread::sleep(std::time::Duration::from_micros(miner.handicap.into()));
        }
        if miner.next_share().is_valid() {
            let nonce = miner.header.unwrap().nonce;
            let time = miner.header.unwrap().time;
            let job_id = miner.job_id.unwrap();
            let version = miner.version;
            share_send
                .try_send((nonce, job_id, version.unwrap(), time))
                .unwrap();
        }
        miner.header.as_mut().map(|h| h.nonce += 1);
    }
}
```

roles/test-utils/mining-device/src/main.rs
```diff
@@ -669,3 +686,74 @@ fn generate_random_32_byte_array() -> [u8; 32] {
     rng.fill(&mut arr[..]);
     arr
 }
+
+fn start_mining_threads(
+    have_new_job: Receiver<()>,
```
plebhash, line 691: this receiver is not being triggered only on the arrival of new jobs

every time `Device` receives any message from upstream (e.g.: `SubmitSharesSuccess`), that is also

roles/test-utils/mining-device/src/main.rs
Fi3: as already stated above this is very hot code and I prefer to keep it like that.

plebhash (approved):

Sjors:
🎉

PR #985, opened 2024-06-20
Add semver check workflow yaml

### LINKED ISSUE #974: Add semver checks to CI
Given that the number of PR and contributor seems to growth having something like https://crates.io/crates/cargo-semver-checks could be very handy

Shourya742:
resolves #974

### DIFF
--- a/.github/workflows/semver-check.yaml
+++ b/.github/workflows/semver-check.yaml
@@ -1,0 +1,131 @@
+name: Semver Check
+
+on:
+  push:
+    branches:
+      - "main"
+      - "dev"
+  pull_request:
+    branches:
+      - "main"
+      - "dev"
+
+jobs:
+  semver-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install Rust 1.75.0
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.75.0
+          override: true
+
+      - name: Cache Cargo registry
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-registry-
+
+      - name: Cache Cargo index
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/git
+          key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-index-
+
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y cmake
+
+      - name: Install cargo-semver-checks
+        run: cargo install cargo-semver-checks --locked
+
+      - name: Run semver checks for common
+        working-directory: common
+        run: cargo semver-checks
+
+      - name: Run semver checks for utils/buffer
+        working-directory: utils/buffer
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/binary-sv2/no-serde-sv2/codec
+        working-directory: protocols/v2/binary-sv2/no-serde-sv2/codec
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/binary-sv2/serde-sv2
+        working-directory: protocols/v2/binary-sv2/serde-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/binary-sv2/binary-sv2
+        working-directory: protocols/v2/binary-sv2/binary-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/const-sv2
+        working-directory: protocols/v2/const-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/framing-sv2
+        working-directory: protocols/v2/framing-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/noise-sv2
+        working-directory: protocols/v2/noise-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/codec-sv2
+        working-directory: protocols/v2/codec-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/subprotocols/common-messages
+        working-directory: protocols/v2/subprotocols/common-messages
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/subprotocols/job-declaration
+        working-directory: protocols/v2/subprotocols/job-declaration
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/subprotocols/mining
+        working-directory: protocols/v2/subprotocols/mining
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/subprotocols/template-distribution
+        working-directory: protocols/v2/subprotocols/template-distribution
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/sv2-ffi
+        working-directory: protocols/v2/sv2-ffi
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v2/roles-logic-sv2
+        working-directory: protocols/v2/roles-logic-sv2
+        run: cargo semver-checks
+
+      - name: Run semver checks for protocols/v1
+        working-directory: protocols/v1
+        run: cargo semver-checks
+
+      - name: Run semver checks for utils/bip32-key-derivation
+        working-directory: utils/bip32-key-derivation
+        run: cargo semver-checks
+
+      - name: Run semver checks for utils/error-handling
+        working-directory: utils/error-handling
+        run: cargo semver-checks
+
+      - name: Run semver checks for utils/key-utils
+        working-directory: utils/key-utils
+        run: cargo semver-checks
+
+      - name: Run semver checks for roles/roles-utils/network-helpers
+        working-directory: roles/roles-utils/network-helpers
+        run: cargo semver-checks
+
+      - name: Run semver checks for roles/roles-utils/rpc
+        working-directory: roles/roles-utils/rpc
+        run: cargo semver-checks


jbesraa (commented):
How would this work in practice? Lets say I added a new breaking change, will the PR CI keep failing until I change the version accordingly?

Shourya742:
> How would this work in practice? Lets say I added a new breaking change, will the PR CI keep failing until I change the version accordingly?

The semver check in our CI pipeline ensures that all modifications to the codebase comply with semantic versioning rules. If a breaking change is introduced without the corresponding version increment, the CI will fail.

jbesraa:
IIUC, we should run this only on versioned branches. I think our release branch is `main`, so maybe we should run this only when a PR base is `main` ? Not sure though,  Ill let @Fi3 chime in as he opened the issue

plebhash:
here's a summary of our current strategy for semver quality control (which is admitedly suboptimal, so this contribution is very desired)

we have [`check-versioning-lib-release.sh`](https://github.com/stratum-mining/stratum/blob/main/check-versioning-lib-release.sh) which is called from [`.github/workflows/release-libs.yaml`](https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/.github/workflows/release-libs.yaml#L25) every time we are trying to merge from `dev` to `main`

if the script returns `1`, CI breaks and merging from `dev` to `main` is blocked

the script will return `1` if:
- some crate was changed since the latest release
- its version wasn't bumped

basically this is a safety check for releases, but implemented in a very suboptimal way

one of the undesired consequences is that even small cosmetic changes (e.g.: [PR #915](https://github.com/stratum-mining/stratum/pull/915)) impose that crates versioning must be bumped, which is not ideal

in summary, whenever we're merging from `dev` to `main`, our goal is:
- breaking changes to any of the crates published to `crates.io` (list can be found at `.github/workflows/release-libs.yaml`) must follow SemVer 2.0
- non-breaking changes (e.g.: `README.md`, code formatting, docs, etc) are not taken into consideration

I believe we will be able to achieve this vision by replacing `check-versioning-lib-release.sh` by leveraging `cargo semver-checks` over this new CI `.github/workflows/semver-check.yaml`

but this is just a comment about the conceptual approach, I still haven't reviewed the contents of the PR in details

jbesraa:
Thanks @plebhash thats very helpful.  @Shourya742 So I think you are half way through, lets just replace the old check with the new one as @plebhash mentioned.

jbesraa:
Note here: `cargo-semver` relies on the result of `rustdoc` in order to notify about semver breaches, i.e. it reads the docs we write and emit notifications accordingly. As our docs are not great yet, `cargo-semver` might not produce the results we want. Nonetheless, I think we should add it, use it as an additional check, and once we have robust docs we can rely on it solely.  cc @Fi3

plebhash:
> Note here: `cargo-semver` relies on the result of `rustdoc` in order to notify about semver breaches, i.e. it reads the docs we write and emit notifications accordingly. As our docs are not great yet, `cargo-semver` might not produce the results we want. Nonetheless, I think we should add it, use it as an additional check, and once we have robust docs we can rely on it solely. cc @Fi3

in other words: this is currently blocked by https://github.com/stratum-mining/stratum/issues/845

Shourya742:
> > Note here: `cargo-semver` relies on the result of `rustdoc` in order to notify about semver breaches, i.e. it reads the docs we write and emit notifications accordingly. As our docs are not great yet, `cargo-semver` might not produce the results we want. Nonetheless, I think we should add it, use it as an additional check, and once we have robust docs we can rely on it solely. cc @Fi3
> 
> in other words: this is currently blocked by #845


`cargo-semver` is designed to ensure semantic versioning (semver) compatibility by analyzing the public API of a Rust crate.  . Importantly, cargo-semver does not rely on rustdoc or the quality of the documentation to detect these breaches. In our case, because the j`d-server` and `jd-client` crates in roles are not published on crates.io, we utilize the cargo-semver-checks --baseline-root . command. This command allows us to specify the local directory as the baseline for semver checks, enabling us to manage version compatibility internally for our unpublished crates.

rrybarczyk (changes requested):
Some suggestions to help improve efficiency and try and avoid issues when non-code logic is changed (like `README.md`s).

rrybarczyk:
Do we want to add this workflow to the `.githooks/pre-push` script?

jbesraa:
> > > Note here: `cargo-semver` relies on the result of `rustdoc` in order to notify about semver breaches, i.e. it reads the docs we write and emit notifications accordingly. As our docs are not great yet, `cargo-semver` might not produce the results we want. Nonetheless, I think we should add it, use it as an additional check, and once we have robust docs we can rely on it solely. cc @Fi3
> > 
> > 
> > in other words: this is currently blocked by #845
> 
> `cargo-semver` is designed to ensure semantic versioning (semver) compatibility by analyzing the public API of a Rust crate. . Importantly, cargo-semver does not rely on rustdoc or the quality of the documentation to detect these breaches. In our case, because the j`d-server` and `jd-client` crates in roles are not published on crates.io, we utilize the cargo-semver-checks --baseline-root . command. This command allows us to specify the local directory as the baseline for semver checks, enabling us to manage version compatibility internally for our unpublished crates.

https://github.com/obi1kenobi/cargo-semver-checks/blob/main/README.md#how-is-cargo-semver-checks-similar-to-and-different-from-other-tools (third paragraph)

`--baseline-root` based on the docs this flag is used for getting the older version number

Fi3:
> based on the docs

not sure that I get the issue here. It say that the default option is to use the version of what is published on crates.io

Fi3:
> > > > Note here: `cargo-semver` relies on the result of `rustdoc` in order to notify about semver breaches, i.e. it reads the docs we write and emit notifications accordingly. As our docs are not great yet, `cargo-semver` might not produce the results we want. Nonetheless, I think we should add it, use it as an additional check, and once we have robust docs we can rely on it solely. cc @Fi3
> > > 
> > > 
> > > in other words: this is currently blocked by #845
> > 
> > 
> > `cargo-semver` is designed to ensure semantic versioning (semver) compatibility by analyzing the public API of a Rust crate. . Importantly, cargo-semver does not rely on rustdoc or the quality of the documentation to detect these breaches. In our case, because the j`d-server` and `jd-client` crates in roles are not published on crates.io, we utilize the cargo-semver-checks --baseline-root . command. This command allows us to specify the local directory as the baseline for semver checks, enabling us to manage version compatibility internally for our unpublished crates.
> 
> https://github.com/obi1kenobi/cargo-semver-checks/blob/main/README.md#how-is-cargo-semver-checks-similar-to-and-different-from-other-tools (third paragraph)
> 
> `--baseline-root` based on the docs this flag is used for getting the older version number

not sure that I get the issue here. It say that the default option is to use the version of what is published on crates.io

rrybarczyk (approved):
lgtm

jbesraa:
> > > > > Note here: `cargo-semver` relies on the result of `rustdoc` in order to notify about semver breaches, i.e. it reads the docs we write and emit notifications accordingly. As our docs are not great yet, `cargo-semver` might not produce the results we want. Nonetheless, I think we should add it, use it as an additional check, and once we have robust docs we can rely on it solely. cc @Fi3
> > > > 
> > > > 
> > > > in other words: this is currently blocked by #845
> > > 
> > > 
> > > `cargo-semver` is designed to ensure semantic versioning (semver) compatibility by analyzing the public API of a Rust crate. . Importantly, cargo-semver does not rely on rustdoc or the quality of the documentation to detect these breaches. In our case, because the j`d-server` and `jd-client` crates in roles are not published on crates.io, we utilize the cargo-semver-checks --baseline-root . command. This command allows us to specify the local directory as the baseline for semver checks, enabling us to manage version compatibility internally for our unpublished crates.
> > 
> > 
> > https://github.com/obi1kenobi/cargo-semver-checks/blob/main/README.md#how-is-cargo-semver-checks-similar-to-and-different-from-other-tools (third paragraph)
> > `--baseline-root` based on the docs this flag is used for getting the older version number
> 
> not sure that I get the issue here. It say that the default option is to use the version of what is published on crates.io

The question is how it knows there is a major breaking change between lets say 1.0 and 1.1 and notify about it. its doing so by inspecting the docs.

Fi3:
> > @jbesraa Yes, the semver check analyzes the documentation to understand the previous API behavior and detect any breaking changes in the current version. If you introduce a breaking change, the PR CI will fail until you update the version number accordingly, usually by incrementing the major version.
> 
> Yup. which makes this not great for this project(at this point) because we dont have docs and the API is not 'stablised' yet

I don't know the internals of it. But rust doc are automatically generated, and you get an entrance for everything that is exported by the library. I'm pretty sure that it do not need us to write actual docs in order to work. 

API is stabilized and if we change it we MUST updated the version accordingly.

Shourya742:
@jbesraa , that's precisely the point. We've already done a major release (version 1), and the APIs should not change. If there is any breaking change, a version bump is necessary. However, I believe @plebhash might be the right person to ask about this.

Fi3:
> @jbesraa , that's precisely the point. We've already done a major release (version 1), and the APIs should not change. If there is any breaking change, a version bump is necessary. However, I believe @plebhash might be the right person to ask about this.

any patch / bug that do not introduce new feature or remove old one MUST be a patch version bump
any new feature MUST be a minor version bump
anything that break compatibility MUST be a majior version bump

This is very important, there are other project out there that use SRI, and since cargo when building a project follow semver, if we do not do it will have build errors!

jbesraa:
Sorry maybe I am not explaining myself here

What iam trying to say is that this tool relies on rust-docs(we should write the docs) and while we dont have enough/good docs we might get alot of false alarms is the code is ahead of the code docs..

Fi3:
> Sorry maybe I am not explaining myself here
> 
> What iam trying to say is that this tool relies on rust-docs(we should write the docs) and while we dont have enough/good docs we might get alot of false alarms because the code is ahead of the code docs..

docs are automatically generated you don't need to write them

.github/workflows/semver-check.yaml
```diff
@@ -0,0 +1,121 @@
...
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y cmake
+
+      - name: Install cargo-semver-checks
+        run: cargo install cargo-semver-checks --locked
+
+      - name: Identify changed crates
+        id: changes
+        run: |
+          echo "::set-output name=changed_crates::$(git diff --name-only HEAD^ HEAD | grep '^crates/' | cut -d'/' -f2 | sort | uniq)"
```
plebhash, line 107: `git diff --name-only HEAD^ HEAD:` 

this lists all the files that have changed between the most recent commit (`HEAD`) and its immediate predecessor (`HEAD^`).

If this workflow is only going to be executed when we merge from `dev` to `main`, I think this is not a reasonable assumption.

There could have been multiple commits that introduced changes before `HEAD^`

.github/workflows/semver-check.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      - name: Run semver checks
+        if: steps.changes.outputs.changed_crates
+        run: |
+          for crate in ${{ steps.changes.outputs.changed_crates }}; do
```
plebhash, line 112: see https://github.com/stratum-mining/stratum/pull/985/files#r1672500430

I feel pre-filtering crates into `steps.changes` is overkill and introduces unnecessary complexity into the workflow

plebhash:
I feel I should point out that I'm not 100% confident on what I said here: https://github.com/stratum-mining/stratum/pull/985#issuecomment-2183480020 so I hid the comment to avoid making things more confusing than necessary

I'd like to ask to everyone (especially @jbesraa ) to take this comment with a grain of salt, it might be the case that we're not necessarily blocked by #845

it seems there's a lot of confusion amongst all of us with regards to how `cargo-semver-checks` works and how it will fill our needs.

I'm taking some time to study the following resources, which hopefully will allow for more well-informed decisions:
- https://www.youtube.com/watch?v=VArNQtYBC6Y
- https://docs.google.com/presentation/d/1gVYM9-YLrBmRYSuHqW8roE5Rt2FLAS0cijjgFU7ztxY/edit?usp=sharing

plebhash:
I just finished watching the youtube video listed on my previous comment

the entire talk is quite interesting, but the most relevant section starts [at minute 21:06 and is titled "cargo semver-checks lints are database queries in disguise"](https://youtu.be/VArNQtYBC6Y?si=I5Xsb9FaBfoAxycY&t=1266)

I believe this is where @jbesraa 's misconceptions will be clarified. In a nutshell, it doesn't matter if the code has comments that will eventually populate the Rust Docs. The relevant point here is that under the hood, [this specific tool called `rustdoc`](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html) is leveraged to create database entries that are used to check for breaking changes.

---

Here's a way to look at it from a different perspective: 

Start a new crate, but never write any comments on the code.

If you run `cargo doc --open`, your browser will still render the Rust Docs for this crate and it will contain links for all meaningful objects (e.g.: structs, enums, traits, functions, etc). There will be no explanation for those objects, since there's no comments on the code.

Now, if you introduce a breaking change to this crate and generate docs again, your browser is going to show the new APIs (still without any explanation). The point is that under the hood, `rustdoc` had to parse all the APIs and create an internal representation of them, and that is precisely what `cargo-semver-checks` is leveraging.

---

So as the final takeaway is: #845 is not a blocker for this PR. We can safely introduce `cargo-semver-checks` into our CI and keep #845 as a WIP without any concerns.

plebhash (changes requested):
I left some comments on the code diff. I feel we are trying to optimize for execution time with some undesired penalties on readability and maintainability.

.github/workflows/semver-check.yaml
Shourya742: Running semver checks on each crate takes a lot of time. Having a git-based filtering mechanism is important because we don't want to check every crate each time we push, especially when no changes have been made to a crate in a commit. I will update the logic for identifying changed crates: instead of checking only the immediate previous commit, I will check the diff against the branch the PR is being merged into. This way, all commits in the PR will be considered, and we can accurately determine the modified crates.

I am not sure if cargo-semver-checks relies on the verification of git history under the hood.

Shourya742: @plebhash

.github/workflows/semver-check.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+      - name: Identify target branch
+        id: target_branch
+        run: echo "::set-output name=branch::$(git rev-parse --abbrev-ref --symbolic-full-name @{u} | cut -d'/' -f2)"
+
+      - name: Identify changed crates
+        id: changes
+        run: |
+          TARGET_BRANCH=$(echo "${{ steps.target_branch.outputs.branch }}")
+          echo "::set-output name=changed_crates::$(git diff --name-only origin/$TARGET_BRANCH...HEAD | grep '^crates/' | cut -d'/' -f2 | sort | uniq)"
```
plebhash, line 117: I'm doing some tests on my fork: https://github.com/plebhash/stratum/actions/runs/9962822250

it seems the entire matrix is running, regardless of this

image

plebhash: also:
image

plebhash: and with this, I come back to my original point:

we should avoid footguns by trying to overengineer and overoptimize things prematurely

by attempting to make things run faster, we ended up with something that was actually behaving in a way that completely diverged from the intented goal

I'm simplifying this workflow on my fork, I'll share it here once it's ready to be cherry-picked into this PR

plebhash: @Shourya742 here's a simplified approach https://github.com/plebhash/stratum/commit/11717da74006ad518a900e6bf86f044e24f1ffa7

here's a test run: https://github.com/plebhash/stratum/actions/runs/9963466632/job/27529566883

execution wasn't lightning fast and it didn't finish in 5s, but it gets the job done in a straightforward way

it behaves perfectly as expected: `framing_sv2` broke the CI because we recently updated its APIs and we still haven't bumped its version

my suggestion for this PR is to cherry-pick this commit... if we eventually realize this CI is prohibitively slow, we can investigate strategies for optimization

Shourya742: Okay, I will let go of the filtering.

Shourya742:
@plebhash  Done. I cherry-picked the commit. Thanks for the detailed review and testing.

plebhash:
@Shourya742 there's also another issue that we need to look into on this PR

currently you cannot see it on the CI, because `framing_sv2` version still hasn't been bumped, so CI breaks before we get to `roles_logic_sv2`

but while doing some tests on my fork (where I bumped `framing_sv2` to unblock CI), I noticed that `cargo semver-checks` breaks in `roles_logic_sv2` for some very strange reasons

luckily, we don't need to reproduce it in CI, since it can also be reproduced manually:
```
$ cd develop/stratum/protocols/v2 
$ cargo semver-checks
     Parsing roles_logic_sv2 v1.1.0 (current)
error: running cargo-doc on crate roles_logic_sv2 failed with output:
    Checking framing_sv2 v1.1.0 (/Users/plebhash/develop/stratum/protocols/v2/framing-sv2)
    Checking job_declaration_sv2 v1.0.0 (/Users/plebhash/develop/stratum/protocols/v2/subprotocols/job-declaration)
    Checking template_distribution_sv2 v1.0.0 (/Users/plebhash/develop/stratum/protocols/v2/subprotocols/template-distribution)
error[E0463]: can't find crate for `serde`
  --> /Users/plebhash/develop/stratum/protocols/v2/framing-sv2/src/header.rs:18:17
   |
18 | #[derive(Debug, Serialize, Deserialize, Copy, Clone)]
   |                 ^^^^^^^^^ can't find crate
   |
   = note: this error originates in the derive macro `Serialize` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0463]: can't find crate for `serde`
  --> /Users/plebhash/develop/stratum/protocols/v2/framing-sv2/src/header.rs:18:28
   |
18 | #[derive(Debug, Serialize, Deserialize, Copy, Clone)]
   |                            ^^^^^^^^^^^ can't find crate
   |
   = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0463`.
error: could not compile `framing_sv2` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error[E0433]: failed to resolve: use of undeclared type `Vec`
   --> /Users/plebhash/develop/stratum/protocols/v2/subprotocols/template-distribution/src/new_template.rs:199:28
    |
199 |             let mut path = Vec::new();
    |                            ^^^ use of undeclared type `Vec`
    |
help: consider importing one of these items
    |
7   + use alloc::vec::Vec;
    |
7   + use crate::vec::Vec;
    |

error[E0433]: failed to resolve: use of undeclared type `Vec`
   --> /Users/plebhash/develop/stratum/protocols/v2/subprotocols/template-distribution/src/new_template.rs:220:34
    |
220 |             coinbase_tx_outputs: Vec::new().try_into().unwrap(),
    |                                  ^^^ use of undeclared type `Vec`
    |
help: consider importing one of these items
    |
7   + use alloc::vec::Vec;
    |
7   + use crate::vec::Vec;
    |

error[E0599]: no method named `try_into` found for struct `Vec<u8>` in the current scope
   --> /Users/plebhash/develop/stratum/protocols/v2/subprotocols/template-distribution/src/new_template.rs:185:66
    |
185 |         let coinbase_prefix: binary_sv2::B0255 = coinbase_prefix.try_into().unwrap();
    |                                                                  ^^^^^^^^ method not found in `Vec<u8>`
    |
   ::: /Users/plebhash/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/convert/mod.rs:610:8
    |
610 |     fn try_into(self) -> Result<T, Self::Error>;
    |        -------- the method is available for `Vec<u8>` here
    |
    = help: items from traits can only be used if the trait is in scope
    = note: 'core::convert::TryInto' is included in the prelude starting in Edition 2021
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
1   + use core::convert::TryInto;
    |

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `template_distribution_sv2` (lib) due to 3 previous errors

-----

error: failed to build rustdoc for crate roles_logic_sv2 v1.1.0
note: this is usually due to a compilation error in the crate,
      and is unlikely to be a bug in cargo-semver-checks
note: the following command can be used to reproduce the compilation error:
      cargo new --lib example &&
          cd example &&
          echo '[workspace]' >> Cargo.toml &&
          cargo add --path /Users/plebhash/develop/stratum/protocols/v2/roles-logic-sv2 --no-default-features --features disable_nopanic,prop_test,serde,with_serde &&
          cargo check

error: aborting due to failure to build rustdoc for crate roles_logic_sv2 v1.1.0
```

this seems somewhat related to `serde`, but using feature flags via `cargo semver-checks --features with_serde` produces the exact same result.

plebhash (approved):

PR #987, opened 2024-06-20
Handle `jd-client` config error

### LINKED ISSUE #863: JDC is not logging any error on wrong TP configuration
JDC is not generating errors when a wrong TP configuration is used.
To reproduce it, you can simply change the port number randomly in `tp_address` parameter, and run it.
A discussion about this here: https://discord.com/channels/950687892169195530/1012371967992672321/1229368703737663508

jbesraa:
resolves #863

### DIFF
--- a/roles/jd-client/src/main.rs
+++ b/roles/jd-client/src/main.rs
@@ -107,7 +107,10 @@ async fn main() {
 
     let proxy_config = match process_cli_args() {
         Ok(p) => p,
-        Err(_) => return,
+        Err(e) => {
+            error!("Failed to read config file: {}", e);
+            return;
+        }
     };
 
     loop {
@@ -205,7 +208,13 @@ async fn initialize_jd_as_solo_miner(
     task_collector: Arc<Mutex<Vec<AbortHandle>>>,
     timeout: Duration,
 ) {
-    let proxy_config = process_cli_args().unwrap();
+    let proxy_config = match process_cli_args() {
+        Ok(p) => p,
+        Err(e) => {
+            error!("Failed to read config file: {}", e);
+            return;
+        }
+    };
     let miner_tx_out = lib::proxy_config::get_coinbase_output(&proxy_config).unwrap();
 
     // When Downstream receive a share that meets bitcoin target it transformit in a


plebhash:
thanks @jbesraa 

we have a WIP refactor PR that touches this part of the code https://github.com/stratum-mining/stratum/pull/937

@rrybarczyk is working on some adjustments via https://github.com/plebhash/stratum/pull/61

I suggest we revert #987 to draft and wait until #937 is merged

plebhash (approved):

roles/jd-client/src/lib/job_declarator/mod.rs
```diff
@@ -87,23 +87,23 @@ impl JobDeclarator {
...
-        SetupConnectionHandler::setup(&mut receiver, &mut sender, proxy_address)
-            .await
-            .unwrap();
+        match SetupConnectionHandler::setup(&mut receiver, &mut sender, proxy_address).await {
+            Ok(_) => {}
+            Err(e) => {
+                error!("Error setting up connection: {:?}", e);
+                return Err(Error::ConnectionSetup);
+            }
+        };
```
Shourya742, line 106: Since the `Ok` case doesn't do anything, using `if let` is cleaner and makes the code easier to read.
```suggestion
   if let Err(e) = SetupConnectionHandler::setup(&mut receiver, &mut sender, proxy_address).await {
    error!("Error setting up connection: {:?}", e);
    return Err(Error::ConnectionSetup);
}
```

Shourya742 (commented):
Concept ACK

roles/jd-client/src/lib/job_declarator/mod.rs
jbesraa: Ah, good point. I left the `match` as is and moved the `success` log into the `ok` arm. thank you

Shourya742 (approved):
ACK bcbf6a02117d996e4b52a32f1e52544ded962cf6

rrybarczyk:
> thanks @jbesraa
> 
> we have a WIP refactor PR that touches this part of the code #937
> 
> @rrybarczyk is working on some adjustments via [plebhash#61](https://github.com/plebhash/stratum/pull/61)
> 
> I suggest we revert #987 to draft and wait until #937 is merged

I think all config/args in the roles should be refactored as there is a lot of duplicate logic/logic that is in the wrong place. I did not end up including this refactor in the review for @plebhash's #813 PR because it started to feel out of scope. I think an immediate fix to make sure the JD config is properly configured is fine if we are time-pressed on this, but if it is not pressing, let's address all of this logic in a single issue/PR. It will fall into the "roles refactor" category.

jbesraa:
@rrybarczyk I dont think its time pressed, it is part of the `1.0.2` milestone tho.
The issue(s) you mentioned includes a good a mount of refactoring and would take I think more time to be merged while this can be a small quick win(to the users)

jbesraa:
Deleted unrelated commit

PR #990, opened 2024-06-21
config examples testnet4

### LINKED ISSUE #989: update config examples for `testnet4`
the SRI community VPS is hosting a TP on `testnet4`

that implies changing some things on the config examples:
- RPC port is now `48332`
- TP authority key is now `9cRiuTwzFCEUDATcZ5pidbQCvdQucBeWni8oCG9N4nY3D7p7iQW` (although ideally this should be customizable, as described in https://github.com/Sjors/bitcoin/issues/46)

plebhash:
close #989

### DIFF
--- a/roles/jd-server/config-examples/jds-config-hosted-example.toml
+++ b/roles/jd-server/config-examples/jds-config-hosted-example.toml
@@ -18,8 +18,8 @@ coinbase_outputs = [
 # SRI Pool JD config
 listen_jd_address = "0.0.0.0:34264"
 # RPC config for mempool (it can be also the same TP if correctly configured)
-core_rpc_url =  "http://127.0.0.1"
-core_rpc_port = 18332
+core_rpc_url =  "http://75.119.150.111"
+core_rpc_port = 48332
 core_rpc_user =  "username"
 core_rpc_pass =  "password"
 # Time interval used for JDS mempool update 
--- a/roles/jd-server/config-examples/jds-config-local-example.toml
+++ b/roles/jd-server/config-examples/jds-config-local-example.toml
@@ -19,7 +19,7 @@ coinbase_outputs = [
 listen_jd_address = "127.0.0.1:34264"
 # RPC config for mempool (it can be also the same TP if correctly configured)
 core_rpc_url =  "http://127.0.0.1"
-core_rpc_port = 18332
+core_rpc_port = 48332
 core_rpc_user =  "username"
 core_rpc_pass =  "password"
 # Time interval used for JDS mempool update 


roles/jd-server/config-examples/jds-config-hosted-example.toml
```diff
@@ -18,8 +18,8 @@ coinbase_outputs = [
 # SRI Pool JD config
 listen_jd_address = "0.0.0.0:34264"
 # RPC config for mempool (it can be also the same TP if correctly configured)
-core_rpc_url =  "http://127.0.0.1"
```
plebhash, line 21: @GitGab19 why was this `127.0.0.1`?

roles/jd-server/config-examples/jds-config-hosted-example.toml
```diff
@@ -18,8 +18,8 @@ coinbase_outputs = [
...
-core_rpc_port = 18332
+core_rpc_url =  "http://75.119.150.111"
```
plebhash, line 21: this is not working as expected:
```
cargo run -- -c config-examples/jds-config-hosted-example.toml
    Finished dev [optimized + debuginfo] target(s) in 0.09s
     Running `/home/plebhash/develop/stratum/roles/target/debug/jd_server -c config-examples/jds-config-hosted-example.toml`
2024-06-21T12:20:52.931525Z  INFO jd_server: Jds INITIALIZING with config: "config-examples/jds-config-hosted-example.toml"
2024-06-21T12:20:52.931639Z  INFO jd_server::lib::job_declarator: JD INITIALIZED
2024-06-21T12:20:53.173286Z ERROR jd_server::lib::mempool::error: Rpc(Http("client error (Connect)"))
2024-06-21T12:20:53.173304Z ERROR jd_server::lib::mempool::error: Unable to establish RPC connection with Template Provider (possible reasons: not fully synced, down)
2024-06-21T12:20:53.173378Z ERROR jd_server: SHUTDOWN from Upstream: Mempool error: `Rpc(Http("client error (Connect)"))`
Try to reconnecting or connecting to a new upstream
```

roles/jd-server/config-examples/jds-config-hosted-example.toml
GitGab19: Because `jds-config-hosted-example.toml` has been always used just by us on our VPS. We don't expose our RPCs to externals. So this file was only meant to be used by us.
But we can change this, and start serving RPCs to internet, so that everybody can spin up a Pool, a JDS, and everything using our hosted TP. What do you think?

roles/jd-server/config-examples/jds-config-hosted-example.toml
GitGab19: That's because we don't expose RPC's to internet. We need to change our bitcoin.conf if we want to achieve [this](https://github.com/stratum-mining/stratum/pull/990#discussion_r1648905833)

plebhash: can you give that a try?

I looked into `.bitcoin/bitcoin.conf` file on the VPS but I see there's many configs I'm not familiar with, so maybe it would be better if someone else with more familiarity did that

GitGab19:
@plebhash I just set our hosted TP authority key as explained [here](https://github.com/Sjors/bitcoin/issues/46#issuecomment-2182734139) by @Sjors. So there's no need to change it in this PR.
I'm going to try exposing our RPCs to internet as well

roles/jd-server/config-examples/jds-config-hosted-example.toml
GitGab19: I managed to expose RPCs to internet, but the problem is that JDS is not performing encryption in the client_rpc which is used to manage rpc calls as far as I can tell. And that explains also why there's no authority_pub_key for the node which JDS connects to. Could you confirm @lorbax?

In this case, I would suggest to remove the `jds-config-hosted-example.toml` in order to avoid confusion. We will just modify that single line on our own for the VPS.

plebhash: but as far as I understand, the SV2 encryption spec doesn't cover RPC, I don't think we should introduce that

GitGab19: I was forgetting about binding the rpc port, sorry.
Now I just tried again and it's working properly. So if you use `jds-config-hosted-example.toml` you can now run a local JDS which uses our hosted TP for RPCs.

GitGab19 (approved):

PR #993, opened 2024-06-21
Add `JobDeclaratorClient` struct

### LINKED ISSUE #992: Move JDC initialization from `main` to `lib`
I am trying to write some tests running the whole flow, and noticed that JDC initialization is happening inside `main.rs` rather than `lib.rs` which makes it not very modular. I propose to move the initialization to `lib` so it could be exported and used in different context, like testing

jbesraa:
resolves #992 

With the goal to write more tests, this commit moves JDC initialisation logic from `main.rs` to `lib.rs` so its easier to test and also separate the actual `lib` code from the binary.


plebhash (approved):

Shourya742 (approved):
ACK 1bdf1326befe4a8bc53038ff3d182bd0156eff61

jbesraa:
@Shourya742 @plebhash 
Fixed up documentation(taken from https://stratumprotocol.org/docs/) for `JobDeclaratorClient`

```shell
diff --git a/roles/jd-client/src/lib/mod.rs b/roles/jd-client/src/lib/mod.rs
index 3ae1d1c1..1979eecb 100644
--- a/roles/jd-client/src/lib/mod.rs
+++ b/roles/jd-client/src/lib/mod.rs
@@ -47,7 +47,15 @@ use tracing::{error, info};
 ///    between all the contexts is not necessary.
 pub static IS_NEW_TEMPLATE_HANDLED: AtomicBool = AtomicBool::new(true);

+/// Job Declarator Client (or JDC) is the role which is Miner-side, in charge of creating new
+/// mining jobs from the templates received by the Template Provider to which it is connected. It
+/// declares custom jobs to the JDS, in order to start working on them.
+/// JDC is also responsible for putting in action the Pool-fallback mechanism, automatically
+/// switching to backup Pools in case of declared custom jobs refused by JDS (which is Pool side).
+/// As a solution of last-resort, it is able to switch to Solo Mining until new safe Pools appear
+/// in the market.
 pub struct JobDeclaratorClient {
+    /// Configuration of the proxy server [`JobDeclaratorClient`] is connected to.
     config: ProxyConfig,
 }
 ```

plebhash:
@jbesraa it looks like merging https://github.com/stratum-mining/stratum/pull/987 introduced some conflics here

jbesraa:
@plebhash resolved conflict

plebhash:
rebased for merging

jbesraa:
@Fi3 we need to make `dev` a default branch otherwise we are gonna keep having commits like the one in this PR. @plebhash need to rebase and co-commit even tho there are no conflicts.

Issue #988, opened 2024-06-20
Fix flaky test `downstream_sv1::diff_management::test::test_diff_management`


jbesraa:
Noisy test. Produces a lot of false alarms

https://github.com/stratum-mining/stratum/actions/runs/9601943467/job/26481865946?pr=968#step:7:422
```shell
 running 4 tests
test downstream_sv1::downstream::tests::gets_difficulty_from_target ... ok
test proxy::bridge::test::test_version_bits_insert ... ok
test downstream_sv1::diff_management::test::test_converge_to_spm_from_low ... ok
test downstream_sv1::diff_management::test::test_diff_management ... FAILED
failures:
---- downstream_sv1::diff_management::test::test_diff_management stdout ----
thread 'downstream_sv1::diff_management::test::test_diff_management' panicked at translator/src/lib/downstream_sv1/diff_management.rs:349:9:
Calculated shares per minute are outside the 99% confidence interval. Error: 4776.132, Error margin: 189.73665961010275, 5776.132
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failures:
    downstream_sv1::diff_management::test::test_diff_management
test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 16.02s
error: test failed, to rerun pass `-p translator_sv2 --lib`
```

plebhash:
closed via #955

jbesraa:
@plebhash I think we should reopen this issue as the test is not fixed. I would suggest to `#[ignore]` the test for now.

Fi3:
about this test I use a z_score of 2.576 total runtime of 40 seconds and measure hash rate of 10 seconds.
With this values on my machine it pass almost all the time, I saw it fail just one time. 
z_score I use for 99%, from what I saw more then 10 second do not improve the estimation, I always get the same error range also if I double or quadruple it on average. 40 seconds empirically are the shortest amount of time that almost never fail, ofc that on my machine dunno but maybe is relevant.

pavlenex:
Do you guys want me to re-open this issue?

plebhash:
https://github.com/stratum-mining/stratum/pull/1074 is disabling the test as a temporary solution

we should eventually try to come up with a more sustainable fix to this problem

GitGab19:
Reopening it, as agreed [here](https://github.com/stratum-mining/stratum/issues/988#issuecomment-2313021875)

GitGab19:
[...]

PR #995, opened 2024-06-23
Ignore `test_diff_management` test

jbesraa:
I think as long as #988 is not fixed, there is no point in leaving this test on.

Its currently producing too much false alarms which can block new pull requests as well as make it harder to find issues in new pull requests.

### DIFF
--- a/roles/translator/src/lib/downstream_sv1/diff_management.rs
+++ b/roles/translator/src/lib/downstream_sv1/diff_management.rs
@@ -320,7 +320,7 @@ mod test {
     #[test]
     fn test_diff_management() {
         let expected_shares_per_minute = 1000.0;
-        let total_run_time = std::time::Duration::from_secs(11);
+        let total_run_time = std::time::Duration::from_secs(30);
         let initial_nominal_hashrate = measure_hashrate(5);
         let target = match roles_logic_sv2::utils::hash_rate_to_target(
             initial_nominal_hashrate,
@@ -342,13 +342,13 @@ mod test {
         }
 
         let calculated_share_per_min = count as f32 / (elapsed.as_secs_f32() / 60.0);
-        // This is the error margin for a confidence of 99% given the expect number of shares per
+        // This is the error margin for a confidence of 99.99...% given the expect number of shares per
         // minute TODO the review the math under it
         let error_margin = get_error(expected_shares_per_minute);
         let error = (calculated_share_per_min - expected_shares_per_minute as f32).abs();
         assert!(
             error <= error_margin as f32,
-            "Calculated shares per minute are outside the 99% confidence interval. Error: {:?}, Error margin: {:?}, {:?}", error, error_margin,calculated_share_per_min
+            "Calculated shares per minute are outside the 99.99...% confidence interval. Error: {:?}, Error margin: {:?}, {:?}", error, error_margin,calculated_share_per_min
         );
     }
 


plebhash:
we should make sure we disable unit tests on `cargo llvm-cov` on the MG CI scripts:

https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/interop-jd-translator/interop-jd-translator.sh

plebhash:
@lorbax  I wonder if we really need `cargo llvm-cov` at all on on the MG CI shell scripts?

they were introduced via https://github.com/stratum-mining/stratum/pull/905 but I don't know about rationale behind them

would it be ok if we did something like this?
```diff
- cargo llvm-cov --no-report -p pool_sv2
- cargo llvm-cov --no-report -p jd_server
- cargo llvm-cov --no-report -p jd_client
- cargo llvm-cov --no-report -p translator_sv2
+ cargo build -p pool_sv2
+ cargo build --no-report -p jd_server
+ cargo build --no-report -p jd_client
+ cargo build --no-report -p translator_sv2
```

lorbax:
> @lorbax I wonder if we really need `cargo llvm-cov` at all on on the MG CI shell scripts?
> 
> they were introduced via #905 but I don't know about rationale behind them
> 
> would it be ok if we did something like this?
> 
> ```diff
> - cargo llvm-cov --no-report -p pool_sv2
> - cargo llvm-cov --no-report -p jd_server
> - cargo llvm-cov --no-report -p jd_client
> - cargo llvm-cov --no-report -p translator_sv2
> + cargo build -p pool_sv2
> + cargo build --no-report -p jd_server
> + cargo build --no-report -p jd_client
> + cargo build --no-report -p translator_sv2
> ```

Unfortunately yes I think. Many MG tests launch some applications and most of the cases this is done using llvm-cov. So, if you build with cargo alone, the application will use other binaries and recompile, making the script useless.
Unless we want to give up con code coverage (but I don't like it). Do you have any prior experience with code coverage?

plebhash:
another alternative is to modify `test_diff_management`:

https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/roles/translator/src/lib/downstream_sv1/diff_management.rs#L321-L358

it seems to impose a 99% confidence range over the `get_error` function.

@GitGab19 @Fi3 what is the rationale behind `z_lambda_99`? if we wanted 95% condifence, which value should we use?

GitGab19:
> another alternative is to modify `test_diff_management`:
> 
> https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/roles/translator/src/lib/downstream_sv1/diff_management.rs#L321-L358
> 
> it seems to impose a 99% confidence range over the `get_error` function.
> 
> @GitGab19 @Fi3 what is the rationale behind `z_lambda_99`? if we wanted 95% condifence, which value should we use?

I remember that @Fi3 introduced that confidence into this test. I would disable only this test from MG CI execution, since it pass on unit-tests CI. Is it possible to exclude only this?

plebhash:
> I would disable only this test from MG CI execution, since it pass on unit-tests CI. Is it possible to exclude it?

That's the point, from my research on `cargo llvm-cov`, I couldn't figure out a way to do this.

So my previous comment is exploring the possibility of reducing the confidence interval to 95%, which is still reasonable but would hopefully reduce the incidence of these broken MG CI

GitGab19:
> > I would disable only this test from MG CI execution, since it pass on unit-tests CI. Is it possible to exclude it?
> 
> That's the point, from my research on `cargo llvm-cov`, I couldn't figure out a way to do this.
> 
> So my previous comment is exploring the possibility of reducing the confidence interval to 95%, which is still reasonable but would hopefully reduce the incidence of these broken MG CI

According to chatgpt, this should move the confidence from 99% to 95%:
```
fn get_error(lambda: f64) -> f64 {
    let z_score_95 = 1.96;
    z_score_95 * lambda.sqrt()
}
```

plebhash:
ok that makes sense, apparently this is leveraging [Statistical Z-test](https://en.wikipedia.org/wiki/Z-test)... I also asked ChatGPT and here's the answer:

> In your code, z_score_99 is used in the get_error function to calculate the error margin based on the expected number of shares per minute, which appears to be based on a statistical confidence interval. Specifically, z_score_99 represents the z-score for a 99% confidence level.

> Understanding the Rationale Behind z_score_99

> Z-score in Statistics:

>	•	A z-score indicates how many standard deviations an element is from the mean. For confidence intervals, z-scores correspond to the desired confidence level.
>	•	For a 99% confidence interval, the z-score is approximately 2.576. This value can be found in z-score tables or by using statistical software.

it also pointed out that `6` is overkill, and results in confidence of `99.9999... %`

plebhash:
@jbesraa to give you a bit more context: this test is asserting the correctness of the translator proxy's difficulty management.

if we allowed share submission of **any** difficulty, then anyone with a high hashrate would essentially DoS the upstream connections with extremely high network traffic

therefore, it is desirable to establish a difficulty threshold under which shares are allowed to be submitted

this difficulty threshold is proportional to the hashrate, and is purely statistical

this PR aims to fix a test on the algorithm that establishes such threshold, and the Z-test is currently miscalibrated

basically, we are expecting 99.9999% statistical correctness, which will obviously produce instability

so the proper fix here is not to disable this unit test, but only to properly calibrate its internal statistical parameters

as @GitGab19 pointed out, we can simply set `z_score_95 = 1.96`

plebhash:
it looks like we need more research here... CI is still breaking:
```
failures:

---- downstream_sv1::diff_management::test::test_diff_management stdout ----
thread 'downstream_sv1::diff_management::test::test_diff_management' panicked at translator/src/lib/downstream_sv1/diff_management.rs:349:9:
Calculated shares per minute are outside the 95% confidence interval. Error: 96.74011, Error margin: 61.98064213930023, 1096.7401
```

jbesraa:
ugh not sure whats the action point should be here. why do we have the time constraints in the test? maybe we can improve those?

GitGab19:
We could try to increase `total_run_time` to 30s and see if it gets better. Since this test is based on producing shares, it's not deterministic at all. So if we make the time a little bit longer it should theoretically improve

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -320,7 +320,7 @@ mod test {
     #[test]
     fn test_diff_management() {
         let expected_shares_per_minute = 1000.0;
-        let total_run_time = std::time::Duration::from_secs(11);
+        let total_run_time = std::time::Duration::from_secs(30);
         let initial_nominal_hashrate = measure_hashrate(5);
```
plebhash, line 324: @GitGab19 mentioned on our call that we could try to increase `5` to improve things

plebhash:
I'm doing some experiments locally. CPU is a `i7-4770 CPU @ 3.40GHz`.

First, I added this shell script to `roles/translator`:
```sh
#!/bin/sh

MAX_RUNS=20
FAIL_COUNT=0

for ((i=1; i<=$MAX_RUNS; i++)); do
    echo "i = $i"
    cargo test lib::downstream_sv1::diff_management::test::test_diff_management 2>/dev/null
    if [ $? -ne 0 ]; then
        FAIL_COUNT=$((FAIL_COUNT + 1))
    fi
done

if [ $FAIL_COUNT -gt 0 ]; then
    echo "$FAIL_COUNT run(s) failed. Please check the test."
    # Add notification logic here (e.g., sending an email or Slack message)
else
    echo "All runs passed successfully."
fi
```

This allowed me to execute the test repeatedly and get some statistical sense of how different parameters affect the incidence of false alarms.

Then, I started tweaking parameters. As I changed one specific parameter `X`, I made sure all other parameters were set to the same original values we currently have on `dev`.

Here's a summary of results:
| parameter           | X              | test failures |
|---------------------|----------------|---------------|
| total_run_time = X  | 11s (original) | 1/20          |
| total_run_time = X  | 1s (smaller)   | 10/20         |
| total_run_time = X  | 30s (bigger)   | **0/20**          |
| measure_hashrate(X) | 5s (original)  | 1/20          |
| measure_hashrate(X) | 1s (smaller)   | 0/20          |
| measure_hashrate(X) | 30s (bigger)   | 1/20          |
| z_score_99 = X      | 6 (original)   | 1/20          |
| z_score_95 = X      | 1.96 (smaller) | 6/20          |
|  z_score_99 = X     | 10 (bigger)    | 1/20          |

Here's my conclusions on each parameter:
- `measure_hashrate(X)`: no significant differences by either increasing or decreasing it
- `z_score_99`: decreasing to 95% it made it significantly worse (which actually proves my previous suggestion was wrong), while increasing even further above 99.9999...% made no significant difference
- `total_run_time`: increasing it seems to be the final solution to our problem

---

TLDR: I think we can finally fix this issue by:
- keeping `z_score_99 = 6.0` (and maybe just adapt the comments so it more accurately says `99.9999...%` instead of `99%`)
- increasing `total_run_time` to 30s
- keeping `measure_hashrate(X)` with its current value

plebhash (approved):

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -342,13 +342,13 @@ mod test {
         }
 
         let calculated_share_per_min = count as f32 / (elapsed.as_secs_f32() / 60.0);
-        // This is the error margin for a confidence of 99% given the expect number of shares per
+        // This is the error margin for a confidence of 99.99% given the expect number of shares per
```
plebhash, line 345:
```suggestion
        // This is the error margin for a confidence of 99.99...% given the expect number of shares per
```

99.99% still gives some impression of exactness, which is not the case

we should use a (`...`) to imply recurring decimals and make it clear we are really overshooting

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -342,13 +342,13 @@ mod test {
...
         // minute TODO the review the math under it
         let error_margin = get_error(expected_shares_per_minute);
         let error = (calculated_share_per_min - expected_shares_per_minute as f32).abs();
         assert!(
             error <= error_margin as f32,
-            "Calculated shares per minute are outside the 99% confidence interval. Error: {:?}, Error margin: {:?}, {:?}", error, error_margin,calculated_share_per_min
+            "Calculated shares per minute are outside the 99.99% confidence interval. Error: {:?}, Error margin: {:?}, {:?}", error, error_margin,calculated_share_per_min
```
plebhash, line 351:
```suggestion
            "Calculated shares per minute are outside the 99.99...% confidence interval. Error: {:?}, Error margin: {:?}, {:?}", error, error_margin,calculated_share_per_min
```

jbesraa:
@plebhash done. thanks

Issue #844, opened 2024-04-12
JDC is not doing fallback to solo-mining


GitGab19:
During testing, I noticed that sometimes JDC disconnects from upstreams because of some shares refused by the pool (this needs to be investigated better).
In this case, our default config for JDC doesn't contain a second upstream, so JDC should fallback to solo-mining.
But it's not doing that, instead it just hangs there.

GitGab19:
For a better context, here's a screenshot I did when I noticed the bug during testing:
![Uploading Screenshot 2024-04-08 at 15.24.09.png…]()

GitGab19:
To reproduce the issue, just try to fake the prev_hash used by the pool to check shares sent by downstream, by changing line 1159 in [protocols/v2/roles-logic-sv2/src/channel-logic/channel_factory.rs](https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs#L1159) with this:

//let prev_blockhash = crate::utils::u256_to_block_hash(referenced_job.prev_hash.clone());
let prev_blockhash = crate::utils::u256_to_block_hash([220, 157, 233, 68, 198, 232, 73, 141, 133, 12, 97, 197, 97, 244, 132, 146, 127, 171, 248, 91, 23, 159, 215, 209, 2, 0, 0, 0, 0, 0, 0, 0].into());

plebhash:
> To reproduce the issue, just try to fake the prev_hash used by the pool to check shares sent by downstream, by changing line 1159 in [protocols/v2/roles-logic-sv2/src/channel-logic/channel_factory.rs](https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs#L1159) with this:
> 
> //let prev_blockhash = crate::utils::u256_to_block_hash(referenced_job.prev_hash.clone()); let prev_blockhash = crate::utils::u256_to_block_hash([220, 157, 233, 68,

it would be good to reproduce this with MG

plebhash:
partially solved by https://github.com/stratum-mining/stratum/issues/1004

PR #1001, opened 2024-06-24
Translator restart if disconnected from upstream

lorbax:
When the jdc changes upstream, it disconnects the TProxy.
This PR addresses this issue by making the TProxy reconnecting to the JDC.

Removed the part related to the MG test.

How to test this PR:
The idea is to have a pool that rejects a share. It is possible to do that by tweaking the part of the code needed to check the target, forcing the pool to calculate a wrong hash for a share. 
Following @GitGab19 [suggestion](https://github.com/stratum-mining/stratum/issues/844#issuecomment-2118864024), the part of the code responsible for that is  [`roles-logc-sv2::channel_logic::channel_factory::on_submit_shares_extended(...)`](https://github.com/stratum-mining/stratum/blob/dev/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs#L1146). Since other roles check the share using the same function, directly modifying this would have the consequence that no share at all is ever sent to the pool. If I understood correctly, the way to fix this issue is to apply @GitGab19 suggestion to a clone of function `on_submit_shares_extended`, which can be called call `on_submit_shares_extended_pool`. After that, modify the implementation of [`handle_submit_share_extended`](https://github.com/stratum-mining/stratum/blob/dev/roles/pool/src/lib/mining_pool/message_handler.rs#L151) trait function  of `ParseDownstreamMiningMessages` trait in the following way
```rust
fn handle_submit_shares_extended(
        &mut self,
        m: SubmitSharesExtended,
    ) -> Result<SendTo<()>, Error> {
        let res = self
            .channel_factory
            //.safe_lock(|cf| cf.on_submit_shares_extended(m.clone()))
            .safe_lock(|cf| cf.on_submit_shares_extended_pool(m.clone()))
            .map_err(|e| roles_logic_sv2::Error::PoisonLock(e.to_string()))?;
        ...
    }
```
 - run a "good" pool on port 34254, this is the pool that the jdc will fallback to.
 - run a "bad" pool with the modification above on port 44254, this pool will refuse the first share submitted to the upstream (recall to adjust the channel hashrate in the tproxy config)
 - run a jdc client with these upstreams in the config
```toml
[[upstreams]]
authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
pool_address = "127.0.0.1:44254"
jd_address = "127.0.0.1:34264"
pool_signature = "Stratum v2 SRI Pool"
[[upstreams]]
authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
pool_address = "127.0.0.1:34254"
jd_address = "127.0.0.1:34264"
pool_signature = "Stratum v2 SRI Pool"
```
 - run the downstreams (usually tproxy + sv1 miner)

### DIFF
--- a/roles/translator/src/lib/downstream_sv1/downstream.rs
+++ b/roles/translator/src/lib/downstream_sv1/downstream.rs
@@ -13,7 +13,7 @@ use async_std::{
 };
 use error_handling::handle_result;
 use futures::FutureExt;
-use tokio::sync::broadcast;
+use tokio::{sync::broadcast, task::AbortHandle};
 
 use super::{kill, DownstreamMessages, SubmitShareWithChannelId, SUBSCRIBE_TIMEOUT_SECS};
 
@@ -110,6 +110,7 @@ impl Downstream {
         host: String,
         difficulty_config: DownstreamDifficultyConfig,
         upstream_difficulty_config: Arc<Mutex<UpstreamDifficultyConfig>>,
+        task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
     ) {
         let stream = std::sync::Arc::new(stream);
 
@@ -150,11 +151,12 @@ impl Downstream {
         let rx_shutdown_clone = rx_shutdown.clone();
         let tx_shutdown_clone = tx_shutdown.clone();
         let tx_status_reader = tx_status.clone();
+        let task_collector_mining_device = task_collector.clone();
         // Task to read from SV1 Mining Device Client socket via `socket_reader`. Depending on the
         // SV1 message received, a message response is sent directly back to the SV1 Downstream
         // role, or the message is sent upwards to the Bridge for translation into a SV2 message
         // and then sent to the SV2 Upstream role.
-        let _socket_reader_task = task::spawn(async move {
+        let socket_reader_task = tokio::task::spawn(async move {
             let reader = BufReader::new(&*socket_reader);
             let mut messages = FramedRead::new(
                 async_compat::Compat::new(reader),
@@ -205,15 +207,22 @@ impl Downstream {
             kill(&tx_shutdown_clone).await;
             warn!("Downstream: Shutting down sv1 downstream reader");
         });
+        let _ = task_collector_mining_device.safe_lock(|a| {
+            a.push((
+                socket_reader_task.abort_handle(),
+                "socket_reader_task".to_string(),
+            ))
+        });
 
         let rx_shutdown_clone = rx_shutdown.clone();
         let tx_shutdown_clone = tx_shutdown.clone();
         let tx_status_writer = tx_status.clone();
         let host_ = host.clone();
 
+        let task_collector_new_sv1_message_no_transl = task_collector.clone();
         // Task to receive SV1 message responses to SV1 messages that do NOT need translation.
         // These response messages are sent directly to the SV1 Downstream role.
-        let _socket_writer_task = task::spawn(async move {
+        let socket_writer_task = tokio::task::spawn(async move {
             loop {
                 select! {
                     res = receiver_outgoing.recv().fuse() => {
@@ -242,11 +251,18 @@ impl Downstream {
                 &host_
             );
         });
+        let _ = task_collector_new_sv1_message_no_transl.safe_lock(|a| {
+            a.push((
+                socket_writer_task.abort_handle(),
+                "socket_writer_task".to_string(),
+            ))
+        });
 
         let tx_status_notify = tx_status;
         let self_ = downstream.clone();
 
-        let _notify_task = task::spawn(async move {
+        let task_collector_notify_task = task_collector.clone();
+        let notify_task = tokio::task::spawn(async move {
             let timeout_timer = std::time::Instant::now();
             let mut first_sent = false;
             loop {
@@ -329,10 +345,14 @@ impl Downstream {
                 &host
             );
         });
+
+        let _ = task_collector_notify_task
+            .safe_lock(|a| a.push((notify_task.abort_handle(), "notify_task".to_string())));
     }
 
     /// Accept connections from one or more SV1 Downstream roles (SV1 Mining Devices) and create a
     /// new `Downstream` for each connection.
+    #[allow(clippy::too_many_arguments)]
     pub fn accept_connections(
         downstream_addr: SocketAddr,
         tx_sv1_submit: Sender<DownstreamMessages>,
@@ -341,8 +361,11 @@ impl Downstream {
         bridge: Arc<Mutex<crate::proxy::Bridge>>,
         downstream_difficulty_config: DownstreamDifficultyConfig,
         upstream_difficulty_config: Arc<Mutex<UpstreamDifficultyConfig>>,
+        task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
     ) {
-        task::spawn(async move {
+        let task_collector_downstream = task_collector.clone();
+
+        let accept_connections = tokio::task::spawn(async move {
             let downstream_listener = TcpListener::bind(downstream_addr).await.unwrap();
             let mut downstream_incoming = downstream_listener.incoming();
 
@@ -369,6 +392,7 @@ impl Downstream {
                             host,
                             downstream_difficulty_config.clone(),
                             upstream_difficulty_config.clone(),
+                            task_collector_downstream.clone(),
                         )
                         .await;
                     }
@@ -377,6 +401,12 @@ impl Downstream {
                     }
                 }
             }
+        });
+        let _ = task_collector.safe_lock(|a| {
+            a.push((
+                accept_connections.abort_handle(),
+                "accept_connections".to_string(),
+            ))
         });
     }
 
--- a/roles/translator/src/main.rs
+++ b/roles/translator/src/main.rs
@@ -3,13 +3,13 @@ mod args;
 mod lib;
 
 use args::Args;
+use async_channel::{bounded, unbounded};
 use error::{Error, ProxyResult};
+use futures::{select, FutureExt};
 use lib::{downstream_sv1, error, proxy, proxy_config, status, upstream_sv2};
 use proxy_config::ProxyConfig;
+use rand::Rng;
 use roles_logic_sv2::utils::Mutex;
-
-use async_channel::{bounded, unbounded};
-use futures::{select, FutureExt};
 use std::{
     net::{IpAddr, SocketAddr},
     str::FromStr,
@@ -17,11 +17,11 @@ use std::{
 };
 
 use ext_config::{Config, File, FileFormat};
-use tokio::{sync::broadcast, task};
+use tokio::{sync::broadcast, task, task::AbortHandle, time::Duration};
 use v1::server_to_client;
 
 use crate::status::{State, Status};
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 /// Process CLI args, if any.
 #[allow(clippy::result_large_err)]
 fn process_cli_args<'a>() -> ProxyResult<'a, ProxyConfig> {
@@ -54,22 +54,121 @@ async fn main() {
         Ok(p) => p,
         Err(e) => panic!("failed to load config: {}", e),
     };
-    info!("PC: {:?}", &proxy_config);
+    info!("Proxy Config: {:?}", &proxy_config);
 
     let (tx_status, rx_status) = unbounded();
+
+    let target = Arc::new(Mutex::new(vec![0; 32]));
+
+    // Sender/Receiver to send SV1 `mining.notify` message from the `Bridge` to the `Downstream`
+    let (tx_sv1_notify, _rx_sv1_notify): (
+        broadcast::Sender<server_to_client::Notify>,
+        broadcast::Receiver<server_to_client::Notify>,
+    ) = broadcast::channel(10);
+
+    let task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>> = Arc::new(Mutex::new(Vec::new()));
+
+    start(
+        tx_sv1_notify.clone(),
+        target.clone(),
+        tx_status.clone(),
+        task_collector.clone(),
+        proxy_config.clone(),
+    )
+    .await;
+
+    debug!("Starting up signal listener");
+    let task_collector_ = task_collector.clone();
+
+    let mut interrupt_signal_future = Box::pin(tokio::signal::ctrl_c().fuse());
+    debug!("Starting up status listener");
+    // Check all tasks if is_finished() is true, if so exit
+    loop {
+        let task_status = select! {
+            task_status = rx_status.recv().fuse() => task_status,
+            interrupt_signal = interrupt_signal_future => {
+                match interrupt_signal {
+                    Ok(()) => {
+                        info!("Interrupt received");
+                    },
+                    Err(err) => {
+                        error!("Unable to listen for interrupt signal: {}", err);
+                        // we also shut down in case of error
+                    },
+                }
+                break;
+            }
+        };
+        let task_status: Status = task_status.unwrap();
+
+        match task_status.state {
+            // Should only be sent by the downstream listener
+            State::DownstreamShutdown(err) => {
+                error!("SHUTDOWN from: {}", err);
+                break;
+            }
+            State::BridgeShutdown(err) => {
+                error!("SHUTDOWN from: {}", err);
+                break;
+            }
+            State::UpstreamShutdown(err) => {
+                error!("SHUTDOWN from: {}", err);
+                break;
+            }
+            State::UpstreamTryReconnect(err) => {
+                error!("SHUTDOWN from: {}", err);
+
+                // wait a random amount of time between 0 and 3000ms
+                // if all the downstreams try to reconnect at the same time, the upstream may fail
+                let mut rng = rand::thread_rng();
+                let wait_time = rng.gen_range(0..=3000);
+                tokio::time::sleep(Duration::from_millis(wait_time)).await;
+
+                // kill al the tasks
+                let task_collector_aborting = task_collector_.clone();
+                kill_tasks(task_collector_aborting.clone());
+
+                warn!("Trying reconnecting to upstream");
+                start(
+                    tx_sv1_notify.clone(),
+                    target.clone(),
+                    tx_status.clone(),
+                    task_collector_.clone(),
+                    proxy_config.clone(),
+                )
+                .await;
+            }
+            State::Healthy(msg) => {
+                info!("HEALTHY message: {}", msg);
+            }
+        }
+    }
+}
+
+fn kill_tasks(task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>) {
+    let _ = task_collector.safe_lock(|t| {
+        while let Some(handle) = t.pop() {
+            handle.0.abort();
+            warn!("Killed task: {:?}", handle.1);
+        }
+    });
+}
+
+async fn start<'a>(
+    tx_sv1_notify: broadcast::Sender<server_to_client::Notify<'static>>,
+    target: Arc<Mutex<Vec<u8>>>,
+    tx_status: async_channel::Sender<Status<'static>>,
+    task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
+    proxy_config: ProxyConfig,
+) {
+    // Sender/Receiver to send a SV2 `SubmitSharesExtended` from the `Bridge` to the `Upstream`
+    // (Sender<SubmitSharesExtended<'static>>, Receiver<SubmitSharesExtended<'static>>)
+    let (tx_sv2_submit_shares_ext, rx_sv2_submit_shares_ext) = bounded(10);
 
     // `tx_sv1_bridge` sender is used by `Downstream` to send a `DownstreamMessages` message to
     // `Bridge` via the `rx_sv1_downstream` receiver
     // (Sender<downstream_sv1::DownstreamMessages>, Receiver<downstream_sv1::DownstreamMessages>)
     let (tx_sv1_bridge, rx_sv1_downstream) = unbounded();
-
-    // Sender/Receiver to send a SV2 `SubmitSharesExtended` from the `Bridge` to the `Upstream`
-    // (Sender<SubmitSharesExtended<'static>>, Receiver<SubmitSharesExtended<'static>>)
-    let (tx_sv2_submit_shares_ext, rx_sv2_submit_shares_ext) = bounded(10);
-
-    // Sender/Receiver to send a SV2 `SetNewPrevHash` message from the `Upstream` to the `Bridge`
-    // (Sender<SetNewPrevHash<'static>>, Receiver<SetNewPrevHash<'static>>)
-    let (tx_sv2_set_new_prev_hash, rx_sv2_set_new_prev_hash) = bounded(10);
 
     // Sender/Receiver to send a SV2 `NewExtendedMiningJob` message from the `Upstream` to the
     // `Bridge`
@@ -80,13 +179,10 @@ async fn main() {
     // passed to the `Downstream` upon a Downstream role connection
     // (Sender<ExtendedExtranonce>, Receiver<ExtendedExtranonce>)
     let (tx_sv2_extranonce, rx_sv2_extranonce) = bounded(1);
-    let target = Arc::new(Mutex::new(vec![0; 32]));
 
-    // Sender/Receiver to send SV1 `mining.notify` message from the `Bridge` to the `Downstream`
-    let (tx_sv1_notify, _rx_sv1_notify): (
-        broadcast::Sender<server_to_client::Notify>,
-        broadcast::Receiver<server_to_client::Notify>,
-    ) = broadcast::channel(10);
+    // Sender/Receiver to send a SV2 `SetNewPrevHash` message from the `Upstream` to the `Bridge`
+    // (Sender<SetNewPrevHash<'static>>, Receiver<SetNewPrevHash<'static>>)
+    let (tx_sv2_set_new_prev_hash, rx_sv2_set_new_prev_hash) = bounded(10);
 
     // Format `Upstream` connection address
     let upstream_addr = SocketAddr::new(
@@ -96,7 +192,7 @@ async fn main() {
     );
 
     let diff_config = Arc::new(Mutex::new(proxy_config.upstream_difficulty_config.clone()));
-
+    let task_collector_upstream = task_collector.clone();
     // Instantiate a new `Upstream` (SV2 Pool)
     let upstream = match upstream_sv2::Upstream::new(
         upstream_addr,
@@ -109,6 +205,7 @@ async fn main() {
         status::Sender::Upstream(tx_status.clone()),
         target.clone(),
         diff_config.clone(),
+        task_collector_upstream,
     )
     .await
     {
@@ -118,12 +215,12 @@ async fn main() {
             return;
         }
     };
-
+    let task_collector_init_task = task_collector.clone();
     // Spawn a task to do all of this init work so that the main thread
     // can listen for signals and failures on the status channel. This
     // allows for the tproxy to fail gracefully if any of these init tasks
     //fail
-    task::spawn(async move {
+    let task = task::spawn(async move {
         // Connect to the SV2 Upstream role
         match upstream_sv2::Upstream::connect(
             upstream.clone(),
@@ -163,6 +260,7 @@ async fn main() {
             async_std::task::sleep(std::time::Duration::from_millis(100)).await;
         }
 
+        let task_collector_bridge = task_collector_init_task.clone();
         // Instantiate a new `Bridge` and begins handling incoming messages
         let b = proxy::Bridge::new(
             rx_sv1_downstream,
@@ -174,6 +272,7 @@ async fn main() {
             extended_extranonce,
             target,
             up_id,
+            task_collector_bridge,
         );
         proxy::Bridge::start(b.clone());
 
@@ -183,6 +282,7 @@ async fn main() {
             proxy_config.downstream_port,
         );
 
+        let task_collector_downstream = task_collector_init_task.clone();
         // Accept connections from one or more SV1 Downstream roles (SV1 Mining Devices)
         downstream_sv1::Downstream::accept_connections(
             downstream_addr,
@@ -192,49 +292,8 @@ async fn main() {
             b,
             proxy_config.downstream_difficulty_config,
             diff_config,
+            task_collector_downstream,
         );
     }); // End of init task
-
-    debug!("Starting up signal listener");
-    let mut interrupt_signal_future = Box::pin(tokio::signal::ctrl_c().fuse());
-    debug!("Starting up status listener");
-
-    // Check all tasks if is_finished() is true, if so exit
-    loop {
-        let task_status = select! {
-            task_status = rx_status.recv().fuse() => task_status,
-            interrupt_signal = interrupt_signal_future => {
-                match interrupt_signal {
-                    Ok(()) => {
-                        info!("Interrupt received");
-                    },
-                    Err(err) => {
-                        error!("Unable to listen for interrupt signal: {}", err);
-                        // we also shut down in case of error
-                    },
-                }
-                break;
-            }
-        };
-        let task_status: Status = task_status.unwrap();
-
-        match task_status.state {
-            // Should only be sent by the downstream listener
-            State::DownstreamShutdown(err) => {
-                error!("SHUTDOWN from: {}", err);
-                break;
-            }
-            State::BridgeShutdown(err) => {
-                error!("SHUTDOWN from: {}", err);
-                break;
-            }
-            State::UpstreamShutdown(err) => {
-                error!("SHUTDOWN from: {}", err);
-                break;
-            }
-            State::Healthy(msg) => {
-                info!("HEALTHY message: {}", msg);
-            }
-        }
-    }
+    let _ = task_collector.safe_lock(|t| t.push((task.abort_handle(), "init task".to_string())));
 }
--- a/roles/translator/Cargo.toml
+++ b/roles/translator/Cargo.toml
@@ -34,10 +34,10 @@ error_handling = { version = "1.0.0", path = "../../utils/error-handling" }
 key-utils = { version = "^1.0.0", path = "../../utils/key-utils" }
 tokio-util = { version = "0.7.10", features = ["codec"] }
 async-compat = "0.2.1"
+rand = "0.8.4"
 
 
 [dev-dependencies]
-rand = "0.8.4"
 sha2 = "0.10.6"
 
 [features]
--- a/roles/translator/src/lib/proxy/bridge.rs
+++ b/roles/translator/src/lib/proxy/bridge.rs
@@ -1,5 +1,4 @@
 use async_channel::{Receiver, Sender};
-use async_std::task;
 use roles_logic_sv2::{
     channel_logic::channel_factory::{ExtendedChannelKind, ProxyExtendedChannelFactory, Share},
     mining_sv2::{
@@ -9,7 +8,7 @@ use roles_logic_sv2::{
     utils::{GroupId, Mutex},
 };
 use std::sync::Arc;
-use tokio::sync::broadcast;
+use tokio::{sync::broadcast, task::AbortHandle};
 use v1::{client_to_server::Submit, server_to_client, utils::HexU32Be};
 
 use super::super::{
@@ -64,6 +63,7 @@ pub struct Bridge {
     last_p_hash: Option<SetNewPrevHash<'static>>,
     target: Arc<Mutex<Vec<u8>>>,
     last_job_id: u32,
+    task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
 }
 
 impl Bridge {
@@ -79,6 +79,7 @@ impl Bridge {
         extranonces: ExtendedExtranonce,
         target: Arc<Mutex<Vec<u8>>>,
         up_id: u32,
+        task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
     ) -> Arc<Mutex<Self>> {
         let ids = Arc::new(Mutex::new(GroupId::new()));
         let share_per_min = 1.0;
@@ -107,6 +108,7 @@ impl Bridge {
             last_p_hash: None,
             target,
             last_job_id: 0,
+            task_collector,
         }))
     }
 
@@ -162,10 +164,12 @@ impl Bridge {
     /// Receives a `DownstreamMessages` message from the `Downstream`, handles based on the
     /// variant received.
     fn handle_downstream_messages(self_: Arc<Mutex<Self>>) {
+        let task_collector_handle_downstream =
+            self_.safe_lock(|b| b.task_collector.clone()).unwrap();
         let (rx_sv1_downstream, tx_status) = self_
             .safe_lock(|s| (s.rx_sv1_downstream.clone(), s.tx_status.clone()))
             .unwrap();
-        task::spawn(async move {
+        let handle_downstream = tokio::task::spawn(async move {
             loop {
                 let msg = handle_result!(tx_status, rx_sv1_downstream.clone().recv().await);
 
@@ -184,6 +188,12 @@ impl Bridge {
                     }
                 };
             }
+        });
+        let _ = task_collector_handle_downstream.safe_lock(|a| {
+            a.push((
+                handle_downstream.abort_handle(),
+                "handle_downstream_message".to_string(),
+            ))
         });
     }
     /// receives a `SetDownstreamTarget` and updates the downstream target for the channel
@@ -367,6 +377,8 @@ impl Bridge {
     /// corresponding `job_id` has already been received. If this is not the case, an error has
     /// occurred on the Upstream pool role and the connection will close.
     fn handle_new_prev_hash(self_: Arc<Mutex<Self>>) {
+        let task_collector_handle_new_prev_hash =
+            self_.safe_lock(|b| b.task_collector.clone()).unwrap();
         let (tx_sv1_notify, rx_sv2_set_new_prev_hash, tx_status) = self_
             .safe_lock(|s| {
                 (
@@ -377,7 +389,7 @@ impl Bridge {
             })
             .unwrap();
         debug!("Starting handle_new_prev_hash task");
-        task::spawn(async move {
+        let handle_new_prev_hash = tokio::task::spawn(async move {
             loop {
                 // Receive `SetNewPrevHash` from `Upstream`
                 let sv2_set_new_prev_hash: SetNewPrevHash =
@@ -396,6 +408,12 @@ impl Bridge {
                     .await
                 )
             }
+        });
+        let _ = task_collector_handle_new_prev_hash.safe_lock(|a| {
+            a.push((
+                handle_new_prev_hash.abort_handle(),
+                "handle_new_prev_hash".to_string(),
+            ))
         });
     }
 
@@ -460,6 +478,8 @@ impl Bridge {
     /// `SetNewPrevHash` `job_id`, an error has occurred on the Upstream pool role and the
     /// connection will close.
     fn handle_new_extended_mining_job(self_: Arc<Mutex<Self>>) {
+        let task_collector_new_extended_mining_job =
+            self_.safe_lock(|b| b.task_collector.clone()).unwrap();
         let (tx_sv1_notify, rx_sv2_new_ext_mining_job, tx_status) = self_
             .safe_lock(|s| {
                 (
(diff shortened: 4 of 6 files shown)


Shourya742 (changes requested):
Concept ACK
This approach can be optimized. Currently, we are spawning a new task immediately following the task we intend to terminate. A more efficient method would involve incorporating the receiver as one of the branches in the select! block within the task we wish to terminate. This adjustment would eliminate the need for an additional asynchronous task, thereby facilitating a more graceful shutdown rather than an abrupt one.
A possible code snippet for one such task can be: 
```
let socket_reader_task = tokio::task::spawn(async move {
    let reader = BufReader::new(&*socket_reader);
    let mut messages = FramedRead::new(
        async_compat::Compat::new(reader),
        LinesCodec::new_with_max_length(MAX_LINE_LENGTH),
    );

    loop {
        tokio::select! {
            res = messages.next().fuse() => {
                match res {
                    Some(Ok(incoming)) => {
                        debug!("Receiving from Mining Device {}: {:?}", &host_, &incoming);
                        let incoming: json_rpc::Message = handle_result!(tx_status_reader, serde_json::from_str(&incoming));

                        if let v1::Message::StandardRequest(standard_req) = incoming {
                            if let Ok(Submit{..}) = standard_req.try_into() {
                                handle_result!(tx_status_reader, Self::save_share(self_.clone()));
                            }
                        }

                        let res = Self::handle_incoming_sv1(self_.clone(), incoming).await;
                        handle_result!(tx_status_reader, res);
                    },
                    Some(Err(_)) => {
                        handle_result!(tx_status_reader, Err(Error::Sv1MessageTooLong));
                    },
                    None => {
                        handle_result!(tx_status_reader, Err(
                            std::io::Error::new(
                                std::io::ErrorKind::ConnectionAborted,
                                "Connection closed by client"
                            )
                        ));
                        break;                    
                      }
                }
            },
            _ = cancellation_token_mining_device.cancelled().fuse() => {
                warn!("Cancellation token triggered: Initiating shutdown of sv1 downstream reader");
                break; 
            },
            _ = rx_shutdown_clone.recv().fuse() => {
                warn!("Shutdown signal received: Initiating shutdown of sv1 downstream reader");
                break;
            }
        }
    }
    warn!("Downstream: Reader task is terminating. Performing cleanup.");
    kill(&tx_shutdown_clone).await; 
});
```

roles/translator/src/main.rs
```diff
@@ -77,6 +82,120 @@ async fn main() {
...
+    rx_sv2_extranonce: Receiver<(ExtendedExtranonce, u32)>,
+    rx_sv2_set_new_prev_hash: Receiver<SetNewPrevHash<'static>>,
+    rx_sv2_new_ext_mining_job: Receiver<NewExtendedMiningJob<'static>>,
+    rx_sv1_downstream: Receiver<DownstreamMessages>,
+    tx_sv1_bridge: Sender<DownstreamMessages>,
+    tx_sv1_notify: broadcast::Sender<server_to_client::Notify<'static>>,
+    target: Arc<Mutex<Vec<u8>>>,
+    tx_status: async_channel::Sender<Status<'static>>,
+    cancellation_token: CancellationToken,
+) {
```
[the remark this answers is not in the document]
lorbax: Yes but at the end you have to construct the struct with a `new()` method or some `add_data_1()`, `add_data_2()` ecc. but eventually that amount of data is gonna needed, one way or another. 
Am I missing something?

lorbax:
> Concept ACK This approach can be optimized. Currently, we are spawning a new task immediately following the task we intend to terminate. A more efficient method would involve incorporating the receiver as one of the branches in the select! block within the task we wish to terminate. This adjustment would eliminate the need for an additional asynchronous task, thereby facilitating a more graceful shutdown rather than an abrupt one. A possible code snippet for one such task can be:
> 
> ```
> let socket_reader_task = tokio::task::spawn(async move {
>     let reader = BufReader::new(&*socket_reader);
>     let mut messages = FramedRead::new(
>         async_compat::Compat::new(reader),
>         LinesCodec::new_with_max_length(MAX_LINE_LENGTH),
>     );
> 
>     loop {
>         tokio::select! {
>             res = messages.next().fuse() => {
>                 match res {
>                     Some(Ok(incoming)) => {
>                         debug!("Receiving from Mining Device {}: {:?}", &host_, &incoming);
>                         let incoming: json_rpc::Message = handle_result!(tx_status_reader, serde_json::from_str(&incoming));
> 
>                         if let v1::Message::StandardRequest(standard_req) = incoming {
>                             if let Ok(Submit{..}) = standard_req.try_into() {
>                                 handle_result!(tx_status_reader, Self::save_share(self_.clone()));
>                             }
>                         }
> 
>                         let res = Self::handle_incoming_sv1(self_.clone(), incoming).await;
>                         handle_result!(tx_status_reader, res);
>                     },
>                     Some(Err(_)) => {
>                         handle_result!(tx_status_reader, Err(Error::Sv1MessageTooLong));
>                     },
>                     None => {
>                         handle_result!(tx_status_reader, Err(
>                             std::io::Error::new(
>                                 std::io::ErrorKind::ConnectionAborted,
>                                 "Connection closed by client"
>                             )
>                         ));
>                         break;                    
>                       }
>                 }
>             },
>             _ = cancellation_token_mining_device.cancelled().fuse() => {
>                 warn!("Cancellation token triggered: Initiating shutdown of sv1 downstream reader");
>                 break; 
>             },
>             _ = rx_shutdown_clone.recv().fuse() => {
>                 warn!("Shutdown signal received: Initiating shutdown of sv1 downstream reader");
>                 break;
>             }
>         }
>     }
>     warn!("Downstream: Reader task is terminating. Performing cleanup.");
>     kill(&tx_shutdown_clone).await; 
> });
> ```

That is a particular task, but how would you do with this task? I could not find a better solution.
```rust
        let task = tokio::task::spawn(async move {
            loop {
                let msg = handle_result!(tx_status, rx_sv1_downstream.clone().recv().await);

                match msg {
                    DownstreamMessages::SubmitShares(share) => {
                        handle_result!(
                            tx_status,
                            Self::handle_submit_shares(self_.clone(), share).await
                        );
                    }
                    DownstreamMessages::SetDownstreamTarget(new_target) => {
                        handle_result!(
                            tx_status,
                            Self::handle_update_downstream_target(self_.clone(), new_target)
                        );
                    }
                };
            }
        });
        tokio::task::spawn(async move {
            tokio::select! {
                _ = cancellation_token_handle_downstream.cancelled() => {
                    task.abort();
                    warn!("Shutting down handle_result task");
                },
            }
        });
```
and what do you think about using a mutex as JoinaHandler` collector for spawned tasks and close all of them in task in the main?

roles/translator/src/lib/downstream_sv1/downstream.rs
```diff
@@ -205,15 +208,24 @@ impl Downstream {
...
             warn!("Downstream: Shutting down sv1 downstream reader");
         });
+        tokio::task::spawn(async move {
+            tokio::select! {
+                _ = cancellation_token_mining_device.cancelled() => {
+                    socket_reader_task.abort();
+                    warn!("Shutting down sv1 downstream reader");
+                },
+            }
+        });
```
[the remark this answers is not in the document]
lorbax: I apply this change to every task

Shourya742:
> > Concept ACK This approach can be optimized. Currently, we are spawning a new task immediately following the task we intend to terminate. A more efficient method would involve incorporating the receiver as one of the branches in the select! block within the task we wish to terminate. This adjustment would eliminate the need for an additional asynchronous task, thereby facilitating a more graceful shutdown rather than an abrupt one. A possible code snippet for one such task can be:
> > ```
> > let socket_reader_task = tokio::task::spawn(async move {
> >     let reader = BufReader::new(&*socket_reader);
> >     let mut messages = FramedRead::new(
> >         async_compat::Compat::new(reader),
> >         LinesCodec::new_with_max_length(MAX_LINE_LENGTH),
> >     );
> > 
> >     loop {
> >         tokio::select! {
> >             res = messages.next().fuse() => {
> >                 match res {
> >                     Some(Ok(incoming)) => {
> >                         debug!("Receiving from Mining Device {}: {:?}", &host_, &incoming);
> >                         let incoming: json_rpc::Message = handle_result!(tx_status_reader, serde_json::from_str(&incoming));
> > 
> >                         if let v1::Message::StandardRequest(standard_req) = incoming {
> >                             if let Ok(Submit{..}) = standard_req.try_into() {
> >                                 handle_result!(tx_status_reader, Self::save_share(self_.clone()));
> >                             }
> >                         }
> > 
> >                         let res = Self::handle_incoming_sv1(self_.clone(), incoming).await;
> >                         handle_result!(tx_status_reader, res);
> >                     },
> >                     Some(Err(_)) => {
> >                         handle_result!(tx_status_reader, Err(Error::Sv1MessageTooLong));
> >                     },
> >                     None => {
> >                         handle_result!(tx_status_reader, Err(
> >                             std::io::Error::new(
> >                                 std::io::ErrorKind::ConnectionAborted,
> >                                 "Connection closed by client"
> >                             )
> >                         ));
> >                         break;                    
> >                       }
> >                 }
> >             },
> >             _ = cancellation_token_mining_device.cancelled().fuse() => {
> >                 warn!("Cancellation token triggered: Initiating shutdown of sv1 downstream reader");
> >                 break; 
> >             },
> >             _ = rx_shutdown_clone.recv().fuse() => {
> >                 warn!("Shutdown signal received: Initiating shutdown of sv1 downstream reader");
> >                 break;
> >             }
> >         }
> >     }
> >     warn!("Downstream: Reader task is terminating. Performing cleanup.");
> >     kill(&tx_shutdown_clone).await; 
> > });
> > ```
> 
> That is a particular task, but how would you do with this task? I could not find a better solution.
> 
> ```rust
>         let task = tokio::task::spawn(async move {
>             loop {
>                 let msg = handle_result!(tx_status, rx_sv1_downstream.clone().recv().await);
> 
>                 match msg {
>                     DownstreamMessages::SubmitShares(share) => {
>                         handle_result!(
>                             tx_status,
>                             Self::handle_submit_shares(self_.clone(), share).await
>                         );
>                     }
>                     DownstreamMessages::SetDownstreamTarget(new_target) => {
>                         handle_result!(
>                             tx_status,
>                             Self::handle_update_downstream_target(self_.clone(), new_target)
>                         );
>                     }
>                 };
>             }
>         });
>         tokio::task::spawn(async move {
>             tokio::select! {
>                 _ = cancellation_token_handle_downstream.cancelled() => {
>                     task.abort();
>                     warn!("Shutting down handle_result task");
>                 },
>             }
>         });
> ```
> 
> and what do you think about using a mutex as JoinaHandler` collector for spawned tasks and close all of them in task in the main?
Ah, I missed it! In that case, aggregating JoinHandles with simultaneous termination would make more sense.

lorbax:
Applied the changes suggested by @Shourya742

lorbax:
> > > Concept ACK This approach can be optimized. Currently, we are spawning a new task immediately following the task we intend to terminate. A more efficient method would involve incorporating the receiver as one of the branches in the select! block within the task we wish to terminate. This adjustment would eliminate the need for an additional asynchronous task, thereby facilitating a more graceful shutdown rather than an abrupt one. A possible code snippet for one such task can be:
> > > ```
> > > let socket_reader_task = tokio::task::spawn(async move {
> > >     let reader = BufReader::new(&*socket_reader);
> > >     let mut messages = FramedRead::new(
> > >         async_compat::Compat::new(reader),
> > >         LinesCodec::new_with_max_length(MAX_LINE_LENGTH),
> > >     );
> > > 
> > >     loop {
> > >         tokio::select! {
> > >             res = messages.next().fuse() => {
> > >                 match res {
> > >                     Some(Ok(incoming)) => {
> > >                         debug!("Receiving from Mining Device {}: {:?}", &host_, &incoming);
> > >                         let incoming: json_rpc::Message = handle_result!(tx_status_reader, serde_json::from_str(&incoming));
> > > 
> > >                         if let v1::Message::StandardRequest(standard_req) = incoming {
> > >                             if let Ok(Submit{..}) = standard_req.try_into() {
> > >                                 handle_result!(tx_status_reader, Self::save_share(self_.clone()));
> > >                             }
> > >                         }
> > > 
> > >                         let res = Self::handle_incoming_sv1(self_.clone(), incoming).await;
> > >                         handle_result!(tx_status_reader, res);
> > >                     },
> > >                     Some(Err(_)) => {
> > >                         handle_result!(tx_status_reader, Err(Error::Sv1MessageTooLong));
> > >                     },
> > >                     None => {
> > >                         handle_result!(tx_status_reader, Err(
> > >                             std::io::Error::new(
> > >                                 std::io::ErrorKind::ConnectionAborted,
> > >                                 "Connection closed by client"
> > >                             )
> > >                         ));
> > >                         break;                    
> > >                       }
> > >                 }
> > >             },
> > >             _ = cancellation_token_mining_device.cancelled().fuse() => {
> > >                 warn!("Cancellation token triggered: Initiating shutdown of sv1 downstream reader");
> > >                 break; 
> > >             },
> > >             _ = rx_shutdown_clone.recv().fuse() => {
> > >                 warn!("Shutdown signal received: Initiating shutdown of sv1 downstream reader");
> > >                 break;
> > >             }
> > >         }
> > >     }
> > >     warn!("Downstream: Reader task is terminating. Performing cleanup.");
> > >     kill(&tx_shutdown_clone).await; 
> > > });
> > > ```
> > 
> > 
> > That is a particular task, but how would you do with this task? I could not find a better solution.
> > ```rust
> >         let task = tokio::task::spawn(async move {
> >             loop {
> >                 let msg = handle_result!(tx_status, rx_sv1_downstream.clone().recv().await);
> > 
> >                 match msg {
> >                     DownstreamMessages::SubmitShares(share) => {
> >                         handle_result!(
> >                             tx_status,
> >                             Self::handle_submit_shares(self_.clone(), share).await
> >                         );
> >                     }
> >                     DownstreamMessages::SetDownstreamTarget(new_target) => {
> >                         handle_result!(
> >                             tx_status,
> >                             Self::handle_update_downstream_target(self_.clone(), new_target)
> >                         );
> >                     }
> >                 };
> >             }
> >         });
> >         tokio::task::spawn(async move {
> >             tokio::select! {
> >                 _ = cancellation_token_handle_downstream.cancelled() => {
> >                     task.abort();
> >                     warn!("Shutting down handle_result task");
> >                 },
> >             }
> >         });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and what do you think about using a mutex as JoinaHandler` collector for spawned tasks and close all of them in task in the main?
> 
> Ah, I missed it! In that case, aggregating JoinHandles with simultaneous termination would make more sense.

We can discuss it at the call, when I realized using a mutex as a joinhandler collector for the tasks, I liked it but I do not want to change it if not necessary.

test/message-generator/test/change-upstream/change-upstream.json
plebhash, line 1: I'm confused. I don't see MG starting any role here, which usually I see on the `setup_commands` field.

Is this testing tProxy? if so, how? What am I missing?

Also, there's no shell script for invoking this MG test, and it would also be good to add it to CI.

lorbax: To replicate the issue and to test the fix is info contained in the messages of the two commits.

GitGab19: @plebhash I think that @lorbax is referring this: https://github.com/stratum-mining/stratum/pull/1001/commits/c54cb76cd7c92ad7444d3a4092cc823a5bf75fe8

lorbax: My idea is to include a MG test in the CI, testing both this scenario and when there is solo fallback (which should be ready soon). Also, would be good to add a new functionality to the MG, namely the possibility to ignore all the messages until one of particular type is received. 
This is because at the same moment the translator can send a SetCustomMiningJob, an UpdateChannel and a SubmitShare to the upstream. Right now there is no way to this "hey ignore the first two message types and send back an error whenever you receive a SubmitShare".

GitGab19:
@lorbax is it ready for a final review?
Did you complete the changes we discussed last week during dev call?

lorbax:
> @lorbax is it ready for a final review? Did you complete the changes we discussed last week during dev call?

yes!

test/message-generator/test/change-upstream/change-upstream.json
plebhash: > My idea is to include a MG test in the CI, testing both this scenario and when there is solo fallback (which should be ready soon). Also, would be good to add a new functionality to the MG, namely the possibility to ignore all the messages until one of particular type is received.

is that going to be done on a separate PR?

lorbax: I think that it is a good idea. This PR has priority in my opinion. We can either wait to have the proper feature on the MG side, for me is the same.

roles/translator/src/lib/downstream_sv1/downstream.rs
```diff
@@ -242,11 +251,18 @@ impl Downstream {
                 &host_
             );
         });
+        let _ = task_collector_new_sv1_message_no_transl.safe_lock(|a| {
+            a.push((
+                socket_writer_task.abort_handle(),
+                "socket_writer_task".to_string(),
+            ))
```
Shourya742, line 258: Any specific reason adding context to abort handle?

roles/translator/config-examples/tproxy-config-local-jdc-example.toml
plebhash, line 1: it looks like this file was committed by mistake

roles/translator/src/lib/downstream_sv1/downstream.rs
plebhash: it seems the purpose of these strings is so that `kill_tasks` can print them to the console, like this:
```
2024-08-01T23:36:40.372202Z  WARN translator_sv2: Killed task: "notify_task"
2024-08-01T23:36:40.372229Z  WARN translator_sv2: Killed task: "socket_writer_task"
2024-08-01T23:36:40.372233Z  WARN translator_sv2: Killed task: "socket_reader_task"
2024-08-01T23:36:40.372237Z  WARN translator_sv2: Killed task: "accept_connections"
2024-08-01T23:36:40.372240Z  WARN translator_sv2: Killed task: "handle_downstream_message"
2024-08-01T23:36:40.372244Z  WARN translator_sv2: Killed task: "handle_new_extended_mining_job"
2024-08-01T23:36:40.372247Z  WARN translator_sv2: Killed task: "handle_new_prev_hash"
2024-08-01T23:36:40.372251Z  WARN translator_sv2: Killed task: "handle_submit"
2024-08-01T23:36:40.372258Z  WARN translator_sv2: Killed task: "parse_incoming"
2024-08-01T23:36:40.372360Z  WARN translator_sv2: Killed task: "start_diff_management"
2024-08-01T23:36:40.372365Z  WARN translator_sv2: Killed task: "init task"
```

plebhash:
yesterday we had a PR review club, where we dived into #1001 (which aims to fix #1004)

we found some issues (`UpstreamShutdown(UpstreamIncoming(UnexpectedMessage(1)))`), and after further investigation I reported it into #1079 and fixed via #1080

after that, I was able to manually reproduce #1004, but only with the trick described by @GitGab19 on this comment: https://github.com/stratum-mining/stratum/issues/844#issuecomment-2118864024

basically I launched one pool with this patch (listening on port 44254), and another pool without any patches (listening on port 34254)

then I modified JDC config so it would have two `[[upstreams]]` sections:
```
[[upstreams]]
authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
pool_address = "127.0.0.1:44254"
jd_address = "127.0.0.1:34264"
# Pool signature (string to be included in coinbase tx)
pool_signature = "Stratum v2 SRI Pool"

[[upstreams]]
authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
pool_address = "127.0.0.1:34254"
jd_address = "127.0.0.1:34264"
# Pool signature (string to be included in coinbase tx)
pool_signature = "Stratum v2 SRI Pool"
```

finally, I was able to verify that this PR is allowing tProxy to restart after the patched pool rejects a valid share

---

unfortunately the MG mock for a bad pool isn't working reliably... I tried following the instructions from 3d8793a3fc5aa087737dc3d207a6b54cf68858f0 multiple times but the behavior isn't deterministic

given that:
- #1059 was de-prioritized
- the instructions on 3d8793a3fc5aa087737dc3d207a6b54cf68858f0 are already highly manual, without deterministic automation
- the manual instructions above are sufficient for reproducing

I think we can probably drop 3d8793a3fc5aa087737dc3d207a6b54cf68858f0 and only use this PR to fix the bug

lorbax:
> yesterday we had a PR review club, where we dived into this PR
> 
> we found some issues (`UpstreamShutdown(UpstreamIncoming(UnexpectedMessage(1)))`), and after further investigation I reported it into #1079 and fixed via #1080
> 
> after that, I was able to manually reproduce #1004, but only with the trick described by @GitGab19 here: [#844 (comment)](https://github.com/stratum-mining/stratum/issues/844#issuecomment-2118864024)
> 
> basically I launched one pool with this patch (listening on port 44254), and another pool without any patches (listening on port 34254)
> 
> then I modified JDC config so it would have two `[[upstreams]]` sections:
> 
> ```
> [[upstreams]]
> authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
> pool_address = "127.0.0.1:44254"
> jd_address = "127.0.0.1:34264"
> # Pool signature (string to be included in coinbase tx)
> pool_signature = "Stratum v2 SRI Pool"
> 
> [[upstreams]]
> authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
> pool_address = "127.0.0.1:34254"
> jd_address = "127.0.0.1:34264"
> # Pool signature (string to be included in coinbase tx)
> pool_signature = "Stratum v2 SRI Pool"
> ```
> 

I already had included a jdc config for testing the PR
[https://github.com/lorbax/stratum/blob/translator-restart-if-disconnected-from-upstream/test/config/change-upstream/jdc-config-local-example-change-upstream.toml](https://github.com/lorbax/stratum/blob/translator-restart-if-disconnected-from-upstream/test/config/change-upstream/jdc-config-local-example-change-upstream.toml)
so that you do have to modify it. It is described in the PR that all the configs to be used are in `test/config/change-upstream/`

> unfortunately the MG mock for a bad pool isn't working reliably... I tried following the instructions from [3d8793a](https://github.com/stratum-mining/stratum/commit/3d8793a3fc5aa087737dc3d207a6b54cf68858f0) multiple times but the behavior isn't deterministic

The only thing is that the MG should need a new [MG feature](https://github.com/stratum-mining/stratum/issues/1059) in order to work flawlessly, but I used the MG test yesterday to replicate the issue and to test the fix (also after the review club) and it does the job. Can you provide more information about the difficulties or inconsistencies that you got using that test? "isn't working reliably" is a bit vague.

plebhash:
apologies for the lack of detail

I started all roles by following the instructions on 3d8793a3fc5aa087737dc3d207a6b54cf68858f0 and using all config files under `test/config/change-upstream`

the mock pool is killed immediately after the `SubmitShares.Error` is triggered, and as a consequence JDC does not try to fallback, it just gets stuck on trying to reconnect to the first pool

here's the JDC logs:
```
2024-08-02T15:33:49.535566Z  INFO jd_client::lib::upstream_sv2::upstream: Set custom mining job success 1
2024-08-02T15:33:58.853289Z  INFO roles_logic_sv2::handlers::mining: Received UpdateChannel->Extended message
2024-08-02T15:34:19.304315Z ERROR roles_logic_sv2::handlers::mining: Received SubmitSharesError with error code invalid-channel-id
2024-08-02T15:34:19.307149Z ERROR network_helpers_sv2::noise_connection_tokio: Disconnected from client while reading : early eof - 127.0.0.1:44254
2024-08-02T15:34:19.307180Z ERROR jd_client::lib::status: Error: ChannelErrorReceiver(RecvError)
2024-08-02T15:34:19.307225Z ERROR jd_client::lib: SHUTDOWN from: Channel receive error: `RecvError`
2024-08-02T15:34:20.319042Z ERROR network_helpers_sv2::noise_connection_tokio: Disconnecting from client due to error receiving: receiving from an empty and closed channel - 127.0.0.1:8442
2024-08-02T15:34:20.321992Z ERROR network_helpers_sv2::noise_connection_tokio: Disconnected from client while reading : early eof - 127.0.0.1:8442
2024-08-02T15:34:21.321514Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:26.331207Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:31.335578Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:36.341290Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:41.343409Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:46.346232Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:51.348015Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:34:56.353763Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:35:01.357632Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
2024-08-02T15:35:06.361785Z ERROR jd_client::lib::upstream_sv2::upstream: Failed to connect to Upstream role at 127.0.0.1:44254, retrying in 5s: Connection refused (os error 61)
```

am I missing something?

roles/translator/src/main.rs
```diff
@@ -77,6 +81,131 @@ async fn main() {
...
+    tx_sv1_bridge: Sender<DownstreamMessages>,
+    tx_sv1_notify: broadcast::Sender<server_to_client::Notify<'static>>,
+    target: Arc<Mutex<Vec<u8>>>,
+    tx_status: async_channel::Sender<Status<'static>>,
+    task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
+) {
+    let proxy_config = match process_cli_args() {
+        Ok(p) => p,
+        Err(e) => panic!("failed to load config: {}", e),
+    };
```
GitGab19, line 207: Why do you need this here?
Where was it before your changes in this PR?

GitGab19 (commented):
LGTM. 
Since it seems not reliable (without the patch you mentioned us some days ago @lorbax), and we decided to move those integration tests in another place, I would just remove the MG test and the relative configs from this PR.
Tomorrow I will properly test it and review it again.

roles/translator/src/main.rs
lorbax: I dont need that there. I changed it passing to the start function

lorbax:
> onnection refused (os error 61)

this is the first time that I see this error. Usually the channel ID is fixed for the entire duration of the connection and starts from 1, but also in the MG test the message `SubmitShareError` has channel id is 1.

plebhash (approved):

roles/translator/src/main.rs
```diff
@@ -77,6 +81,129 @@ async fn main() {
...
+                break;
+            }
+            State::BridgeShutdown(err) => {
+                error!("SHUTDOWN from: {}", err);
+                break;
+            }
+            State::UpstreamShutdown(err) => {
+                error!("SHUTDOWN from: {}", err);
+
+                // wait a random amount of time between 0 and 3000ms
```
jbesraa, line 142: Why do you need to introduce rand as a dep for this? just go for the 3000ms ?

roles/translator/src/main.rs
```diff
@@ -77,6 +81,129 @@ async fn main() {
...
+    let _ = task_collector.safe_lock(|t| {
+        while let Some(handle) = t.pop() {
+            handle.0.abort();
+            warn!("Killed task: {:?}", handle.1);
+        }
+    });
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn start<'a>(
```
jbesraa, line 190: ummm ?? could you please why all of these inputs?  how are you expecting other users to call this function?

jbesraa (changes requested):
nack. I dont think the translator should look like that. the `start` function is far from anything we can work with. 
please look here https://github.com/stratum-mining/stratum/blob/fbb846450a3053f7e57c9bdf90d407d4e10449da/roles/translator/src/lib/mod.rs I think this should be more or less the approach.
If you need to persist data, then we should add write/read to (disk?) abilities and save the data and recover it when we need it later.

roles/translator/src/main.rs
plebhash: as the comment states:

> if all the downstreams try to reconnect at the same time, the upstream may fail

if I understand correctly, this is trying to avoid a situation where multiple tProxies attempt to reconnect to the same Upstream at the same time, which could potentially cause it to fail

so it's attempting to introduce a randomness to this process, so that different tProxies will try to reconnect at different moments

I agree with the overall sentiment that this approach feels a bit hacky.

@lorbax do you know for a fact that the upstream will fail under these circumstances, or is it still just speculation?

if we're not sure this will happen, I'd say we should avoid solutions to imaginary problems

if we eventually see this happening, we can try to come up with a more sustainable solution

roles/translator/src/main.rs
plebhash: I agree with the criticism, but I wonder if this couldn't be addressed in #1066 , since there's already some refactoring going on there?

plebhash (changes requested):
I approved this earlier but then I noticed two MG tests are failing deterministically, so unfortunately I have to revert my approval.

Basically, now we're making tProxy attempt to reconnect to `Upstream` every time there's a `State::UpstreamShutdown`, regardless of what caused it.

## `pool-sri-test-close-channel`

To be honest, I'm not sure what this MG tries to achieve in the first place.

There's nothing [in the specs](https://stratumprotocol.org/specification/05-Mining-Protocol/#539-closechannel-client---server-server---client) stating that a proxy must close the entire connection when it receives a `CloseChannel` from `Upstream`.

This feels like a naive test. I would lean towards completely removing this test from the repo (ideally on a separate PR, to be merged before this one), but it would be good to get feedback from others.

## `translation-proxy-broke-pool`

This MG test asserts that the tProxy closes the connection when it receives a bad extranonce from `Upstream`.

That feels like a reasonable behavior, but it kind of conflicts with the current approach of this PR, because a bad extrononce also triggers a `State::UpstreamShutdown`, which now causes tProxy to attempt to reconnect. 

So if `Upstream` sends a bad extranonce, should tProxy really keep trying to reconnect forever?

I'm not sure how to proceed here.

---

@GitGab19 @lorbax please let me know your thoughts.

roles/translator/src/main.rs
Fi3: this is best practice, when you have several client that can disconnect and then reconnect automatically you have to add a little randomness to the process. Is not hacky is the standard way to do this kind of thing

plebhash: what if an attacker deploys multiple malicious tProxies and intentionally resets all the connections at the same time?

plebhash:
after further investigating the MG tests described above, I came up with a solution: https://github.com/plebhash/stratum/commit/e63c249bec12081115b7a41fcc999d5b97c5af99

basically, tProxy should try to reconnect if the error coming from `Upstream` is `Error::ChannelErrorReceiver(_)` or `Error::TokioChannelErrorRecv(_)`, which indicate an abrupt interruption on the channel managed by `network_helpers_sv2`

with this approach, both MG tests listed above are passing

I still think that `pool-sri-test-close-channel` is a bit weird and deserves a deeper investigation, but we can leave that for https://github.com/stratum-mining/stratum/issues/1081

@lorbax please have a look at the solution on the suggested commit... if you agree, feel free to just port it to your branch and squash it into the single commit of this PR, no need to worry about credit/authorship

roles/translator/src/main.rs
lorbax: > ummm ?? could you please why all of these inputs? how are you expecting other users to call this function?

I do not expect other users to call this function.

roles/translator/src/main.rs
Fi3: is there to prevent to ddos yourself or the pool that you are using

roles/translator/src/main.rs
jbesraa: This is a `start` function to the `translator` role, how are u not expecting other users to call it?
I mean, clippy is already screaming and you decide to disable it, I dont think this is the best approach

roles/translator/src/main.rs
jbesraa: @Fi3 could you please explain further? I dont understand why this is needed

Fi3: you have 1M tproxy connected to a pool. The pool have to close the connection, all the tproxy try to reconnect in the same exact moment. Not good

lorbax:
> after further investigating the MG tests described above, I came up with a solution: [plebhash@e63c249](https://github.com/plebhash/stratum/commit/e63c249bec12081115b7a41fcc999d5b97c5af99)
> 
> basically, tProxy should only try to reconnect if the error coming from `Upstream` is `Error::ChannelErrorReceiver(_)` or `Error::TokioChannelErrorRecv(_)`, which indicates an abrupt interruption on the channel managed by `network_helpers_sv2`
> 
> with this approach, both MG tests listed above are passing
> 
> I still think that `pool-sri-test-close-channel` is a bit weird and deserves a deeper investigation, but we can leave that for #1081
> 
> @lorbax please have a look at the solution on the suggested commit... if you agree, feel free to just port it to your branch and squash it into the single commit of this PR, no need to worry about credit/authorship

looks good

roles/translator/src/main.rs
plebhash: I'm not strictly opposed to having this here, I just don't think there any substantial benefit

sure, we are making sure SRI tProxy is a "benign" client that intentionally avoids DDoSing the upstream connection

but in production scenarios, this is not really protecting upstream from malicious users

as discussed in the call, the only sustainable solution for this is load balancers and similar techniques to be deployed on the server infrastructure

Fi3: again this is industry best practice everyone does it. And there is a reason. This have nothing to do with protecting upstream from malicious users.  https://www.google.com/search?client=firefox-b-d&q=random+delay+backoff

Fi3: when you have an automatic reconnection or you retry to reconnect with exponetial backoff you add a random delay to avoid have all the client reconnect in the same instant.

Fi3: One thing is not ddos yourself another thing is protecting against a ddos

pavlenex: @Fi3 Were this concerns addressed in https://github.com/stratum-mining/stratum/pull/1001#issuecomment-2270026282 it seems @lorbax agrees with the proposal, but it's not yet been committed?

Fi3: I think so. I did not reviewed it, I was only explaining why we should use randomized delay when we reconnect.

pavlenex: @lorbax can you commit what @plebhash suggested if it addresses the concern? cc @plebhash 

Then we may be able to finally push this PR through and start testing it, and if it works https://github.com/stratum-mining/stratum/pull/1001#discussion_r1704210248 can be done in a follow up.

lorbax:
> after further investigating the MG tests described above, I came up with a solution: [plebhash@e63c249](https://github.com/plebhash/stratum/commit/e63c249bec12081115b7a41fcc999d5b97c5af99)
> 
> basically, tProxy should only try to reconnect if the error coming from `Upstream` is `Error::ChannelErrorReceiver(_)` or `Error::TokioChannelErrorRecv(_)`, which indicates an abrupt interruption on the channel managed by `network_helpers_sv2`
> 
> with this approach, both MG tests listed above are passing
> 
> I still think that `pool-sri-test-close-channel` is a bit weird and deserves a deeper investigation, but we can leave that for #1081
> 
> @lorbax please have a look at the solution on the suggested commit... if you agree, feel free to just port it to your branch and squash it into the single commit of this PR, no need to worry about credit/authorship

Hi!
I had a closer look to your proposal. When the Translator gets disconnected, it receives 
```rust
translator_sv2::error::Error::ChannelErrorReceiver
```
but you suggests to send `UpstreamTryReconnect`. Did you have an actual case in which the jdc disconnects the Translator and you get 
```rust
translator_sv2::error::Error::TokioChannelErrorRecv
```
?
why it is not enough the following?
```rust
        Sender::Upstream(tx) => {
            tx.send(Status {
            match e {
                state: State::UpstreamShutdown(e),
                Error::ChannelErrorReceiver(_) => {
            })
                    tx.send(Status {
            .await
                        state: State::UpstreamTryReconnect(e),
            .unwrap_or(());
                    })
                        .await
                        .unwrap_or(());
                },
                _ => {
                    tx.send(Status {
                        state: State::UpstreamShutdown(e),
                    })
                        .await
                        .unwrap_or(());
                },
            }
        }
    }

**Comment by plebhash** [comment]
@lorbax `network_helpers_sv2` offers two flavors of channels: `async_std` and `tokio`: https://github.com/stratum-mining/stratum/tree/dev/roles/roles-utils/network-helpers/src

so I just tried to cover for both flavors

but feel free to only do `ChannelErrorReceiver` if you think that's overkill

**Comment by lorbax** [comment]
> @lorbax `network_helpers_sv2` offers two flavors of channels: `async_std` and `tokio`: https://github.com/stratum-mining/stratum/tree/dev/roles/roles-utils/network-helpers/src
> 
> so I just tried to cover for both flavors
> 
> but feel free to only do `ChannelErrorReceiver` if you think that's overkill

my idea is to introduce changes only when strictly needed in order to avoid unexpected behavior. What do you think about?
again, I stress also that we are using two types of channels, from tokio and from async_std. It seems a red flag to me, what do you think also about this?

**Comment by plebhash** [comment]
> > @lorbax `network_helpers_sv2` offers two flavors of channels: `async_std` and `tokio`: https://github.com/stratum-mining/stratum/tree/dev/roles/roles-utils/network-helpers/src
> > so I just tried to cover for both flavors
> > but feel free to only do `ChannelErrorReceiver` if you think that's overkill
> 
> my idea is to introduce changes only when strictly needed in order to avoid unexpected behavior. What do you think about? again, I stress also that we are using two types of channels, from tokio and from async_std. It seems a red flag to me, what do you think also about this?

yes those are very good points and I agree it's kind of a red flag... I'm not sure what are the historical motivations for having two different flavors of channels, but I think that when we start making [deeper refactoring into the application layer](https://github.com/stratum-mining/stratum/discussions/1069#discussioncomment-10290635) we should discuss the possibility of getting rid of some of the two kinds of channels in favor of standardization

but that's a discussion for the future

for now, on the context of this PR, I agree we should move forward only covering `ChannelErrorReceiver` on the `match` arms and just ignore the `TokioChannelErrorRecv`, since `tokio` channels are not even enabled on the `network_helpers_sv2` feature flags:

https://github.com/stratum-mining/stratum/blob/2541703f2f5371052110f740f7eab9409a10d0f1/roles/translator/Cargo.toml#L22

**Review by plebhash — APPROVED** [review]

File: roles/translator/src/main.rs [hunk 7]
```diff
@@ -54,22 +54,122 @@ async fn main() {
...
+            State::UpstreamShutdown(err) => {
+                error!("SHUTDOWN from: {}", err);
+                break;
+            }
+            State::UpstreamTryReconnect(err) => {
+                error!("SHUTDOWN from: {}", err);
+
+                // wait a random amount of time between 0 and 3000ms
+                // if all the downstreams try to reconnect at the same time, the upstream may fail
+                tokio::time::sleep(Duration::from_millis(1000)).await;
```
GitGab19, line 123: Isn't this line redundant?
You already wait a random time below

roles/translator/src/main.rs
```diff
@@ -54,22 +54,122 @@ async fn main() {
...
+                let mut rng = rand::thread_rng();
+                let wait_time = rng.gen_range(0..=3000);
+                tokio::time::sleep(Duration::from_millis(wait_time)).await;
+
+                // kill al the tasks
+                let task_collector_aborting = task_collector_.clone();
+                kill_tasks(task_collector_aborting.clone());
+
+                error!("Trying recconnecting to upstream");
```
GitGab19, line 132: Shouldn't be a warning?
Also, there's a typo: it should be "Trying reconnecting to upstream"

GitGab19 (commented):
Tested twice and it works as expected, good work @lorbax !
Just a few minor nits and it's ready to go

lorbax:
> > > @lorbax `network_helpers_sv2` offers two flavors of channels: `async_std` and `tokio`: https://github.com/stratum-mining/stratum/tree/dev/roles/roles-utils/network-helpers/src
> > > so I just tried to cover for both flavors
> > > but feel free to only do `ChannelErrorReceiver` if you think that's overkill
> > 
> > 
> > my idea is to introduce changes only when strictly needed in order to avoid unexpected behavior. What do you think about? again, I stress also that we are using two types of channels, from tokio and from async_std. It seems a red flag to me, what do you think also about this?
> 
> yes those are very good points and I agree it's kind of a red flag... I'm not sure what are the historical motivations for having two different flavors of channels, but I think that when we start making [deeper refactoring into the application layer](https://github.com/stratum-mining/stratum/discussions/1069#discussioncomment-10290635) we should discuss the possibility of getting rid of some of the two kinds of channels in favor of standardization
> 
> but that's a discussion for the future
> 
> for now, on the context of this PR, I agree we should move forward only covering `ChannelErrorReceiver` on the `match` arms and just ignore the `TokioChannelErrorRecv`, since `tokio` channels are not even enabled on the `network_helpers_sv2` feature flags:
> 
> https://github.com/stratum-mining/stratum/blob/2541703f2f5371052110f740f7eab9409a10d0f1/roles/translator/Cargo.toml#L22

with channel I mean channels that allows a task to send data to another task, like mpsc channels.
In the translator we are using async_std channels
https://docs.rs/async-std/latest/async_std/channel/fn.bounded.html
but the error variant also take in consideration the mpmc channel in tokio
https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html

These are not things that are exported by network-helpers (or at least it appears to me like this). Perhaps is a good idea to make it uniform and substitute the usage of  `async_std::channel` with `tokio::sync::mpsc::channel`. But in this case we must recall to change it also in `send_status` function

GitGab19 (approved):

lorbax:
> > > > @lorbax `network_helpers_sv2` offers two flavors of channels: `async_std` and `tokio`: https://github.com/stratum-mining/stratum/tree/dev/roles/roles-utils/network-helpers/src
> > > > so I just tried to cover for both flavors
> > > > but feel free to only do `ChannelErrorReceiver` if you think that's overkill
> > > 
> > > 
> > > my idea is to introduce changes only when strictly needed in order to avoid unexpected behavior. What do you think about? again, I stress also that we are using two types of channels, from tokio and from async_std. It seems a red flag to me, what do you think also about this?
> > 
> > 
> > yes those are very good points and I agree it's kind of a red flag... I'm not sure what are the historical motivations for having two different flavors of channels, but I think that when we start making [deeper refactoring into the application layer](https://github.com/stratum-mining/stratum/discussions/1069#discussioncomment-10290635) we should discuss the possibility of getting rid of some of the two kinds of channels in favor of standardization
> > but that's a discussion for the future
> > for now, on the context of this PR, I agree we should move forward only covering `ChannelErrorReceiver` on the `match` arms and just ignore the `TokioChannelErrorRecv`, since `tokio` channels are not even enabled on the `network_helpers_sv2` feature flags:
> > https://github.com/stratum-mining/stratum/blob/2541703f2f5371052110f740f7eab9409a10d0f1/roles/translator/Cargo.toml#L22
> 
> with channel I mean channels that allows a task to send data to another task, like mpsc channels. In the translator we are using async_std channels https://docs.rs/async-std/latest/async_std/channel/fn.bounded.html but the error variant also take in consideration the mpmc channel in tokio https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html
> 
> These are not things that are exported by network-helpers (or at least it appears to me like this). Perhaps is a good idea to make it uniform and substitute the usage of `async_std::channel` with `tokio::sync::mpsc::channel`. But in this case we must recall to change it also in `send_status` function

jbesraa:
@GitGab19 lets not merge unless we have all CI checks in the green please

PR #1007, opened 2024-06-26
Refactor ping pong encrypted example

plebhash:
similarly to https://github.com/stratum-mining/stratum/pull/977, this PR is refactoring the `examples/ping-pong-with-noise` example into `examples/ping-pong-encrypted`.

I wrote it from scratch as an exercise to get familiar with `protocols` crates.

The overall idea is quite similar to the previous implementation, but with some simplifications to make the example easier to understand.

---

`ping-pong-encrypted` is an example of how to do encrypted SV2 traffic while leveraging the following crates:
- [`binary_sv2`](http://docs.rs/binary_sv2)
- [`codec_sv2`](http://docs.rs/codec_sv2)
- [`framing_sv2`](http://docs.rs/framing_sv2) (which is actually just re-exported by `codec_sv2`)
- [`noise_sv2`](http://docs.rs/noise_sv2)

We establish a simple `Ping`-`Pong` protocol with a server and a client communicating over a TCP socket.

The server expects to receive a `Ping` message encoded as a SV2 binary frame.
The `Ping` message contains a `nonce`, which is a `u8` generated randomly by the client.

The client expects to get a `Pong` message in response, also encoded as a SV2 binary frame, with the same `nonce`.

The messages are assigned arbitrary values for binary encoding:
```rust
pub const PING_MSG_TYPE: u8 = 0xfe;
pub const PONG_MSG_TYPE: u8 = 0xff;
``` 

All communication is encrypted with [SV2 Noise Protocol](https://stratumprotocol.org/specification/04-Protocol-Security/).


Shourya742 (changes requested):
tACK e37608ee3afecc097ccdd43c499ae8fe88dc560b. There are just a few nits.

examples/ping-pong-encrypted/src/client.rs
```diff
@@ -0,0 +1,77 @@
...
+        Connection::new(stream, HandshakeRole::Initiator(initiator)).await?;
+
+    // create Ping message
+    let ping = Ping::new()?;
+    let ping_nonce = ping.get_nonce();
+    let message = Message::Ping(ping);
+
+    // create Ping frame
+    let ping_frame =
+        StandardSv2Frame::<Message>::from_message(message.clone(), PING_MSG_TYPE, 0, false)
```
marathon-gary, line 31: https://stratumprotocol.org/specification/09-Extensions/
Is a 0 extension type unique in some way?

marathon-gary (approved):
LGTM.

examples/ping-pong-encrypted/src/client.rs
plebhash: since there are no extensions defined yet, using `0` is just the void/null case

plebhash:
reverting back to draft, since it could be affected by #903 (and other similar issues that are output from #845 )

rrybarczyk:
Note: Wait to merge until the following is merged:
- #1049 
- #1033

plebhash:
> reverting back to draft, since it could be affected by #903 (and other similar issues that are output from #845 )

> Note: Wait to merge until the following is merged:
> 
> * [`Framing` refactor: test `Sv2Frame` and `HandShakeFrame` #1049](https://github.com/stratum-mining/stratum/pull/1049)
> * [`Framing` refactor: `framing.rs` api #1033](https://github.com/stratum-mining/stratum/pull/1033)

the tasks mentioned above have been de-prioritized

@Shourya742 is working to remove `async_std` support from SRI, and this PR helps achieve that goal

therefore, I'm removing the draft status so we can merge this soon

GitGab19 (approved):
tACK

Issue #484, opened 2023-03-09
mining-proxy needs graceful death instead of panic


bcral:
The mining-proxy seems to be panicking on closure.  This makes test coverage data impossible to retrieve with llvm-cov.

This same issue used to exist on the pool and TProxy until it was fixed by #432

bcral:
Please tag as "good first issue"!

Priceless-P:
@bcral i’d like to work on this.

Priceless-P:
@bcral Just to be sure, [this](https://github.com/stratum-mining/stratum/blob/main/roles/v2/mining-proxy/src/main.rs) is the file that needs to be worked on for this issue right?

darricksee:
I haven't and probably won't get to it soon.

darricksee:
Just a heads-up @Priceless-P  - to implement this the `status-channel` also needs to be added to the mining-proxy before getting to this issue. You can certainly add both with this issue - [take a look at how the `status-channel` is used in the pool role](https://github.com/stratum-mining/stratum/blob/main/roles/v2/pool/src/main.rs#L226). It's basically a mechanism for the application to signal the main thread to terminate if one of the tasks has an error.

Priceless-P:
Okay. Thank you for letting me know. @darricksee

SumantxD:
@bcral i would like to work on this issue, can you assign it to me.

Priceless-P:
Hi @SumantxD, I am currently working on it.

SumantxD:
@Priceless-P ok fine

mihir1739:
Hey is anyone working on this issue ? If not please assign this to me .

pavlenex:
Hey @mihir1739 it seems @Priceless-P is working on it here https://github.com/stratum-mining/stratum/pull/518

plebhash:
@Priceless-P what is the status on this? why was #518 closed without being merged?

Priceless-P:
@plebhash Yes.

johnnyasantoss:
Is this up for grabs? @Priceless-P

PR #1021, opened 2024-06-30
Graceful shutdown for mining-proxy

johnnyasantoss:
Add graceful shutdown for mining proxy fixing #484 

PS: I'm not sure if I should merge against dev or main

### DIFF
--- a/roles/mining-proxy/src/lib/downstream_mining.rs
+++ b/roles/mining-proxy/src/lib/downstream_mining.rs
@@ -1,7 +1,4 @@
-#![allow(dead_code)]
-
-use core::convert::TryInto;
-use std::sync::Arc;
+use std::{convert::TryInto, sync::Arc};
 
 use async_channel::{Receiver, SendError, Sender};
 use tokio::{net::TcpListener, sync::oneshot::Receiver as TokioReceiver};
@@ -39,7 +36,6 @@ pub struct DownstreamMiningNode {
     receiver: Receiver<EitherFrame>,
     sender: Sender<EitherFrame>,
     pub status: DownstreamMiningNodeStatus,
-    pub prev_job_id: Option<u32>,
     upstream: Option<Arc<Mutex<UpstreamMiningNode>>>,
 }
 
@@ -61,14 +57,6 @@ pub enum Channel {
     DownstreamHomUpstreamExtended {
         data: CommonDownstreamData,
         channel_id: u32,
-        group_id: u32,
-    },
-    // Below variant is not supported cause do not have much sense
-    // DowntreamNonHomUpstreamGroup { data: CommonDownstreamData, group_ids: Vec<u32>, extended_ids: Vec<u32>},
-    DowntreamNonHomUpstreamExtended {
-        data: CommonDownstreamData,
-        group_ids: Vec<u32>,
-        extended_ids: Vec<u32>,
     },
 }
 
@@ -119,42 +107,18 @@ impl DownstreamMiningNodeStatus {
         }
     }
 
-    fn open_channel_for_down_hom_up_extended(&mut self, channel_id: u32, group_id: u32) {
+    fn open_channel_for_down_hom_up_extended(&mut self, channel_id: u32, _group_id: u32) {
         match self {
             DownstreamMiningNodeStatus::Initializing => panic!(),
             DownstreamMiningNodeStatus::Paired(data) => {
                 let channel = Channel::DownstreamHomUpstreamExtended {
                     data: *data,
                     channel_id,
-                    group_id,
                 };
                 let self_ = Self::ChannelOpened(channel);
                 let _ = std::mem::replace(self, self_);
             }
             DownstreamMiningNodeStatus::ChannelOpened(..) => panic!("Channel already opened"),
-        }
-    }
-
-    fn add_extended_from_non_hom_for_up_extended(&mut self, id: u32) {
-        match self {
-            DownstreamMiningNodeStatus::Initializing => panic!(),
-            DownstreamMiningNodeStatus::Paired(data) => {
-                let channel = Channel::DowntreamNonHomUpstreamExtended {
-                    data: *data,
-                    group_ids: vec![],
-                    extended_ids: vec![id],
-                };
-                let self_ = Self::ChannelOpened(channel);
-                let _ = std::mem::replace(self, self_);
-            }
-            DownstreamMiningNodeStatus::ChannelOpened(
-                Channel::DowntreamNonHomUpstreamExtended { extended_ids, .. },
-            ) => {
-                if !extended_ids.contains(&id) {
-                    extended_ids.push(id)
-                }
-            }
-            _ => panic!(),
         }
     }
 }
@@ -179,16 +143,12 @@ impl DownstreamMiningNode {
         self.status
             .open_channel_for_down_hom_up_extended(channel_id, group_id);
     }
-    pub fn add_extended_from_non_hom_for_up_extended(&mut self, id: u32) {
-        self.status.add_extended_from_non_hom_for_up_extended(id);
-    }
 
     pub fn new(receiver: Receiver<EitherFrame>, sender: Sender<EitherFrame>, id: u32) -> Self {
         Self {
             receiver,
             sender,
             status: DownstreamMiningNodeStatus::Initializing,
-            prev_job_id: None,
             upstream: None,
             id,
         }
@@ -430,12 +390,6 @@ impl
                 let res = UpstreamMiningNode::handle_std_shr(remote.clone(), m).unwrap();
                 Ok(SendTo::Respond(res))
             }
-            DownstreamMiningNodeStatus::ChannelOpened(
-                Channel::DowntreamNonHomUpstreamExtended { .. },
-            ) => {
-                // unreachable cause the proxy do not support this kind of channel
-                unreachable!();
-            }
         }
     }
 
@@ -537,9 +491,6 @@ impl IsDownstream for DownstreamMiningNode {
                 data,
                 ..
             }) => data,
-            DownstreamMiningNodeStatus::ChannelOpened(
-                Channel::DowntreamNonHomUpstreamExtended { data, .. },
-            ) => data,
             DownstreamMiningNodeStatus::ChannelOpened(Channel::DownstreamHomUpstreamExtended {
                 data,
                 ..
--- a/roles/mining-proxy/src/lib/downstream_mining.rs
+++ b/roles/mining-proxy/src/lib/downstream_mining.rs
@@ -1,7 +1,14 @@
 #![allow(dead_code)]
 
-use super::upstream_mining::{StdFrame as UpstreamFrame, UpstreamMiningNode};
+use core::convert::TryInto;
+use std::sync::Arc;
+
 use async_channel::{Receiver, SendError, Sender};
+use tokio::{net::TcpListener, sync::oneshot::Receiver as TokioReceiver};
+use tracing::{info, warn};
+
+use codec_sv2::{StandardEitherFrame, StandardSv2Frame};
+use network_helpers_sv2::plain_connection_tokio::PlainConnection;
 use roles_logic_sv2::{
     common_messages_sv2::{SetupConnection, SetupConnectionSuccess},
     common_properties::{CommonDownstreamData, IsDownstream, IsMiningDownstream},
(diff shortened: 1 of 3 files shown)


roles/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -487,40 +487,49 @@ use network_helpers_sv2::plain_connection_tokio::PlainConnection;
...
-                node.clone(),
-                message_type,
-                payload,
-                routing_logic,
-            ) {
-                Ok(SendToCommon::RelayNewMessageToRemote(_, message)) => {
-                    let message = match message {
-                        roles_logic_sv2::parsers::CommonMessages::SetupConnectionSuccess(m) => m,
+    let mut should_continue = true;
+    while should_continue {
```
jbesraa, line 496: Could we use `loop` and `break` in the `Err` arm?

roles/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -487,40 +487,49 @@ use network_helpers_sv2::plain_connection_tokio::PlainConnection;
...
+                        }
                         _ => panic!(),
-                    };
-                    DownstreamMiningNode::start(node, message).await
-                }
-                _ => panic!(),
+                    }
+                });
+            },
+            _ = &mut shutdown_rx => {
```
jbesraa, line 528: Would be useful to log the error here

roles/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -487,40 +487,49 @@ use network_helpers_sv2::plain_connection_tokio::PlainConnection;
...
+                info!("Closing listener on {}", address);
```
jbesraa, line 529: s/info!/error!

roles/mining-proxy/src/main.rs
```diff
@@ -137,5 +138,18 @@ async fn main() {
...
     info!("PROXY INITIALIZED");
-    crate::lib::downstream_mining::listen_for_downstream_mining(socket).await
+
+    let (shutdown_tx, shutdown_rx) = oneshot::channel();
+
+    tokio::select! {
+        _ = lib::downstream_mining::listen_for_downstream_mining(socket, shutdown_rx) => {
+            warn!("Downstream mining listener exited prematurely");
+        },
+        _ = tokio::signal::ctrl_c() => {
```
jbesraa, line 148: Could we move this internally to `listen_for_downstream_minining` and avoid nesting `tokio::select!`?

jbesraa (commented):
Thanks for looking into this, added few small comments

roles/mining-proxy/src/lib/downstream_mining.rs
johnnyasantoss: I tried that first, but as the code that listens for the shutdown message is inside a macro, it can't break from there. At least I couldn't get it to compile with `break` from inside `select!`, not even using named loops

roles/mining-proxy/src/lib/downstream_mining.rs
johnnyasantoss: This is not an error but a message to stop accepting new connections and exit cleanly

roles/mining-proxy/src/lib/downstream_mining.rs
johnnyasantoss: This is the expected behavior, to stop and return from that function, leading to clean exit.

roles/mining-proxy/src/main.rs
johnnyasantoss: We could, but it seemed to me better to leave the `main` fn handling signal and other executable related stuff while leaving `listen_for_downstream_minining` cancellable but only focused on accepting connections.
Anyway, if you feel strongly about this, I'll change :)

plebhash:
@johnnyasantoss can you write some instructions on how to reproduce the original issue?

I still haven't been able to reproduce it

johnnyasantoss:
> @johnnyasantoss can you write some instructions on how to reproduce the original issue?
> 
> I still haven't been able to reproduce it

Try `cargo run -- -c config-examples/proxy-config-example.toml` followed by `echo $?`, and ctrl-c on both `dev` and on this branch

Ideally it should be `0` and not `130` so that's easier to script

plebhash (approved):

johnnyasantoss:
Should I keep rebasing or only rebase once there's enough reviews?

plebhash:
> Should I keep rebasing or only rebase once there are enough reviews?

this is tricky, we discussed it a bit on the last minutes of our call yesterday

basically, every PR that gets merged before this would mean you need to rebase it, which is very boring and doesn't really scale well

we established the following convention:
- if PR author needs to update the PR, or solve some conflict, they should obviously rebase it
- if PR is ready to be merged but it's not urgent, we wait a few days for PR author to rebase
- if PR is ready to be merged and it's urgent, some core dev with merging rights rebases it, leaves a comment for context (since now the core dev will co-author all commits in the PR), and merges it

pavlenex:
> > Should I keep rebasing or only rebase once there are enough reviews?
> 
> this is tricky, we discussed it a bit on the last minutes of our call yesterday
> 
> basically, every PR that gets merged before this would mean you need to rebase it, which is very boring and doesn't really scale well
> 
> we established the following convention:
> 
> * if PR author needs to update the PR, or solve some conflict, they should obviously rebase it
> * if PR is ready to be merged but it's not urgent, we wait a few days for PR author to rebase
> * if PR is ready to be merged and it's urgent, some core dev with merging rights rebases it, leaves a comment for context (since now the core dev will co-author all commits in the PR), and merges it

Should we add this to contributing.md ?

johnnyasantoss:
> > Should I keep rebasing or only rebase once there are enough reviews?
> 
> this is tricky, we discussed it a bit on the last minutes of our call yesterday
> 
> basically, every PR that gets merged before this would mean you need to rebase it, which is very boring and doesn't really scale well
> 
> we established the following convention:
> 
>     * if PR author needs to update the PR, or solve some conflict, they should obviously rebase it
> 
>     * if PR is ready to be merged but it's not urgent, we wait a few days for PR author to rebase
> 
>     * if PR is ready to be merged and it's urgent, some core dev with merging rights rebases it, leaves a comment for context (since now the core dev will co-author all commits in the PR), and merges it

I would also add about the commit strategy. Should we use `merge commits` + `squash merge PR` or `rebase` + `merge commit PR`?
Also, the PR reviewer can perform minor updates (like clicking in update branch here on GH) to the PR, in my humble opinion. That would speed up the maintainer's task of closing PRs.

jbesraa (commented):
I have been looking into the code a bit and it would be good to do a couple of things here: 
It would be very nice if you could add unit test to test function `start_mining_proxy`, that I think can look something like this: 
```rust
async fn start_mining_proxy(config: Config) {
    let group_id = Arc::new(Mutex::new(GroupId::new()));
    lib::ROUTING_LOGIC
        .set(Mutex::new(
            lib::initialize_r_logic(&config.upstreams, group_id, config.clone()).await,
        ))
        .expect("BUG: Failed to set ROUTING_LOGIC");
    info!("PROXY INITIALIZING");
    lib::initialize_upstreams(config.min_supported_version, config.max_supported_version).await;

    // Wait for downstream connection
    let socket = SocketAddr::new(
        config.listen_address.parse().unwrap(),
        config.listen_mining_port,
    );


    let (shutdown_tx, mut shutdown_rx) = oneshot::channel();
    let listener = TcpListener::bind(socket).await.unwrap();
    info!("Listening on {}", socket.to_string());

    tokio::select! {
        _ = lib::downstream_mining::listen_for_downstream_mining(listener) => {
            warn!("Downstream mining listener exited prematurely");
        },
        _ = &mut shutdown_rx => {
            info!("Closing listener on {}", socket.to_string());
            let _ = shutdown_tx.send(());
        }
        _ = tokio::signal::ctrl_c() => {
            let _ = shutdown_tx.send(());
            info!("Interrupt received");
        }
    }

    info!("Shutdown done");

}

```

Notice how I changed the `tokio::select` but I dont have strong opinion about it, but I do think we should try to avoid nested `select's`. 

This would also require to change `listen_for_downstream_mining` signature to accept `TcpListener` rather than `SocketAddr`. 

This way we separate the binary stuff(getting the config), and the actual service initiating function and we can test it more easily to verify your solution. 

If you think this is too much, I am happy to pick it up in a later PR.

Here is a commit with a bit of changes https://github.com/jbesraa/stratum/commit/8753830468ad01026d0c09eeb9e02a40554806a1

plebhash:
> > > Should I keep rebasing or only rebase once there are enough reviews?
> > 
> > 
> > this is tricky, we discussed it a bit on the last minutes of our call yesterday
> > basically, every PR that gets merged before this would mean you need to rebase it, which is very boring and doesn't really scale well
> > we established the following convention:
> > ```
> > * if PR author needs to update the PR, or solve some conflict, they should obviously rebase it
> > 
> > * if PR is ready to be merged but it's not urgent, we wait a few days for PR author to rebase
> > 
> > * if PR is ready to be merged and it's urgent, some core dev with merging rights rebases it, leaves a comment for context (since now the core dev will co-author all commits in the PR), and merges it
> > ```
> 
> I would also add about the commit strategy. Should we use `merge commits` + `squash merge PR` or `rebase` + `merge commit PR`? Also, the PR reviewer can perform minor updates (like clicking in update branch here on GH) to the PR, in my humble opinion. That would speed up the maintainer's task of closing PRs.

@marathon-gary started working on #1030 so we should move this discussion to that PR

jbesraa (commented):
cool! this looks much better. Ill create a followup pr to cover another couple of things in `mining-proxy` (like remove the `dead_code`, adding tests..)

could you please tide up the commit message?

johnnyasantoss:
> cool! this looks much better. Ill create a followup pr to cover another couple of things in `mining-proxy` (like remove the `dead_code`, adding tests..)
> 
> could you please tide up the commit message?

@jbesraa tide up? Is there a pattern that we must follow? Something like conventional commits? Asking because I didn't see any docs about this and this is my first PR here.

I still need to reformat and add the tests you mentioned. I was fighting tooling yesterday. Apparently, Rust has gotten a lot of tooling nowadays, so I need to configure properly my dev env.

jbesraa:
> > cool! this looks much better. Ill create a followup pr to cover another couple of things in `mining-proxy` (like remove the `dead_code`, adding tests..)
> > could you please tide up the commit message?
> 
> @jbesraa tide up? Is there a pattern that we must follow? Something like conventional commits? Asking because I didn't see any docs about this and this is my first PR here.
> 
> I still need to reformat and add the tests you mentioned. I was fighting tooling yesterday. Apparently, Rust has gotten a lot of tooling nowadays, so I need to configure properly my dev env.

Commit guidelines: https://cbea.ms/git-commit/

sounds good, feel free to bing me once its ready for review

jbesraa:
@johnnyasantoss could you please cherry-pick this commit https://github.com/jbesraa/stratum/commit/f91be81b129b3f2008d0168fc1599881094f40ce here?

johnnyasantoss:
@jbesraa the permission to write is enabled (you mentioned on the dev call)

See 
image

roles/mining-proxy/src/lib/downstream_mining.rs
```diff
@@ -483,44 +481,56 @@ impl
...
+    loop {
+        tokio::select! {
+            accept_result = listener.accept() => {
+                match accept_result {
+                    Ok((stream, _)) => {
+                let (receiver, sender): (Receiver<EitherFrame>, Sender<EitherFrame>) =
+                    PlainConnection::new(stream).await;
+                let node = DownstreamMiningNode::new(receiver, sender, ids.next());
+
+                task::spawn(async move {
```
jbesraa, line 453: I wonder if this `task::spawn` is needed here as we are already in a `select!`

jbesraa (approved):
I think we can merge this. Ill do a bit of testing with `mining-proxy` as part of #1066

jbesraa:
@johnnyasantoss This needs a rebase

roles/mining-proxy/src/lib/downstream_mining.rs
Shourya742: Nice catch! We can remove this `task::spawn` since the `tokio::select!` block already provides an asynchronous context for handling operations concurrently.

johnnyasantoss: I've removed `spawn` and simplified the code a bit

jbesraa (commented):
@johnnyasantoss Can you please organize your commits so they are grouped correctly? for example the `spawn` thing shouldn't be in its own commit, same for the typos. 
The first commits has refactor + small fixes + the actual solution, so I would organize that as well.  In this PR you can end-up with the following commits:

1 - all the typos/format/ordering imports(you can also break this into multiple commits if one of the changes is too big)
2 - code refactor(same, you can decide to break this)
3 - actual code for graceful shutdown

Could you also remove 2e48ac16576536d189b05a510de2e7670f5c8774 as we currently don't commit lock files

Please refer to this https://cbea.ms/git-commit/ for a better understanding to what we are trying to achieve in the commit history.

jbesraa (approved):

PR #1022, opened 2024-06-30
Redefining lastDeclareJob 

Shourya742:
The PR is opened to address the suggested changes by @Fi3 on PR #904. The comments are as follows:

- [Better management of LastDeclareJobs - no more wrong fallback-system activations #904 (comment)](https://github.com/stratum-mining/stratum/pull/904#discussion_r1609437598): Panic in safe lock state not advisable.
- [Better management of LastDeclareJobs - no more wrong fallback-system activations #904 (comment)](https://github.com/stratum-mining/stratum/pull/904#discussion_r1609469048): We are now using an array of size 2, as there should not be more than 2 Request id with corresponding jobs at any given point. If an anomaly arises, the process will error out.

### DIFF
--- a/roles/jd-client/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-client/src/lib/job_declarator/message_handler.rs
@@ -55,12 +55,20 @@ impl ParseServerJobDeclarationMessages for JobDeclarator {
     ) -> Result<SendTo, Error> {
         let tx_list = self
             .last_declare_mining_jobs_sent
-            .get(&message.request_id)
-            .unwrap()
-            .clone()
-            .unwrap()
-            .tx_list
-            .into_inner();
+            .iter()
+            .find_map(|entry| {
+                if let Some((id, last_declare_job)) = entry {
+                    if *id == message.request_id {
+                        Some(last_declare_job.clone().tx_list.into_inner())
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .ok_or_else(|| Error::UnknownRequestId(message.request_id))?;
+
         let unknown_tx_position_list: Vec<u16> = message.unknown_tx_position_list.into_inner();
         let missing_transactions: Vec<binary_sv2::B016M> = unknown_tx_position_list
             .iter()
--- a/roles/jd-client/src/lib/job_declarator/mod.rs
+++ b/roles/jd-client/src/lib/job_declarator/mod.rs
@@ -146,6 +146,10 @@ impl JobDeclarator {
             .unwrap()
     }
 
+    /// We maintain a window of 2 jobs. If more than 2 blocks are found,
+    /// the ordering will depend on the request ID. Only the 2 most recent request
+    /// IDs will be kept in memory, while the rest will be discarded.
+    /// More information can be found here: https://github.com/stratum-mining/stratum/pull/904#discussion_r1609469048
     fn update_last_declare_job_sent(
         self_mutex: &Arc<Mutex<Self>>,
         request_id: u32,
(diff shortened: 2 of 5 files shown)


GitGab19 (commented):
LGTM.
@Fi3 could you give a round of review as well since these changes were the ones you did request to me some months ago?

roles/jd-client/src/lib/job_declarator/message_handler.rs
```diff
@@ -55,12 +55,20 @@ impl ParseServerJobDeclarationMessages for JobDeclarator {
     ) -> Result<SendTo, Error> {
         let tx_list = self
             .last_declare_mining_jobs_sent
-            .get(&message.request_id)
-            .unwrap()
-            .clone()
-            .unwrap()
-            .tx_list
-            .into_inner();
+            .iter()
```
Fi3, line 58: I don't get why we should iter over the keys rather then use get

Fi3: btw not really important as last_declare_mining_jobs_sent is supposed to be very small

Fi3: a ok you changed the definition of itignore my prev comments

roles/jd-client/src/lib/job_declarator/mod.rs
```diff
@@ -55,7 +55,7 @@ pub struct JobDeclarator {
     req_ids: Id,
     min_extranonce_size: u16,
     // (Sent DeclareMiningJob, is future, template id, merkle path)
-    last_declare_mining_jobs_sent: HashMap<u32, Option<LastDeclareJob>>,
+    last_declare_mining_jobs_sent: [Option<(u32, LastDeclareJob)>; 2],
```
Fi3, line 58: Can you explain why is ok to have at max 2 last_declare_mining_job and at min 0. I would like to see in which cases we have 0, when 1 when 2 and most important why more then 2 is impossible

Shourya742: f the system detects more than 2 jobs, it should shut down to avoid potential problems. This helps prevent issues like downstream components sending shares for outdated jobs, which can lead to inefficiencies and financial losses. I'm not entirely certain about this, but we don't want miners wasting energy if something unusual occurs. I am still looking for concrete proof. This assertion is based on your response here: [GitHub Discussion](https://github.com/stratum-mining/stratum/pull/904#discussion_r1609469048).

Fi3: Can you add a comment in the code for it? It would be nice to state also way we can't have more then 2 jobs. (I don't remember why, if you can't figure out I can take a look)

Fi3: Maybe @GitGab19 have more context?

GitGab19: The context is the one already referenced. If JDC is using more than 2 jobs it means something wrong is happening. The edge case I'm thinking about is the one in which more than 2 blocks are found in some seconds. So new templates and new declared mining jobs are created by JDC. How this edge case is handled in this PR?

GitGab19: Perfect, so I think that you just need to put this as a comment in the code and this PR is ready to go

Shourya742: @GitGab19, I have added the comment.

Fi3 (approved):

GitGab19 (approved):

PR #1024, opened 2024-07-01
add `ActionResult::SustainConnection` to MG

### LINKED ISSUE #1023: MG action result that asserts connection was not dropped
MG currently has an action result that asserts that the connection was closed (`CloseConnection`)

https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/utils/message-generator/src/main.rs#L183-L195

However, it is also desirable to have an action result that asserts the opposite: the connection was **not** closed.

That would be particularly important to avoid #912 happening on future SV2 implementations.

plebhash:
close #1023 and unblock #1025 

this allows for MG to verify that a connection was not closed

it also refactors some parts of `Executor::execute` by moving some logic to the arms of `match result`, which improves code readability

### DIFF
--- a/utils/message-generator/src/executor.rs
+++ b/utils/message-generator/src/executor.rs
@@ -199,35 +199,21 @@ impl Executor {
                     result
                 );
 
-                // If the connection should drop at this point then let's just break the loop
-                // Can't do anything else after the connection drops.
-                if *result == ActionResult::CloseConnection {
-                    info!(
-                        "Waiting 1 sec to make sure that remote have time to close the connection"
-                    );
-                    tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
-                    recv.recv()
-                        .await
-                        .expect_err("Expecting the connection to be closed: wasn't");
-                    success = true;
-                    break;
-                }
-
-                let message = match recv.recv().await {
-                    Ok(message) => message,
-                    Err(_) => {
-                        success = false;
-                        error!("Connection closed before receiving the message");
-                        break;
-                    }
-                };
-
-                let mut message: Sv2Frame<AnyMessage<'static>, _> = message.try_into().unwrap();
-                debug!("RECV {:#?}", message);
-                let header = message.get_header().unwrap();
-                let payload = message.payload();
                 match result {
                     ActionResult::MatchMessageType(message_type) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
+                                success = false;
+                                error!("Connection closed before receiving the message");
+                                break;
+                            }
+                        };
+
+                        let message: Sv2Frame<AnyMessage<'static>, _> = message.try_into().unwrap();
+                        debug!("RECV {:#?}", message);
+                        let header = message.get_header().unwrap();
+
                         if header.msg_type() != *message_type {
                             error!(
                                 "WRONG MESSAGE TYPE expected: {} received: {}",
@@ -245,6 +231,20 @@ impl Executor {
                         message_type,
                         field_data, // Vec<(String, Sv2Type)>
                     )) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
+                                success = false;
+                                error!("Connection closed before receiving the message");
+                                break;
+                            }
+                        };
+
+                        let mut message: Sv2Frame<AnyMessage<'static>, _> =
+                            message.try_into().unwrap();
+                        debug!("RECV {:#?}", message);
+                        let header = message.get_header().unwrap();
+                        let payload = message.payload();
                         if subprotocol.as_str() == "CommonMessages" {
                             match (header.msg_type(), payload).try_into() {
                                 Ok(roles_logic_sv2::parsers::CommonMessages::SetupConnection(m)) => {
@@ -532,6 +532,20 @@ impl Executor {
                         message_type: _,
                         fields,
                     } => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
+                                success = false;
+                                error!("Connection closed before receiving the message");
+                                break;
+                            }
+                        };
+
+                        let mut message: Sv2Frame<AnyMessage<'static>, _> =
+                            message.try_into().unwrap();
+                        debug!("RECV {:#?}", message);
+                        let header = message.get_header().unwrap();
+                        let payload = message.payload();
                         if subprotocol.as_str() == "CommonMessages" {
                             match (header.msg_type(), payload).try_into() {
                                 Ok(parsers::CommonMessages::SetupConnection(m)) => {
@@ -730,6 +744,19 @@ impl Executor {
                         };
                     }
                     ActionResult::MatchMessageLen(message_len) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
+                                success = false;
+                                error!("Connection closed before receiving the message");
+                                break;
+                            }
+                        };
+
+                        let mut message: Sv2Frame<AnyMessage<'static>, _> =
+                            message.try_into().unwrap();
+                        debug!("RECV {:#?}", message);
+                        let payload = message.payload();
                         if payload.len() != *message_len {
                             error!(
                                 "WRONG MESSAGE len expected: {} received: {}",
@@ -741,6 +768,18 @@ impl Executor {
                         }
                     }
                     ActionResult::MatchExtensionType(ext_type) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
+                                success = false;
+                                error!("Connection closed before receiving the message");
+                                break;
+                            }
+                        };
+
+                        let message: Sv2Frame<AnyMessage<'static>, _> = message.try_into().unwrap();
+                        debug!("RECV {:#?}", message);
+                        let header = message.get_header().unwrap();
                         if header.ext_type() != *ext_type {
                             error!(
                                 "WRONG EXTENSION TYPE expected: {} received: {}",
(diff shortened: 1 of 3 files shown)


Fi3:
Can you add something about this new check in the readme? How and when to use it?

plebhash:
> Can you add something about this new check in the readme? How and when to use it?

I'm looking around but I cannot find any place where the other Action Results are described. Should we create a new section on [`utils/message-generator/README.md`](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md)?

Fi3:
> > Can you add something about this new check in the readme? How and when to use it?
> 
> I'm looking around but I cannot find any place where the other Action Results are described. Should we create a new section on [`utils/message-generator/README.md`](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md)?

yep I would add it in the [action section](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md#actions)

plebhash:
> > > Can you add something about this new check in the readme? How and when to use it?
> > 
> > 
> > I'm looking around but I cannot find any place where the other Action Results are described. Should we create a new section on [`utils/message-generator/README.md`](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md)?
> 
> yep I would add it in the [action section](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md#actions)

great, I added a new issue for this, since I don't think it makes sense to document all possible Action Results under this PR:

https://github.com/stratum-mining/stratum/issues/1027

I also wrote [a comment about rationale and usage of `SustainConnection`](https://github.com/stratum-mining/stratum/issues/1027#issuecomment-2203502658) there for future reference.

For now, @Fi3 would you mind carrying on with the review on this PR without those docs? It would be good to get it merged in order to unblock #1025

Fi3:
> > > > Can you add something about this new check in the readme? How and when to use it?
> > > 
> > > 
> > > I'm looking around but I cannot find any place where the other Action Results are described. Should we create a new section on [`utils/message-generator/README.md`](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md)?
> > 
> > 
> > yep I would add it in the [action section](https://github.com/stratum-mining/stratum/blob/main/utils/message-generator/README.md#actions)
> 
> great, I added a new issue #1027 for this, since I don't think it makes sense to document all possible Action Results under this PR.
> 
> I also wrote [a comment about rationale and usage of `SustainConnection`](https://github.com/stratum-mining/stratum/issues/1027#issuecomment-2203502658) there for future reference.
> 
> For now, @Fi3 would you mind carrying on with the review on this PR without those docs? It would be good to get it merged in order to unblock #1025

My only concern was adding something in readme about why we need it and how to use. Is ok with me to add it later. Everything else LGTM.

Fi3 (approved):

utils/message-generator/src/executor.rs
```diff
@@ -203,558 +203,600 @@ impl Executor {
...
                     success = true;
                     break;
-                }
 
-                let message = match recv.recv().await {
-                    Ok(message) => message,
-                    Err(_) => {
+                // Conversely, if the connection should sustain but the channel is closed,
+                // We also break since the test has failed
+                } else if *result == ActionResult::SustainConnection {
```
jbesraa, line 217: Why not have this and `CloseConnection` as part of the match in L241?

jbesraa (commented):
A bit tricky to review with all the things moving in `uitls/mg/src/executor.rs`, added one small comment

utils/message-generator/src/executor.rs
plebhash: I see your point!

I initially avoided this approach for the same reason I moved all those lines inside the `else` case

the original scenario that motivated this PR is one where the role being tested:
- receives a message
- does not close the connection (it was closing, which is the bug we're aiming to fix)
- does not send any message back in response

if we had the logic from this `else if` moved down to the `match` in L241, while unconditionally running `let message = match recv.recv().await`, the test would end up hanging on the scenario described above.

but now your suggestion inspired me to explore an alternative approach

I committed it on a different branch: https://github.com/plebhash/stratum/commit/e0164a8bc8c99cccc902063db57eab4770b385d5

there, I replicated the `recv.recv().await` on the different `match` arms where a response message is expected, and avoided it on those where it wouldn't necessarily happen, which also avoids the hanging situation

it does feel a bit cleaner, although with the tradeoff of some code repetition

if you prefer, I could bring this alternative approach over to this PR

jbesraa: new commit looks great, lets use it

plebhash: done

utils/message-generator/src/executor.rs
```diff
@@ -199,35 +199,21 @@ impl Executor {
...
-
-                let mut message: Sv2Frame<AnyMessage<'static>, _> = message.try_into().unwrap();
-                debug!("RECV {:#?}", message);
-                let header = message.get_header().unwrap();
-                let payload = message.payload();
                 match result {
                     ActionResult::MatchMessageType(message_type) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
```
jbesraa, line 206: We could use the error in the `error!` macro

utils/message-generator/src/executor.rs
```diff
@@ -245,6 +231,20 @@ impl Executor {
                         message_type,
                         field_data, // Vec<(String, Sv2Type)>
                     )) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
```
jbesraa, line 236: same as above

utils/message-generator/src/executor.rs
```diff
@@ -752,7 +791,26 @@ impl Executor {
...
+                            error!("Expected connection to close, but it didn't. Test failed.");
+                            success = false;
+                            break;
+                        }
+                    }
+                    ActionResult::SustainConnection => {
+                        info!(
+                            "Waiting 1 sec to make sure that remote has time to close the connection"
+                        );
+                        tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
```
jbesraa, line 808: you can use `from_secs`

utils/message-generator/src/executor.rs
```diff
@@ -752,7 +791,26 @@ impl Executor {
...
+                        if recv.is_closed() {
+                            error!("Expected connection to sustain, but it didn't. Test failed.");
```
jbesraa, line 810: whats the `Test failed` ?

utils/message-generator/src/executor.rs
```diff
@@ -741,6 +768,18 @@ impl Executor {
                         }
                     }
                     ActionResult::MatchExtensionType(ext_type) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
```
jbesraa, line 773: ^^

utils/message-generator/src/executor.rs
```diff
@@ -730,6 +744,19 @@ impl Executor {
                         };
                     }
                     ActionResult::MatchMessageLen(message_len) => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
```
jbesraa, line 749: ^^

utils/message-generator/src/executor.rs
```diff
@@ -532,6 +532,20 @@ impl Executor {
                         message_type: _,
                         fields,
                     } => {
+                        let message = match recv.recv().await {
+                            Ok(message) => message,
+                            Err(_) => {
```
jbesraa, line 537: ^^

jbesraa (commented):
Ah nice, this looks great. Added few small comments

utils/message-generator/src/executor.rs
plebhash: can you elaborate? I don't fully understand the suggestion

plebhash: I originally didn't understand the suggestion, but now I get where you are going, but I think you're missing some context.

If we check the [docs on `async_channel::Receiver::recv` method](https://docs.rs/async-channel/latest/async_channel/struct.Receiver.html#method.recv), we see that the only situation where this method returns an error is when the channel has been closed. And on this specific context, a closed channel can only mean one thing: the TCP connection was closed and no message was received, which is already being properly conveyed in the current error message.

If we print the actual value of `Err(e)` here we get: `receiving from an empty and closed channel`. Printing out this string into the logs wouldn't improve the UX. In fact, it would make things more confusing to the user, since there's no need for them to know that we're using `async_channel::Receiver` under the hood. With the current implementation, the user already gets all info they need to know.

I think we can leave `Err(_)` with no penalty on code quality.

utils/message-generator/src/executor.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1024#discussion_r1666094287

utils/message-generator/src/executor.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1024#discussion_r1666094287

utils/message-generator/src/executor.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1024#discussion_r1666094287

utils/message-generator/src/executor.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1024#discussion_r1666094287

rrybarczyk (approved):
Lgtm. Definitely more readable.

utils/message-generator/src/executor.rs
jbesraa: Couldnt the error explain why we are 'receiving from an empty and closed channel' ?

General q about this MG executor, is it only used in tests or  also somewhere else?

plebhash: > General q about this MG executor, is it only used in tests or also somewhere else?

The entire implementation of MG (`utils/message-generator/*`) is only used in MG tests, and no other circumstance. 

Every `*.rs` file exists with the [purpose of being called by `utils/message-generator/src/main.rs` when MG is being run](https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/utils/message-generator/src/main.rs#L419), and that includes both `utils/message-generator/src/executor.rs` + `utils/message-generator/src/executor_sv1.rs`.

There's some small nuance with regards to real tests vs mocks. Each mock means that a new MG child process (with its own executor) is created and managed by the root executor (via `cargo run`). See [this](https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/test/message-generator/test/jds-do-not-fail-on-wrong-tsdatasucc/jds-do-not-fail-on-wrong-tsdatasucc.json#L42) + [this](https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/utils/message-generator/src/executor.rs#L54) as an example.

> Couldnt the error explain why we are 'receiving from an empty and closed channel' ?

If `executor.rs` was meant to be exposed as an API for external consumption in some specific use-cases, maybe that would make sense, since we would need to bubble-up the error message to whatever code was consuming this.

That is not the case however. The only scenario where this `error!` macro provides some meaningful information is when the MG test is being executed. And the current information already provides everything the user needs to know.

jbesraa (approved):

PR #1025, opened 2024-07-01
avoid JDS dropping connection on `SubmitSolution` with missing txs 

### LINKED ISSUE #912: JDS can't submit block if some transactions are missing
As you can see from the logs attached here, this is the messages flow which causes the issue:
- `NewTemplate` received by JDC
- `RequestTransactionData` from JDC to TP
- `RequestTransactionDataSuccess` from TP to JDC
- `DeclareMiningJob` from JDC to JDS
- `ProvideMissingTransactions` from JDS to JDC
- In the meantime that JDC is answering with `ProvideMissingTransactionsSuccess` a block is found, so there's a `SubmitSolution` from JDC to JDS
- JDS doesn't have txs that were asked to JDC through `ProvideMissingTransaction` message, so errors are generated

JDS logs:
Screenshot 2024-05-17 at 12 13 18

JDC logs:
Screenshot 2024-05-17 at 12 13 46

The big problem here is that connection between JDS and JDC is closed in this case, so JDC stops working properly after that.
We should manage this situation better, avoiding the connection to be closed.

plebhash:
this PR:
- fixes the bug described in #912
- adds a MG test to assert the correct JDS behavior under those specific conditions

### DIFF
--- a/roles/jd-server/src/lib/job_declarator/mod.rs
+++ b/roles/jd-server/src/lib/job_declarator/mod.rs
@@ -280,23 +280,18 @@ impl JobDeclaratorDownstream {
                                         match Self::collect_txs_in_job(self_mutex.clone()) {
                                             Ok(_) => {
                                                 info!("All transactions in downstream job are recognized correctly by the JD Server");
-                                                let hexdata =
-                                                    match JobDeclaratorDownstream::get_block_hex(
-                                                        self_mutex.clone(),
-                                                        message,
-                                                    ) {
-                                                        Ok(inner) => inner,
-                                                        Err(e) => {
-                                                            error!(
-                                                                "Received solution but encountered error: {:?}",
-                                                                e
-                                                            );
-                                                            recv.close();
-                                                            //TODO should we brake it?
-                                                            break;
-                                                        }
-                                                    };
-                                                let _ = new_block_sender.send(hexdata).await;
+                                                match JobDeclaratorDownstream::get_block_hex(
+                                                    self_mutex.clone(),
+                                                    message,
+                                                ) {
+                                                    Ok(hexdata) => {
+                                                        let _ =
+                                                            new_block_sender.send(hexdata).await;
+                                                    }
+                                                    Err(e) => {
+                                                        handle_result!(tx_status, Err(*e))
+                                                    }
+                                                };
                                             }
                                             Err(error) => {
                                                 error!("Missing transactions: {:?}", error);
@@ -316,22 +311,20 @@ impl JobDeclaratorDownstream {
                                                     .unwrap();
                                                 tokio::select! {
                                                     _ = JDsMempool::add_tx_data_to_mempool(mempool, retrieve_transactions) => {
-                                                        let hexdata = match JobDeclaratorDownstream::get_block_hex(
+                                                        match JobDeclaratorDownstream::get_block_hex(
                                                             self_mutex.clone(),
                                                             message.clone(),
                                                         ) {
-                                                            Ok(inner) => inner,
+                                                            Ok(hexdata) => {
+                                                                let _ = new_block_sender.send(hexdata).await;
+                                                            },
                                                             Err(e) => {
-                                                                error!(
-                                                                    "Error retrieving transactions: {:?}",
-                                                                    e
+                                                                handle_result!(
+                                                                    tx_status,
+                                                                    Err(*e)
                                                                 );
-                                                                recv.close();
-                                                                //TODO should we brake it?
-                                                                break;
                                                             }
                                                         };
-                                                        let _ = new_block_sender.send(hexdata).await;
                                                     }
                                                     _ = tokio::time::sleep(Duration::from_secs(60)) => {}
                                                 };
--- a/test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
+++ b/test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
@@ -1,0 +1,146 @@
+{
+  "version": "2",
+  "doc": [
+    "This test does",
+    "Mock JDC",
+    "Send SetupConnection and await for SetupConnection.Success",
+    "Send AllocateMiningJobToken0 and await for AllocateMiningJobToken.Success0",
+    "Send AllocateMiningJobToken1 and await for AllocateMiningJobToken.Success1",
+    "Send DeclareMiningJob and await for ProvideMissingTransactions",
+    "Send SubmitSolution and assert that connection was sustained"
+  ],
+  "job_declaration_messages": [
+    {
+      "message": {
+        "type": "AllocateMiningJobToken",
+        "user_identifier": "",
+        "request_id": 0,
+        "coinbase_tx_outputs": []
+      },
+      "id": "allocate_mining_job_token0"
+    },
+    {
+      "message": {
+        "type": "AllocateMiningJobToken",
+        "user_identifier": "",
+        "request_id": 1,
+        "coinbase_tx_outputs": []
+      },
+      "id": "allocate_mining_job_token1"
+    },
+    {
+      "message": {
+        "type": "DeclareMiningJob",
+        "request_id": 0,
+        "mining_job_token": [1, 0, 0, 0],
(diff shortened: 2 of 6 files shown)


jbesraa:
^^

jbesraa (approved):
Looks good. Would be good if someone with more knowledge about MG tests than myself can chime in as well.

test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
```diff
@@ -0,0 +1,121 @@
...
+      "message_ids": ["allocate_mining_job_token"],
+      "role": "client",
+      "results": [
+        {
+          "type": "match_message_type",
+          "value": "0x51"
+        }
+      ],
+      "actiondoc": "This action sends AllocateMiningJobToken and awaits for a AllocateMiningJobToken.Success"
+    },
```
Shourya742, line 89: Shouldn't we add two blocks of 'allocate mining job token' to comply with the standard? This would prevent confusion for anyone using this test as a reference in the future.

test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
```diff
@@ -0,0 +1,121 @@
+{
+  "version": "2",
+  "doc": [
+    "This test does",
+    "Mock JDC",
+    "Send SetupConnection and await for SetupConnection.Success",
+    "Send AllocateMiningJobToken and await for AllocateMiningJobToken.Success",
+    "Send DeclareMiningJob and await for ProvideMissingTransactions",
+    "Send SubmitSolution and xxx"
```
Shourya742, line 9: Can we provide more detail here?

Shourya742 (changes requested):
ACK d4510a5dde5f4ebd772311987cca830e2c6e7e90. Just few nits

test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
plebhash: oops this was left over from the drafting phase, thanks for catching that up

already updated!

test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
plebhash: added

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -280,23 +280,21 @@ impl JobDeclaratorDownstream {
...
                                                 info!("All transactions in downstream job are recognized correctly by the JD Server");
-                                                let hexdata =
-                                                    match JobDeclaratorDownstream::get_block_hex(
-                                                        self_mutex.clone(),
-                                                        message,
-                                                    ) {
-                                                        Ok(inner) => inner,
-                                                        Err(e) => {
-                                                            error!(
+                                                match JobDeclaratorDownstream::get_block_hex(
```
jbesraa, line 283: Inside `get_block_hex` we call `collect_txs_in_job`(again, after the call in L280) which uses the mutex heavily, that could be something to improve in a future commit

jbesraa (approved):

roles/jd-server/src/lib/job_declarator/mod.rs
plebhash: hmm I see your point but I don't think there's an easy way out of this

in theory we could do something like this:
```diff
    fn get_block_hex(
        self_mutex: Arc<Mutex<Self>>,
+       transactions_list: Vec<bitcoin::Transaction>
        message: SubmitSolutionJd,
    ) -> Result<String, Box<JdsError>> {
        let (last_declare_, _, _) = self_mutex
            .clone()
            .safe_lock(|x| x.declared_mining_job.clone())
            .map_err(|e| Box::new(JdsError::PoisonLock(e.to_string())))?;
        let last_declare = last_declare_.ok_or(Box::new(JdsError::NoLastDeclaredJob))?;
-         let transactions_list = Self::collect_txs_in_job(self_mutex)?;
        let block: Block =
            roles_logic_sv2::utils::BlockCreator::new(last_declare, transactions_list, message)
                .into();
        Ok(hex::encode(serialize(&block)))
    }
```

and then just take advantage of the fact that this call to `get_block_hex` happens on this context:
```diff
                                        match Self::collect_txs_in_job(self_mutex.clone()) {
-                                             Ok(_) => {
+                                             Ok(transactions_list) => {
                                                info!("All transactions in downstream job are recognized correctly by the JD Server");
                                                match JobDeclaratorDownstream::get_block_hex(
                                                    self_mutex.clone(),
+                                                   transactions_list
                                                    message,
                                                ) {
```

the problem is that there's yet another call to `get_block_hex` here:
https://github.com/stratum-mining/stratum/blob/5d569bbdb0f6d1dc7bc97a62abff9d6ded4b425d/roles/jd-server/src/lib/job_declarator/mod.rs#L319

and if we applied the changes above to `get_block_hex`, then we would need an extra call to `collect_txs_in_job` here, which brings us back to the original point

plebhash: I do agree this is meaningful technical debt, but I'm afraid that mitigating it on this PR would snowball into a massive change that would probably fall out of scope

unless we can find a simple fix, it should become something to be addressed when we move onto refactoring the application layer (which should only be our focus after we're done with the low-level/protocols layer)

test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
```diff
@@ -0,0 +1,146 @@
+{
+  "version": "2",
+  "doc": [
+    "This test does",
+    "Mock JDC",
+    "Send SetupConnection and await for SetupConnection.Success",
+    "Send AllocateMiningJobToken0 and await for AllocateMiningJobToken.Success0",
+    "Send AllocateMiningJobToken1 and await for AllocateMiningJobToken.Success1",
+    "Send DeclareMiningJob and await for ProvideMissingTransactions",
+    "Send SubmitSolution and assert that connection was sustained"
```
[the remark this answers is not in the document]
plebhash: hmm maybe the mock JDC could send a `ProvideMissingTransactions.Success` and expect a `DeclareMiningJob.Success`

I'm not 100% sure now, I'll have to check this again later

roles/jd-server/src/lib/job_declarator/mod.rs
jbesraa: Yea, I agree that it should be handled in a different PR

test/message-generator/mock/jdc-mock-jds-receive-solution-while-processing-declared-job.json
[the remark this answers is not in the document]
plebhash: during the call we agreed this is not necessary

lorbax:
Did you consider to remove a placeholder for the inner error her?
https://github.com/stratum-mining/stratum/blob/dev/roles/jd-server/src/lib/status.rs#L122
If you put something like this
```rust
        JdsError::MempoolError(error_inner) => {
            match error_inner {
                EmptyMempool => {...}
                NoClient => {...}
                Rpc(RpcError) => {...}
                PoisonLock(String) => {...}
            }
        }
```
and return 
```rust
 send_status(sender, e, error_handling::ErrorBranch::Break).await
 ```
 or 
 ```rust
  send_status(sender, e, error_handling::ErrorBranch::continue).await
 ```
 according on when you want the JDS to continue or break

jbesraa (approved):
Nice, I learned about `handle_result`. 

I think 3rd commit should be squashed into the first one

Fi3:
> Nice, I learned about `handle_result`.
> 
> I think 3rd commit should be squashed into the first one

I think that in the long term we should get rid of it. I usually prefer handling the task as close as possible from where I start it, I also think that for production code abort the task on dropping is better, since it easier to understand if there is a bug

GitGab19 (approved):

PR #1026, opened 2024-07-02
MG Test on quick templates

### LINKED ISSUE #923: Add MG test for false fallback activation
This is more a reminder for me than an issue, but I preferred to open an issue to better track it.
We need to add a MG test for specific scenario described in issue #901, addressing suggestions given under PR #904, that has been merged yesterday.

Shourya742:
Closes #923

Work in Progress:
1. Added JDS mock
2. Added TP mock
3. Added TProxy mock

Current State:
The process currently encounters an issue where it gets stuck at the declareMiningJob step. Debugging in progress.


test/message-generator/mock/template-provider-mock-quick-template.json
```diff
@@ -0,0 +1,215 @@
...
+        {
+            "message_ids": [],
+            "role": "server",
+            "results": [
+                {
+                    "type": "match_message_type",
+                    "value": "0x70"
+                }
+            ],
+          "actiondoc": "This action checks that a message SetCoinbaseDataSize is received"
```
plebhash, line 155:
```suggestion
          "actiondoc": "This action checks that a message CoinbaseOutputDataSize is received"
```

test/message-generator/mock/template-provider-mock-quick-template.json
```diff
@@ -0,0 +1,215 @@
...
+                    "value": "0x00"
+                }
+            ],
+            "actiondoc": "(pool) This action checks that a SetupConnection message is received"
+        },
+        {
+            "message_ids": ["setup_connection_success_template_distribution"],
+            "role": "server",
+            "results": [],
+            "actiondoc": "(pool) This action sends SetupConnection.Success with flag 2 (TD protocol)"
```
plebhash, line 144: the mock TP should expect 3 `SetupConnection` messages:
- pool
- mock JDS
- JDC

currently, it is only receiving one, and then moving onwards to the next messages

as a consequence, templates are sent downstream too fast, even before JDC has even started

test/message-generator/mock/template-provider-mock-quick-template.json
```diff
@@ -0,0 +1,215 @@
...
+        {
+            "message_ids": [],
+            "role": "server",
+            "results": [
+                {
+                    "type": "match_message_type",
+                    "value": "0x73"
+                }
+            ],
+            "actiondoc": "This action checks RequestTransactionData received"
```
plebhash, line 172: as mentioned above, the mock TP sends `NewTemplate` and `SetNewPrevHash` before the JDC has had the time to bootstrap and establish a connection

as a consequence, JDC is never aware of neither of these messages, and it never really sends a `RequestTransactionData`, so this action keeps hanging and the mock TP doesn't progress

test/message-generator/mock/jds-mock-quick-templates.json
```diff
@@ -0,0 +1,172 @@
...
+                }
+            ],
+            "actiondoc": "This action receives allocate mining job"
+        },
+        {
+            "message_ids": ["allocate_mining_job_token_success_1"],
+            "role": "server",
+            "results": [],
+            "actiondoc": "This action sends back a AllocateMiningJobToken.Success (allocate_mining_job_token_success_1)"
+        }
```
plebhash, line 152: I see there's still no action for receiving a `DeclareMiningJob`

you probably declared it locally, but you noticed the message wasn't arriving and you never committed it

please see the comments on TP mock, where I explain a possible root cause

test/message-generator/mock/tproxy-mock-quick-template.json
plebhash, line 1: you probably started this file in preparation to the fact that JDC will need to receive a share from downstream before it can send it to the pool

however, I would call it `downstream-mock*` instead of `tproxy-mock`, because this share could come from not only from a tproxy, but also:
- some ASIC
- some SV2 proxy

so using a `downstream` terminology is more generic and doesn't impose any assumptions on where the share is coming from

test/message-generator/mock/jds-mock-quick-templates.json
Shourya742: Exactly, I removed it manually... I will revert back once the TP stuff resolves..

test/message-generator/mock/template-provider-mock-quick-template.json
Shourya742: I am now spawning another TP at port 8443 to connect to JDC. Using the same file to connect to both the Pool and JDC doesn't seem to work. The overall flow looks good; only a few minor issues remain.

test/message-generator/mock/tproxy-mock-quick-template.json
Shourya742: Makes sense. Modified

rrybarczyk:
Closing as MG test does not solve for this, we need integration tests for this (#1077).

PR #1033, opened 2024-07-04
`Framing` refactor: `framing.rs` api

### LINKED ISSUE #903: refactor: `framing_sv2` crate

jbesraa:
`framing-sv2` Crate refactor p3, previous prs #982 #976 
Partially resolves #903 
In this pr I started working on the exposed API from `framing.rs` by handling the errors in a nicer way and also be more consistent with the naming

### DIFF
--- a/protocols/v2/framing-sv2/src/framing.rs
+++ b/protocols/v2/framing-sv2/src/framing.rs
@@ -82,12 +82,11 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// This function is only intended as a fast way to get a reference to an
     /// already serialized payload. If the frame has not yet been
     /// serialized, this function should never be used (it will panic).
-    pub fn payload(&mut self) -> &mut [u8] {
+    pub fn payload(&mut self) -> Option<&mut [u8]> {
         if let Some(serialized) = self.serialized.as_mut() {
-            &mut serialized.as_mut()[Header::SIZE..]
+            Some(&mut serialized.as_mut()[Header::SIZE..])
         } else {
-            // panic here is the expected behaviour
-            panic!("Sv2Frame is not yet serialized.")
+            None
         }
     }
 
--- a/protocols/v2/framing-sv2/src/framing.rs
+++ b/protocols/v2/framing-sv2/src/framing.rs
@@ -44,11 +44,9 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> From<Sv2Frame<T, B>> 
 
 /// Abstraction for a SV2 Frame.
 #[derive(Debug, Clone)]
-pub struct Sv2Frame<T, B> {
-    header: Header,
-    payload: Option<T>,
-    /// Serialized header + payload
-    serialized: Option<B>,
+pub enum Sv2Frame<T, B> {
+    Payload { header: Header, payload: T },
+    Raw { header: Header, serialized: B },
 }
 
 impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
@@ -57,23 +55,23 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// When called on a non serialized frame, it is not so cheap (because it serializes it).
     #[inline]
     pub fn serialize(self, dst: &mut [u8]) -> Result<(), Error> {
-        if let Some(mut serialized) = self.serialized {
-            dst.swap_with_slice(serialized.as_mut());
-            Ok(())
-        } else if let Some(payload) = self.payload {
-            #[cfg(not(feature = "with_serde"))]
-            to_writer(self.header, dst).map_err(Error::BinarySv2Error)?;
-            #[cfg(not(feature = "with_serde"))]
-            to_writer(payload, &mut dst[Header::SIZE..]).map_err(Error::BinarySv2Error)?;
-            #[cfg(feature = "with_serde")]
-            to_writer(&self.header, dst.as_mut()).map_err(Error::BinarySv2Error)?;
-            #[cfg(feature = "with_serde")]
-            to_writer(&payload, &mut dst.as_mut()[Header::SIZE..])
-                .map_err(Error::BinarySv2Error)?;
-            Ok(())
-        } else {
-            // Sv2Frame always has a payload or a serialized payload
-            panic!("Impossible state")
+        match self {
+            Sv2Frame::Raw { mut serialized, .. } => {
+                dst.swap_with_slice(serialized.as_mut());
+                Ok(())
+            }
+            Sv2Frame::Payload { header, payload } => {
+                #[cfg(not(feature = "with_serde"))]
+                to_writer(header, dst).map_err(Error::BinarySv2Error)?;
+                #[cfg(not(feature = "with_serde"))]
+                to_writer(payload, &mut dst[Header::SIZE..]).map_err(Error::BinarySv2Error)?;
+                #[cfg(feature = "with_serde")]
+                to_writer(&header, dst.as_mut()).map_err(Error::BinarySv2Error)?;
+                #[cfg(feature = "with_serde")]
+                to_writer(&payload, &mut dst.as_mut()[Header::SIZE..])
+                    .map_err(Error::BinarySv2Error)?;
+                Ok(())
+            }
         }
     }
 
@@ -83,16 +81,18 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// already serialized payload. If the frame has not yet been
     /// serialized, this function should never be used (it will panic).
     pub fn payload(&mut self) -> Option<&mut [u8]> {
-        if let Some(serialized) = self.serialized.as_mut() {
-            Some(&mut serialized.as_mut()[Header::SIZE..])
-        } else {
-            None
+        match self {
+            Sv2Frame::Raw { serialized, .. } => Some(&mut serialized.as_mut()[Header::SIZE..]),
+            Sv2Frame::Payload { .. } => None,
         }
     }
 
     /// `Sv2Frame` always returns `Some(self.header)`.
     pub fn header(&self) -> crate::header::Header {
-        self.header
+        match self {
+            Self::Payload { header, .. } => *header,
+            Self::Raw { header, .. } => *header,
+        }
     }
 
     /// Tries to build a `Sv2Frame` from raw bytes, assuming they represent a serialized `Sv2Frame` frame (`Self.serialized`).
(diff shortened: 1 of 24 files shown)


protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -1,57 +1,56 @@
...
-        Sv2Frame {
-            header,
-            payload,
-            serialized,
+/// A wrapper to be used in a context we need a generic reference to a frame
+/// but it doesn't matter which kind of frame it is (`Sv2Frame` or `HandShakeFrame`)
+#[derive(Debug)]
+pub enum Frame<T, B>
+where
+    T: Serialize + GetSize,
```
Fi3, line 17: Best practice is to constrain generics only where you need it.

Fi3: In the impl block where you have function that need the generics to be constrained

jbesraa: Having naked generics makes it really hard to read the enum

Fi3: I don't get it. Why is harder?

Fi3: Maybe you want to use InnerType and Buffer instead of T and B ?

Fi3: Btw we shouldn't restrain generics if there is no need

jbesraa: previously it was only `Frame<T,B>{ HandShake(HandShakeFrame), Sv2(Sv2Frame<T,B>)` without any info about T and B

roles/jd-client/src/lib/upstream_sv2/upstream.rs
```diff
@@ -231,13 +231,14 @@ impl Upstream {
...
         // Gets the binary frame message type from the message header
-        let message_type = if let Some(header) = incoming.get_header() {
-            header.msg_type()
-        } else {
-            return Err(framing_sv2::Error::ExpectedHandshakeFrame.into());
-        };
+        let message_type = incoming.header().msg_type();
         // Gets the message payload
-        let payload = incoming.payload();
+        let payload = match incoming.payload() {
```
Fi3, line 236: This seems like a place where clippy would complain to use if let Some . But if clippy is happy I'm as well. Maybe since you are at it you could just remove the above comment.

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -200,11 +200,11 @@ impl Downstream {
...
     pub async fn next(self_mutex: Arc<Mutex<Self>>, mut incoming: StdFrame) -> PoolResult<()> {
-        let message_type = incoming
-            .get_header()
-            .ok_or_else(|| PoolError::Custom(String::from("No header set")))?
-            .msg_type();
-        let payload = incoming.payload();
+        let message_type = incoming.header().msg_type();
+        let payload = match incoming.payload() {
+            Some(p) => p,
+            None => return Err(PoolError::Custom(String::from("No payload set"))),
```
Fi3, line 206: just out of curiosity. Why here you use String::from and in other places you use .to_string() . From what I concern you can just resolve this conversation and ignore my comment.

Fi3: Also here I would just have use expect since if there is an issue we are defenetly in an unrecoverable situation. But more then ok then merging it like that

protocols/v2/framing-sv2/src/framing.rs
Fi3: I went trough the code everything look very good to me.

roles/pool/src/lib/mining_pool/mod.rs
jbesraa: tbh everywhere outside the `framing` crate I was just following whatever is already their with minimum effort to improve/change it because otherwise those prs become to big to review

plebhash:
@jbesraa it seems we're breaking `sv2-ffi`

jbesraa:
oops missed this, it should be fixed now @plebhash

plebhash:
@jbesraa FYI I solved some conflicts here, which is why I'm co-authoring the commits

protocols/v2/framing-sv2/src/framing.rs
plebhash: what is the outcome here?

@Fi3 do you still feel need we shouldn't be constraining generics here? or looking at the rest of the PR change your mind?

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -44,11 +44,9 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> From<Sv2Frame<T, B>>
 
 /// Abstraction for a SV2 Frame.
```
plebhash, line 45: we should expand this comment to make it clear what each `enum` variant is

something like:
```
/// Both `Payload` and `Raw` variants carry the `Header` as metadata
/// `Payload` means the `Sv2Frame` carries a payload, but it has not yet been serialized
/// `Raw` means the `Sv2Frame` has been fully serialized, where the `serialized` field contains both `header` + `payload`
```

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -57,23 +55,23 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// When called on a non serialized frame, it is not so cheap (because it serializes it).
```
plebhash, line 55: we should update these comments so that they are coherent with the new behavior of this function

updating the comments is especially important given that these are used on the Rust Docs, so they need to follow these changes accordingly

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -83,16 +81,18 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// already serialized payload. If the frame has not yet been
```
plebhash, line 81: we should update these comments so that they are coherent with the new behavior of this function

updating the comments is especially important given that these are used on the Rust Docs, so they need to follow these changes accordingly

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -83,16 +81,18 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
...
     /// serialized, this function should never be used (it will panic).
```
plebhash, line 82: we should update these comments so that they are coherent with the new behavior of this function

updating the comments is especially important given that these are used on the Rust Docs, so they need to follow these changes accordingly

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -91,8 +91,8 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     }
 
     /// `Sv2Frame` always returns `Some(self.header)`.
```
plebhash, line 90: we should update these comments so that they are coherent with the new behavior of this function

updating the comments is especially important given that these are used on the Rust Docs, so they need to follow these changes accordingly

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -113,10 +113,9 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     pub fn from_bytes_unchecked(mut bytes: B) -> Self {
```
plebhash, line 113: Github doesn't allow me to add this comment on the lines above. Ideally this comment should be placed on line 98, but placing it here should be sufficient

the comments on `from_bytes` are outdated with regards to the new `enum`-based `Sv2Frame`

we should update these comments so that they are coherent with the new behavior of this function

updating the comments is especially important given that these are used on the Rust Docs, so they need to follow these changes accordingly

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -150,13 +149,9 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// otherwise, returns the length of `self.payload`.
```
plebhash, line 149: we should update these comments so that they are coherent with the new behavior of this function

updating the comments is especially important given that these are used on the Rust Docs, so they need to follow these changes accordingly

plebhash (changes requested):
strong ACK on converting `Sv2Frame` from a `struct` into an `enum`

the previous APIs around the old `struct` were very confusing for the user, and also had some panics on edge cases, which IMHO is not the best way to design an API

so this is making things much better!

however, we need to also update all comments accordingly, otherwise we will be generating outdated Rust Docs

protocols/v2/framing-sv2/src/framing.rs
```diff
@@ -83,16 +81,18 @@ impl<T: Serialize + GetSize, B: AsMut<[u8]> + AsRef<[u8]>> Sv2Frame<T, B> {
     /// already serialized payload. If the frame has not yet been
     /// serialized, this function should never be used (it will panic).
     pub fn payload(&mut self) -> Option<&mut [u8]> {
```
plebhash, line 83: shouldn't we rename this method?

on this commit we are sort of establishing a convention where the `Payload` variant signals that `Sv2Frame` has not yet been serialized... so calling a method called `payload` for a `Sv2Frame::Payload` and getting a `None` is extremely counter intuitive!

I feel we would create a more sane user experience if this method was called `raw` or `serialized`

jbesraa:
> @jbesraa FYI I solved some conflicts here, which is why I'm co-authoring the commits

Thanks for doing that. Please next time just write a comment and I'll rebase. It's gonna be tricky for me now to understand what you did or what were the changes in dev that caused conflicts.

protocols/v2/framing-sv2/src/framing.rs
Fi3: constrain generics where not needed is bad, so my opinion is the same. If you are concerned about readability you can use Frame<InnerType,Buffer> instead of Frame<T,B>

plebhash:
> > @jbesraa FYI I solved some conflicts here, which is why I'm co-authoring the commits
> 
> Thanks for doing that. Please next time just write a comment and I'll rebase. It's gonna be tricky for me now to understand what you did or what were the changes in dev that caused conflicts.

it was some small conflicts on `roles/mining-proxy/src/lib/downstream_mining.rs`, which came from merging #1021

basically I just had to adjust some calls to `payload` and `header` so they followed the changes introduced on this PR

but I understand how that was potentially counter-productive on your side, and I will avoid doing this again in the future. apologies again!

plebhash:
let's revert this to draft for now, we should aim to finish docs and have a refined and wide collective understanding of `protocols` crates, which will be addressed by the documentation issues listed on #845

rrybarczyk:
Converted to draft until all documentation issues in #845 are addressed, all refactor issues are complete, and we have team discussions/consensus on path forwards to `protocols` refactoring.

PR #918, opened 2024-05-20
Enhance SetupConnection Handling and Implement Protocol-Specific Flag Checks

xyephy:
This PR works on addressing the discussions in issue #853

PR #1035, opened 2024-07-05
Enhance SetupConnection Handling and Implement Job declaration Flag Checks 

### LINKED ISSUE #853: Issue with flags checking and related roles behaviour
In this [todo!](https://github.com/stratum-mining/stratum/blob/dev/protocols/v2/subprotocols/common-messages/src/setup_connection.rs#L88) we see that the check_flags panics if used in any other protocol except `MininProtocol`. Indeed, this method is used in `IsUpstream` trait, that does not have an extension (possibly empty) to the `JobDeclaration` protocol which is implemented for JDS
As a result, apparently there is no way for the JDS to check the bit `REQUIRES_ASYNC_JOB_MINING` of `SetupConnection.flag` in JobDeclaration protocol, making the SRI slightly no specs compliant.

1. Things to do: test with MG if different flags trigger the desired behaviour in the particular case above.
2. Check this issue affects also other parts of the code.

Shourya742:
This PR closes #853:

1. Completed MG Test.
2. Add flag check during SetupConnection call.

### DIFF
--- a/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
+++ b/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
@@ -84,8 +84,23 @@ impl<'decoder> SetupConnection<'decoder> {
 
                 work_selection && version_rolling
             }
-            // TODO
-            _ => todo!(),
+            Protocol::JobDeclarationProtocol => {
+                let available = available_flags.reverse_bits();
+                let required = required_flags.reverse_bits();
+
+                let requires_async_job_mining_passed = (required >> 31) & 1 > 0;
+                let requires_async_job_mining_self = (available >> 31) & 1 > 0;
+
+                let specific_flags_check =
+                    !requires_async_job_mining_self || requires_async_job_mining_passed;
+                let general_flags_check = (available & required) == required;
+
+                specific_flags_check && general_flags_check
+            }
+            Protocol::TemplateDistributionProtocol | Protocol::JobDistributionProtocol => {
+                // Assuming these protocols do not define flags
+                false
+            }
         }
     }
 
@@ -391,12 +406,30 @@ mod test {
     #[test]
     fn test_check_flag() {
         let protocol = crate::Protocol::MiningProtocol;
-        let flag_available = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
+        let flag_avaiable = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
         let flag_required = 0b_0000_0000_0000_0000_0000_0000_0000_0001;
         assert!(SetupConnection::check_flags(
             protocol,
-            flag_available,
+            flag_avaiable,
             flag_required
+        ));
+
+        let protocol = crate::Protocol::JobDeclarationProtocol;
+
+        let available_flags = 0b_1000_0000_0000_0000_0000_0000_0000_0000;
+        let required_flags = 0b_1000_0000_0000_0000_0000_0000_0000_0000;
+        assert!(SetupConnection::check_flags(
+            protocol,
+            available_flags,
+            required_flags
+        ));
+
+        let available_flags = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
+        let required_flags = 0b_1000_0000_0000_0000_0000_0000_0000_0000;
+        assert!(!SetupConnection::check_flags(
+            protocol,
+            available_flags,
+            required_flags
         ));
     }
 
--- a/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
+++ b/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
@@ -54,11 +54,11 @@ pub struct SetupConnection<'decoder> {
 
 impl<'decoder> SetupConnection<'decoder> {
     pub fn set_requires_standard_job(&mut self) {
-        self.flags |= 0b_0000_0000_0000_0000_0000_0000_0000_0001
+        self.flags |= 0b_0000_0000_0000_0000_0000_0000_0000_0001;
     }
 
     pub fn set_async_job_nogotiation(&mut self) {
-        self.flags |= 0b_0000_0000_0000_0000_0000_0000_0000_0001
+        self.flags |= 0b_0000_0000_0000_0000_0000_0000_0000_0001;
     }
 
     /// Check if passed flags support self flag
@@ -71,11 +71,11 @@ impl<'decoder> SetupConnection<'decoder> {
             Protocol::MiningProtocol => {
                 let available = available_flags.reverse_bits();
                 let required_flags = required_flags.reverse_bits();
-                let requires_work_selection_passed = (required_flags >> 30) > 0;
-                let requires_version_rolling_passed = (required_flags >> 29) > 0;
+                let requires_work_selection_passed = required_flags >> 30 > 0;
+                let requires_version_rolling_passed = required_flags >> 29 > 0;
 
-                let requires_work_selection_self = (available >> 30) > 0;
-                let requires_version_rolling_self = (available >> 29) > 0;
+                let requires_work_selection_self = available >> 30 > 0;
+                let requires_version_rolling_self = available >> 29 > 0;
 
                 let work_selection =
                     !requires_work_selection_self || requires_work_selection_passed;
@@ -98,7 +98,7 @@ impl<'decoder> SetupConnection<'decoder> {
                 specific_flags_check && general_flags_check
             }
             Protocol::TemplateDistributionProtocol | Protocol::JobDistributionProtocol => {
-                // Assuming these protocols do not define flags
+                // These protocols do not define flags for setting up a connection.
                 false
             }
         }
@@ -325,7 +325,7 @@ impl<'decoder> binary_sv2::Decodable<'decoder> for Protocol {
     fn get_structure(
         _: &[u8],
     ) -> core::result::Result<alloc::vec::Vec<FieldMarker>, binary_sv2::Error> {
-        let field: FieldMarker = 0_u8.into();
+        let field: FieldMarker = (0_u8).into();
         Ok(alloc::vec![field])
     }
     fn from_decoded_fields(
@@ -406,11 +406,11 @@ mod test {
     #[test]
     fn test_check_flag() {
         let protocol = crate::Protocol::MiningProtocol;
-        let flag_avaiable = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
+        let flag_available = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
         let flag_required = 0b_0000_0000_0000_0000_0000_0000_0000_0001;
         assert!(SetupConnection::check_flags(
             protocol,
-            flag_avaiable,
+            flag_available,
             flag_required
         ));
 
--- a/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
+++ b/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
@@ -69,6 +69,17 @@ impl<'decoder> SetupConnection<'decoder> {
             // [1] [1] -> true
             // [0] [1] -> false
             Protocol::MiningProtocol => {
+                // Evaluates protocol requirements based on flag bits.
+                //
+                // Checks if the current protocol meets the required flags for work selection and version rolling
+                // by reversing the bits of `available_flags` and `required_flags`. It extracts the 30th and 29th
+                // bits to determine if work selection and version rolling are needed.
+                //
+                // Returns `true` if:
+                // - The work selection requirement is satisfied or not needed.
+                // - The version rolling requirement is satisfied or not needed.
+                //
+                // Otherwise, returns `false`.
                 let available = available_flags.reverse_bits();
                 let required_flags = required_flags.reverse_bits();
                 let requires_work_selection_passed = required_flags >> 30 > 0;
@@ -85,6 +96,13 @@ impl<'decoder> SetupConnection<'decoder> {
                 work_selection && version_rolling
             }
             Protocol::JobDeclarationProtocol => {
+                // Determines if asynchronous job mining is required based on flag bits.
+                //
+                // Reverses the bits of `available_flags` and `required_flags`, extracts the 31st bit from each,
+                // and evaluates if the condition is met using these bits. Returns `true` or `false` based on:
+                // - True if `requires_async_job_mining_self` is true, or both are true.
+                // - False if `requires_async_job_mining_self` is false and `requires_async_job_mining_passed` is true.
+                // - True otherwise.
                 let available = available_flags.reverse_bits();
                 let required = required_flags.reverse_bits();
 
--- a/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
+++ b/protocols/v2/subprotocols/common-messages/src/setup_connection.rs
@@ -91,11 +91,15 @@ impl<'decoder> SetupConnection<'decoder> {
                 let requires_async_job_mining_passed = (required >> 31) & 1 > 0;
                 let requires_async_job_mining_self = (available >> 31) & 1 > 0;
 
-                let specific_flags_check =
-                    !requires_async_job_mining_self || requires_async_job_mining_passed;
-                let general_flags_check = (available & required) == required;
-
-                specific_flags_check && general_flags_check
+                match (
+                    requires_async_job_mining_self,
+                    requires_async_job_mining_passed,
+                ) {
+                    (true, true) => true,
+                    (true, false) => true,
+                    (false, true) => false,
+                    (false, false) => true,
+                }
             }
             Protocol::TemplateDistributionProtocol | Protocol::JobDistributionProtocol => {
                 // These protocols do not define flags for setting up a connection.
@@ -419,14 +423,6 @@ mod test {
         let available_flags = 0b_1000_0000_0000_0000_0000_0000_0000_0000;
         let required_flags = 0b_1000_0000_0000_0000_0000_0000_0000_0000;
         assert!(SetupConnection::check_flags(
-            protocol,
-            available_flags,
-            required_flags
-        ));
-
-        let available_flags = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
-        let required_flags = 0b_1000_0000_0000_0000_0000_0000_0000_0000;
-        assert!(!SetupConnection::check_flags(
             protocol,
             available_flags,
             required_flags
--- a/roles/jd-server/config-examples/jds-config-local-example.toml
+++ b/roles/jd-server/config-examples/jds-config-local-example.toml
@@ -1,3 +1,6 @@
+# Async Job Support
+async_mining_allowed = true
+
 # SRI Pool config
 authority_public_key = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
 authority_secret_key = "mkDLTBBRxdBv998612qipDYoTK3YUrqLe8uWw7gu3iXbSrn2n"
--- a/roles/jd-server/src/lib/job_declarator/mod.rs
+++ b/roles/jd-server/src/lib/job_declarator/mod.rs
@@ -3,12 +3,15 @@ use super::{error::JdsError, mempool::JDsMempool, status, Configuration, EitherF
 use async_channel::{Receiver, Sender};
 use binary_sv2::{B0255, U256};
 use codec_sv2::{HandshakeRole, Responder};
+use core::panic;
 use error_handling::handle_result;
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey, SignatureService};
 use network_helpers_sv2::noise_connection_tokio::Connection;
 use nohash_hasher::BuildNoHashHasher;
 use roles_logic_sv2::{
-    common_messages_sv2::SetupConnectionSuccess,
+    common_messages_sv2::{
+        Protocol, SetupConnection, SetupConnectionError, SetupConnectionSuccess,
+    },
     handlers::job_declaration::{ParseClientJobDeclarationMessages, SendTo},
     job_declaration_sv2::{DeclareMiningJob, SubmitSolutionJd},
     parsers::{JobDeclaration, PoolMessages as JdsMessages},
@@ -175,8 +178,10 @@ impl JobDeclaratorDownstream {
         for transaction in job_transactions {
             match transaction {
                 TransactionState::PresentInMempool(txid) => known_transactions.push(txid),
-                TransactionState::Missing => continue,
-            };
+                TransactionState::Missing => {
+                    continue;
+                }
+            }
         }
         known_transactions
     }
@@ -230,26 +235,26 @@ impl JobDeclaratorDownstream {
                             Ok(SendTo::Respond(m)) => {
                                 match m {
                                     JobDeclaration::AllocateMiningJobToken(_) => {
-                                        error!("Send unexpected message: AMJT")
+                                        error!("Send unexpected message: AMJT");
                                     }
                                     JobDeclaration::AllocateMiningJobTokenSuccess(_) => {
-                                        debug!("Send message: AMJTS")
+                                        debug!("Send message: AMJTS");
                                     }
                                     JobDeclaration::DeclareMiningJob(_) => {
                                         error!("Send unexpected message: DMJ");
                                     }
                                     JobDeclaration::DeclareMiningJobError(_) => {
-                                        debug!("Send nmessage: DMJE")
+                                        debug!("Send nmessage: DMJE");
                                     }
                                     JobDeclaration::DeclareMiningJobSuccess(_) => {
                                         debug!("Send message: DMJS. Updating the JDS mempool.");
                                         Self::send_txs_to_mempool(self_mutex.clone()).await;
                                     }
                                     JobDeclaration::IdentifyTransactions(_) => {
-                                        debug!("Send  message: IT")
+                                        debug!("Send  message: IT");
                                     }
                                     JobDeclaration::IdentifyTransactionsSuccess(_) => {
-                                        error!("Send unexpected message: ITS")
+                                        error!("Send unexpected message: ITS");
                                     }
                                     JobDeclaration::ProvideMissingTransactions(_) => {
                                         debug!("Send message: PMT. Updating the JDS mempool.");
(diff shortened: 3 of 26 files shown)


jbesraa:
@Shourya742 could you please tide up the commit history?

Shourya742:
> @Shourya742 could you please tide up the commit history?

Ah, I'm not sure if we should do that. This PR is based on the remaining tasks left from one of the stale PRs (#918). I'm hesitant to undermine the previous contributor's commit.

Shourya742:
@jbesraa  made the commits modular and relevant to the context. Please review and suggest any changes.

jbesraa (commented):
Looks good. Just added two nits.
Could you also please take out the `Cargo.lock` file changes?

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -450,25 +453,62 @@ impl JobDeclarator {
...
-                info!(
-                    "Setup connection message from proxy: {:?}",
-                    setup_message_from_proxy_jd
-                );
-
-                let setup_connection_success_to_proxy = SetupConnectionSuccess {
-                    used_version: 2,
-                    // Setup flags for async_mining_allowed
-                    flags: 0b_0000_0000_0000_0000_0000_0000_0000_0001,
+                let address = addr.unwrap();
```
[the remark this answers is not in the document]
Shourya742: These unwraps are safe, and since we are the ones creating the response, I don't think there is a need for additional error handling as this will always succeed. However, I can add an expect block for better readability in case of an error.

test/message-generator/mock/job-declarator-mock-invalid-flag.json
```diff
@@ -19,11 +19,10 @@
             "results": [
                 {
                     "type": "match_message_type",
-                    "value": "0x01"
+                    "value": "0x02"
                 }
-
             ],
-            "actiondoc": "This action sends SetupConnection and checks that .Success"
```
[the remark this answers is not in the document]
Shourya742: Yes, it should be. Can we let it go this time?

test/message-generator/test/job-declarator-flag-test/job-declarator-flag-test.json
```diff
@@ -0,0 +1,71 @@
...
+            "command": "cargo",
+            "args": [
+                        "llvm-cov",
+                        "--no-report",
+                        "run",
+                        "-p",
+                        "jd_server",
+                        "--",
+                        "-c",
+                        "../test/config/jds-do-not-fail-on-wrong-txdatasucc/jds-config.toml"
```
plebhash, line 22: hmm are you repurposing the config file from a different test?

I'd lean towards copying this directory into a new one with a more meaningful name. Even if the contents are identical and it ends up being somewhat redundant, if feels kinda dirty to have something here that is completely unrelated to the test.

test/message-generator/test/job-declarator-flag-test/job-declarator-flag-test.json
plebhash, line 1: `job-declarator` used to be the old name for JDS, before we introduced the conceptual distinction between JDS vs JDC

I would avoid using this generic naming, even if we have some legacy tests that still follow this convention

it seems this test is focusing on JDS, so we should call it `jds-*`

test/message-generator/mock/job-declarator-mock-invalid-flag.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": "2",
+    "doc": [
+        "This test does",
+        "Soft mock of JD",
```
plebhash, line 5:
```suggestion
        "Soft mock of JDC",
```

test/message-generator/mock/job-declarator-mock-invalid-flag.json
plebhash, line 1: this comment is similar to what I said about the `job-declarator-flag-test.json` file

I'd rename this file to ``jdc-mock-for-jds-setupconnection-flag-test.json``

test/message-generator/messages/common_messages.json
```diff
@@ -48,6 +48,22 @@
...
+                "max_version": 2,
+                "flags": 6,
+                "endpoint_host": "",
+                "endpoint_port": 0,
+                "vendor": "",
+                "hardware_version": "",
+                "firmware": "",
+                "device_id": ""
+            },
+            "id": "setup_connection_job_declarator_with_no_async_flag"
```
plebhash, line 65: can you explain the rationale on this?
why is `"flags": 6`?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -84,8 +84,23 @@ impl<'decoder> SetupConnection<'decoder> {
...
+                let requires_async_job_mining_passed = (required >> 31) & 1 > 0;
+                let requires_async_job_mining_self = (available >> 31) & 1 > 0;
+
+                let specific_flags_check =
+                    !requires_async_job_mining_self || requires_async_job_mining_passed;
+                let general_flags_check = (available & required) == required;
+
+                specific_flags_check && general_flags_check
+            }
+            Protocol::TemplateDistributionProtocol | Protocol::JobDistributionProtocol => {
```
plebhash, line 100: why is `Protocol::JobDistributionProtocol` on this line?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -84,8 +84,23 @@ impl<'decoder> SetupConnection<'decoder> {
...
+                // Assuming these protocols do not define flags
```
plebhash, line 101: @Fi3 can you confirm Template Distribution Protocol does not set any flags for `SetupConnection`?

I cannot find anything on [the specs](https://stratumprotocol.org/specification/07-Template-Distribution-Protocol/) but just to make sure.

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -391,13 +406,31 @@ mod test {
     #[test]
     fn test_check_flag() {
         let protocol = crate::Protocol::MiningProtocol;
-        let flag_available = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
+        let flag_avaiable = 0b_0000_0000_0000_0000_0000_0000_0000_0000;
```
plebhash, line 409: it seems we're changing this line just to introduce a typo?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -391,13 +406,31 @@ mod test {
...
         let flag_required = 0b_0000_0000_0000_0000_0000_0000_0000_0001;
         assert!(SetupConnection::check_flags(
             protocol,
-            flag_available,
+            flag_avaiable,
```
plebhash, line 413: it seems we're changing this line just to introduce a typo?

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -450,25 +453,62 @@ impl JobDeclarator {
...
+                    Err(e) => {
+                        error!("Error receiving message: {:?}", e);
+                        continue;
+                    }
                 };
-                let sv2_frame: StdFrame =
-                    JdsMessages::Common(setup_connection_success_to_proxy.into())
+
+                let payload = incoming.payload();
+                let flag = payload[5] as u32;
```
plebhash, line 476: shouldn't we do some kind of verification to make sure this is indeed a `SetupConnection` that JDS just received?

what if the client sent some nonsense and we're blindly parsing this payload as if it was a `SetupConnection` flag?

plebhash: also, what is the rationale behind `payload[5]`?

plebhash (changes requested):
this PR has a solid foundation and demonstrates some great work by @Shourya742 

but we still need to improve some things before we are able to merge it

roles/jd-server/src/lib/job_declarator/mod.rs
Shourya742: I completely overlooked this possibility. It makes sense to have this approach rather than parsing based on my understanding. The reason for payload[5] was that the flag is a 4-byte value encoded in the payload in little-endian format, so the 5th byte must contain the flag value. Since the current range for the flag value is up to 3 bits max, it fits easily within 8 bits or the 1st byte. However, we still have the possibility of increasing the flag bit range in the future.

For any other type of message except the setupConnection, we should not respond. Instead, we should continue listening to our other processes, as these messages might be some sort of attack to slow down our JDS server. In such cases, we should avoid sending any response to the connection request.

But ya thanks for pointing this out. @plebhash

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
Shourya742: Will correct it out

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
Shourya742: will correct it out

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
Shourya742: Because for both of these cases, we dont receive any flags so no point of checking for them

test/message-generator/messages/common_messages.json
Shourya742: Okay, so the idea is that a setupConnection request from JDC to JDS needs to have a flag with the 0th bit set. This flag value have both the 1st and 2nd bits set but not the 0th. In such a case, the expected behavior is a setupConnection error.

test/message-generator/test/job-declarator-flag-test/job-declarator-flag-test.json
Shourya742: Yup makes sense

test/message-generator/test/job-declarator-flag-test/job-declarator-flag-test.json
Shourya742: Okay, I will create a separate configuration file. It makes sense to have different configuration files for different contexts.

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
Fi3: I confirm

roles/jd-server/src/lib/job_declarator/mod.rs
plebhash: > The reason for payload[5] was that the flag is a 4-byte value encoded in the payload in little-endian format, so the 5th byte must contain the flag value.

we have APIs for parsing messages from frame payloads, so we should use them for safety and readability

> For any other type of message except the setupConnection, we should not respond. Instead, we should continue listening to our other processes, as these messages might be some sort of attack to slow down our JDS server. In such cases, we should avoid sending any response to the connection request.

the attack vector is a good point that I also overlooked in my previous message.

given that the context of this code is a `while let Ok((stream, _)) = listner.accept().await { ... }`, I wonder if it doesn't make sense to just drop the connection for every case where the first incoming message is not a `SetupConnection`?

cc @Fi3

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
plebhash: ah ok, for some reason my brain parsed `Protocol::JobDeclarationProtocol` and `Protocol::JobDistributionProtocol`  as if they were the same thing

which is actually a good opportunity to highlight a potential footgun:

I'm about 95% sure that `JobDistributionProtocol` is just some legacy terminology that we forgot to remove after some spec changes

if we check [stratumprotocol.org/specification](https://stratumprotocol.org/specification/) we'll only find 3 subprotocols:
- Mining
- Job Declaration
- Template Distribution

@Fi3 do you confirm that we should clean up these remaining `JobDistributionProtocol` from the code?

if that's the case, we should create an issue so we can clean this up in the future

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
plebhash: ok so @Shourya742 we should change this comment

instead of saying we are assuming, we should confidently state this scenario does not have any meaningful flags for `SetupConnection` message

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
Fi3: if we are not going to add it the spec I would remove it

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -84,8 +84,23 @@ impl<'decoder> SetupConnection<'decoder> {
...
+            Protocol::JobDeclarationProtocol => {
+                let available = available_flags.reverse_bits();
+                let required = required_flags.reverse_bits();
+
+                let requires_async_job_mining_passed = (required >> 31) & 1 > 0;
+                let requires_async_job_mining_self = (available >> 31) & 1 > 0;
+
+                let specific_flags_check =
+                    !requires_async_job_mining_self || requires_async_job_mining_passed;
+                let general_flags_check = (available & required) == required;
```
lorbax, line 96: I don't understand semantic of this.
You just have to check the 31st bit, why you do the & bitwise between the whole `available` and `required`?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -84,8 +84,23 @@ impl<'decoder> SetupConnection<'decoder> {
...
-            _ => todo!(),
+            Protocol::JobDeclarationProtocol => {
+                let available = available_flags.reverse_bits();
+                let required = required_flags.reverse_bits();
+
+                let requires_async_job_mining_passed = (required >> 31) & 1 > 0;
+                let requires_async_job_mining_self = (available >> 31) & 1 > 0;
+
+                let specific_flags_check =
+                    !requires_async_job_mining_self || requires_async_job_mining_passed;
```
lorbax, line 95: seems that this produces false when the async mining job is available in the upstream, but it is not required by the downstream. In my opinion this should produce true, not false. 
Also, this produces true when async mining job is required by the downstrem bu not available by the upstream. I think that this should produce false.
Passed  $  Self  $  !Self || Passed
      T       $    F    $            T
      T       $    T    $            T
      F       $    T    $            F
      F       $    F    $            T

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -450,39 +453,77 @@ impl JobDeclarator {
...
-                    mempool.clone(),
-                    // each downstream has its own sender (multi producer single consumer)
-                    sender_add_txs_to_mempool.clone(),
-                )));
+                let payload = incoming.payload();
+
+                if let Ok(value) = binary_sv2::from_bytes::<SetupConnection>(payload) {
+                    let flag = value.flags;
+                    let is_valid =
+                        SetupConnection::check_flags(Protocol::JobDeclarationProtocol, flag, 1);
```
lorbax, line 479: here the JDS offers async job mining, so if the downstream does not requires it, the flags checking should produce `True`.
Aslo, according to the fact that the client set or not the flag `async mining job`, we espect different behaviours from the JDS. Perhaps this is material for a second PR.

test/message-generator/messages/common_messages.json
plebhash: why not simply `0`?

1st and 2nd bits will be shifted away and ignored inside `check_flags`, so it feels like there's no real upside for using `6` here, it just makes it more confusing to understand

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -437,54 +451,98 @@ impl JobDeclarator {
...
 
-                let jddownstream = Arc::new(Mutex::new(JobDeclaratorDownstream::new(
-                    receiver.clone(),
-                    sender.clone(),
-                    &config,
-                    mempool.clone(),
-                    // each downstream has its own sender (multi producer single consumer)
-                    sender_add_txs_to_mempool.clone(),
-                )));
+                            let sv2_frame: StdFrame = if is_valid {
```
plebhash, line 484: I would change this line, and simply do a `if is_valid`

then, declare `sv2_frame: StdFrame` inside the `{}` block, and also move the calls to `JobDeclaratorDownstream::new` + `JobDeclaratorDownstream::start` from line 514 to here, while doing the same for the `else` case

this way, we can avoid checking `is_valid` 2x, which feels a bit redundant

roles/jd-server/src/lib/job_declarator/mod.rs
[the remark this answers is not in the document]
plebhash: I'm not a fan of `unwrap` either, but there's many instances where getting rid of them requires changing breaking APIs, which drastically change the nature of the task at hand

this kind of thing should be addressed during the refactor efforts

for now, I think replacing `unwrap` with `expect` where possible feels like a good start

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
plebhash: @lorbax has a point, here's a refined truth table:
| available | required | check |
|-----------|----------|-------|
| T         | T        | T     |
| T         | F        | T     |
| F         | T        | F    |
| F         | F        | T     |

I'm also writing some suggestions on the code

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -437,54 +451,94 @@ impl JobDeclarator {
...
-                sender.send(sv2_frame).await.unwrap();
+                        if let Ok(setup_connection) =
+                            binary_sv2::from_bytes::<SetupConnection>(payload)
+                        {
+                            let flag = setup_connection.flags;
+                            let is_valid = SetupConnection::check_flags(
+                                Protocol::JobDeclarationProtocol,
+                                flag,
+                                1,
+                            );
```
plebhash, line 482: it seems we were mixing up the order of the parameters:

second parameter is `available_flags`, which signals that JDS supports async jobs, and therefore should be `1`

third parameter is `required_flags`,  which signals if JDC requested async jobs or not, so that one should be `flag`
```suggestion
                            let is_valid = SetupConnection::check_flags(
                                Protocol::JobDeclarationProtocol,
                                flag,
                                1,
                            );
```

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
plebhash, line 95:
```suggestion
                match (requires_async_job_mining_self, requires_async_job_mining_passed) {
                    (true, true) => true,
                    (true, false) => true,
                    (false, true) => false,
                    (false, false) => true,
                }
```

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
plebhash: I agree this line is unnecessary, and should be replaced by the `match` suggestion above
```suggestion
```

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -84,8 +84,23 @@ impl<'decoder> SetupConnection<'decoder> {
...
+                let general_flags_check = (available & required) == required;
+
+                specific_flags_check && general_flags_check
```
plebhash, line 98: the `match` suggestion above already returns the `bool`
```suggestion
```

plebhash:
@lorbax provided some valuable feedback https://github.com/stratum-mining/stratum/pull/1035#discussion_r1679624474 where basically @Shourya742 and I were missing the fact that some deeper changes to JDS were actually necessary

if we carefully read #853, a logical conclusion is that JDS needs a new config parameter, where the user chooses if JDS should support async jobs or not

that was achieved on this commit: 996adc37532d246c8d2594360c7c8b9eacdf56ac

---

also, we need MG tests that will cover this entire truth table:
| async jobs allowed | async jobs requested | `SetupConnection.Success` |
|-----------|----------|-------|
| T         | T        | T     |
| T         | F        | T     |
| F         | T        | F    |
| F         | F        | T     |

that was achieved on this commit: 35b1e30ded5f78760e5fea37d155e2c3a815e5ff

---

I suggest @Shourya742 reviews these two commits, and if he agrees, cherry-pick them into this PR

then, @lorbax can verify if now we finally have a sane solution to #853

roles/jd-server/src/lib/job_declarator/mod.rs
plebhash: you are right!

and I think we should fix that on this PR, otherwise this PR would be introducing a broken feature into JDS

I elaborate more here https://github.com/stratum-mining/stratum/pull/1035#issuecomment-2237516417

plebhash (approved):

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
jbesraa: @Shourya742 I think as well this shouldn't be part of this PR, or at least it should in its own commits explaining what/why

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -71,11 +71,11 @@ impl<'decoder> SetupConnection<'decoder> {
             Protocol::MiningProtocol => {
                 let available = available_flags.reverse_bits();
```
jbesraa, line 72: Could you please add an explanation above the L72 describing what are the checks we are doing?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -84,8 +84,27 @@ impl<'decoder> SetupConnection<'decoder> {
 
                 work_selection && version_rolling
             }
-            // TODO
-            _ => todo!(),
+            Protocol::JobDeclarationProtocol => {
+                let available = available_flags.reverse_bits();
```
jbesraa, line 88: I think here as well(above L88) it would be useful to add explaination

jbesraa (changes requested):
This looks very close. Added few comments.
Could you please rearrange the commits as well?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
Shourya742: Aren't these self-explanatory, as in taking the first bit and checking the condition based on whether it's set or not? However, I will still add documentation if it's unclear to others as well.

Shourya742:
> This looks very close. Added few comments. Could you please rearrange the commits as well?

Yes, I'll squash a few of my commits and add descriptions according to the context

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
jbesraa: Yea, I think it would be better to document those. Not necessarily line by line, be the general approach

jbesraa:
> > This looks very close. Added few comments. Could you please rearrange the commits as well?
> 
> Yes, I'll squash a few of my commits and add descriptions according to the context

Thanks, this might be helpful for when writing commits  https://cbea.ms/git-commit/

Shourya742:
@jbesraa I’ve added the suggested changes. Please review and let me know if any further modifications are needed.

pavlenex:
@Shourya742 can you fix this https://github.com/stratum-mining/stratum/actions/runs/10062714214/job/27906911880 ? Everything else is passing.

pavlenex:
This one is waiting for @lorbax's review according to the last dev call and then it's good to go.

lorbax:
I think that now the logic of the flag checking is fine.
Nevertheless, perhaps the author should open an issue regarding the fact that the standard behavior is to have the JDS work with that flag unset, and `async_mining_allowed` should offered optionally. Not having this may expose the pool to this [this](https://github.com/stratum-mining/stratum/issues/1053) attack (pointed out in [this](https://github.com/stratum-mining/stratum/issues/1053#issuecomment-2230607082)  comment).

roles/jd-server/config-examples/jds-config-local-example.toml
```diff
@@ -1,3 +1,6 @@
+# Async Job Support
+async_mining_allowed = true
```
plebhash, line 2: as suggested by @lorbax :https://github.com/stratum-mining/stratum/pull/1035#issuecomment-2252334208

> the standard behavior is to have the JDS work with that flag unset, and async_mining_allowed should offered optionally. Not having this may expose the pool to this https://github.com/stratum-mining/stratum/issues/1053 attack (pointed out in https://github.com/stratum-mining/stratum/issues/1053#issuecomment-2230607082 comment).

```suggestion
async_mining_allowed = false
```

Fi3:
> I think that now the logic of the flag checking is fine. Nevertheless, perhaps the author should open an issue regarding the fact that the standard behavior is to have the JDS work with that flag unset, and `async_mining_allowed` should offered optionally. Not having this may expose the pool to this [this](https://github.com/stratum-mining/stratum/issues/1053) attack (pointed out in [this](https://github.com/stratum-mining/stratum/issues/1053#issuecomment-2230607082) comment).

if the flag is set this is not an attack but the desired behavior (by flag definition). For example if you save everything and you calculate the miner reward in a later moment or somewhere else and not each time that you receive a share (what I expect most pool will do) setting this flag is ok.

lorbax:
> > I think that now the logic of the flag checking is fine. Nevertheless, perhaps the author should open an issue regarding the fact that the standard behavior is to have the JDS work with that flag unset, and `async_mining_allowed` should offered optionally. Not having this may expose the pool to this [this](https://github.com/stratum-mining/stratum/issues/1053) attack (pointed out in [this](https://github.com/stratum-mining/stratum/issues/1053#issuecomment-2230607082) comment).
> 
> if the flag is set this is not an attack but the desired behavior (by flag definition). For example if you save everything and you calculate the miner reward in a later moment or somewhere else and not each time that you receive a share (what I expect most pool will do) setting this flag is ok.
> 
> Also I do not thing that the issue that you refer should be considered an issue see my comment: [#1053 (comment)](https://github.com/stratum-mining/stratum/issues/1053#issuecomment-2238880037)

I agree that the issue does not exists, I misread it.
Nevertheless, JDS works only with that flag set, while that flag set indicates a feature.

plebhash:
> Nevertheless, JDS works only with that flag set, while that flag set indicates a feature.

hopefully this PR fixes this

@lorbax do you think it's ready?

Issue #1038, opened 2024-07-05
`pre-push` git hooks errors + should be limited to formatting only


rrybarczyk:
# Background
`act` is a tool that works with Docker to run GitHub Actions workflows locally, designed to be run before pushing changes. The `.githooks/pre-push` uses `act` to run some GitHub workflows with `% ./.githooks/pre-push origin https://github.com/stratum-mining/stratum.git`.

The currently defined workflows in this `pre-push` script are:

- `message_generator_check` (this is not an existing workflow, probably trying to reference `message-generator-test`)
- `sv2_header_check`
- `fmt`
- `clippy-check`
- `ci`

During the 2024.07.02 dev meet, the team decided that the `pre-push` should only containing more commonly needed workflows, like for formatting. Additionally, because of long run times when using `act`, the usage of `act` should be replaced with calling the bash scripts in the `scripts/` directory in the repo root. This will keep the `pre-push` runtime shorter, and then any more specific workflows needed can be run manually by executing the bash scripts in the `scripts/` directory on the `dev` branch.

# Problem
* The `pre-push` script should only contain workflows that are more commonly run, like for linting.
* The non-existent `message_generator_check` is causing the `pre-push` to fail.
* The run-times associated with `act` are long.

# Solution
In `.githooks/pre-push`, replace the use of `act` with calls to the following bash scripts in the `scripts/` directory on the `dev` branch:

* `clippy-on-all-workspaces.sh`: Runs `cargo clippy`, `cargo test`, and `cargo +nightly fmt`.
* `sv2-header-check.sh`: Ensures that the `sv2.h` file generated by `build_header.sh` is in sync with the commited `protocols/v2/sv2-ffi/sv2.h`.

rrybarczyk:
Here are the run times for the scripts:
* `clippy-on-all-workspaces.sh` containing `cargo clippy`, `cargo test`, and `cargo +nightly fmt`:`./scripts/clippy-on-all-workspaces.sh  58.09s user 1.76s system 145% cpu 41.107 total`
* `clippy-on-all-workspaces.sh` containing `cargo clippy` and `cargo +nightly fmt` (`cargo test` removed): ./scripts/clippy-on-all-workspaces.sh  2.90s user 0.52s system 74% cpu 4.582 total

I will add the `sv2-header-check.sh`, but am experiencing some errors.

rrybarczyk:
I am experiencing the following issues in running `./scripts/sv2-header-check.sh`:

[...]

PR #1039, opened 2024-07-05
`pre-push` githook for fmt only + remove `act`

rrybarczyk:
Addresses #1038.


benches/Cargo.lock
plebhash, line 1: did you intentionally commit these `Cargo.lock` files or was it a mistake?

.githooks/pre-push
```diff
@@ -1,31 +1,28 @@
...
-remote="$1"
-url="$2"
-
-act --job message_generator_check --reuse -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest
-act --job sv2_header_check --reuse -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest
-act --job fmt --reuse -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest
-act --job clippy-check --reuse -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest
-act --job ci --reuse -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest
+# Run clippy, test, and rustfmt on all workspaces
+sh ./scripts/clippy-on-all-workspaces.sh
```
GitGab19, line 15: As suggested [here](https://github.com/stratum-mining/stratum/issues/1038#issuecomment-2211651871), I would rename the script into `clippy-fmt-and-test.sh`

benches/Cargo.lock
rrybarczyk: They were generated/updated when I ran the clippy and sv2 header check script. I was not sure if I should commit them or not. Should I remove?

.githooks/pre-push
Fi3: now that we removed cargo version from the toml files we should enforce cargo 1.75 in the scripts

benches/Cargo.lock
plebhash: yeah I would remove them from this PR

this is probably something more relevant for https://github.com/stratum-mining/stratum/pull/1044

.githooks/pre-push
```diff
@@ -1,31 +1,38 @@
...
-remote="$1"
-url="$2"
+# Enforce cargo version 1.75
+REQUIRED_CARGO_VERSION="1.75.0"
+INSTALLED_CARGO_VERSION=$(cargo --version | awk '{print $2}')
+
+if [ "$INSTALLED_CARGO_VERSION" != "$REQUIRED_CARGO_VERSION" ]; then
+    echo "Error: Cargo version $REQUIRED_CARGO_VERSION is required. Installed version is $INSTALLED_CARGO_VERSION."
+    exit 1
+fi
```
rrybarczyk, line 22: @Fi3, regarding your comment [here](https://github.com/stratum-mining/stratum/pull/1039#discussion_r1671966258),  do you like how I am enforcing the `cargo` version? Open to alternatives

rrybarczyk:
@plebhash, @GitGab19, and @Fi3: This is ready for review again. Please note that when the `pre-push` script is ran, the following `Cargo.locks` get changed:

* `benches/Cargo.lock`
* `roles/Cargo.lock`

I did not commit these changes to this repo (had them in before but remove that commit).

Also, unrelated to this PR, but one thing I noticed was that a couple of the tests do not complete but instead print that they have been running for over 60 seconds. This greatly slows down the runtime of the `pre-push` script, which is annoying:

cargo-test-run-over-60s

Does anyone know why this is happening?

.githooks/pre-push
Fi3: I would just try to run the wanted cargo verision `cargo +1.75.0 ...` and fail if not avaible, the error from cargo is already descriptive. Instead of 1.75.0 you use REQUIRED_CARGO_VERSION

plebhash:
with https://github.com/stratum-mining/stratum/pull/1102 in mind, it might be a good idea to also run this on our githook:
```
cargo build --manifest-path=roles/Cargo.toml --locked
```

.githooks/pre-push
plebhash: I got this error:
```
+ REQUIRED_CARGO_VERSION=1.75.0
++ cargo --version
++ awk '{print $2}'
+ INSTALLED_CARGO_VERSION=1.77.2
+ '[' 1.77.2 '!=' 1.75.0 ']'
+ echo 'Error: Cargo version 1.75.0 is required. Installed version is 1.77.2.'
Error: Cargo version 1.75.0 is required. Installed version is 1.77.2.
+ exit 1
```

I think this should only fail if the installed cargo version is smaller than MSRV, not bigger.

scripts/sv2-header-check.sh
```diff
@@ -9,11 +9,11 @@
...
 #   (4) takes the SHA1 hash of the `sv2.h` file generated by `build_header.sh`
-#   (5) compares the hashes of each `sv2.h`, if they are equal then the GitHub Action passes, otherwise
-#       the GitHub Action fails
-#
-# This script is called by `.github/workflows/sv2-header-check.yaml` on every PR onto the main branch.
+#   (5) compares the hashes of each `sv2.h`, if they are equal then the GitHub Action passes,
+#       otherwise the GitHub Action fails
 #
+# This script is called by `.github/workflows/sv2-header-check.yaml` on every PR onto the main
+# branch.
```
plebhash, line 16:
```suggestion
# branch.

# Check if sha1sum is available on the system
if ! command -v sha1sum >/dev/null 2>&1; then
  echo "Warning: sha1sum is not installed on this system."
  exit 1
fi
```

.githooks/pre-push
rrybarczyk: Update the logic to enforce `1.75.0` as minimum. Checked that `1.74.0` cargo fails this check and that `1.75.0` and `1.77.0` cargo passes this check.

rrybarczyk:
Will change to ready for review to be merged after #1102 and #1044 are merged.

jbesraa:
Does this run automatically on a pre-push or due I need to trigger it automatically?

plebhash:
> Does this run automatically on a pre-push or do I need to trigger it automatically?

You need to do this to enable the pre-push:

https://github.com/stratum-mining/stratum/blob/0f0ee1e32a7cdeb9af83718364e0896b36744824/README-DEV.md?plain=1#L8-L11

After you do this, your local git repo will be configured to always run this script when you do `git push`. In order to bypass it, you will need to add `--no-verify`.

rrybarczyk:
The `roles/Cargo.lock` is not constant and the `cargo build --manifest-path=roles/Cargo.toml --locked` check in the `pre-push` fails with:

```sh
error: the lock file /Users/rachelrybarczyk/Development/StratumV2/Dev/stratum/roles/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
+ echo 'Error: Cargo.lock file in roles crate is out of date. Please run '\''cargo update'\'' in the roles crate.'
Error: Cargo.lock file in roles crate is out of date. Please run 'cargo update' in the roles crate.
+ exit 1
```

This `Cargo.lock` file needs to be fixed before this PR can be merged.

plebhash (approved):

PR #1040, opened 2024-07-05
Rust docs for `protocols::v2::codec-sv2`

rrybarczyk:
Addresses #1012.

See discussion for potential code debt discovered during this documentation effort in t[his discussion](https://github.com/stratum-mining/stratum/discussions/873).


plebhash:
@rrybarczyk is this PR still in draft status for some specific reason or is it ready for review?

we should leave the technical debt for a new issue/PR and constrain this one to just adding Rust Docs.

protocols/v2/codec-sv2/src/error.rs
rrybarczyk, line 1: I did not change any code here, but I did reorder to be in alphabetical order.

jbesraa:
@rrybarczyk let me know when you want a review here

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +71,51 @@ use crate::Error::MissingBytes;
...
+#[cfg(feature = "noise_sv2")]
+pub type StandardNoiseDecoder<T> = WithNoise<Buffer, T>;
+
+/// Standard Sv2 decoder without Noise protocol support.
 pub type StandardDecoder<T> = WithoutNoise<Buffer, T>;
 
+/// Decoder for Sv2 frames with Noise protocol support.
 #[cfg(feature = "noise_sv2")]
 pub struct WithNoise<B: IsBuffer, T: Serialize + binary_sv2::GetSize> {
+    /// Marker for the type of frame being decoded.
```
Fi3, line 90: This comment is redundant and should be removed.

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +71,51 @@ use crate::Error::MissingBytes;
...
+    /// Buffer for holding decrypted Sv2 data.
+    ///
+    /// Stores the decrypted data from the Noise protocol until it is ready to be processed and
+    /// concerted into a Sv2 frame.
     sv2_buffer: B,
 }
 
 #[cfg(feature = "noise_sv2")]
 impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> WithNoise<B, T> {
+    /// Attempts to decode the next frame, returning either a frame or an error indicating how many
```
Fi3, line 117: Can also return other kind of error. So it return either a frame, an error, or an error with missing bytes. I did it like that but I think that maybe here having and enum into the Ok and the error would have been a better interface. Not sure if make sense to change it now.

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -157,17 +233,24 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
         frame.into()
     }
 
+    /// Provides a writable buffer for incoming data.
```
Fi3, line 236: Maybe I would add that the buffer have the right number of bytes needed by the codec to go to the next step. I would also add an example of how to use the returned buffer.

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +71,51 @@ use crate::Error::MissingBytes;
...
 pub type StandardSv2Frame<T> = Sv2Frame<T, <Buffer as IsBuffer>::Slice>;
+
+/// Standard decoder with Noise protocol support.
+#[cfg(feature = "noise_sv2")]
+pub type StandardNoiseDecoder<T> = WithNoise<Buffer, T>;
+
+/// Standard Sv2 decoder without Noise protocol support.
 pub type StandardDecoder<T> = WithoutNoise<Buffer, T>;
 
+/// Decoder for Sv2 frames with Noise protocol support.
```
Fi3, line 87: here we can add an example of how to create and use a codec. If is possible to do it vai a doc test would be fantastic.

protocols/v2/codec-sv2/src/decoder.rs
Fi3: also here an example would be nice. I just stop writing this but I think that when is possible we should add examples

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -157,17 +233,24 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
         frame.into()
     }
 
+    /// Provides a writable buffer for incoming data.
     #[inline]
     pub fn writable(&mut self) -> &mut [u8] {
         self.noise_buffer.get_writable(self.missing_noise_b)
     }
+
+    /// Checks if the buffers are droppable.
```
Fi3, line 242: maybe we can link the Buffer trait docs (now is undocumented but as soon as we will document it we will have a meaningful link

plebhash: https://docs.rs/buffer_sv2/1.1.0/buffer_sv2/

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,40 @@
+//! # Stratum V2 Codec Decoder
+//!
+//! This module provides functionality for decoding Stratum V2 messages, including support for
```
plebhash, line 3:
```suggestion
//! This module provides functionality for encoding and decoding Stratum V2 messages, including support for
```

plebhash:
@rrybarczyk make sure you also add a descriptive `README.md` to the crate, so it's displayed on `crates.io`

protocols/v2/codec-sv2/src/lib.rs
```diff
@@ -1,3 +1,60 @@
...
+//! * **Encoding/Decoding**: Serialize and deserialize Stratum V2 messages.
+//! * **Noise Protocol**: Support for secure communication using the Noise protocol framework.
+//! * **Error Handling**: Comprehensive error handling for various stages of communication and
+//!   encryption.
+//!
+//! ## Modules
+//!
+//! * `decoder`: Provides functionality to decode Stratum V2 messages.
+//! * `encoder`: Provides functionality to encode Stratum V2 messages.
+//! * `error`: Defines error types and result aliases used throughout the crate.
```
plebhash, line 21: a `#Modules` section is automatically generated by Rust docs, so it is redundant to add one here

laos, the `encoder` and `decoder` modules are not exposed

protocols/v2/codec-sv2/src/lib.rs
```diff
@@ -1,3 +1,60 @@
...
+//! `codec-sv2` is essential for implementing the Sv2 protocol correctly and securely. It abstracts
+//! the complexity of message encoding, decoding, and encryption, providing a reliable and
+//! consistent foundation for mining software imperative for interoperability.
+//!
+//! ## Features
+//!
+//! * **Encoding/Decoding**: Serialize and deserialize Stratum V2 messages.
+//! * **Noise Protocol**: Support for secure communication using the Noise protocol framework.
+//! * **Error Handling**: Comprehensive error handling for various stages of communication and
+//!   encryption.
```
plebhash, line 15:
```suggestion
```

[`framing_sv2`](https://docs.rs/framing_sv2/latest/framing_sv2/) has a `# Features` section that is dedicated to feature flags.

protocols/v2/codec-sv2/src/lib.rs
```diff
@@ -1,3 +1,60 @@
...
+//!     }
+//! }
+//! ```
+//!
+//! ## Conditional Compilation
+//!
+//! * The crate uses conditional compilation to enable or disable features based on the `noise_sv2`
+//!   feature flag. When enabled, additional functionality related to the Noise protocol is
+//!   available.
+//!
```
plebhash, line 56:
```suggestion
//! # Features
//! This crate can be built with the following features:
//! - `with_serde`: builds `binary_sv2` and `buffer_sv2` crates with `serde`-based encoding and decoding.
//! - `with_buffer_pool`: uses `buffer_sv2` to provide a more efficient allocation method for `non_std` environments. Please refer to `buffer_sv2` crate docs for more context.
//! - `noise_sv2`: enables encryption via Noise protocol.
//!
//! The `with_serde` feature flag is only used for the Message Generator, and deprecated for any other kind of usage. It will likely be fully deprecated in the future.
```

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,47 @@
...
+//! let mut encoder: Encoder<MyFrameType> = Encoder::new();
+//!
+//! // Encode a frame into a byte slice
+//! let frame = Sv2Frame::new( /* frame details */ );
+//! let encoded_data = encoder.encode(frame).unwrap();
+//!
+//! // Create a Noise encoder for encrypted SV2 frames (requires the `noise_sv2` feature)
+//! #[cfg(feature = "noise_sv2")]
+//! let mut noise_encoder: NoiseEncoder<MyFrameType> = NoiseEncoder::new();
+//! ```
```
plebhash, line 44: this header is not going to be displayed in Rust Docs, because the `encoder` module is not exposed

overall I'd recommend to regularly do `cargo doc --all-features --open` to render the docs

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,43 @@
...
+//!
+//! // Create a standard decoder
+//! let mut decoder: StandardDecoder<MyFrameType> = StandardDecoder::new();
+//!
+//! // Create a noise decoder (requires the `noise_sv2` feature)
+//! #[cfg(feature = "noise_sv2")]
+//! let mut noise_decoder: StandardNoiseDecoder<MyFrameType> = StandardNoiseDecoder::new();
+//! ```
+//!
+
```
plebhash, line 40: this header is not going to be displayed in Rust Docs, because the `decoder` module is not exposed

overall I'd recommend to regularly do `cargo doc --all-features --open` to render the docs so you can have a good idea of what the end-user is going to see

Fi3: I think that we can change how we export thing in order to include it without changing the Api

plebhash: sorry @Fi3 I don't fully understand, can you elaborate?

rrybarczyk: We can add the `--document-private-items` to expose documentation for private modules like:

```sh
cargo doc --document-private-items --all-features --open
```

I left a [comment here](https://github.com/stratum-mining/stratum/issues/1130#issuecomment-2347055483) with some thoughts.

Essentially, I am not sure if you can publish the private crate documentation to crates.io, or if we would want to do that. BUT, we can use it to guide us and view our docs on private modules while we develop the documentation.

Either way, my example here sucks 😝.

protocols/v2/codec-sv2/src/encoder.rs
rrybarczyk: I think we should still impose our documentation standards on private modules, regardless of whether it will be generated. Thoughts?

protocols/v2/codec-sv2/src/decoder.rs
Fi3: yep or we can make the module public in this case: we make codec_sv2::decoder public and in the module we make private all the things that are not exported by lib.rs so we don't change the api.

protocols/v2/codec-sv2/src/encoder.rs
Fi3: no if a module is private we don't have to document it for user but for maintainers so is a different thing. In that case the module should not be private. A rule of thumb is, if you thing that you need to document a private module for the user maybe the module should not be private.

jbesraa: I also think we should not document private modules. If there is some private module that is very vague, and we feel strongly about adding a comment, maybe we can use regular comment and not `///`?

Shourya742, line 44: I'm not entirely sure if writing this is really helpful. It could be misleading for someone referencing it, as they might assume these are public-facing APIs. I believe we should include code snippets only for public-facing modules, rather than for all modules.

protocols/v2/codec-sv2/src/lib.rs
```diff
@@ -1,3 +1,55 @@
...
+//! // Proceed with the handshake steps
+//! match state.step_0() {
+//!     Ok(frame) => {
+//!         // Send the initial handshake frame
+//!     }
+//!     Err(e) => {
+//!         // Handle error
+//!     }
+//! }
+//! ```
```
Shourya742, line 43: Here, we can have a doc test.

Shourya742 (commented):
Few nits.

protocols/v2/codec-sv2/src/encoder.rs
plebhash: I agree with @rrybarczyk in the sense that it's important that we don't fully ignore parts of the code just because they are not exposed. Afterall, one of the main goals of this collective effort is for SRI contributors to refine their understanding and ultimately ownership over these crates.

However, while we write the docs, it's also important for us to have a clear view over what adopters will see vs what is going to be abstracted away from their eyes. This broader perspective will allow us to shape the APIs in a way that lowers friction for adoption.

A good way to achieve this goal, while still documenting (and therefore gaining knowledge and ownership) is to do [Non Doc comments](https://doc.rust-lang.org/reference/comments.html#non-doc-comments) for the code that will not be rendered over Rust Docs (via `cargo doc --all-features --open`).

I see @rrybarczyk also suggested `cargo doc --document-private-items --all-features --open` on the other comment thread. This can be good to see how these will be rendered. But we should be careful because this could create a blindspot where we don't care what the user sees (because we're not seeing ourselves).

plebhash: I formalized my proposal above here https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10697537

plebhash: let's try to reach consensus here https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10697537

protocols/v2/codec-sv2/src/lib.rs
plebhash: assuming we will make `encoder` and `decoder` modules public, we should still remove this `## Modules` section because it's redundant

protocols/v2/codec-sv2/src/decoder.rs
plebhash: I agree, we should make these modules public.

protocols/v2/codec-sv2/src/lib.rs
plebhash: this could be good for CI, but will require more research to figure out the best way to achieve this

plebhash, line 43: if our main example is going to include encryption, we should do two code blocks:
- one example where `Initiator` does the full handshake cycle, then encodes a message, and sends it
- one example where `Receiver` does the full handshake cycle, then decodes a received message

as an alternative, we could also just simplify things by doing an unencrypted example

I'm working on something, I'll propose it here once I know it compiles.

plebhash: I wrote a replit with an atomic example: https://replit.com/@plebhash/GummyLividClicks

it is mainly inspired by #977

we have a good chunk of ~80 lines of code there, and a fully functional example including encryption would probably take a lot more lines

so I would lean towards either:
- showing an unencrypted example code on the top of `codec_sv2` Rust Docs
- now showing any example code on the top of `codec_sv2` Rust Docs

and then for encrypted examples we could maybe create an issue so we eventually start a `protocols/v2/codec_sv2/examples` directory with more complex examples

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +74,51 @@ use crate::Error::MissingBytes;
...
 
+/// Decoder for Sv2 frames with Noise protocol support.
 #[cfg(feature = "noise_sv2")]
 pub struct WithNoise<B: IsBuffer, T: Serialize + binary_sv2::GetSize> {
+    /// Used to maintain type information for the generic parameter `T`, which represents the type
+    /// of frames being decoded.
+    ///
+    /// `T` refers to a type that implements the necessary traits for serialization
+    /// (`binary_sv2::Serialize`), deserialization (`binary_sv2::Deserialize`), and size
+    /// calculation (`binary_sv2::GetSize`).
```
plebhash, line 98: private fields are never exposed to reader eyes

it's good to document this for maintainer eyes, but we should use Non Doc comments, as proposed in https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10697537

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +74,51 @@ use crate::Error::MissingBytes;
...
     frame: PhantomData<T>,
+
+    /// Number of missing bytes needed to complete the Noise header or payload.
+    ///
+    /// Keeps track of how many more bytes are required to fully decode the current Noise frame.
```
plebhash, line 103: ditto

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +74,51 @@ use crate::Error::MissingBytes;
...
     missing_noise_b: usize,
+
+    /// Buffer for holding encrypted Noise data.
+    ///
+    /// Stores the incoming encrypted data until it is ready to be decrypted and processed.
```
plebhash, line 108: ditto

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -34,22 +74,51 @@ use crate::Error::MissingBytes;
...
     noise_buffer: B,
+
+    /// Buffer for holding decrypted Sv2 data.
+    ///
+    /// Stores the decrypted data from the Noise protocol until it is ready to be processed and
+    /// concerted into a Sv2 frame.
```
plebhash, line 114: ditto

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -25,28 +69,61 @@ use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory as Buffer};
 #[cfg(feature = "with_buffer_pool")]
 use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory, BufferPool};
 
+/// The buffer type for holding intermediate data during encoding.
+///
+/// When the `with_buffer_pool` feature is enabled, `Buffer` is a pool-allocated buffer type
+/// (`BufferPool`), which allows for more efficient memory management. Otherwise, it defaults to
+/// `BufferFromSystemMemory`.
```
plebhash, line 76:
```suggestion
/// When the `with_buffer_pool` feature is enabled, `Buffer` is a pool-allocated buffer type
/// (`BufferPool`), which allows for more efficient memory management. Otherwise, it defaults to
/// `binary_sv2::BufferFromSystemMemory`.
```

plebhash: we should reproduce this comment on `decoder.rs` because it also has a type alias for `Buffer`

plebhash (changes requested):
great progess!

left some comments

protocols/v2/codec-sv2/src/lib.rs
plebhash: note for clarity: I'm not arguing for us to use replit links in our docs

the point there is just that it is a good prototyping tool that helps us assert for code validity while we share examples with each other

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,43 @@
+//! # Stratum V2 Codec Decoder
```
plebhash, line 1: "Stratum V2 Codec Decoder" feels a bit redundant.

```suggestion
//! # Stratum V2 Codec Decoder
//! # Decoder
```

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,47 @@
+//! # Stratum V2 Codec Encoder
```
plebhash, line 1: "Stratum V2 Codec Encoder" feels a bit redundant.

```suggestion
//! # Encoder
```

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,47 @@
...
+//!
+//! This module provides functionality for encoding Stratum V2 messages, including support for the
+//! Noise protocol for secure communication.
```
plebhash, line 4: https://github.com/stratum-mining/stratum/pull/1111#pullrequestreview-2248938300

```suggestion
//! Provides functionality for encoding Stratum V2 messages, including support for the
//! Noise protocol for secure communication.
```

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,43 @@
...
+//!
+//! This module provides functionality for decoding Stratum V2 messages, including support for
```
plebhash, line 3: following suggestion number 5 of https://github.com/stratum-mining/stratum/pull/1111#pullrequestreview-2248938300

> Avoid including language like "This struct performs ABC" or "This is a trait defining XYZ", just say "Performs ABC" or "Defines XYZ".

```suggestion
//! Provides functionality for decoding Stratum V2 messages, including support for
```

protocols/v2/codec-sv2/src/lib.rs
plebhash: proposal for standardizing examples: https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10755951

plebhash:
@rrybarczyk one important detail to keep in mind here

we will also need to update `protocols/v2/sv2-ff2/sv2.h` on this PR

see https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10766163 for more details

protocols/v2/codec-sv2/src/lib.rs
Shourya742: When we run `cargo test`, all the unit tests and doc tests are executed, and our CI already includes them.

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,47 @@
...
+//! ## Features
+//!
+//! Two primary encoders are defined:
+//! * **Standard Encoder**:
+//! * **Noise Encoder**:
+//!
+//! ## Types
+//!
+//! * `Encoder`: Encoder for Sv2 frames with Noise protocol support.
+//! * `NoiseEncoder`: Encoder for Sv2 frames without Noise protocol support (requires the
```
Shourya742, line 15: We should move them out of types.

Shourya742 (changes requested):

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,47 @@
...
+//!   `noise_sv2` feature).
+//! * `Item<T>`: Represents an SV2 frame that is to be encoded and optionally encrypted using the
+//!   Noise protocol.
+//! * `Buffer`: Holds intermediate data during encoding, which can be either a pool-allocated
+//!   buffer or a system memory buffer depending on the feature flags.
+//! * `Slice`: Stores the serialized SV2 frame data before transmission.
```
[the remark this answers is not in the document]
plebhash: no, we should not

also, the `## Types` section is redundant

when we render the docs via `cargo doc` command, all of those items are automatically rendered

IMHO we should avoid redundant docs as much as possible, because they can easily be overlooked when changes are applied to the code in the future, which will lead to outdated stuff.

https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10799306

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,43 @@
...
+//! the Noise protocol for secure communication.
+//!
+//! ## Features
```
plebhash, line 6: I usually feel the word "Feature" can be confusing, because in the context of Rust it usually means "feature flag" for `cargo`.

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,47 @@
+//! # Stratum V2 Codec Encoder
+//!
+//! This module provides functionality for encoding Stratum V2 messages, including support for the
+//! Noise protocol for secure communication.
+//!
+//! ## Features
```
plebhash, line 6: I usually feel the word "Feature" can be confusing, because in the context of Rust it usually means "feature flag" for `cargo`.

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: Agree. What do you like instead? Optional Functionality, Configuration Options, Build Options, something else?

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,43 @@
...
+//!
+//! * **Standard Decoder**: Decodes Stratum V2 frames without encryption.
+//! * **Noise Decoder**: Decodes Stratum V2 frames with Noise protocol encryption.
+//!
+//! ## Types
+//!
+//! * `WithoutNoise`: Decoder for Sv2 frames without Noise protocol support.
+//! * `WithNoise`: Decoder for Sv2 frames with Noise protocol support (requires the `noise_sv2`
+//!   feature).
```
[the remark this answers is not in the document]
rrybarczyk: Removed.

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: Sorry, "ditto" to which other comment?

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: Went with "Build Options".

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: Agree. Added as TODO in #1129.

protocols/v2/codec-sv2/src/encoder.rs
rrybarczyk: Agree. Added to #1129 as TODO.

protocols/v2/codec-sv2/examples/encrypted.rs
```diff
@@ -0,0 +1,167 @@
...
+    // Read the payload into the decoder buffer
+    stream_receiver
+        .read_exact(decoder_buf)
+        .expect("Failed to read the encoded frame payload");
+
+    // INCORRECT PANIC HERE (DO NOT MERGE):
+    // Decode the frame
+    let _decoded_frame = decoder
+        .next_frame(&mut receiver_state)
+        .expect("Failed to decode the frame");
```
rrybarczyk, line 156: This panics with:

```sh
thread 'main' panicked at v2/codec-sv2/examples/encrypted.rs:156:10:
Failed to decode the frame: MissingBytes(18)
[Finished running. Exit status: 101]
```

@Shourya742, are you able to identify if the noise protocol has any errors that would cause this panic?

@fi3, are you able to identify any errors in the codec and framing related logic that would cause this panic?

protocols/v2/codec-sv2/Cargo.toml
```diff
@@ -16,9 +16,10 @@ const_sv2 = { version = "2.0.0", path = "../../../protocols/v2/const-sv2"}
...
 tracing = { version = "0.1"}
 
-
+[dev-dependencies]
+key-utils = { version = "^1.0.0", path = "../../../utils/key-utils" }
 
 [features]
 with_serde = ["binary_sv2/with_serde", "serde", "framing_sv2/with_serde", "buffer_sv2/with_serde"]
 with_buffer_pool = ["framing_sv2/with_buffer_pool"]
-no_std = []
```
jbesraa, line 24: is this EOF line removal intended?

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,13 @@
+//! # Decoder
+//!
+//! Provides functionality for decoding Stratum V2 messages, including support for
+//! the Noise protocol for secure communication.
+//!
+//! ## Usage
+//!
+//! This module is designed to be used to decode incoming Sv2 messages, with optional Noise
+//! protocol encryption support for secure communication.
```
jbesraa, line 9:
```suggestion
//! protocol decryption support for secure communication.
```

jbesraa, line 9: I am not sure the mention of `for secure communication` is relevant here or in L3, IIUC the codec does not care or know anything about secure/not secure. it is just able to decode/encode sv2 messages with or without noise. In regards to what noise is, I think that that should be part of `noise-sv2`

jbesraa: if this is decoder, should it not be doing decryption?

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -23,6 +33,14 @@ use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory as Buffer};
 
 #[cfg(feature = "with_buffer_pool")]
 use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory, BufferPool};
+
+// The buffer type for holding intermediate data during decoding.
+//
+// When the `with_buffer_pool` feature is enabled, `Buffer` is a pool-allocated buffer type
```
jbesraa, line 39: where is the definition of `Buffer` without the feature flag `with_buffer_pool` ?

GitGab19:
Hey @rrybarczyk, is this ready for another round of reviews?

protocols/v2/codec-sv2/src/error.rs
```diff
@@ -1,69 +1,105 @@
...
 
+/// A type alias for results returned by the `codec_sv2` modules.
+///
+/// `Result` is a convenient wrapper around the `core::result::Result` type,
+/// using the `Error` enum defined in this crate as the error type.
 pub type Result<T> = core::result::Result<T, Error>;
 
+/// Enumeration of possible errors in the `codec_sv2` module.
+///
+/// This enum represents various errors that can occur within the `codec_sv2` module, including
```
jbesraa, line 22: I think using `codec_sv2 crate` instead of `.. module` would be more consistent.

Also, you can make the crate names clickable in docs.rs by doing the following:
```md
/// Enumeration of possible errors in the [`codec_sv2`] module.
///
/// This enum represents various errors that can occur within the [`codec_sv2`] module, including
/// errors from related crates like [`binary_sv2`], [`framing_sv2`], and [`noise_sv2`].
///
/// [`codec_sv2`]: https://docs.rs/codec_sv2
/// [`binary_sv2`]: https://docs.rs/binary_sv2
/// [`framing_sv2`]: https://docs.rs/framing_sv2
/// [`noise_sv2`]: https://docs.rs/noise_sv2
```

jbesraa (commented):
Very comprehensive overall, just added few nits

protocols/v2/codec-sv2/examples/encrypted.rs
rrybarczyk: @fi3 you are correct. The examples have been updated to read from the TCP stream in a loop. This fixed the panic. 

However, when I run the examples, they work. But @plebhash has pulled down the code and ran them and they fail with [this stderr ](https://github.com/user-attachments/files/17297047/message.1.txt).

Are you able to run them successfully? Any idea has to why the errors are occurring?

protocols/v2/codec-sv2/Cargo.toml
```diff
@@ -16,9 +16,10 @@ const_sv2 = { version = "2.0.0", path = "../../../protocols/v2/const-sv2"}
 buffer_sv2 = { version = "1.0.0", path = "../../../utils/buffer"}
 tracing = { version = "0.1"}
 
-
```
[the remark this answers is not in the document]
rrybarczyk: I am not sure why the change appears this way on the GitHub UI. What was added is the `key-utils` crate as a dev dependency to expose the `key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};` primitives for pub/priv key handling.

protocols/v2/codec-sv2/Cargo.toml
rrybarczyk: Again, I do not know what the GitHub UI is showing this was removed. This line is still present in the toml.

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: I made a lot of changes to the doc cmts. Give it another look and let me know what you think.

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: I added some extra language here with the last commit. Can you review now? I am a bit confused on how the code knows to default to `BufferFromSystemMemory` when the feature flag is disabled. I would expect there to be a type like:

```rs
#[cfg(not(feature = "with_buffer_pool"))]
type Buffer = Vec<u8>;
```

@fi3, do you have insights here?

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -1,3 +1,26 @@
...
+//!
+//! ### Buffer Management
+//!
+//! The encoders rely on buffers to hold intermediate data during the encoding process.
+//!
+//! - When the `with_buffer_pool` feature is enabled, the internal `Buffer` type is backed by a
+//!   pool-allocated buffer (`binary_sv2::BufferPool`), providing more efficient memory usage,
+//!   particularly in high-throughput scenarios.
+//! - If the feature is not enabled, a system memory buffer (`binary_sv2::BufferFromSystemMemory`)
+//!   is used for simpler applications where memory efficiency is less critical.
```
rrybarczyk, line 22: @fi3, can you review my buffer understanding?

protocols/v2/codec-sv2/src/encoder.rs
```diff
@@ -25,28 +48,94 @@ use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory as Buffer};
...
+// may not offer the same memory efficiency as the pool-allocated version (`BufferPool`).
 #[cfg(not(feature = "with_buffer_pool"))]
 type Slice = Vec<u8>;
 
+// A buffer slice used for holding serialized Sv2 frame data before transmission.
+//
+// When the `with_buffer_pool` feature is enabled, `Slice` defaults to a `buffer_sv2::Slice`, which
+// serves as a slice of the `Buffer` that stores the encoded data. It holds the frame's serialized
+// bytes temporarily, ensuring the data is ready for transmission or encryption, depending on
+// whether Noise protocol support is enabled.
```
rrybarczyk, line 78: @fi3, can you review my buffer and slice understanding?

protocols/v2/codec-sv2/README.md
```diff
@@ -0,0 +1,50 @@
...
+This crate provides two examples demonstrating how to encode and decode Sv2 frames:
+
+1. **[Unencrypted Example](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/codec-sv2/examples/unencrypted.rs)**:
+   Encode and decode standard Sv2 frames, detailing the message serialization and transmission
+   process for unencrypted communications.
+
+2. **[Encrypted Example](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/codec-sv2/examples/encrypted.rs)**:
+   Encode and decode Sv2 frames with Noise protocol encryption, detailing the entire encryption
+   handshake and transport phase and serialization and transmission process for encrypted
+   communications.
```
rrybarczyk, line 50: @plebhash is this overkill for the example section or do we like it?

plebhash: looks good to me!

protocols/v2/codec-sv2/examples/encrypted.rs
rrybarczyk: Nvm. I fixed with with appropriate usage of `#[cfg(feature = noise_sv2)]` decorators.

protocols/v2/codec-sv2/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Usage
+
+To use this crate, add it to your `Cargo.toml`:
+
+```toml
+[dependencies]
+codec_sv2 = "1.2.1"
+```
```
plebhash, line 28: see https://github.com/stratum-mining/stratum/pull/1111#discussion_r1791426219

rrybarczyk:
> Hey @rrybarczyk, is this ready for another round of reviews?

It is ready for another round of reviews!

protocols/v2/codec-sv2/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+This crate can be built with the following features:
```
plebhash, line 30:
```suggestion
This crate can be built with the following feature flags:
```

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: Added `examples/` and documentation surrounding the `MissingBytes` error situation.

protocols/v2/codec-sv2/src/decoder.rs
rrybarczyk: It happens in the import statement:

https://github.com/stratum-mining/stratum/blob/e65cb320173b0048ec65dc4e042d8013a7fa962d/protocols/v2/codec-sv2/src/decoder.rs#L21-L22

I have added a note in the `codec-sv2` technical debt issue #1129 to define this as a type rather than rely on the implication of the import statement.

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -1,3 +1,28 @@
...
+//!
+//! ### Buffer Management
+//!
+//! The decoders rely on buffers to hold intermediate data during the decoding process.
+//!
+//! - When the `with_buffer_pool` feature is enabled, the internal `Buffer` type is backed by a
+//!   pool-allocated buffer (`binary_sv2::BufferPool`), providing more efficient memory usage,
+//!   particularly in high-throughput scenarios.
+//! - If this feature is not enabled, a system memory buffer (`binary_sv2::BufferFromSystemMemory`)
+//!   is used for simpler applications where memory efficiency is less critical.
```
plebhash, line 24: the module is private, which means this header is for maintainer eyes only

whenever we make the module public we can change this, but for now let's document things as they are
```suggestion
// # Decoder
//
// Provides utilities for decoding messages held by Sv2 frames, with or without Noise protocol
// support.
//
// It includes primitives to both decode encoded standard Sv2 frames and to decrypt
// and decode Noise-encrypted encoded Sv2 frames, ensuring secure communication when required.
//
// ## Usage
// All messages passed between Sv2 roles are encoded as Sv2 frames. These frames are decoded using
// primitives in this module. There are two types of decoders for reading these frames: one for
// regular Sv2 frames (`StandardDecoder`), and another for Noise-encrypted frames
// (`StandardNoiseDecoder`). Both decoders manage the deserialization of incoming data and, when
// applicable, the decryption of the data upon receiving the transmitted message.
//
// ### Buffer Management
//
// The decoders rely on buffers to hold intermediate data during the decoding process.
//
// - When the `with_buffer_pool` feature is enabled, the internal `Buffer` type is backed by a
//   pool-allocated buffer (`binary_sv2::BufferPool`), providing more efficient memory usage,
//   particularly in high-throughput scenarios.
// - If this feature is not enabled, a system memory buffer (`binary_sv2::BufferFromSystemMemory`)
//   is used for simpler applications where memory efficiency is less critical.
```

protocols/v2/codec-sv2/src/encoder.rs
plebhash, line 22: the module is private, which means this header is for maintainer eyes only

whenever we make the module public we can change this, but for now let's document things as they are.

```suggestion
// # Encoder
//
// Provides utilities for encoding messages into Sv2 frames, with or without Noise protocol
// support.
//
// ## Usage
//
// All messages passed between Sv2 roles are encoded as Sv2 frames using primitives in this
// module. There are two types of encoders for creating these frames: one for regular Sv2 frames
// (`Encoder`), and another for Noise-encrypted frames (`NoiseEncoder`). Both encoders manage the
// serialization of outgoing data and, when applicable, the encryption of the data before
// transmission.
//
// ### Buffer Management
//
// The encoders rely on buffers to hold intermediate data during the encoding process.
//
// - When the `with_buffer_pool` feature is enabled, the internal `Buffer` type is backed by a
//   pool-allocated buffer (`binary_sv2::BufferPool`), providing more efficient memory usage,
//   particularly in high-throughput scenarios.
// - If the feature is not enabled, a system memory buffer (`binary_sv2::BufferFromSystemMemory`)
//   is used for simpler applications where memory efficiency is less critical.
```

protocols/v2/codec-sv2/examples/unencrypted.rs
```diff
@@ -0,0 +1,135 @@
...
+//! * Create an arbitrary custom message type (`CustomMessage`) for encoding and decoding.
+//! * Encode the message into a Sv2 frame.
+//! * Send the encoded frame over a TCP connection.
+//! * Decode the Sv2 frame on the receiving side and extract the original message.
+//!
+//! ## Run
+//!
+//! ```
+//! cargo run --example unencrypted
+//! ```
```
plebhash, line 15:
```suggestion
// # Using Sv2 Codec Without Encryption
//
// This example demonstrates how to use the `codec-sv2` crate to encode and decode Sv2 frames
// without encryption over a TCP connection. It showcases how to:
//
// * Create an arbitrary custom message type (`CustomMessage`) for encoding and decoding.
// * Encode the message into a Sv2 frame.
// * Send the encoded frame over a TCP connection.
// * Decode the Sv2 frame on the receiving side and extract the original message.
//
// ## Run
//
// ```
// cargo run --example unencrypted
// ```
```

protocols/v2/codec-sv2/src/encoder.rs
plebhash: `BufferPool` uses unsafe Rust, which is probably an important detail to be mentioned here.

plebhash: aside from this detail on unsafe Rust, the description we have here is pretty aligned with my current understanding of `with_buffer_pool`

plebhash: another improvement here is that we could embed these links:
- https://docs.rs/buffer_sv2/latest/buffer_sv2/struct.BufferPool.html
- https://docs.rs/buffer_sv2/latest/buffer_sv2/struct.BufferFromSystemMemory.html

not only here, but also in the other references to `Buffer` and `BufferPool`

jbesraa:
@rrybarczyk  This PR is looking kinda complete, could you please squash the commits to something similar to what we have here https://github.com/stratum-mining/stratum/pull/1111/commits so it is easier to give this some final reviews?

protocols/v2/codec-sv2/src/error.rs
```diff
@@ -1,69 +1,105 @@
...
-    /// Error if Noise protocol state is not as expected
-    UnexpectedNoiseState,
+    InvalidStepForInitiator,
+
+    /// Invalid step for responder in the Noise protocol.
     #[cfg(feature = "noise_sv2")]
     InvalidStepForResponder,
+
+    /// Missing bytes in the Noise protocol.
+    MissingBytes(usize),
```
GitGab19, line 48: Shouldn't this have a reference to Noise in its name? If it's just used to indicate missing bytes in the Noise protocol

protocols/v2/codec-sv2/README.md
```diff
@@ -0,0 +1,48 @@
+# `codec_sv2`
+
+[![crates.io](https://img.shields.io/crates/v/codec_sv2.svg)](https://crates.io/crates/codec_sv2)
+[![docs.rs](https://docs.rs/codec_sv2/badge.svg)](https://docs.rs/codec_sv2)
+[![rustc+](https://img.shields.io/badge/rustc-1.75.0%2B-lightgrey.svg)](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html)
+[![license](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg)](https://github.com/stratum-mining/stratum/blob/main/LICENSE.md)
+
```
GitGab19, line 7: Add the badge here (cc @Shourya742)

GitGab19 (commented):
LGTM, docs are very clear 👏 
I left some nits, the major one is about adding the `code-cov` badge into `README.md`

protocols/v2/codec-sv2/src/error.rs
```diff
@@ -1,69 +1,105 @@
...
+    AeadError(AeadError),
+
+    /// Binary Sv2 data format error.
     BinarySv2Error(binary_sv2::Error),
+
+    /// Framing Sv2 error.
+    FramingError(FramingError),
+
+    /// Framing Sv2 error.
     FramingSv2Error(framing_sv2::Error),
```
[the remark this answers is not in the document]
rrybarczyk: Yep, I have it listed in issue #1129 to address.

protocols/v2/codec-sv2/README.md
```diff
@@ -0,0 +1,49 @@
+# `codec_sv2`
+
+[![crates.io](https://img.shields.io/crates/v/codec_sv2.svg)](https://crates.io/crates/codec_sv2)
+[![docs.rs](https://docs.rs/codec_sv2/badge.svg)](https://docs.rs/codec_sv2)
+[![rustc+](https://img.shields.io/badge/rustc-1.75.0%2B-lightgrey.svg)](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html)
+[![license](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg)](https://github.com/stratum-mining/stratum/blob/main/LICENSE.md)
+[![codecov](https://codecov.io/gh/stratum-mining/stratum/branch/main/graph/badge.svg?flag=codec_sv2-coverage)](https://codecov.io/gh/stratum-mining/stratum)
```
rrybarczyk, line 7: @Shourya742 is this right? Looks like it is showing 0%, but it should be 10.61% I think.

Shourya742 (changes requested):
Really well-written! It cleared up most of my doubts just by going through the documentation. Just a few minor nits, but it's green from my side.

protocols/v2/codec-sv2/src/lib.rs
```diff
@@ -34,20 +68,74 @@ pub use buffer_sv2;
...
+    /// stores a `Responder` object, which contains the necessary state and cryptographic materials
+    /// for the responder's part in the Noise handshake.
+    Responder(Box<noise_sv2::Responder>),
+}
+
+/// Represents the state of the Noise protocol codec during different phases: initialization,
+/// handshake, or transport mode, where encryption and decryption are fully operational.
+///
+/// The state transitions from initialization (`NotInitialized`) to handshake (`HandShake`)
+/// and finally to transport mode (`Transport`) as the encryption handshake is completed.
```
jbesraa, line 102: Variables here and else where are not linked correctly and not consistently. For example (`Transport`) is referenced is different ways that also usually dont link to the actual object. 

If we leave the variables like this, in docs.rs the variables wont be clickable(they will be gray instead of blue).


I suggest to change it to the following:
```suggestion
/// and finally to transport mode [`State::Transport`] as the encryption handshake is completed.
///
/// [`State::Transport`]: crate::State::Transport
```

jbesraa (commented):
Added few nits.
The main thing I noticed is inconsistent-ency on how variables/structs/others are being referenced in comments.  There are a lot of (`Variable`)  instances that I don't think are doing much in docs.rs and not sure whats the goal behind this way of referencing.

protocols/v2/codec-sv2/src/lib.rs
```diff
@@ -1,3 +1,37 @@
...
+//! - `with_buffer_pool`: Enables buffer pooling for more efficient memory management.
+//! - `with_serde`: builds [`binary_sv2`] and [`buffer_sv2`] crates with `serde`-based encoding and
+//!   decoding. Note that this feature flag is only used for the Message Generator, and deprecated
+//!   for any other kind of usage. It will likely be fully deprecated in the future.
+//!
+//! ## Examples
+//!
+//! See the examples for more information:
+//!
+//! - [Unencrypted Example](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/codec-sv2/examples/unencrypted.rs)
```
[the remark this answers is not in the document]
plebhash: what is broken?

it works fine for me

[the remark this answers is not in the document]
GitGab19: Yeah I think @jbesraa is referring to the link. It is broken, but it will work after the merge

protocols/v2/codec-sv2/README.md
plebhash: assuming the URL will remain the same, let's save one extra PR and just leave it here

protocols/v2/codec-sv2/examples/unencrypted.rs
```diff
@@ -0,0 +1,135 @@
...
+        .expect("Failed to get the frame header");
+    let decoded_msg: CustomMessage = binary_sv2::from_bytes(decoded_frame.payload())
+        .expect("Failed to extract the message from the payload");
+
+    // Assert that the decoded message is as expected
+    assert_eq!(decoded_frame_header.msg_type(), CUSTOM_MSG_TYPE);
+    assert_eq!(decoded_msg.nonce, nonce);
+}
+
+fn sender_side(
```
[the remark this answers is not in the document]
rrybarczyk: Can you add this to the #1129 tracker issue?

GitGab19 (approved):

pavlenex:
Hey @rrybarczyk, your PR cannot be rebased due to conflicts. Could you resolve them, please?

Shourya742 (approved):

PR #1043, opened 2024-07-08
optimize: call function once and destructure

marathon-gary:
Perhaps the compiler would catch this but methinks not so made this small change and saved CPU cycles for people.


marathon-gary:
Maybe not given all the benchmarking results?

jbesraa (approved):

Shourya742 (approved):
ACK 950d9d87708f2293e58cd9c8d88fd819d0822503

PR #1044, opened 2024-07-09
Ignore `Cargo.lock` files

jbesraa:
add `Cargo.lock` files to `.gitignore`


Shourya742 (approved):
Strong ACK. This is much needed.

marathon-gary:
Ack. 
But for consideration leaving this here: https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control

plebhash:
rebasing to get this merged

plebhash (approved):

johnnyasantoss:
I don't think this is a good idea. Lockfiles exist for a reason, and if you read the link that @marathon-gary sent, you will see why. 
There are multiple dependencies on the workspace that are unlocked. Clearly, it wasn't committed along with the dependency changes.
See it in action: `fd -e toml Cargo -x sh -c 'cd {//} && cargo generate-lockfile'` this will run the cargo command for every subfolder with a Cargo.toml file. After running that, you should see a lot of lockfiles generated/updated on the git working tree.

1) This can lead to flaky CI/CD pipelines, as cargo will not know the exact version to look for.
2) Can lead to software supply chain attack (there's no shortage in those - [1](https://snyk.io/blog/polyfill-supply-chain-attack-js-cdn-assets/) [2](https://www.crowdstrike.com/blog/cve-2024-3094-xz-upstream-supply-chain-attack/) [3](https://snyk.io/blog/npm-faker-package-open-source-libraries/) [4](https://snyk.io/blog/peacenotwar-malicious-npm-node-ipc-package-vulnerability/))
3) Will most likelly lead to *runs on my machine* type of issue, preventing debugging of hard problems, git bisecting to find a regression/bug, etc.
4) impossible to have reproducible builds (not sure if this is goal)
5) impossible auditing

johnnyasantoss:
Btw, we already ignore a few lockfiles which is not ideal.

Regarding inter-workspace dependency management, I think we can stop using the version+path definition on `Cargo.toml` files as that just leads to unwanted changes on lock files (which this is the root of the problem - imo).
For example: commit 63b8d23b51378a781d951320eab79f46ee4f7026 updated the `sv1_api` version but didn't relock the other projects, leading to unwanted changes every time you checkout `dev` and use an IDE/LSP that checks for project correctness

jbesraa:
I am aware of the link you shared but dont agree with you. I think its enough to have msrv check on each pr that makes sure the pr is always building on at least 1.75.

Regrding your points: 
1- why flaky tests? Cargo.toml still limits the different crates versions you know. I dont see any reason why the ci would be flaky.
2- this is not necessarily about lock files but management and inspecting of new crates added to the project.
The last three points are theoretical and I dont have much to say about that.

Anyway, my feeling is you just pouring here what you read in the link but in practice things are quite different. It would be good to try and show how this is failing rather than "it might fail"..

johnnyasantoss:
@jbesraa 
> Regrding your points: 1- why flaky tests? Cargo.toml still limits the different crates versions you know. I dont see any reason why the ci would be flaky.

Cargo.toml does not **limit** crates versions, it does specify the range that we [**want**](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio). Without a lock, it will perform dependency selection every time a fetch is run, leading to flaky CI. Also, crates can be yanked from registry. 
We don't use `=<version>` so all our ranges are [caret range](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) which means for all crates above `1.0.0` when declared on Cargo.toml as `1.0.1` are interpreted as `>=1.0.1, <2.0.0` and so it will select the latest on that range.

And, in practice, if you checkout `dev` (5d569bbdb0f6d1dc7bc97a62abff9d6ded4b425d) and run `fd -e toml Cargo -j1 -x sh -c 'cd {//} && cargo +1.75 fetch && cargo +1.75 build'` you will see *already* a few range issues (internal, imagine external deps).
```log
error: failed to select a version for the requirement `network_helpers_sv2 = "^0.1.0"`
candidate versions found which didn't match: 2.0.0
location searched: /path/to/stratum/roles/roles-utils/network-helpers
required by package `template-provider-test v0.1.0 (/path/to/stratum/examples/template-provider-test)`
```

On 2) I agree with you. Lockfiles are not magical we still have to inspect dependencies before adding them to the project but to ensure that we always get the version that was inspected we need a lockfile.

Don't get me wrong, I'm not saying we can't merge this. I'm just saying this is counterproductive.

jbesraa:
I don't have strong opinion about this PR and I would prefer it being closed rather than trying to get it merged. It could be a good convince for devs and also for PR review process, but nothing crucial.

jbesraa:
Re-opening this as I am not able to rebase regularly on my local machine. 
Each time I `git rebase -i dev` the rebase stops somewhere complaining about the `Cargo.lock` stuff. If anyone have a strong opinion about not merging this please come forward, otherwise I think we should land this as the dev experience is currently kinda shit

plebhash:
I'm doing some research on this topic. Here's some findings:

## `protocols/Cargo.lock`

I believe we can safely delete and gitignore `protocols/Cargo.lock`. All those crates are only libs, and there's no reason to keep a `Cargo.lock` file for libs. Some references:
- [StackOverflow Q&A](https://stackoverflow.com/questions/62861623/should-cargo-lock-be-committed-when-the-crate-is-both-a-rust-library-and-an-exec)
- [LDK/`rust-lightning`](https://github.com/lightningdevkit/rust-lightning) doesn't keep any `Cargo.lock` files
- [`rust-bitcoin`](https://github.com/rust-bitcoin/rust-bitcoin) doesn't keep any `Cargo.lock` files

## `roles/Cargo.lock`

The crates on `roles` do generate binaries, which means it's a good engineering practice to keep `roles/Cargo.lock` on our git history.

The root cause of our issues on this specific file comes from the fact that we often bump dependency versions on `roles/*/Cargo.toml` without committing the same updates to `roles/Cargo.lock`.

That could be fixed by:
- https://github.com/stratum-mining/stratum/pull/1103 updates `roles/Cargo.lock` according to the current state of the repo
- https://github.com/stratum-mining/stratum/pull/1102 which introduces a new CI action that blocks PRs that change dependencies on the `roles` workspace from being merged if they don't update `roles/Cargo.lock` accordingly.

## `benches/Cargo.lock`

I don't have much insight into this crate. Given that we have https://github.com/stratum-mining/benchmarking-tool, I wonder if this isn't obsolete and the crate could simply be removed as a whole? cc @GitGab19

rrybarczyk:
> I'm doing some research on this topic. Here's some findings:
> 
> ## `protocols/Cargo.lock`
> I believe we can safely delete and gitignore `protocols/Cargo.lock`. All those crates are only libs, and there's no reason to keep a `Cargo.lock` file for libs. Some references:
> 
> * [StackOverflow Q&A](https://stackoverflow.com/questions/62861623/should-cargo-lock-be-committed-when-the-crate-is-both-a-rust-library-and-an-exec)
> * [LDK/`rust-lightning`](https://github.com/lightningdevkit/rust-lightning) doesn't keep any `Cargo.lock` files
> * [`rust-bitcoin`](https://github.com/rust-bitcoin/rust-bitcoin) doesn't keep any `Cargo.lock` files
> 
> ## `roles/Cargo.lock`
> The crates on `roles` do generate binaries, which means it's a good engineering practice to keep `roles/Cargo.lock` on our git history.
> 
> The root cause of our issues on this specific file comes from the fact that we often bump dependency versions on `roles/*/Cargo.toml` without committing the same updates to `roles/Cargo.lock`.
> 
> That could be fixed by:
> 
> * [update `roles/Cargo.lock` #1103](https://github.com/stratum-mining/stratum/pull/1103) updates `roles/Cargo.lock` according to the current state of the repo
> * [add `roles-lockfile.yaml` #1102](https://github.com/stratum-mining/stratum/pull/1102) which introduces a new CI action that blocks PRs that change dependencies on the `roles` workspace from being merged if they don't update `roles/Cargo.lock` accordingly.
> 
> ## `benches/Cargo.lock`
> I don't have much insight into this crate. Given that we have https://github.com/stratum-mining/benchmarking-tool, I wonder if this isn't obsolete and the crate could simply be removed as a whole? cc @GitGab19
> 
> if we decide to keep it, a similar strategy to `roles/Cargo.lock` could be applied here

This looks good to me. Basically library crates should NOT have a `Cargo.lock`. If it is a binary, it SHOULD have the `Cargo.lock`.

Thinking longer term, this now gets us back to the need to separate the library from binaries in the repo itself...the `roles` crate really should be a library. And then in another repo, we should have the binary applications (with the `roles`'s toml config files there). 

For now, @plebhash's suggestions make sense to me.

Issue #902, opened 2024-05-09
IdentifyTransactions message not implemented


lorbax:
IdentifyTransactions and .Succes are not implemented. The JDS just ignores the cases where it necessary to send this message (collisions with short hashes or impossible to reconstruct the `tx_hash_list_hash`). If the JDC receives such a message, [it responds with an empty vector](https://github.com/stratum-mining/stratum/blob/dev/roles/jd-client/src/lib/job_declarator/message_handler.rs#L45 t).

Good first issue

rockcoolsaint:
@lorbax I don't quite understand what is to be done here.
What is the expected code deliverable?

What does this statement mean? `The JDS just ignores the cases where it necessary to send this message (collisions with short hashes or impossible to reconstruct the tx_hash_list_hash)`

marathon-gary:
https://stratumprotocol.org/specification/06-Job-Declaration-Protocol/#617-identifytransactions-server-client
> Sent by the Server in response to a DeclareMiningJob message indicating it detected a collision in the `tx_short_hash_list`, or was unable to reconstruct the `tx_hash_list_hash`.

The client receives an IdentifyTransactions message from the server, indicating a collision or inability to reconstruct the hash list of tx's. Currently, there is [just an empty vec returned.](https://github.com/stratum-mining/stratum/blob/dev/roles/jd-client/src/lib/job_declarator/message_handler.rs#L45) 

According to spec it should be providing:
> The full list of transaction data hashes used to build the mining job in the corresponding DeclareMiningJob message

The original `DeclareMiningJob` only contains `tx_short_hash_list` and `tx_hash_list_hash`. An `IdentifyTransactions` message indicates there was a collision in the `short hash` list or the `hash list hash` couldn't be recreated.

I assume the client just looks up the original list of tx's and then sends it back in the `.Success` message.

I will take a pass at a PR for this

lorbax:
> https://stratumprotocol.org/specification/06-Job-Declaration-Protocol/#617-identifytransactions-server-client
> 
> > Sent by the Server in response to a DeclareMiningJob message indicating it detected a collision in the `tx_short_hash_list`, or was unable to reconstruct the `tx_hash_list_hash`.
> 
[...]

PR #1050, opened 2024-07-12
impl: IdentifyTransactions.Success for jdc

marathon-gary:
We scrub the last declared mining jobs and collect a transaction list from the JD. Then mapping over the found transactions we convert the txid in the binary_sv2::U256 type. From this we can create a new Seq064K<U256> to send back from the JDC to the JDS that encountered a hash collision or was unable to recreate the tx_hash_list_hash.

Probably some code conciseness clean up I could do here but wanted to push the PR to gather feedback.

### DIFF
--- a/roles/jd-client/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-client/src/lib/job_declarator/message_handler.rs
@@ -1,4 +1,5 @@
 use super::JobDeclarator;
+use binary_sv2::{Seq064K, U256};
 use roles_logic_sv2::{
     handlers::{job_declaration::ParseServerJobDeclarationMessages, SendTo_},
     job_declaration_sv2::{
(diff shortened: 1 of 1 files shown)


plebhash:
I haven't dived into the code in detail, but a few points worth noting:

- #902 describes the fact that neither `IdentifyTransactions` nor `IdentifyTransactions.Success` are implemented. So it doesn't make sense to only implement `IdentifyTransactions.Success` in isolation.
- We need to assert the implementation of both messages for correctness, which can only be done via MG tests. [Here](https://www.youtube.com/watch?v=Gic7qrmWdU4&list=PL4okm1s0ZpT9YveJBiU51gaaOmNJviugi)'s a tutorial with some foundational knowledge on how to get started.
- Given that the PR is still WIP, it should be set as a Draft.

I don't think the `good-first-issue` was an accurate label and we probably made a mistake when we applied it to this task. Writing MG tests and making sure the implementation is sane is not necessarily trivial.

But I don't want to discourage you @marathon-gary , just letting you know that the challenge is probably bigger than originally perceived.

marathon-gary:
Also Ack on both sides of the messaging not being implemented, I definitely misread the issue. Will keep working on this to learn the codebase.

plebhash:
> Is there a good code example of the MG testing to reference? I'll look for some but referrals for starting points are always welcome.

Watch the youtube videos I embedded on my previous comment. It's the closest thing we have to some educational material on MG tests so far.

rrybarczyk:
Closing until we address roles refactor after protocols refactor.

PR #1055, opened 2024-07-16
Fix JDC fallback to solo-mining

lorbax:
If the JDC switches to solo_mining, then it initializes as solo mining (using `initialize_jd_as_solo_miner`).
This function initialize a `DownstreamMiningNode` with `jd` field set to `None`. This is the right behaviour, but makes panic this [unwrap](https://github.com/stratum-mining/stratum/blob/dev/roles/jd-client/src/lib/downstream.rs#L379).
Since we are doing solo mining, it should be care of the downstream to change the coinbase downstream, instead of wrapping a None, the `coinbase_tx_prefix` and `suffix` are ignored. 

Perhaps this is not the best way to do that, feedback needed.

In the commit message is writtend how to test this PR.

This PR  is on top of [PR1001](https://github.com/stratum-mining/stratum/pull/1001).
This PR closes [issue 844](https://github.com/stratum-mining/stratum/issues/844)

How to test this PR:
The testing environment is very similar to the one described on PR1001. The only difference is that here we do not have a "good" pool to fallback into. So apply suggestions on PR1001 with the only difference that the jdc config is 
```toml
[[upstreams]]
authority_pubkey = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
pool_address = "127.0.0.1:44254"
jd_address = "127.0.0.1:34264"
pool_signature = "Stratum v2 SRI Pool"
```
(you can avoid running a "good" pool then)

EDIT: the MG test is removed

### DIFF
--- a/roles/jd-client/src/lib/downstream.rs
+++ b/roles/jd-client/src/lib/downstream.rs
@@ -47,7 +47,7 @@ pub struct DownstreamMiningNode {
     miner_coinbase_output: Vec<TxOut>,
     // used to retreive the job id of the share that we send upstream
     last_template_id: u64,
-    jd: Option<Arc<Mutex<JobDeclarator>>>,
+    pub jd: Option<Arc<Mutex<JobDeclarator>>>,
 }
 
 #[allow(clippy::large_enum_variant)]
@@ -376,12 +376,13 @@ impl DownstreamMiningNode {
         let to_send = to_send.into_values();
         for message in to_send {
             let message = if let Mining::NewExtendedMiningJob(job) = message {
-                let jd = self_mutex.safe_lock(|s| s.jd.clone()).unwrap().unwrap();
-                jd.safe_lock(|jd| jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone())
+                if let Some(jd) = self_mutex.safe_lock(|s| s.jd.clone()).unwrap() {
+                    jd.safe_lock(|jd| {
+                        jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone();
+                        jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone();
+                    })
                     .unwrap();
-                jd.safe_lock(|jd| jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone())
-                    .unwrap();
-
+                }
                 Mining::NewExtendedMiningJob(job)
             } else {
                 message
@@ -514,7 +515,7 @@ impl
 
     fn handle_update_channel(
         &mut self,
-        _: UpdateChannel,
+        m: UpdateChannel,
     ) -> Result<SendTo<UpstreamMiningNode>, Error> {
         if !self.status.is_solo_miner() {
             // Safe unwrap alreay checked if it cointains upstream with is_solo_miner
@@ -522,7 +523,16 @@ impl
                 self.status.get_upstream().unwrap(),
             ))
         } else {
-            todo!()
+            let maximum_target =
+                roles_logic_sv2::utils::hash_rate_to_target(m.nominal_hash_rate.into(), 10.0)?;
+            self.status
+                .get_channel()
+                .update_target_for_channel(m.channel_id, maximum_target.clone().into());
+            let set_target = SetTarget {
+                channel_id: m.channel_id,
+                maximum_target,
+            };
+            Ok(SendTo::Respond(Mining::SetTarget(set_target)))
         }
     }
 
--- a/benches/Cargo.lock
+++ b/benches/Cargo.lock
@@ -373,7 +373,7 @@ dependencies = [
 
 [[package]]
 name = "buffer_sv2"
-version = "1.0.0"
+version = "1.1.0"
 dependencies = [
  "aes-gcm",
 ]
@@ -497,7 +497,7 @@ checksum = "98cc8fbded0c607b7ba9dd60cd98df59af97e84d24e49c8557331cfc26d301ce"
 
 [[package]]
 name = "codec_sv2"
-version = "1.1.0"
+version = "1.2.0"
 dependencies = [
  "binary_sv2",
  "buffer_sv2",
@@ -1564,7 +1564,7 @@ checksum = "81cdd64d312baedb58e21336b31bc043b77e01cc99033ce76ef539f78e965ebc"
 
 [[package]]
 name = "sv1_api"
-version = "1.0.0"
+version = "1.0.1"
 dependencies = [
  "binary_sv2",
  "bitcoin_hashes 0.3.2",
--- a/roles/jd-client/src/lib/template_receiver/mod.rs
+++ b/roles/jd-client/src/lib/template_receiver/mod.rs
@@ -57,9 +57,16 @@ impl TemplateRx {
         test_only_do_not_send_solution_to_tp: bool,
     ) {
         let mut encoded_outputs = vec![];
-        miner_coinbase_outputs
-            .consensus_encode(&mut encoded_outputs)
-            .expect("Invalid coinbase output in config");
+        // jd is set to None in initialize_jd_as_solo_miner (in this case we need to take the first output as done by JDS)
+        if jd.is_none() {
+            miner_coinbase_outputs[0]
+                .consensus_encode(&mut encoded_outputs)
+                .expect("Invalid coinbase output in config");
+        } else {
+            miner_coinbase_outputs
+                .consensus_encode(&mut encoded_outputs)
+                .expect("Invalid coinbase output in config");
+        }
         let stream = tokio::net::TcpStream::connect(address).await.unwrap();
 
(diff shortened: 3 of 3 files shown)


GitGab19:
Since this PR is built on PR #1001, it's also blocked by #1059

roles/jd-client/src/lib/downstream.rs
```diff
@@ -376,12 +376,12 @@ impl DownstreamMiningNode {
...
-                jd.safe_lock(|jd| jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone())
-                    .unwrap();
-                jd.safe_lock(|jd| jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone())
-                    .unwrap();
-
+                if let Some(jd) = self_mutex.safe_lock(|s| s.jd.clone()).unwrap() {
+                    jd.safe_lock(|jd| jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone())
+                        .unwrap();
+                    jd.safe_lock(|jd| jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone())
+                        .unwrap();
```
Shourya742, line 383:
```suggestion
                     jd.safe_lock(|jd| {
                        jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone();
                        jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone();
                    }).unwrap();
```

Shourya742 (changes requested):
utack. Few nits. Can we remove these lock files also.

roles/jd-client/src/lib/downstream.rs
lorbax: I don't commit this suggestion because of the policies on how nice commit history must be, if you need credits for this mod I can add it in the commit message. Let me know.

Shourya742: No credit needed. I see you included the suggestion in the latest push, and it works perfectly.

roles/translator/src/lib/downstream_sv1/downstream.rs
```diff
@@ -110,6 +110,7 @@ impl Downstream {
         host: String,
         difficulty_config: DownstreamDifficultyConfig,
         upstream_difficulty_config: Arc<Mutex<UpstreamDifficultyConfig>>,
+        task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>,
```
Shourya742, line 113:
```suggestion
        task_collector: Arc<Mutex<Vec<AbortHandle>>>,
```
Any specific reason for adding string here?

test/config/change-upstream/jdc-config-local-example-change-upstream-solo-fallback.toml
```diff
@@ -0,0 +1,70 @@
+# SRI JDC config
```
Shourya742, line 1: This file can be removed now

roles/translator/src/lib/proxy/bridge.rs
```diff
@@ -470,7 +490,7 @@ impl Bridge {
             })
```
Shourya742, line 490: Can be clubbed:
```
let (task_collector_new_extended_mining_job, tx_sv1_notify, rx_sv2_new_ext_mining_job, tx_status) = self_
    .safe_lock(|s| {
        (
            s.task_collector.clone(),
            s.tx_sv1_notify.clone(),
            s.rx_sv2_new_ext_mining_job.clone(),
            s.tx_status.clone(),
        )
    })
    .unwrap();

```

Shourya742 (changes requested):
Utack. A few more nits. The rest looks decent code-wise. I will test the functionality and get back to you soon.

roles/translator/src/lib/downstream_sv1/downstream.rs
plebhash: this commit will be merged via #1001

the answer to this question is the same as here: https://github.com/stratum-mining/stratum/pull/1001#discussion_r1701015868

roles/translator/src/lib/proxy/bridge.rs
lorbax: this is relative to PR 1001

roles/jd-client/src/lib/downstream.rs
```diff
@@ -376,12 +376,13 @@ impl DownstreamMiningNode {
...
             let message = if let Mining::NewExtendedMiningJob(job) = message {
-                let jd = self_mutex.safe_lock(|s| s.jd.clone()).unwrap().unwrap();
-                jd.safe_lock(|jd| jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone())
-                    .unwrap();
-                jd.safe_lock(|jd| jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone())
+                if let Some(jd) = self_mutex.safe_lock(|s| s.jd.clone()).unwrap() {
+                    jd.safe_lock(|jd| {
+                        jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone();
+                        jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone();
+                    })
```
GitGab19, line 383: I might have misunderstood reading the PR description, but in this way, is the jd is updating the coinbase tx right after it's initialized as a solo miner? I would guess it's not, until a new job is created by the JDC.

GitGab19 (commented):
tACK 👏 
I left just one comment, if you could clarify it it would be great, thanks @lorbax

roles/jd-client/src/lib/downstream.rs
lorbax: if initialized as solo miner, then `self_mutex.safe_lock(|s| s.jd.clone()).unwrap()` is `None` and the coinbase is not updated. If you test dev in order to reproduce the error, you should see that the jdc panics exactly because it is unwrapping a `None`. 
BTW like it is works, but if jdc switches to solo mining the coinbase output is the same, and basically the jdc downstreams are mining for the pool for free. Moreover, for each upstream there should be a different coinbase output, with a special coinbase output reserved for solo mining.
Have I misunderstood something or this is a real issue?

lorbax: # List of coinbase outputs used to build the coinbase tx in case of Solo Mining (as last-resort solution of the pools fallback system)

is literally written in the config. Sorry, didn't read!

GitGab19: When JDC does fallback to solo-mining, it should change the coinbase tx output, by using the one specified in config file of the JDC (https://github.com/stratum-mining/stratum/blob/a09084426092a6dfef8862571bc4a0098a05e935/roles/jd-client/config-examples/jdc-config-hosted-example.toml#L33). But if jd is being initialised as None, does it mean it's not going to update the coinbase tx output? Maybe I'm just missing some details

lorbax: > When JDC does fallback to solo-mining, it should change the coinbase tx output, by using the one specified in config file of the JDC
> 
> https://github.com/stratum-mining/stratum/blob/a09084426092a6dfef8862571bc4a0098a05e935/roles/jd-client/config-examples/jdc-config-hosted-example.toml#L33
> But if jd is being initialised as None, does it mean it's not going to update the coinbase tx output? Maybe I'm just missing some details

yes perhaps you are right. I should investigate it!

GitGab19: What's the purpose of this segment?
https://github.com/lorbax/stratum/blob/58c70a48e0b1c5fbb9f153700ccf867f425c41b8/roles/jd-client/src/lib/downstream.rs#L379

lorbax: the `jd` field of `DownstreamMiningNode` is an `Option<Arc<Mutex<JobDeclarator>>>` and it is `None` if jd is initialized as solo mining mode ([here](https://github.com/stratum-mining/stratum/blob/dev/roles/jd-client/src/lib/mod.rs#L177) the function `listen_for_downstream_mining` is returning a `DownstreamMiningNode.jd: None`). Before was like this
`
                let jd = self_mutex.safe_lock(|s| s.jd.clone()).unwrap().unwrap();
`
and caused a panic.

GitGab19: I meant the purpose of this segment:
```
if let Some(jd) = self_mutex.safe_lock(|s| s.jd.clone()).unwrap() {
                    jd.safe_lock(|jd| {
                        jd.coinbase_tx_prefix = job.coinbase_tx_prefix.clone();
                        jd.coinbase_tx_suffix = job.coinbase_tx_suffix.clone();
                    })
                    .unwrap();
                }
```

Why is it needed to set in that way the jd.coinbase_prefix and suffix? I guess that this is done only in the in which JDC is acting normally with a Pool. But why is that needed?

plebhash (approved):

PR #1060, opened 2024-07-17
add `Aggregate MG Test Results` job to MG CI 

plebhash:
Github only allows branch protection rules based on specific CI jobs. So we are adding a new job that just takes all the other MG test jobs as a prerequisite, so we can use this as an umbrella rule for branch protection.

when some MG test fails:
image

when all MG tests pass:
image


GitGab19 (approved):

plebhash:
it turns out this PR was not enough

a failing MG Test job only resulted in skipping the `mg-aggregate-results` job, which we wanted to use for branch protection

so this PR was followed by #1064, which finally achieved the desired goal

Issue #1059, opened 2024-07-17
New feature for the Message Generator


lorbax:
In the Message Generator, an `action` is an operation that is performed against the tested software. An action sends a list of messages (that may be empty) and perform a list of checks (that may be empty) on the response message from the tested software. These checks are called `action_results`. One example of action result is `MatchMessageType`, which checks that the received message has a specific type. For example, if we are testing a certain upstream, a typical action sends a `SetupConnection` and checks that the response has the message type corresponding to `SetupConnectionSuccess`.
There are some cases in which the tested software can answer with two or more allowed messages; for example, after a `NewExtendedMiningJob`, a downstream can send either a `SubmitSharesExtended` or an `Updatechannel`. The reception of any of these messages fall within the correct behaviour of a SRI compliant software.
Nevertheless, there is no way right now to check this situation. If we set an action that sends `NewExtendedMiningJob` and waits for an `UpdateChannel` the test will fail id receives a `SubmitSharesExtended`. Similarly, If we set an action that sends `NewExtendedMiningJob` and waits for an `SubmitSharesExtended` the test will fail id receives a `UpdateChannel`.
This issue presents itself in this during the checking of the result of [this action](https://github.com/stratum-mining/stratum/blob/dev/test/message-generator/test/translation-proxy/translation-proxy.json#L145).

It is needed a feature that signal that a result that checks a received message has to be performed either at the first message received, or at the first message received of a certain type (i.e. waits until the reception of a specific message, and exit with a success when this happens).

I am already working on this issue.

plebhash:
this is great!

we get many MG CI false alarms from this missing feature, as it was already described here: https://github.com/stratum-mining/stratum/issues/928#issuecomment-2127937555

> EXECUTING ../../test/message-generator/test/translation-proxy/translation-proxy.json
> ...
> WRONG MESSAGE TYPE expected: 27 received: 22

I'm excited to bring one more enhancement to our MG test stack!

jbesraa:
> Nevertheless, there is no way right now to check this situation. If we set an action that sends `NewExtendedMiningJob` and waits for an `UpdateChannel` the test will fail id receives a `SubmitSharesExtended`.

[...]

PR #1068, opened 2024-07-19
WaitUntil feature on MG

lorbax:
fix to issue [1059](https://github.com/stratum-mining/stratum/issues/1059).
To see a description of the changes apported to this PR, look at the commit message and the README changes.
The feature introduced is backward compatible, as it is introduced optionally.

How to test it:

easy way. look at `translation-proxy.json` MG test. YOu can see that it is launched TProxy. Go the config used and change it in order that the `channel_nominal_hashrate` is few magnitude order higher than the `min_individual_miner_hashrate`. This forces the channel to be updated (possibly more than once) and an `UpdateChannel`. Then you should see `RECEIVED 22, WAITING MESSAGE TYPE 27` several times (i.e. `UpdateChannel`) and finally `MATCHED WAITED  MESSAGE TYPE 27`, namely `SubmitSharesExtended`, resulting in success of the MG test. Reducing the `miner_num_submits_before_update` to 1 may help testing. The hashrates depend on your hardware, on my PC this worked
```toml
min_individual_miner_hashrate=1000000.0
channel_nominal_hashrate = 50000000000.0

```

hard way. test PR [1001](https://github.com/stratum-mining/stratum/pull/1001): checkit out and rebase it on top of the current PR.  Then test it following instructions in the two top commits. Finally change this [MG-test](https://github.com/lorbax/stratum/blob/translator-restart-if-disconnected-from-upstream/test/message-generator/test/change-upstream/change-upstream.json) with the following
```json
{
    "version": "2",
    "doc": [
        "This test does",
        "Acts like a pool that refuses share with JobDeclatarion",
        "Waits for SetupConnection",
        "Sends a SetupConnection.Success and waits that a SetCustomMiningJob is received",
        "Sends a SetCustomMiningJob.Success and waits that UpdateChannel is received",
        "Waits that SubmitShare is received",
        "Responds with SubmitShare.Error"
    ],
    "frame_builders": [
        {
            "type": "automatic",
            "message_id": "test/message-generator/messages/mining_messages.json::open_extended_mining_channel_success"
        },
        {
            "type": "automatic",
            "message_id": "test/message-generator/messages/common_messages.json::setup_connection_success_tproxy"
        },
        {
            "type": "automatic",
            "message_id": "test/message-generator/messages/mining_messages.json::set_custom_mining_job_success"
        },
        {
            "type": "automatic",
            "message_id": "test/message-generator/messages/mining_messages.json::submit_shares_error"
        }
    ],
    "actions": [
       {
            "message_ids": [],
            "role": "server",
            "results": [
                {
                    "type": "match_message_type",
                    "value": "0x00"
                }
            ],
            "actiondoc": "This action checks that a Setupconnection message is received"
        },
        {
            "message_ids": ["setup_connection_success_tproxy"],
            "role": "server",
            "results": [
                {
                    "type": "match_message_type",
                    "value": "0x13"
                }
            ],
          "actiondoc": "This action sends SetupConnection.Success and check that a OpenExtendedfMiningChannel is received"
        },
        {
            "message_ids": ["open_extended_mining_channel_success"],
            "role": "server",
            "results": [
                {
                    "type": "get_message_field",
                    "value": [
                        "MiningProtocol",
                        "SetCustomMiningJob",
                        [
                            [
                                "request_id",
                                "custom_job_req_id"
                            ]
                        ]
                    ]
                }
            ],
          "actiondoc": "This action sends open_extended_mining_channel_success, and a wait for a SetCustomMiningJoband get the req id"
        },
        {
            "message_ids": [],
            "role": "server",
            "results": [
                {
                    "type": "match_message_type",
                    "value": "0x1b",
                    "condition": {"WaitUntil": {"WaitUntilConfig": {"timeout": 120, "allowed_messages": ["0x1b", "0x16", "0x22"] } } }
                }
            ],
            "actiondoc": "This action check submit_shares is received, allowed messages: update channel, submit shares, set custom miningjob"
        },
        {
            "message_ids": ["submit_shares_error"],
            "role": "server",
            "results": [],
            "actiondoc": "This action sends a submit_shares_error"
        }
    ],
    "setup_commands": [
    ],
    "execution_commands": [
    ],
    "cleanup_commands": [
        {
            "command": "sleep",
            "args": ["100000000000"],
            "conditions": "None"
        }
    ],
    "role": "server",
    "upstream": {
        "ip": "127.0.0.1",
        "port": 44254,
        "pub_key": "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72",
        "secret_key": "mkDLTBBRxdBv998612qipDYoTK3YUrqLe8uWw7gu3iXbSrn2n"
    }
}
```

### DIFF
--- a/utils/message-generator/src/executor.rs
+++ b/utils/message-generator/src/executor.rs
@@ -3,13 +3,14 @@ use crate::{
     into_static::into_static,
     net::{setup_as_downstream, setup_as_upstream},
     parser::sv2_messages::ReplaceField,
-    Action, ActionResult, Command, Role, SaveField, Sv2Type, Test,
+    Action, ActionResult, Command, Condition, Role, SaveField, Sv2Type, Test,
 };
 use async_channel::{Receiver, Sender};
 use binary_sv2::Serialize;
 use codec_sv2::{StandardEitherFrame as EitherFrame, Sv2Frame};
 use roles_logic_sv2::parsers::{self, AnyMessage};
 use std::{collections::HashMap, convert::TryInto, sync::Arc};
+use tokio::{pin, time::sleep};
 
 use tracing::{debug, error, info};
 
@@ -200,32 +201,105 @@ impl Executor {
                 );
 
                 match result {
-                    ActionResult::MatchMessageType(message_type) => {
-                        let message = match recv.recv().await {
-                            Ok(message) => message,
-                            Err(_) => {
+                    ActionResult::MatchMessageType(message_type, condition) => match condition {
+                        None => {
+                            let message = match recv.recv().await {
+                                Ok(message) => message,
+                                Err(_) => {
+                                    success = false;
+                                    error!("Connection closed before receiving the message");
+                                    break;
+                                }
+                            };
+
+                            let message: Sv2Frame<AnyMessage<'static>, _> =
+                                message.try_into().unwrap();
+                            debug!("RECV {:#?}", message);
+                            let header = message.get_header().unwrap();
+
+                            if header.msg_type() != *message_type {
+                                error!(
+                                    "WRONG MESSAGE TYPE expected: {} received: {}",
+                                    message_type,
+                                    header.msg_type()
+                                );
                                 success = false;
-                                error!("Connection closed before receiving the message");
                                 break;
+                            } else {
+                                info!("MATCHED MESSAGE TYPE {}", message_type);
                             }
-                        };
-
-                        let message: Sv2Frame<AnyMessage<'static>, _> = message.try_into().unwrap();
-                        debug!("RECV {:#?}", message);
-                        let header = message.get_header().unwrap();
-
-                        if header.msg_type() != *message_type {
-                            error!(
-                                "WRONG MESSAGE TYPE expected: {} received: {}",
-                                message_type,
-                                header.msg_type()
-                            );
-                            success = false;
-                            break;
-                        } else {
-                            info!("MATCHED MESSAGE TYPE {}", message_type);
                         }
-                    }
+                        Some(condition_inner) => {
+                            match condition_inner {
+                                Condition::WaitUntil(wait_until_config) => {
+                                    let duration = std::time::Duration::from_secs(
+                                        wait_until_config.timeout as u64,
+                                    );
+                                    let timer = sleep(duration);
+                                    pin!(timer);
+                                    let async_block = async {
+                                        loop {
+                                            let message = recv.recv().await;
+                                            let message = match message {
+                                                Ok(message) => message,
+                                                Err(_) => {
+                                                    success = false;
+                                                    error!("Connection closed before receiving the message");
+                                                    break;
+                                                }
+                                            };
+                                            let message: Sv2Frame<AnyMessage<'static>, _> =
+                                                message.try_into().unwrap();
+                                            debug!("RECV {:#?}", message);
+                                            let header = match message.get_header() {
+                                                Some(header_) => header_,
+                                                None => {
+                                                    error!("Failed to get message header");
+                                                    success = false;
+                                                    break;
+                                                }
+                                            };
+                                            let allowed_messages =
+                                                &wait_until_config.allowed_messages;
+                                            let received_message_type = header.msg_type();
+                                            if allowed_messages.contains(&received_message_type) {
+                                                if received_message_type == *message_type {
+                                                    info!(
+                                                        "MATCHED WAITED  MESSAGE TYPE {}",
+                                                        received_message_type
+                                                    );
+                                                    break;
+                                                } else {
+                                                    info!(
+                                                        "RECEIVED {}, WAITING MESSAGE TYPE {}",
+                                                        received_message_type, message_type
+                                                    );
+                                                    continue;
+                                                }
+                                            } else {
+                                                error!(
+                                                    "RECEIVED MESSAGE OF NOT ALLOWED TYPE {:?}",
+                                                    received_message_type
+                                                );
+                                                success = false;
+                                                break;
+                                            }
+                                        }
+                                    };
+                                    tokio::select! {
+                                        _ = &mut timer => {
+                                            error!("Timer has elapsed before wanted message has arrived");
+                                            success = false;
+                                            break;
+                                        },
+                                        _ = async_block => {
+                                            info!("MATCHED WAITED  MESSAGE TYPE {}", message_type);
+                                        }
+                                    }
+                                }
+                            };
+                        }
+                    },
                     ActionResult::MatchMessageField((
                         subprotocol,
                         message_type,
--- a/utils/message-generator/src/main.rs
+++ b/utils/message-generator/src/main.rs
@@ -11,6 +11,7 @@ extern crate load_file;
 use crate::parser::sv2_messages::ReplaceField;
 use binary_sv2::{Deserialize, Serialize};
 use codec_sv2::StandardEitherFrame as EitherFrame;
+use core::fmt::Display;
 use external_commands::*;
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
 use rand::Rng;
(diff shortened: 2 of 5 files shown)


utils/message-generator/src/main.rs
```diff
@@ -179,9 +180,22 @@ pub struct SaveField {
     keyword: String,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub enum Condition {
+    WaitUntil,
+}
```
plebhash, line 186:
```suggestion
// enum as a placeholder for potential conditions to be implemented in the future
#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
pub enum Condition {
    WaitUntil,
}
```

utils/message-generator/src/parser/actions.rs
```diff
@@ -52,7 +54,23 @@ impl Sv2ActionParser {
...
+                                    serde_json::from_value::<Condition>(condition_inner.clone())
+                                        .unwrap();
+                                match condition_ {
+                                    Condition::WaitUntil => {
+                                        action_results.push(ActionResult::MatchMessageType(
+                                            message_type,
+                                            Some(Condition::WaitUntil),
+                                        ))
+                                    }
+                                }
```
plebhash, line 69:
```suggestion
                                match condition_ {
                                    Condition::WaitUntil => {
                                        action_results.push(ActionResult::MatchMessageType(
                                            message_type,
                                            Some(Condition::WaitUntil),
                                        ))
                                    }
                                    // match as a placeholder for potential conditions to be implemented in the future
                                }
```

plebhash, line 69:
```suggestion
                                match serde_json::from_value::<Condition>(condition_inner.clone()) {
                                    Ok(Condition::WaitUntil) => {
                                        action_results.push(ActionResult::MatchMessageType(
                                            message_type,
                                            Some(Condition::WaitUntil),
                                        ))
                                    },
                                    Err(e) => {
                                        panic!("Failed to deserialize condition: {}", e);
                                    }
                                }
```

plebhash (approved):
Looks good to me.

I left some suggestion for small improvements.

---

It's also worth noting that if the current implementation of this new feature introduces the possibility of MG entering an infinite loop.

That would happen if some MG test defined a `match_message_type ` with a `"condition": "WaitUntil"` but the expected message never arrives.

In that scenario, `Executor::execute` will enter an infinite loop and the test will never return `TEST FAILED` nor `TEST OK`.

---

So maybe in the future we might realize we need something like this:
```
pub enum Condition {
    WaitUntil(u32), // timeout
}
```

where the `u32` represents a `timeout` that breaks the infinite loop inside `Executor::execute`

---

but for now, I think this is good to merge so we can unblock the other work that depends on it

lorbax:
> Looks good to me.
> 
> I left some suggestion for small improvements.
> 
> It's also worth noting that the current implementation of this new feature introduces the possibility of MG entering an infinite loop, which is a potential footgun.
> 
> That would happen if some MG test defined a `match_message_type ` with a `"condition": "WaitUntil"` but the expected message never arrived.
> 
> In that scenario, `Executor::execute` would enter an infinite loop and the test would never return `TEST FAILED` nor `TEST OK`.
> 
> So maybe in the future we might realize we need something like this:
> 
> ```
> pub enum Condition {
>     WaitUntil(u32), // timeout
>     ...
> }
> ```
> 
> where the `u32` represents a `timeout` that breaks the infinite loop inside `Executor::execute`
> 
> but for now, I think this is good to merge so we can unblock the other work that depends on it

you are right I thought that there was a timer for the entire MG test. BTW this should be a nice-to-have and an easy fix

plebhash:
> I thought that there was a timer for the entire MG test

you're right @lorbax 

this is documented on the MG `README.md`:

https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/utils/message-generator/README.md?plain=1#L281-L294

so if some MG test (or mock) has the following conditions:
- `command` has `WithConditions` with a `timer_secs` parameter
- action has `match_message_type` with a `"condition": "WaitUntil"`

if the expected message never arrives, eventually the original `command` will run out of time and the test will fail.

but if some MG test (or mock) has the following conditions:
- `command` **does not** have `WithConditions` with a `timer_secs` parameter
- action has `match_message_type` with a `"condition": "WaitUntil"`

then the test could still get stuck.

however, this can be easily solved with good documentation, so that the user is aware that `WaitUntil` can only be used in conjunction with `WithConditions` + `timer_secs`

here's a commit where I add the docs for this new feature, feel free to cherry pick it into your branch so it's included in this PR: 2be0191f4676fe84e9ef8779833ba5862ed85e6c

utils/message-generator/src/parser/actions.rs
plebhash: @lorbax do you agree with this suggestion?

if yes, you can simply clicl the `Commit suggestion` button above, no need to manually implement this on your branch

utils/message-generator/src/main.rs
plebhash: @lorbax do you agree with this suggestion?

if yes, you can simply clicl the `Commit suggestion` button above, no need to manually implement this on your branch

Shourya742 (changes requested):
Shouldn't we also add this for other related ActionResults? I tested it locally and it looks good. I agree with @plebhash's view on having a timeout for this action due to the possibility of staleness.

lorbax:
> > I thought that there was a timer for the entire MG test
> 
> you're right @lorbax
> 
> this is documented on the MG `README.md`:
> 
> https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/utils/message-generator/README.md?plain=1#L281-L294
> 
> so if some MG test (or mock) has the following conditions:
> 
>     * `command` has `WithConditions` with a `timer_secs` parameter
> 
>     * action has `match_message_type` with a `"condition": "WaitUntil"`
> 
>     * the expected message never arrives
> 
> 
> then eventually the original `command` will run out of time and the test will fail.
> 
> but if some MG test (or mock) has the following conditions:
> 
>     * `command` **does not** have `WithConditions` with a `timer_secs` parameter
> 
>     * action has `match_message_type` with a `"condition": "WaitUntil"`
> 
>     * the expected message never arrives
> 
> 
> then the test would get stuck.
> 
> however, this can be easily solved with good documentation, so that the user is aware that `WaitUntil` must be used in conjunction with `WithConditions` + `timer_secs`
> 
> here's a commit where I add the docs for this new feature, feel free to cherry pick it into your branch so it's included in this PR: [plebhash@2be0191](https://github.com/plebhash/stratum/commit/2be0191f4676fe84e9ef8779833ba5862ed85e6c)

This timeout seems to be valid only for commands, while we are talking about a timeout for the action. I proposed also a global timeout, longer then all the other timeouts.

utils/message-generator/src/parser/actions.rs
lorbax: I accept this suggestion but I prefer to force push because I want the commit history to look as it is (single commit). Do you have a particular reason for adding a commit with this small suggestion?

lorbax:
> Shouldn't we also add this for other related ActionResults? I tested it locally and it looks good. I agree with @plebhash's view on having a timeout for this action due to the possibility of staleness.

I think that it make sense with `GetMessageField` and `MatchMessageField`, but I would prefer to introduce this feature for those results when it is needed.

utils/message-generator/src/executor.rs
```diff
@@ -200,30 +200,65 @@ impl Executor {
...
+                                                success = false;
+                                                error!("Connection closed before receiving the message");
+                                                break;
+                                            }
+                                        };
+
+                                        let message: Sv2Frame<AnyMessage<'static>, _> =
+                                            message.try_into().unwrap();
+                                        debug!("RECV {:#?}", message);
+                                        let header = message.get_header().unwrap();
```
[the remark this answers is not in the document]
lorbax: don't know if I can commit this, the `get_header()` produces an `Option<T>`

lorbax:
Pushed a commit for applying the review.
@plebhash @Shourya742

utils/message-generator/src/parser/actions.rs
plebhash: > Do you have a particular reason for adding a commit with this small suggestion?

not really! feel free to force push from your local changes

utils/message-generator/src/main.rs
plebhash: > Do you have a particular reason for adding a commit with this small suggestion?

not really! feel free to force push from your local changes

utils/message-generator/src/executor.rs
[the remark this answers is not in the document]
plebhash: @lorbax can you elaborate? this `match` works exactly on a `Option`, what's wrong with it?

plebhash:
@lorbax we need docs for this new feature, please add something to `utils/message-generator/README.md`

as already stated, I added some to https://github.com/plebhash/stratum/commit/2be0191f4676fe84e9ef8779833ba5862ed85e6c

you can either cherry-pick this commit, or just use it as inspiration to write something on your own words

utils/message-generator/src/executor.rs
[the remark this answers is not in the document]
lorbax: added this part of code

Issue #1065, opened 2024-07-17
remove `JobDistributionProtocol` from `common_messages_sv2 `


plebhash:
during the [review process](https://github.com/stratum-mining/stratum/pull/1035/files#r1677822834) of #1035 we identified that `common_messages_sv2` crate has some references to `JobDistributionProtocol`, which is just some legacy terminology that we forgot to remove after some spec changes

if we check [stratumprotocol.org/specification](https://stratumprotocol.org/specification/) we'll only find 3 subprotocols:

- Mining
- Job Declaration
- Template Distribution

but the `Protocol` enum declaration includes 4 different subprotocols:
https://github.com/stratum-mining/stratum/blob/7841021e3f20280c62af119de5de58f421c6e428/protocols/v2/subprotocols/common-messages/src/setup_connection.rs#L293-L298

all references to `JobDistributionProtocol` should be cleaned up

Shourya742:
@plebhash, can I take this on? This will help me understand the APIs in the protocol, which might be useful during future refactoring.

PR #1072, opened 2024-07-25
Remove job distribution protocol

Shourya742:
Closes: #1065


scripts/build_header.sh
```diff
@@ -1,9 +1,8 @@
 #! /bin/sh
-
 cargo install --version 0.20.0 cbindgen
 
-rm -f ./scripts/sv2.h
-touch ./scripts/sv2.h
+rm -f ./sv2.h
+touch ./sv2.h
```
GitGab19, line 5: Why is this change needed?

Shourya742: We're calling this script from `sv2-header-check.sh`, which changes the PWD to the scripts directory before invoking `build_header.sh`. The buildr_header shell script point to the scripts folder, but since it was already inside the scripts directory, the removal and creation commands couldn't locate the correct path. As a result, it couldn't create the `sv2.h` file at the desired location and instead appended the `cbindgen` output to the older version of `sv2.h`. This caused the hash test to fail.

GitGab19 (approved):

plebhash:
rebasing so we can merge

Issue #1051, opened 2024-07-15
Dig github CI benches execution and fix false alarms


GitGab19:
As requested [here](https://github.com/stratum-mining/stratum/pull/983#issuecomment-2199915659), I need to dig the current execution of the benches in github CI, fixing all the false alarms we are getting in every PR we opened again `dev` branch.

PR #1074, opened 2024-07-26
ignore `test_diff_management`

### LINKED ISSUE #988: Fix flaky test `downstream_sv1::diff_management::test::test_diff_management`

plebhash:
after some failed attempts on #995, this PR disables the `test_diff_management` unit test as a temporary fix to #988

we should not close #988 until a more sustainable fix is found for the long term

we should also be extra careful when modifying `roles/translator/src/lib/downstream_sv1/diff_management.rs`

### DIFF
--- a/roles/translator/src/lib/downstream_sv1/diff_management.rs
+++ b/roles/translator/src/lib/downstream_sv1/diff_management.rs
@@ -317,6 +317,7 @@ mod test {
 
     use crate::downstream_sv1::Downstream;
 
+    #[ignore] // as described in issue #988
     #[test]
     fn test_diff_management() {
         let expected_shares_per_minute = 1000.0;


Fi3:
Have you tried the values suggested here https://github.com/stratum-mining/stratum/issues/988#issuecomment-2250316209 ?

plebhash:
> Have you tried the values suggested here [#988 (comment)](https://github.com/stratum-mining/stratum/issues/988#issuecomment-2250316209) ?

I did some investigation, which is described here: https://github.com/stratum-mining/stratum/pull/995#issuecomment-2209602801

my findings were:
> - `measure_hashrate(X)`: no significant differences by either increasing or decreasing it (1s might seem better, but if this was significant, 30s would be significantly worse, which is not the case, so we just attribute that to luck and conclude this parameter is not important)
> - `z_score_99`: decreasing to 95% it made it significantly worse (which actually proves my previous suggestion was wrong), while increasing even further above 99.9999...% made no significant difference
> - `total_run_time`: increasing to 30s made the errors go away, while decreasing to 1s made it significantly worse, indicating this parameter is sensitive and increasing it seems to be the final solution to our problem

it seems your https://github.com/stratum-mining/stratum/issues/988#issuecomment-2250316209 suggests 40s, but only empirically tested on your machine

in summary, bigger values for `total_run_time` seems to always improve things, so maybe we can overshoot to 60s and hope it will fix this once and for all https://github.com/stratum-mining/stratum/pull/1075

if we merge #1075 and have to re-open #988 again, then we should consider this PR, which completely disables `test_diff_management`

plebhash:
reverting this to draft status in favor of #1075

Fi3:
> > Have you tried the values suggested here [#988 (comment)](https://github.com/stratum-mining/stratum/issues/988#issuecomment-2250316209) ?
> 
> I did some investigation, which is described in detail here: [#995 (comment)](https://github.com/stratum-mining/stratum/pull/995#issuecomment-2209602801)
> 
> my findings were:
> 
> > * `measure_hashrate(X)`: no significant differences by either increasing or decreasing it (1s might seem better, but if this was significant, 30s would be significantly worse, which is not the case, so we just attribute that to luck and conclude this parameter is not important)
> > * `z_score_99`: decreasing to 95% it made it significantly worse (which actually proves my previous suggestion was wrong), while increasing even further above 99.9999...% made no significant difference
> > * `total_run_time`: increasing to 30s made the errors go away, while decreasing to 1s made it significantly worse, indicating this parameter is sensitive and increasing it seems to be the final solution to our problem
> 
> it seems your [#988 (comment)](https://github.com/stratum-mining/stratum/issues/988#issuecomment-2250316209) suggests `total_run_time = 40s`, but only empirically tested on your machine
> 
> in summary, bigger values for `total_run_time` seems to always improve things, so maybe we can overshoot to `60s` and hope it will fix this once and for all, which I'm now proposing via #1075
> 
> if we merge #1075 and have to re-open #988 again, then we should consider this PR, which completely disables `test_diff_management`

yep measure hashrate above 10s seems to not improve significantly anymore. Higher total run time seems to always improve 40s seems to be good enough 60 would be better, but is 20s more every time we  run the test

Fi3:
if we hit that only once in weeks I would leave it there as the advantage of having the test overtake the hassle of re run it very few time. Otherwise we can just remove it.

Fi3:
also why we have all those warnings in CI?

GitGab19:
> also why we have all those warnings in CI?

https://github.com/stratum-mining/stratum/issues/1051
I will fix them asap, but before doing it, I need to know exactly how our branch structure will be. So I will fix them once we defined it. For now I could disable (temporarily) benches from CI since they are generating only false alarms

GitGab19 (approved):

PR #1085, opened 2024-08-09
add missing `impl` for `Seq0255` on `binary_sv2/serde-sv2`

plebhash:
required to unblock https://github.com/stratum-mining/stratum/pull/985#issuecomment-2234328927

### DIFF
--- a/protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
+++ b/protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
@@ -259,6 +259,15 @@ impl<'a> From<Vec<u32>> for Seq0255<'a, u32> {
     }
 }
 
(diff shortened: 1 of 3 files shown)


protocols/v2/binary-sv2/serde-sv2/src/primitives/sequences/seq0255.rs
```diff
@@ -259,6 +259,15 @@ impl<'a> From<Vec<u32>> for Seq0255<'a, u32> {
     }
 }
 
+impl<'a> From<Vec<U256<'a>>> for Seq0255<'a, U256<'a>> {
+    fn from(v: Vec<U256<'a>>) -> Self {
+        Seq0255 {
```
Shourya742, line 264:
```suggestion
        Self {
```

Shourya742 (changes requested):
ACK. One small nit.

rrybarczyk (approved):

PR #1087, opened 2024-08-09
Remove toml dependency

Shourya742:
Closes: #456

### DIFF
--- a/roles/jd-client/src/lib/error.rs
+++ b/roles/jd-client/src/lib/error.rs
@@ -1,3 +1,4 @@
+use ext_config::ConfigError;
 use std::fmt;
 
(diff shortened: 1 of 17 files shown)


roles/jd-client/src/lib/error.rs
```diff
@@ -209,12 +210,6 @@ impl<'a>
     }
 }
 
-impl<'a> From<Vec<u8>> for Error<'a> {
```
plebhash, line 212: I assume this is being removed because it's no longer explicitly used in the deserialization process and started triggering build warnings?

is that correct?

Shourya742: yes

plebhash (approved):

Issue #1093, opened 2024-08-13
Restructure `Roles` as a prepration for integration tests


jbesraa:
This is the first stage of #1120 

While trying to add integration tests we noticed that the roles logic should have a bit of refactor, mainly moving some of the library logic out of `main.rs` to `lib` folder/file in order to make it nicely accessible via testing.

Currently we have four different roles we want to refactor
- [x] Translator - #1092 
- [x] JDServer - #1095 
- [x] JDClient - #993 
- [x] Pool - #1097 
- [x] MiningDevice - https://github.com/stratum-mining/stratum/pull/1163
- [x] Sv1 Mining Device - https://github.com/stratum-mining/stratum/issues/1245

Example for such change covering the Translator role https://github.com/stratum-mining/stratum/pull/1092/files 

Goals:
- Get the roles code structured in a way that they can be composable inside the new integration test framework.
- Avoid breaking current behavior of roles. After this refactoring, it’s important that all roles continue passing the current MG test stack, and can also be launched manually while following the expected behavior.

plebhash:
@jbesraa should we list https://github.com/stratum-mining/stratum/pull/993 on the issue description for the JDClient bullet point?

GitGab19:
Can we close this issue now that #1163 has been merged? @jbesraa @plebhash

jbesraa:
> Can we close this issue now that #1163 has been merged? @jbesraa @plebhash 

Yup all done here

PR #1092, opened 2024-08-13
Move `TranslatorSv2` lib code out of `main.rs`

jbesraa:
Related to #1093

PR #1095, opened 2024-08-13
Move `JDServer` lib code out of `main.rs`

jbesraa:
part of #1093


marathon-gary (commented):
Did you intend to omit the `Cargo.toml` change to specify a `[[bin]]` value?

roles/jd-server/src/lib/core.rs
```diff
@@ -0,0 +1,183 @@
...
+            new_block_receiver,
+        )));
+        let mempool_update_interval = config.mempool_update_interval;
+        let mempool_cloned_ = mempool.clone();
+        let (status_tx, status_rx) = unbounded();
+        let sender = status::Sender::Downstream(status_tx.clone());
+        let mut last_empty_mempool_warning =
+            std::time::Instant::now().sub(std::time::Duration::from_secs(60));
+
+        // TODO if the jd-server is launched with core_rpc_url empty, the following flow is never
```
marathon-gary, line 43: We should migrate this to a GH issue.

roles/jd-server/src/lib/core.rs
```diff
@@ -0,0 +1,183 @@
...
+                        match mempool::JDsMempool::add_tx_data_to_mempool(
+                            mempool_cloned,
+                            add_transactions_to_mempool,
+                        )
+                        .await
+                        {
+                            Ok(_) => (),
+                            Err(err) => {
+                                // TODO
+                                // here there should be a better error management
```
marathon-gary, line 133: Maybe not it's own issue but perhaps an overarching issue for tracking better error handling across the code base? Seems like this could be part of the refactoring documentation/story.
cc: @plebhash @pavlenex @rrybarczyk

roles/jd-server/src/lib/core.rs
```diff
@@ -0,0 +1,183 @@
...
+                        match err {
+                            JdsMempoolError::EmptyMempool => {
+                                if last_empty_mempool_warning.elapsed().as_secs() >= 60 {
+                                    warn!("{:?}", err);
+                                    warn!("Template Provider is running, but its mempool is empty (possible reasons: you're testing in testnet, signet, or regtest)");
+                                    last_empty_mempool_warning = std::time::Instant::now();
+                                }
+                            }
+                            _ => {
+                                // TODO here there should be a better error managmenet
```
marathon-gary, line 96: Strong indications that there is a need for a JDS specific error handling review. 😅

roles/jd-server/src/main.rs
```diff
@@ -108,163 +103,5 @@ async fn main() {
         }
     };
```
marathon-gary, line 105: I think it would be helpful to keep this sort of daemon awareness logging.
```suggestion
info!("Jds INITIALIZING with config: {:?}", &args.config_path);
```

marathon-gary (commented):
LGTM.

Give the interaction with bitcoin core, would it be helpful to use this crate: https://github.com/rust-bitcoin/rust-bitcoincore-rpc
Or perhaps this would need to be updated for compatibility with @sjors branch for the TP changes?

roles/jd-server/src/main.rs
```diff
@@ -108,163 +103,5 @@ async fn main() {
...
-            }
-            status::State::Healthy(msg) => {
-                info!("HEALTHY message: {}", msg);
-            }
-            status::State::DownstreamInstanceDropped(downstream_id) => {
-                warn!("Dropping downstream instance {} from jds", downstream_id);
-            }
-        }
-    }
+    lib::core::JDServer::new(config).start().await;
```
marathon-gary, line 106: Need associated import.
```suggestion
    JDServer::new(config).start().await;
```

Sjors:
@marathon-gary wrote:

> Give the interaction with bitcoin core, would it be helpful to use this crate: https://github.com/rust-bitcoin/rust-bitcoincore-rpc
> Or perhaps this would need to be updated for compatibility with @Sjors branch for the TP changes?

My branch does not change RPC methods, so that crate should not be impacted.

Shourya742 (changes requested):
ACK.  I believe some of the error cases are no longer being triggered. If possible, could you remove those and resolve the other warnings as well.
<details>
<summary>
Warning logs JDS
</summary>
<code>
warning: field `tx` is never read
  --> jd-server/src/lib/mempool/mod.rs:15:9
   |
13 | pub struct TransactionWithHash {
   |            ------------------- field in this struct
14 |     pub id: Txid,
15 |     pub tx: Option<Transaction>,
   |         ^^
   |
   = note: `TransactionWithHash` has derived impls for the traits `Debug` and `Clone`, but these are intentionally ignored during dead code analysis
   = note: `#[warn(dead_code)]` on by default

warning: field `0` is never read
 --> jd-server/src/lib/mempool/error.rs:9:9
  |
9 |     Rpc(RpcError),
  |     --- ^^^^^^^^
  |     |
  |     field in this variant
  |
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
  |
9 |     Rpc(()),
  |         ~~

warning: field `0` is never read
  --> jd-server/src/lib/mempool/error.rs:10:16
   |
10 |     PoisonLock(String),
   |     ---------- ^^^^^^
   |     |
   |     field in this variant
   |
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
   |
10 |     PoisonLock(()),
   |                ~~

</code>
</details>

Shourya742:
When shutting down the entire system, I followed this order: Pool, JDS, JDC, and Translator. However, the shutdown signaling isn't working for JDC and Translator. I'm not sure if this is the right place to mention it, but could someone check?
<details>
<summary>
Log screenshot
</summary>
Screenshot from 2024-08-18 08-30-30
</details>

marathon-gary:
> When shutting down the entire system for this branch, I followed this order: Pool, JDS, JDC, and Translator. However, the shutdown signaling isn't working for JDC and Translator. I'm not sure if this is the right place to mention it, but could someone check?
> 
>  Log screenshot
> Screenshot from 2024-08-18 08-30-30
@Shourya742 
Your screenshot is not showing for me?

jbesraa:
> When shutting down the entire system for this branch, I followed this order: Pool, JDS, JDC, and Translator. However, the shutdown signaling isn't working for JDC and Translator. I'm not sure if this is the right place to mention it, but could someone check?
> 
> Screenshot from 2024-08-18 08-30-30

Could you please explain how did you conduct the tests so I can try to replicate it?

roles/jd-server/src/lib/core.rs
jbesraa: This will need to happen as part of a different PR really. Its good that we have more visibility now about the technical debt in the code though

Shourya742:
> > When shutting down the entire system for this branch, I followed this order: Pool, JDS, JDC, and Translator. However, the shutdown signaling isn't working for JDC and Translator. I'm not sure if this is the right place to mention it, but could someone check?
> > Screenshot from 2024-08-18 08-30-30
> 
> Could you please explain how did you conduct the tests so I can try to replicate it?

@jbesraa Run all the roles in the following order: Pool, JDS, JDC, and Translator. Once they are in sync, terminate them using `Ctrl+C`. You will notice that the JDC and Translator do not respond to the shutdown signal.

rrybarczyk:
@jbesraa this is very much needed, but I think we should wait on this kind of `roles` refactoring until we have a more concrete plan and complete #845. Otherwise we will have a disorganized approach to the refactoring effort, which is causing confusion.

@plebhash thoughts?

plebhash:
> @jbesraa this is very much needed, but I think we should wait on this kind of `roles` refactoring until we have a more concrete plan and complete #845. Otherwise we will have a disorganized approach to the refactoring effort, which is causing confusion.
> 
> @plebhash thoughts?

I think the changes within this PR are reasonable as a requirement for the new Integration Tests framework. The same applies to all PRs tracked under https://github.com/stratum-mining/stratum/issues/1093

If you closely look at the diff, you'll see it's mostly shuffling things around without deep architectural changes. It simply declares a new `struct JDServer` and moves already existing code under some methods inside it.

but I do understand where you're coming from @rrybarczyk, and I agree we should be very conservative and limit the scope of changes to the `roles` to only what is a strict requirement to the Integration Tests framework

I wrote a comment on a similar spirit on another, where (different to here), I believe we are getting a bit out of scope: https://github.com/stratum-mining/stratum/pull/1115#issuecomment-2297341493

plebhash:
> When shutting down the entire system for this branch, I followed this order: Pool, JDS, JDC, and Translator. However, the shutdown signaling isn't working for JDC and Translator. I'm not sure if this is the right place to mention it, but could someone check?
> 
> Screenshot from 2024-08-18 08-30-30

I've seen this behavior before

indeed it's quite annoying, but this is just technical debt that unfortunately fall out of scope for this PR

plebhash:
> Given the interaction with bitcoin core, would it be helpful to use this crate: https://github.com/rust-bitcoin/rust-bitcoincore-rpc Or perhaps this would need to be updated for compatibility with @Sjors branch for the TP changes?

@marathon-gary I agree this is a desirable change, but we should keep in mind that this PR is simply introducing some necessary restructuring as a prerequisite for #1120 

overall, there is technical debt all over the place (e.g.: todos, suboptimal error handling, suboptimal architecture), but we should resist the urge of trying to fix all of them now just because reviewing this PR made us aware of them

roles/jd-server/src/lib/core.rs
plebhash, line 1: hmm do we really want to add a new file here?

this PR + #1097 + #1092 are unifying the role struct patterns, which is very good

but I don't understand why only this PR is creating a new file for it

is there any specific reason we shouldn't do this inside `roles/jd-server/src/lib/mod.rs`?

roles/jd-server/src/lib/mod.rs
```diff
@@ -19,6 +28,177 @@ pub type Message = JdsMessages<'static>;
 pub type StdFrame = StandardSv2Frame<Message>;
 pub type EitherFrame = StandardEitherFrame<Message>;
 
+pub struct JDServer {
```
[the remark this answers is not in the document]
jbesraa: Fixed

roles/jd-server/src/lib/core.rs
jbesraa: Removed `core` file

plebhash (approved):

PR #1066, opened 2024-07-19
`PoolSv2` integration tests

jbesraa:
Blocked by #1155 and #1156

for more context: 
- [PR Review Club 24/07/24](https://drive.google.com/file/d/1qiLZDWWwd2yC-AmwhFvLme-l0iRrwGPt/view?usp=sharing)
- [PR Review Club 07/08/24](https://drive.google.com/file/d/1PN9J250rqkp6FU8S1URuHepZ8r2bEjAI/view?usp=sharing)

PR #1097, opened 2024-08-13
Move`PoolSv2` lib code out of `main.rs`

jbesraa:
part of #1093


marathon-gary:
`[[bin]]`??
https://github.com/stratum-mining/stratum/pull/1095#pullrequestreview-2238272656

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -66,6 +66,15 @@ pub struct CoinbaseOutput {
     output_script_value: String,
 }
 
+impl CoinbaseOutput {
```
marathon-gary, line 69: This struct appears all over. I'll create an issue from this comment to consolidate to a location in the code base that makes sense.
https://github.com/stratum-mining/stratum/pull/1066#discussion_r1714014229
https://github.com/stratum-mining/stratum/pull/1066#discussion_r1713911191
https://github.com/stratum-mining/stratum/pull/1066#discussion_r1714003837

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -96,6 +105,33 @@ pub struct Configuration {
     pub test_only_listen_adress_plain: String,
 }
 
+impl Configuration {
+    pub fn new(
```
marathon-gary, line 109: I don't see this function used anywhere?

roles/pool/src/lib/mod.rs
```diff
@@ -2,3 +2,105 @@ pub mod error;
...
+use tracing::{error, info, warn};
+
+use tokio::select;
+
+pub struct PoolSv2;
+
+impl PoolSv2 {
+    pub async fn start(config: Configuration) {
+        let (status_tx, status_rx) = unbounded();
+        let (s_new_t, r_new_t) = bounded(10);
```
marathon-gary, line 19: nit: naming conventions for channels should be consistent with the rest of the code base.

marathon-gary (commented):
LGTM.
Seems naming of some variables drifts from conventions I've seen elsewhere but that's outside the scope of this PR.

Shourya742 (changes requested):
ACK. Same nits as @marathon-gary mentioned. I am still facing the shutdown signalling error mentioned here: https://github.com/stratum-mining/stratum/pull/1095#issuecomment-2295246988

roles/pool/src/lib/mining_pool/mod.rs
jbesraa: This should be used in tests to initialize config from variables rather than file..

plebhash: it seems clippy is not happy about the number of arguments here: https://github.com/stratum-mining/stratum/actions/runs/10457582329/job/28957531606?pr=1097

roles/pool/src/lib/mod.rs
```diff
@@ -2,3 +2,105 @@ pub mod error;
...
+
+use async_channel::{bounded, unbounded};
+
+use mining_pool::{get_coinbase_output, Configuration, Pool};
+use template_receiver::TemplateRx;
+use tracing::{error, info, warn};
+
+use tokio::select;
+
+pub struct PoolSv2;
```
plebhash, line 14: on https://github.com/stratum-mining/stratum/pull/993, https://github.com/stratum-mining/stratum/pull/1095 and https://github.com/stratum-mining/stratum/pull/1092 we're introducing this pattern:

```
pub struct RoleName {
   config: Config
}
```

it would be good to keep that pattern consistent here

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -96,6 +105,73 @@ pub struct Configuration {
...
+impl AuthorityConfig {
+    pub fn new(public_key: Secp256k1PublicKey, secret_key: Secp256k1SecretKey) -> Self {
+        Self {
+            public_key,
+            secret_key,
+        }
+    }
+}
+
+pub struct PoolConnection {
```
plebhash, line 136: maybe call this `ConnectionConfig`?

just to make it explicitly clear that this `struct` is just a `*Config`, not really doing anything meaningful aside from that

roles/pool/src/lib/mod.rs
```diff
@@ -2,3 +2,105 @@ pub mod error;
...
+
+impl PoolSv2 {
+    pub async fn start(config: Configuration) {
```
plebhash, line 17: on https://github.com/stratum-mining/stratum/pull/993, https://github.com/stratum-mining/stratum/pull/1095 and https://github.com/stratum-mining/stratum/pull/1092 we're introducing this pattern:

```
impl RoleName {
    pub fn new(config: Configuration) -> Self { ... }
    pub async fn start(self) { ... }
}
```

it would be good to keep that pattern consistent here

plebhash (approved):

PR #1102, opened 2024-08-14
add `lockfiles.yaml` + rm and gitignore some unnecessary lockfiles

plebhash:
The crates on `roles` and `utils` generate binaries, which means it's a good engineering practice to keep `{roles, utils}/Cargo.lock` on our git history.

The root cause of the issues described on #1044 comes from the fact that we often bump dependency versions on `Cargo.toml` files without committing the same updates to the respective `Cargo.lock`.

This PR introduces a new CI action that blocks PRs that change dependencies on the binary workspaces from being merged if they don't update `Cargo.lock` files accordingly.

We also remove and gitignore `{protocols,benches/Cargo.lock}` since there's no need to keep them in the repository


rrybarczyk (approved):

rrybarczyk (approved):

rrybarczyk (approved):

plebhash:
https://github.com/stratum-mining/stratum/pull/1044 was closed, even though it was deleting `protocols/Cargo.lock` + `benches/Cargo.lock` and adding them to `.gitignore`

that was still a desirable change as stated on this comment https://github.com/stratum-mining/stratum/pull/1044#issuecomment-2289801195

so I'm also including those changes to this PR

jbesraa:
> #1044 was closed, even though it was deleting `protocols/Cargo.lock` + `benches/Cargo.lock` and adding them to `.gitignore`
> 
> those were still desirable changes, as stated on this comment [#1044 (comment)](https://github.com/stratum-mining/stratum/pull/1044#issuecomment-2289801195)
> 
> so I'm also including those changes to this PR
> 
> @jbesraa please let me know your thoughts here

I am a bit confused tbh.

I think it was mentioned that we are not publishing the binaries anymore right? users just git clone and cargo install & run which I think anyway ignores cargo.lock file(unless you provide `--locked` flag) but I honestly dont understand the need for maintaining the `cargo.lock` now(new ci, new discussions, is this really critical/important now?) while it was outdated for a long time and despite that things were working. I would suggest to remove lock files and when we encounter a problem we can deal with it.

Fi3:
having lock files for binary crates is good practice this is another reason to keep roles and protocols in different workspaces

jbesraa:
@plebhash I dont wanna block this. Feel free to CP or add your own commits to ignore the libraries `.lock` files and add the one needed.  Happy with any solution you end-up with here. 
I do think that in the not so far future we will remove all of those lock files anyway because we are aiming to move things to libraries and have something like https://github.com/stratum-mining/stratum/pull/1115 to manage the user API and thats where we should probably maintain good practices for `.lock` files.

PR #1103, opened 2024-08-14
update `roles/Cargo.lock`

plebhash:
`roles/Cargo.lock` is out of sync with `roles/Cargo.toml`

PR #1102 will prevent this from happening again in the future

### DIFF
--- a/roles/Cargo.lock
+++ b/roles/Cargo.lock
@@ -1942,7 +1942,7 @@ checksum = "adad44e29e4c806119491a7f06f03de4d1af22c3a680dd47f1e6e179439d1f56"
 
 [[package]]
 name = "roles_logic_sv2"
-version = "1.2.0"
+version = "1.2.1"
 dependencies = [
  "binary_sv2",
  "chacha20poly1305",


Shourya742:
Should we add this to the pre-push hook as well? We already have a PR for that here: https://github.com/stratum-mining/stratum/pull/1039.

plebhash:
> Should we add this to the pre-push hook as well? We already have a PR for that here: #1039.

hmm I think you meant to write this comment on #1102 ?

this PR is simply updating the `Cargo.lock` file

Shourya742:
> > Should we add this to the pre-push hook as well? We already have a PR for that here: #1039.
> 
> hmm I think you meant to write this comment on #1102 ?
> 
> this PR is simply updating the `Cargo.lock` file
> 
> anyways, just wrote this as a reminder: [#1039 (comment)](https://github.com/stratum-mining/stratum/pull/1039#issuecomment-2291270656)

Silly me...!! thanks

rrybarczyk (approved):

rrybarczyk (approved):

PR #1110, opened 2024-08-17
use rust 1.75.0 and `cargo-semver-checks` 0.33.0 on semver CI

plebhash:
addressing https://github.com/stratum-mining/stratum/pull/1104#issuecomment-2294794478

### DIFF
--- a/.github/workflows/semver-check.yaml
+++ b/.github/workflows/semver-check.yaml
@@ -44,7 +44,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y cmake
 
       - name: Install cargo-semver-checks
-        run: cargo install cargo-semver-checks --locked
+        run: cargo install cargo-semver-checks --version 0.33.0 --locked
 
       - name: Run semver checks for common
         working-directory: common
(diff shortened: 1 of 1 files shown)


Fi3:
we should use cargo semver latest whith whatever rustc version is needed. This do not means that our MSRV is changed. For example for cargo fmt we use nightly but it do not affect our MSRV

plebhash:
> we should use cargo semver latest whith whatever rustc version is needed

that was exactly the goal of #1104 

I'm confused now. What was the objection to #1104?

Fi3:
If we need to use rustc 1.77 for semver check we should use it (like we use nightly for cargo fmt) without changing MSRV.
One thing is the rustc version that we use to do the release the checks ecc ecc. Another completely different thing is MSRV, MSRV is the min rustc version that a crate that import our libs MUST have. If you import a crate from protocols you don't need 1.77 1.75 is ok, we don't use feature added after 1.75 in the code. Semver check use feature added after 1.75, but if want to use our libraries you don't need to run semver check.

PR #1115, opened 2024-08-19
Add `stratum-cli` project

jbesraa:
**Note: This PR is still a draft and I do not expect it to be merged in less than a month TBH as it depends on the integration test PRs, but I do think its a good point to start the discussion and try and reach a consensus.** 

Currently this only covers the `translator` and `jdclient`. In order to test it locally, you can build the branch `cargo build` and then `./target/debug/stratum-cli translator start-local`. For a full list of commands, try ` ./target/debug/stratum-cli help`

blocked by https://github.com/stratum-mining/stratum/pull/1092, https://github.com/stratum-mining/stratum/pull/1097 and https://github.com/stratum-mining/stratum/pull/1095

.. this should allow us to remove `main.rs` files from all the roles and leave them as pure libraries


plebhash:
as much as I think would be a positive change at a first glance, I feel this scope is starting to move in the direction of optimizations that are not really needed with regards to where the project is right now

we agreed on #1066 (and later https://github.com/stratum-mining/stratum/issues/1093) because that was a strict requirement for the new Integration Test framework, while I'm not really convinced on the need for a unified binary just yet

as I already stated in our dev calls, I feel that deeper changes to the application layer should be left for a more distant future, once we have spent 

don't get me wrong, I do feel a unified CLI would be ideal for the long term, as I proposed https://github.com/stratum-mining/stratum/discussions/1069#discussioncomment-10290201

but for now, I feel our time and efforts would be better spent if we focused on either:
- https://github.com/stratum-mining/stratum/issues/845
- the Integration Tests framework, which will require substantial effort on the message synchronization primitives

@jbesraa can you elaborate on what you see as the immediate benefits of this change? why should we do this sooner rather than later?

jbesraa:
> as much as I think would be a positive change, I feel this scope is starting to move in the direction of optimizations that are not really needed with regards to where the project is right now

Why do you think so? Our project does not have a proper entry point, which is one of the first things usually done in projects.

> 
> we agreed on #1066 (and later #1093) because that was a strict requirement for #1120, while I'm not really convinced on the need for a unified binary just yet

Adding a `cli` tool for a "plug and play" access to our code-base is not a "unified binary", its just an API for the end user to interact with the roles.

> 
> as I already stated in our dev calls, I feel that deeper changes to the application layer should be left for a more distant future, once we are already done with #845 and we have a more clear outline on the ideal architecture for the application layer that are still under conceptual discussion under #1069

How would #845 affect this change? I do not think its related in anyway. Adding a cli app could be discarded if you think there is a better way to expose our libraries to the end user, but I do not see how #845 can be a factor here. Moreover, the work on #845 is gonna take at least a year if not more, I have a few open PRs that are waiting for review for a bit and they are not moving forward. #845 Only changes the protocols code which I think should be in general the lowest priority for the following reasons:
1) Its tricky to review and approve
2) Its working and kinda fine, not as bad as the roles
3) The end user will see minor benefits from months of work of multiple developers

> 
> don't get me wrong, I do feel a unified CLI would be ideal for the long term, as I already proposed via [#1069 (comment)](https://github.com/stratum-mining/stratum/discussions/1069#discussioncomment-10290201)
> 
> but for now, I feel our time and efforts would be better spent if we focused on either:
> 
>     * [Rust Docs + refactor: `protocols` crates #845](https://github.com/stratum-mining/stratum/issues/845)
> 
>     * [SRI Integration Test Framework #1120](https://github.com/stratum-mining/stratum/issues/1120)

Both #845 and #1120 are work in progress and have open PRs that should be reviewed so I am not sure why adding CLI should be blocked by those?

> 
> @jbesraa can you elaborate on what you see as the immediate benefits of this change? why should we focus on this sooner rather than later?

1) Low effort: Creating the initial CLI with just the functionality exposed from the binaries is easy and quick. This PR should just cherry-pick the other commits from the `Move bin code to lib` and add another few cases in the code to cover them. It can be a quick win with minimum(actually, none) footprint on the current code
2) It will allow us to remove all the binary related things in the roles which is quite a distraction
3) In the last few weeks you spent time fixing lock files and other problems related to binary/library, bit-aloo also did a similar effort with removing the `toml` package, our best developers are spending the time putting plasters on a bleeding heart with changes that will be discarded sooner or later.
4) 100x better user experience 
5) No breaking changes
6) No actual code changes in `roles` or anywhere else, this is just adding new crate `stratum-cli`, we can decide to remove the `main.rs` files in a later PR

jbesraa:
In general, it would be better to stop thinking about everything as a refactor. Even when adding this 'stratum-cli' there is none refactoring, its merely a code organization putting things in there correct place.

plebhash:
> Both https://github.com/stratum-mining/stratum/issues/845 and https://github.com/stratum-mining/stratum/issues/1120 are work in progress and have open PRs that should be reviewed so I am not sure why adding CLI should be blocked by those?

it's not a matter of blocking, but a matter of how we strategize about where our collective efforts are focused

the reason progress on #845 and #1120 have been slow is the fact that we are a small number of contributors, and unless we keep our scope under reasonable and manageable constrains, our efforts will be scattered and progress will be slow everywhere

with `1.0.2` out of the way, I hope we will be able to spend more time on those activities and make meaningful progress on #845 and #1120

> In the last few weeks you spent time fixing lock files and other problems related to binary/library

the lockfiles issue could be mitigated via CI, as proposed via https://github.com/stratum-mining/stratum/pull/1102

> we can decide to remove the main.rs files in a later PR

removing `main.rs` files from `roles` will have implications on the entire MG CI stack, which relies heavily on the current CLI structure of the different `roles` binaries

so we can't break the current `roles` CLI until https://github.com/stratum-mining/stratum/issues/1121 is fully addressed, otherwise we will loose our current CI

we would also need to adapt docs and communicate those changes to adopters who are currently experimenting with the current `roles` binaries

---

which brings me back to my original point: this might seem like a good direction at a first glance, but it opens up many cans of worms that will inevitably drain our energy into new unforeseen directions... in an ideal world where all SRI contributors had infinite bandwidth for tackling all those implications, sure, that would be ok... but unfortunately we need to be very strategic about how we focus our efforts

my bottom line here is: I'm not strictly opposed to UI enhancements for `roles`, I just think this is a bit premature, and our efforts will be more optimally allocated if we could move #1120 forward

Issue #1098, opened 2024-08-13
Optimise contributing and release processes


jbesraa:
Conversation moved from #978 to this issue.

## Items to Include
- [x] Consider using release branches and use a single main branch - for example for release `1.0.3` we can maintain only `main` branch and when we are ready for release we create a new branch/tag `1.0.3` and release to it

plebhash:
ideally this issue should be solved by proposing a PR that establishes the new conventions under:
- [`CONTRIBUTING.md`](https://github.com/stratum-mining/stratum/blob/dev/CONTRIBUTING.md)
- [`README-DEV.md`](https://github.com/stratum-mining/stratum/blob/dev/README-DEV.md)
- [`RELEASE.md`](https://github.com/stratum-mining/stratum/blob/dev/RELEASE.md)

jbesraa:
Suggestions from @rrybarczyk 

- [ ] Enforce each PR should have a branch associated with it and linked in GitHub's "Development" section on the Issue/PR.
- [ ] Enforce branch naming `yyyy-mm-branch-name`.
  - [ ] Branch names must start with the year (`yyyy`) and month (`mm`).
  - [ ] Words in branch names must be separated by a `-`.
  - [ ] Branch names must be short but descriptive and in present tense. It should contain the action/what is being added/fixed.
- [ ] Issue names should be the "problem" (as opposed to the solution). For example, "Bug in message generator `SubmitShare` `channel_id` test" or "Translator proxy config fails to parse downstream IP address".
- [ ] Pull request names should be the "solution" (as opposed to the problem). For example, "Increment message generator `SubmitShare` `channel_id` field value" or "Translator parse downstream IP as socket addr immediately on read".

jbesraa:
I would suggest to separate the contrib/release process. I am not very much in favor of changing the current contrib process as I think adding https://github.com/stratum-mining/stratum/pull/991 is enough. its more a matter of enforcing it.

In general iam not a big fan of enforcing branch/issue/pr titles. I try to always read the issue body and if there is a pull request then its commits titles and content.

plebhash:
I think the only change to `CONTRIBUTING.md` would be the references we have to `dev` branch there.

jbesraa:
We should combine `CONTRIBUTING` and `README-DEV`. This is the first time I see `README-DEV` actually exist..

plebhash:
[...]

PR #1124, opened 2024-08-21
Consider using release branches

jbesraa:
and maintain only `main` without `dev` branch.

relates to #1098

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,88 +1,58 @@
-# Versioning
+## Changelog
 
-SRI releases follow [SemVer 2.0.0](https://semver.org/).
+The changelog is a file that contains a curated, chronologically ordered list of
+notable changes for each version of a project. It is a way to keep track of the
+project's progress and to communicate the changes to the users and the
+community. The changelog should be updated for each release and is part of the
+release process.
 
-Given a version number `MAJOR.MINOR.PATCH`, we increment the:
-- `MAJOR` version when incompatible API changes are introduced
-- `MINOR` version when functionality is added in a backward compatible manner
-- `PATCH` version when backward compatible bug fixes are introduced
+Things to take into account while writing a changelog:
 
-SRI has a global version, which uses git tags and keeps track of how the codebase evolves across time as a whole.
+ - General release information.
+ - Breaking changes.
+ - Notable changes.
 
-Every internal SRI crate also follows SemVer 2.0.0, but each crate version is only set on the respective `Cargo.toml`,
-(no git tags), and it evolves independently of the global release version.
 
-Whenever a `PATCH` is introduced, it is applied to all the latest `MAJOR` releases.
-For example: imagine there's releases `v1.0.0`, `v1.1.0`, and `v2.0.0`. A bug is found, dating back all the way to `v1.0.0`.
-A patch is applied such that the following new tags are introduced: `v1.1.1` and `v2.0.1`.
+## Release Process
 
-# Git Branching Strategy
+Try to constrain each development cycle to a fixed time period, after which a
+release is made. After all release tasks are completed, initiate the release
+process by creating a new release branch, named `x.y.z.` referring to the
+version number, from the `main` branch.  The release branch is used as a
+breaking point and future reference for the release and should include the
+changelog entries for the release and any other release-specific tasks. Any bug
+fixes or changes for the release should be done on the release branch. When the
+release branch is ready, create a Pull Request from release branch to the `main`
+branch.  Once the Pull Request is approved and merged, create a new tag for the
+release branch and initiate any publishing tasks.
 
-We follow a simplified [gitflow](https://nvie.com/posts/a-successful-git-branching-model/) branching strategy.
+Usually the release process is as follows:
 
-Although our strategy is very similar to the classic gitflow model, we do not keep release branches.
+1. Create a new release branch from the `main` branch.
+2. Amend the changelog with the new release information.
+3. Create a Pull Request with `main` branch as base.
+4. Merge the Pull Request.
+5. Create a new tag for the release branch.
+6. Publish the release.
 
-![](git-branching.png)
 
-## Principal Branches
+## Versioning
 
-The SRI repo holds two principal branches with an infinite lifetime:
-- `main`
-- `dev`
+Protocol and Utils crates follow SemVer 2.0.0. The version number is stored in
+the `Cargo.toml` file of each crate. If a breaking change is introduced to one
+of the crates, the version number must be updated accordingly, otherwise a
+SemVer CI check would fail. Note that this does not apply to the `roles` and
+other crates in the repository.
 
-We consider `main` to be the branch where the source code of `HEAD` always reflects the latest release.
+The general SRI version does not follow any SemVer rules. You can track the
+releases on Github Releases in the repository.
 
-We consider `dev` to be the branch where the source code of `HEAD` always reflects the latest changes in the development cycle.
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -1,88 +1,37 @@
...
-
-Every PR to `main` needs to increase the version of whatever crate it is touching. Otherwise, we will mess up the dependency chain of whoever is fetching from crates.io
-
-Manually running `cargo release` in the various workspaces helps to prepare the version number before the releases are published.
-
-Every time we bump some crate's version, `release-libs.yaml` needs to be manually triggered in order to update crates.io.
+## Release Process
+
+- Constrain each development cycle to a fixed time period, after which a release
+  is made. Typically, the development cycle is 6-8 weeks long.
```
plebhash, line 4: @pavlenex @GitGab19  what are your thoughts about having a time constrain on release cycles?

RELEASE.md
```diff
@@ -1,88 +1,37 @@
...
+- The release process is initiated by creating a new release branch from the
+  `main` branch. The release branch is named `x.y.z`, where `x.y.z` is the
+  version number of the release.
+- The release branch is used as a breaking point(and future reference) for the
```
rrybarczyk, line 8:
```suggestion
- The release branch is used as a breaking point (and future reference) for the
```

RELEASE.md
```diff
@@ -1,88 +1,37 @@
...
+- If any bugs are found in the release branch, they should be fixed in the
+  release branch.
+- Create a Pull Request from release branch to the `main` branch.
+- The release branch is then tagged(with release notes) with the version number
+  of the release.
+- Release workspace crates to crates.io.
+
+## Versioning
+
+All workspace member crates follow SemVer 2.0.0. The version number is stored in
```
plebhash, line 21: let's just focus on enforcing SemVer 2.0.0 to `protocols` and `utils`

RELEASE.md
Fi3: I think that if we release ONLY on 6/8 weeks we will find ourself rushing to get PR merged nearby the release. That of course will impact the quality of the code. For example if we have a feature that we need and is almost ready and release is in 3 days we will do everything possible to have it in the next release otherwise we will have to wait 8 weeks.

GitGab19: I think that we should aim to a more precise time frame for future releases (even if it's generally hard in a FOSS environment), and 6 to 8 weeks looks reasonable to me. If we will encounter some barriers as the one mentioned by @Fi3 I'm pretty sure we could postpone the specific release a little bit. At the end we will keep asserting the progress on every dev call as usual. As a general guideline, I would leave it

Fi3: IMHO having shorter release cycle solve better this issue

plebhash: I'm not too opinionated on this, but is seems a middle ground would be to:
- aim for 4/6 weeks (shorter cycle)
- be open to exceptions in case specific PRs are taking longer than expected, in which case we postpone those specific PRs to the next release

what do you guys think?

GitGab19: I would wait for @pavlenex input, he's definitely the one with more experience regarding time frames and roadmap definition in FOSS 😉

RELEASE.md
plebhash: https://github.com/stratum-mining/stratum/issues/1128

RELEASE.md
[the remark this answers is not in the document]
jbesraa: umm I am maintaining `textwidth=80` everywhere and this is the first time I get a comment about it in stratum. is the 100 chars mentioned anywhere?

RELEASE.md
Fi3: > if some specific task is taking longer than expected, we postpone it to the next release (so that we're not rushing anything)

yep this is ideal and should be always like that but there is a positive correlation between the length of release cycle and the willinges to ignore best practice and rushing thing to not miss the release.

RELEASE.md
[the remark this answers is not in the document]
Fi3: I use 80 everywhere but in rust code. But I vote for 80.

RELEASE.md
```diff
@@ -1,88 +1,34 @@
...
+SemVer CI check would fail. Note that this does not apply to the `roles` and
+other crates in the repository.
+
+The general SRI version does not follow any SemVer rules. You can track the
+releases on Github Releases in the repository.
+
+## Tags and Branches
+
+- The `main` branch is the default branch and it is always active.
+- The `main` branch is protected and requires a pull request to merge changes
```
plebhash, line 31: should we add some words here about all PRs being merged to `main` via merge commits, and not rebases or squashes?

cc @GitGab19

GitGab19: Yeah, we can add a sentence like:
"Changes to `main` branch are added through a merge commit"

jbesraa: added

jbesraa:
@GitGab19 @plebhash what do you think about adding a `CHANGELOG.md` file in addition to the gh release notes, that should be updated on each release with the following:
- General release information
- Breaking changes(if any)
- Notable changes

GitGab19:
> @GitGab19 @plebhash what do you think about adding a `CHANGELOG.md` file in addition to the gh release notes, that should be updated on each release with the following:
> 
>     * General release information
> 
>     * Breaking changes(if any)
> 
>     * Notable changes

Yes. I think it can definitely add value 👍

plebhash:
> @GitGab19 @plebhash what do you think about adding a `CHANGELOG.md` file in addition to the gh release notes, that should be updated on each release with the following:
> 
> * General release information
> * Breaking changes(if any)
> * Notable changes

sounds like a good idea!

RELEASE.md
```diff
@@ -1,88 +1,60 @@
...
-Although our strategy is very similar to the classic gitflow model, we do not keep release branches.
+Usually the release process is as follows:
 
-![](git-branching.png)
+1. Create a new release branch from the `main` branch.
+2. Amend the changelog with the new release information.
+3. Create a Pull Request from the release branch to the `main` branch.
+4. Merge the Pull Request.
+5. Create a new tag for the release branch.
+6. Publish the release.
```
jbesraa, line 38: @plebhash How is the crate publishing is triggered atm?

plebhash: `release-libs.yaml`

also relevant: https://github.com/stratum-mining/stratum/pull/1142

RELEASE.md
```diff
@@ -1,88 +1,60 @@
...
 
-We follow a simplified [gitflow](https://nvie.com/posts/a-successful-git-branching-model/) branching strategy.
 
-Although our strategy is very similar to the classic gitflow model, we do not keep release branches.
+Usually the release process is as follows:
 
-![](git-branching.png)
+1. Create a new release branch from the `main` branch.
+2. Amend the changelog with the new release information.
+3. Create a Pull Request from the release branch to the `main` branch.
```
plebhash, line 35: I don't understand this step. If the new release branch was just created from `main`, why do we want to merge it back to `main`?

RELEASE.md
plebhash: it seems @pavlenex won't be available to provide his input

for now, let's avoid blocking this PR because of this

let's just remove this phrase about release timeline

GitGab19: > it seems @pavlenex won't be available to provide his input
> 
> for now, let's avoid blocking this PR because of this
> 
> let's just remove this phrase about release timeline

Agree 👍

RELEASE.md
plebhash: ok so the only file changed is the changelog file

GitGab19 (approved):

plebhash (approved):

Issue #1125, opened 2024-08-21
Consider moving to a single workspace


jbesraa:
Currently `roles` and `protocols` are separated and we have two workspaces in addition to `common` and other utils packages scattered around in the root folder. After we have completed https://github.com/stratum-mining/stratum/pull/994 I suggest to do the next step, combining the protocols/roles/common into a single workspace and rename each one to have a `stratum` prefix, i.e., `stratum-protocols` and so on.

Fi3:
My undestanding was that we were ok in put roles and protocols in different git repos (we disscussed it in a call if I remeber correctly) why now you want to put them in the same workspace? Common can fo with protocols I guess

Issue #1120, opened 2024-08-19
SRI Integration Test Framework


plebhash:
SRI Message Generator (MG) was originally conceived as a tool for interoperability tests, where different implementations of the Stratum V2 protocol could be tested against each other.

With time, the SRI community started using MG tests as a tool for Continuous Integration (CI).

There’s been [some discussion in the SRI community](https://github.com/stratum-mining/stratum/discussions/944#discussioncomment-9894276) about the possibility of partially replacing some MG tests with a new framework for Integration Tests. That framework would take inspiration from well established projects in the Bitcoin community such as [LDK](https://github.com/lightningdevkit/ldk-node/tree/main/tests) and Fedimint. 

It would allow MG to be limited to its originally intended purpose of interoperability testing, while also giving more flexibility and agility with regards to how CI is done for development of the SRI application layer.

This issue proposes a plan for how this migration will be executed. First, we propose a roadmap for the actual development of the new Integration Test Framework. Then, we propose a roadmap for the migration of the CI stack from MG tests into Integration Tests.

## SRI Integration Test Framework
This is the proposed roadmap for the creation of the new SRI Integration Test Framework. It is split in multiple stages. Some stages could be potentially executed in parallel (e.g.: 1 + 2 and 2 + 3), while others need to be executed sequentially (e.g.: 1, then 3).

### 1. Initial roles refactoring

This is the first stage of the creation of the new SRI integration test framework. It aims for the following goals:
- Get the roles code structured in a way that they can be composable inside the new integration test framework.
- Avoid breaking current behavior of roles. After this refactoring, it’s important that all roles continue passing the current MG test stack, and can also be launched manually while following the expected behavior.

This stage is currently being tracked by #1093 

### 2. Bitcoind integration
The second stage of the roadmap consists of the following goals:
- Ability to launch a Template Provider (TP) process (SV2-enabled `bitcoind` in `regtest` mode) via the [`bitcoind` crate](https://crates.io/crates/bitcoind). 
- SRI requires that the blockchain has at least 16 blocks, so the creation of those should be automated and abstracted away from the framework user (so it doesn’t have to be done explicitly for every test).

This should be useful for tests where orchestrating a `bitcoind` process in `regtest` mode is sufficient. 

There might be instances however where this is not sufficient. For example, we might want to assert some role’s behavior against an invalid template, which would be rather difficult to generate by simply orchestrating a `bitcoind` process without access to its internal C++ APIs.

[...]

PR #1126, opened 2024-08-21
Merge `protocols` and `roles` workspaces

jbesraa:
#1125 

This commits removes the `protocols` and `roles` workspaces and creates a new shared workspace for both of them. In practice we dont have a need to maintain two different workspaces and no benefit is derived from that.

With moving to a shared workspaces, the naming of the parent folders was changed as will with `roles` becoming `stratum-roles` and `protocols` becoming `stratum-protocols` and same for `common` crate. This required a fix a across the codebase of the imports inside the different `Cargo.toml`s.


Fi3:
see my comment on related issue

jbesraa:
@Fi3 I dont honestly remember when we talked about splitting them into different git repos, but anyhow thats a much bigger task than just combining them into a single workspace. Things are way over-engineered at the moment without any justification.

plebhash:
> I think there is also an open issue about that by @plebhash ?

There is this proposal discussion for a repository split: https://github.com/stratum-mining/stratum/discussions/1069#discussioncomment-10290201

But I'm not in a rush to execute this split just yet, because this has many implications and needs to be properly planned. 

For now, I'd be happy to keep the discussion going on a conceptual level, so we get consensus on the best repository structure and we can eventually plan accordingly for the execution in the future.

> Btw we splitted the workspaces with the goal of putting protocols and roles in separate repos. The plan was somthing like that:
> 1. develop the low level libraries
> 2. develop  a bunch of roles with the intention of (1) test the library (2) show that sv2 could work
> 3. stabilize the libraries
> 4. use the libraries to develop a bunch o prod grade tools (proxies, translator, everything together) whatever we see that is more needed. (I would start with downstream tools since the pools can develop upstream tools for themself but ofc here I'm in confilct of interest :)

I think we are in the middle of 3 right now, via #845

And also #1120 will help bring more stability into the work that was done on 2, because we will be able to get more confidence that the roles implementations are correctly implementing the protocol spec.

Once we move into 4 in the future, we will basically be trying to bring the work that was done in 2 into production readiness. Which ones we will focus first is a future discussion, and hopefully by that time we will also have more feedback from industry adopters, which will help us make those decisions.

---

with regards to this actual PR, I'm not very opinionated

in case I come across something that makes me realize this is needed I'll come back here and elaborate, but for now I'll focus on other tasks

Fi3:
concept nack
as the final goal is still to have different repos for libraries and applications

PR #1142, opened 2024-08-23
publish libs to `crates.io` on release tags

plebhash:
currently, `release-libs.yaml` will publish the libs to `crates.io` every time we merge to `main`, which is the main event for releases

with the new conventions proposed in #1124, we want to avoid that, since not every merge to `main` will be a release

this PR makes `release-libs.yaml` only be able to run:
- once new release tags are added
- via manual trigger

### DIFF
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -11,9 +11,12 @@
 name: Release Libs
 
 on:
+  # Manually run by going to "Actions/Release" in Github and running the workflow
+  workflow_dispatch:
+  # every time a new release tag is created
   push:
-    branches:
-      - main
+    tags:
+      - "v*.*.*"
 
 jobs:
   libs_publish:


jbesraa (commented):
Another option is to run it on every new tag https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

```shell
name: GitHub Release

on:
  push:
    tags:
      - "v*.*.*"
 ```

plebhash:
> Another option is to run it on every new tag https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs
> 
> ```shell
> name: GitHub Release
> 
> on:
>   push:
>     tags:
>       - "v*.*.*"
> ```

thanks! did some more research and found this: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

going for trigger on releases

plebhash:
during the review of #1143, @jbesraa [asked](https://github.com/stratum-mining/stratum/pull/1143#pullrequestreview-2266193936):

> ... why do we have `continue-on-error` ? if one crate does not update while other crates depend on it, and those update, this can break things. is it possible to publish all crates in a single atomic command?

on the header of `release-libs.yaml`, we have:
https://github.com/stratum-mining/stratum/blob/76659580fddfa909aabaaffa5d610b2d12b2091f/.github/workflows/release-libs.yaml#L1-L9

I believe those notes were originally written by @Fi3 . 

The part that really gets my attention is:
> when too many crates are updated crates.io could fail and ask to rerun the action later

@Fi3 have you ever seen this happen in practice? Or have you ever read it anywhere?

Anyways, if that is the case, I think that we should also have a `workflow_dispatch` for manually triggering the workflow in case of failure.

GitGab19 (approved):

PR #1143, opened 2024-08-23
stop publishing some crates

### LINKED ISSUE #1128: stop publishing `roles` to `crates.io`
our `release-libs` workflow is publishing all `roles` crates to `crates.io`

those crates are merely PoCs and not nearly close to production readiness, and won't be for a while

for now, we could simply stop publishing them

https://github.com/stratum-mining/stratum/blob/0f0ee1e32a7cdeb9af83718364e0896b36744824/.github/workflows/release-libs.yaml#L142-L166

plebhash:
close #1128

### DIFF
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -139,28 +139,3 @@ jobs:
         run: |
           cd roles/roles-utils/rpc
           cargo publish
-      - name: Publish crate jd_client
-        continue-on-error: true
-        run: |
-          cd roles/jd-client
-          cargo publish
-      - name: Publish crate jd_server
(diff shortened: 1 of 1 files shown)


GitGab19 (approved):

jbesraa (commented):
unrelated to the change, why do we have `continue-on-error` ? if one crate does not update while other crates depend on it, and those update, this can break things. is it possible to publish all crates in a single atomic command?

plebhash:
> unrelated to the change, why do we have `continue-on-error` ? if one crate does not update while other crates depend on it, and those update, this can break things. is it possible to publish all crates in a single atomic command?

migrated this discussion to #1142

PR #1155, opened 2024-09-04
Add integration tests project

jbesraa:
Adds an initial integration tests project to the `roles` workspace.

This crate should hold tests related to the other crates in the `roles` workspace, allowing us to write integration tests for those.


roles/tests-integration/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "integration-test"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+[[bin]]
+name = "run"
```
jbesraa, line 9: @plebhash try to checkout to this branch and run `cargo build` in `roles` workspace, it should work. but if you remove this bin and the `src` folder, you will see an error. let me know if thats not the behavior you are seeing.

roles/Cargo.toml
```diff
@@ -22,7 +22,7 @@ members = [
     "test-utils/sv1-mining-device",
     "translator",
     "jd-client",
-    "jd-server",
+    "jd-server", "tests-integration",
```
GitGab19, line 25: It's a stupid thing, but could you put it under `jd-server` like all the others?

roles/tests-integration/src/bin/run.rs
GitGab19, line 1: What's the purpose of this run?

roles/tests-integration/Cargo.toml
jbesraa: @GitGab19 ^^

roles/Cargo.toml
jbesraa: done

roles/Cargo.toml
```diff
@@ -23,6 +23,7 @@ members = [
     "translator",
     "jd-client",
     "jd-server",
+    "tests-integration",
```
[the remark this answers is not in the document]
Shourya742: might be misleading? or can be `roles-integration-tests`

roles/tests-integration/Cargo.toml
GitGab19: If I remove `[[bin]]` and `src` I get the error @jbesraa was mentioning.
I think the other way is to specify the `common/mod.rs` as a lib there.

I did some tests where I removed `src`, and I put this into `Cargo.toml`:
```
[lib]
path = "tests/common/mod.rs"
```
In this way everything is compiling, and we are explicitly saying that `integration-test` is not a binary

roles/Cargo.toml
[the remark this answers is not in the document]
GitGab19: Yeah I also prefer `roles-integration-tests`. Also in the relative `Cargo.toml`

roles/tests-integration/Cargo.toml
Shourya742: Following @GitGab19's suggestion, I was able to successfully build the roles. It seems more efficient to have the integration tests as a lib, as this allows us to remove the redundant `run.rs` file. Interestingly, Fedimint follows a similar workspace structure and also organizes their integration tests as libs, as shown here: https://github.com/fedimint/fedimint/tree/master/fedimint-testing-core.

rrybarczyk (changes requested):
This setup does not follow Rust's way of doing integration tests. See my comments on #1120:
- Internal to crate integration test setup [comment](https://github.com/stratum-mining/stratum/issues/1120#issuecomment-2338453729).
- Multiple crate integration test setup [comment](https://github.com/stratum-mining/stratum/issues/1120#issuecomment-2338487323).

If the intention of this PR is to setup the basic scaffolding of these tests, we should follow the aforementioned patterns.

jbesraa:
> This setup does not follow Rust's way of doing integration tests. See my comments on #1120:
> 
>     * Internal to crate integration test setup [comment](https://github.com/stratum-mining/stratum/issues/1120#issuecomment-2338453729).
> 
>     * Multiple crate integration test setup [comment](https://github.com/stratum-mining/stratum/issues/1120#issuecomment-2338487323).
> 
> 
> If the intention of this PR is to setup the basic scaffolding of these tests, we should follow the aforementioned patterns.

Thanks for the review!

Do you have an example repo with the mentioned structure for 'Multiple crate integration test setup'?
From what I understand, a new crate is required in the workspace `Cargo.toml` and then the `tests` folder should live within this new crate. 
If a new `tests` folder is created without a crate, whats the way to invoke the tests inside this folder and how crates  are imported into the tests(internal or external)?
An example from tokio https://github.com/tokio-rs/tokio/tree/master/tests-integration

roles/tests-integration/Cargo.toml
plebhash: I set my `rustc` to 1.75.0 and I still cannot reproduce this build error

anyways, I agree with the suggestions above

using `lib` feels more coherent than having an empty `run.rs` file

plebhash:
> This setup does not follow Rust's way of doing integration tests. See my comments on #1120:
> 
> * Internal to crate integration test setup [comment](https://github.com/stratum-mining/stratum/issues/1120#issuecomment-2338453729).
> * Multiple crate integration test setup [comment](https://github.com/stratum-mining/stratum/issues/1120#issuecomment-2338487323).
> 
> If the intention of this PR is to setup the basic scaffolding of these tests, we should follow the aforementioned patterns.

I agree with the sentiment against adding yet another crate to the codebase.

However, given the fact that all roles are already separate crates under the same workspace, I'm afraid this is the only available path forward.

If we just added a `tests/integration_test.rs` file under the `roles/Cargo.toml` (which is a virtual manifest), we would be violating the fact that virtual manifests are not allowed to have `bin` or `lib` sections.

I tried implementing this locally and got stuck with:
```
error: failed to parse manifest at `/Users/plebhash/develop/stratum/roles/Cargo.toml`

Caused by:
  this virtual manifest specifies a `lib` section, which is not allowed
```

```
error: failed to parse manifest at `/Users/plebhash/develop/stratum/roles/Cargo.toml`

Caused by:
  this virtual manifest specifies a `bin` section, which is not allowed
```

Maybe in the future when we have a unified crate containing all roles (which is not something we're aiming for now), it will be easier to bypass the need for a new crate and we will be able to have all integration tests as part of the same crate (much like [`ldk-node`](https://github.com/lightningdevkit/ldk-node/tree/main/tests)).

For now, I could live with this tradeoff of having one extra crate inside `roles`, which seems hard to get around.

rrybarczyk:
Ahh that is too bad.

Let's do what ldk-node does then.

rrybarczyk:
I think we should consider having integration tests in each `roles` sub-crate though, wherever possible. Maybe post-refactor planning we can see what kind of things we can cover there.

roles/Cargo.toml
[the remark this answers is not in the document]
rrybarczyk: I think it should just be `tests`.

roles/tests-integration/Cargo.toml
rrybarczyk: Do we need a `bin/` at all? Looking at [ldk-node](https://github.com/lightningdevkit/ldk-node/tree/main/tests), it does not have any executable.

plebhash: we need either a `[lib]` or a `[bin]` statement on `Cargo.toml` of the integration tests crate

the debate here is around the fact that @jbesraa added a placeholder`src/bin/run.rs` so that he could have a `[bin]`, but @GitGab19 suggested to use `[lib]` with a file we already have, which is `tests/common/mod.rs`

to elaborate a bit on what I said on this comment https://github.com/stratum-mining/stratum/pull/1155#issuecomment-2339022184, `ldk-node` will be a good reference for these kind of patterns in a hypothetical future where we have all `roles` unified under one single crate

for now, it is not a good reference, because following its patterns would imply that we would need to unify all crates from `roles` into a single crate

roles/Cargo.toml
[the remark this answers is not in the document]
jbesraa: initially  `tests-integration` was chosen because there is already a `tests-utils` folder and I wanted to be consistent. Is it worth deciding on the folder naming now, or when we know better how the role's crates are going to be structured/named?  I prefer to keep it as is and rename it when restructuring the `roles` crates is underway, otherwise happy to change to either `tests` or `roles-integration-tests`, whatever gets more thumbs up.

jbesraa:
> I think we should consider having integration tests in each `roles` sub-crate though, wherever possible. Maybe post-refactor planning we can see what kind of things we can cover there.

Yup that would be great to have. But as mentioned, its really outside of the scope of this Pull Request.

rrybarczyk (approved):
Lgtm

roles/Cargo.toml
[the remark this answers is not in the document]
Shourya742: makes sense!!!

Shourya742 (approved):
ACK

PR #1156, opened 2024-09-04
Add Template Provider to integration test project

jbesraa:
blocked by #1155

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -1,1 +1,152 @@
+use bitcoind::{bitcoincore_rpc::RpcApi, BitcoinD, Conf};
+use flate2::read::GzDecoder;
+use std::{
+    env,
+    fs::{create_dir_all, File},
+    io::{BufReader, Read},
+    path::{Path, PathBuf},
+};
+use tar::Archive;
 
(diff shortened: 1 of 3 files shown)


roles/tests-integration/tests/common/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        if let Ok(file) = entry.path() {
+            if file.ends_with("bitcoind") {
+                entry.unpack_in(destination).unwrap();
+            }
+        }
+    }
+}
+
+fn get_bitcoind_filename(os: &str, arch: &str) -> String {
+    match (os, arch) {
```
Shourya742, line 49: Should we move these to a separate utils module?

jbesraa: That was my initial though as well. But I think we should wait a bit and see if those functions are gonna be used by other roles, otherwise they really should be part of the `TemplateProvier` module in `tests::common::mod::TemplateProvider`.

Shourya742: Yeah, that makes sense. I don't think these utility methods will be used anywhere else except here. Let's keep them here for now, and if we find that another module uses them, we can move them to a separate utility module.

plebhash (approved):

jbesraa:

PR #1158, opened 2024-09-05
Fix bencher false alarms on CI

### LINKED ISSUE #1051: Dig github CI benches execution and fix false alarms

GitGab19:
This PR updates the way branches are managed by [bencher](bencher.dev), greatly improving the possibility of getting false alarms (about performance regressions) by our CI. 
Close #1051

### DIFF
--- a/.github/workflows/run-benchmarks.yaml
+++ b/.github/workflows/run-benchmarks.yaml
@@ -4,7 +4,6 @@ on:
   pull_request:
     branches:
       - main
-      - dev
 
 jobs:
   benchmark_sv1_criterion:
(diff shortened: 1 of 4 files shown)


jbesraa (approved):
Looks good. Did you get the chance to inspect how the 50/100 sample sizes make difference? should the time also be increased in the `measurement_time` ?

We might also in the future want to move the the env's initiated in the `script` section to an `env` section and that could probably allow us to use the envs like `process.env.PR_BASE` across the different `run's`

GitGab19:
Moving from 50 to 100 was the suggestion I got from the `bencher` maintainer, and  I saw some improvements when I switched to 100, during my tests. Regarding the `env` section I agree, but that was not the scope of this PR, and I also prefer to have these workflows as they are now, since they reflect perfectly the actual `bencher` docs

plebhash (approved):

PR #1163, opened 2024-09-12
Separate `lib` code from `bin` in `mining-device`

jbesraa:
Resolving `MiningDevice` in the list here https://github.com/stratum-mining/stratum/issues/1093


roles/test-utils/mining-device/Cargo.toml
```diff
@@ -24,3 +29,6 @@ tracing = { version = "0.1" }
 tracing-subscriber = "0.3"
 sha2 = "0.10.6"
 tokio = "^1.38.0"
+
+[features]
+abort_mining = []
```
plebhash, line 34: what is the purpose of this feature?

jbesraa: I added this to replicate exactly the MG test, which says that the `mining-device` should not start mining https://github.com/stratum-mining/stratum/pull/1163/files#diff-cc06b104e73cde20ec61e683c5b4eb74e5422641e422652485c1026dd72a7a98R255

In practice I do not think its needed tbh because anyway the mining-device wont be able to open a channel

roles/test-utils/mining-device/Cargo.toml
```diff
@@ -1,11 +1,16 @@
 [package]
-name = "mining-device"
+name = "mining_device"
```
rrybarczyk, line 2: Let's keep the `-` here, not an `_`.

roles/test-utils/mining-device/Cargo.toml
```diff
@@ -1,11 +1,16 @@
...
 version = "0.1.1"
 edition = "2018"
 publish = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[lib]
+name = "mining_device"
```
rrybarczyk, line 10: Keep `-`, not `_`.

rrybarczyk (changes requested):

roles/test-utils/mining-device/Cargo.toml
plebhash: ok so why do we have it? can't we get away with not having this?

roles/test-utils/mining-device/Cargo.toml
jbesraa: I should have written why I did that from the first place, sorry. 

Initially I did not change the hyphen, but then I got the following error: 
```
Caused by: library target names cannot contain hyphens
```
There is something about it here https://github.com/rust-lang/cargo/issues/11259 but I did not dig deep into it tbh.

roles/test-utils/mining-device/Cargo.toml
jbesraa: Removed the redundant flag

roles/test-utils/mining-device/Cargo.toml
rrybarczyk: Ok! Let's keep `_`.

rrybarczyk (approved):
Looks good!

roles/test-utils/mining-device/src/lib/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+    }
+
+    fn handle_open_standard_mining_channel_success(
+        &mut self,
+        m: OpenStandardMiningChannelSuccess,
+        _: Option<std::sync::Arc<Mutex<()>>>,
+    ) -> Result<SendTo<()>, Error> {
+        self.channel_opened = true;
+        self.channel_id = Some(m.channel_id);
+        let req_id = m.get_request_id_as_u32();
```
rrybarczyk, line 389: The `cargo clippy` CI fails on this line with:

```sh
warning: unused variable: `req_id`
   --> test-utils/mining-device/src/lib/mod.rs:389:13
    |
389 |         let req_id = m.get_request_id_as_u32();
    |             ^^^^^^ help: if this is intentional, prefix it with an underscore: `_req_id`
    |
    = note: `#[warn(unused_variables)]` on by default
```

The reason it fails is because the `main` branch immediately uses this `req_id` in an `info!` log on [L438](https://github.com/stratum-mining/stratum/blob/ab0ddd625706698c47d44b9b4e4f985c39be18db/roles/test-utils/mining-device/src/main.rs#L438) with:

```rs
        info!(
            "MINING DEVICE: channel opened with: group id {}, channel id {}, request id {}",
            m.group_channel_id, m.channel_id, req_id
        );
```

Why did we remove this log? Are there any other logs that got removed? Are there any other changes in the code in this PR beyond breaking up the logic into the `mod.rs` file?

rrybarczyk (changes requested):
Need to understand why we removed the `info` log containing the `req_id` and then fix the `cargo clippy` error in the CI.

roles/test-utils/mining-device/Cargo.toml
plebhash: > Removed the redundant flag

I don't understand, `abort_mining` seems to be still there

roles/test-utils/mining-device/src/lib/mod.rs
plebhash: @jbesraa we need to make sure that this PR is strictly moving lines of code to different files but without losing logic, which seems to be the root cause of this

plebhash:
> CI is currently broken, it is fixed at #1170

@jbesraa FYI this was also necessary to fix the `sv2.h Header Check` https://github.com/stratum-mining/stratum/pull/1172

please rebase with this

jbesraa:

GitGab19 (approved):

plebhash (approved):

rrybarczyk (approved):

PR #1164, opened 2024-09-12
Add integration test for auth failure between `mining_device` and a `pool`

### LINKED ISSUE #1165: Migrate pool downstream auth rejection from MG to integration test
MG test: https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/pool-sri-test-reject-auth/pool-sri-test-reject-auth.json

This should a good example on how to approach asserting a role state/behavior after message exchange with other role.

jbesraa:
Blocked by #1163 #1066 #1155 #1156
Resolves #1165 

Note that majority of the commits here will be hopefully merged in the blocking PRs mentioned above. This PR should end up with the last three commits.

This should make the following test redundant https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/pool-sri-test-reject-auth/pool-sri-test-reject-auth.json


protocols/v2/roles-logic-sv2/src/errors.rs
```diff
@@ -61,6 +61,7 @@ pub enum Error {
     HashrateError(InputError),
     LogicErrorMessage(std::boxed::Box<AllMessages<'static>>),
     JDSMissingTransactions,
+    MiningChannelError(u32),
```
rrybarczyk, line 64: Why do we need to add this error type? Also, let's not prepend the variant with the word `Error` as it is redundant.

```suggestion
    MiningChannel(u32),
```

protocols/v2/roles-logic-sv2/src/errors.rs
```diff
@@ -153,6 +154,7 @@ impl Display for Error {
             HashrateError(e) => write!(f, "Impossible to get Hashrate: {:?}", e),
             LogicErrorMessage(e) => write!(f, "Message is well formatted but can not be handled: {:?}", e),
             JDSMissingTransactions => write!(f, "JD server cannot propagate the block: missing transactions"),
+            MiningChannelError(e) => write!(f, "Mining channel error: {:?}", e),
```
rrybarczyk, line 157: Could we come up with a more descriptive error message here?

```suggestion
            MiningChannel(e) => write!(f, "Mining channel error: {:?}", e),
```

jbesraa:
will close this for now and wait for #1066 to get merged and then maybe re-open this..

Issue #1048, opened 2024-07-10
Cargo build produces several warning if used with recent toolchain


lorbax:
I have toolchain with rustc 1.78
```

~/s/s/protocols (dev) [1]> rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/lorban/.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu (default)
nightly-x86_64-unknown-linux-gnu
1.75.0-x86_64-unknown-linux-gnu
1.75-x86_64-unknown-linux-gnu

active toolchain
----------------

stable-x86_64-unknown-linux-gnu (default)
rustc 1.78.0 (9b00956e5 2024-04-29)
```
If I compile `protocols` crate I get the following warnings
```
warning: trait `Variable` is never used
  --> v2/binary-sv2/no-serde-sv2/codec/src/codec/mod.rs:37:11
   |
37 | pub trait Variable {
   |           ^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: trait `IntoOwned` is never used
 --> v2/binary-sv2/no-serde-sv2/codec/src/datatypes/non_copy_data_types/mod.rs:7:7
  |
7 | trait IntoOwned {
  |       ^^^^^^^^^

warning: `binary_codec_sv2` (lib) generated 2 warnings
warning: `binary_codec_sv2` (lib) generated 2 warnings (2 duplicates)
warning: methods `into_aesg` and `into_chacha` are never used
  --> v2/noise-sv2/src/cipher_state.rs:26:8
   |
7  | pub trait CipherState<Cipher_: AeadCipher>
   |           ----------- methods in this trait
...
26 |     fn into_aesg(mut self) -> Option<Cipher<Aes256Gcm>> {
   |        ^^^^^^^^^
...
33 |     fn into_chacha(mut self) -> Option<Cipher<ChaCha20Poly1305>> {
   |        ^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: associated items `name`, `hkdf_3`, and `ecdh` are never used
   --> v2/noise-sv2/src/handshake.rs:10:8
    |
9   | pub trait HandshakeOp<Cipher: AeadCipher>: CipherState<Cipher> {
    |           ----------- associated items in this trait
10  |     fn name(&self) -> String;
    |        ^^^^
...
69  |     fn hkdf_3(
    |        ^^^^^^
...
[...]

PR #1168, opened 2024-09-18
Use version 1.75.0 of toolchain in scripts 

lorbax:


plebhash (approved):

jbesraa (approved):

jbesraa:
@lorbax did you have any problem with a rust toolchain?

lorbax:
> @lorbax did you have any problem with a rust toolchain?

Have we done anything about this issue?
https://github.com/stratum-mining/stratum/issues/1048
if no, the same issues still remain. These scripts are used to ensure that at least with 1.75 there are no clippy warnings.

PR #1177, opened 2024-10-02
use fixed `temp_dir` for TP on integration tests

### LINKED ISSUE #1176: integration test TP is downloaded every time
we are downloading and unpacking TP every time we run the test, which is a bit redundant

https://github.com/stratum-mining/stratum/blob/ba4c93508421fc1533893dd642a739bc41320255/roles/tests-integration/tests/common/mod.rs#L67-L70

 https://github.com/stratum-mining/stratum/blob/ba4c93508421fc1533893dd642a739bc41320255/roles/tests-integration/tests/common/mod.rs#L88-L106

for some unknown reason, it seems sometimes the test is getting stuck on the step of downloading/unpacking the TP

regardless of the root cause, I think we should simply use a static `"/tmp/.template-provider"` for `path_name`... no point in downloading and unpacking everything every time

making `path_name` into a static string made the issue go away

if it keeps happening on the initial bootstrapping of TP, we should investigate more

plebhash:
fix #1176

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -66,10 +66,11 @@ pub struct TemplateProvider {
 
 impl TemplateProvider {
     pub fn start(port: u16) -> Self {
-        let path_name = format!("/tmp/.template-provider-{}", port);
-        let temp_dir = PathBuf::from(&path_name);
+        let temp_dir = PathBuf::from("/tmp/.template-provider");
 
         let mut conf = Conf::default();
+        let staticdir = format!(".bitcoin-{}", port);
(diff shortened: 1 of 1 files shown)


plebhash:
> I think this will make all tests run against the same bitcoind instance which I am not sure it is a good idea

not really... they will use the same executable from disk but each test will launch its own `bitcoind` process

jbesraa:
Ok now with the `static_dir` change i think it makes more sense.. will test this tomorrow and review again

jbesraa (approved):
Tested this locally, with running multiple tests utilising the TP and all is good. it actually made things slightly faster.

GitGab19 (approved):

Issue #1178, opened 2024-10-02
`framing-sv2` doc cmts do not conform to standard


rrybarczyk:
As a part of the `protocols` documentation and refactor effort in #845, Rust doc comments were added to the `framing-sv2` crate in #848. However, this was done before the team fully finalized the standardization for the Rust doc comments in discussion #970.

The content of the `framing-sv2` documentation post-#845 is sound, only the formatting needs to be updated. A `README.md` will also be added, following the established standard.

Issue #1135, opened 2024-08-21
Refactor: `const-sv2` crate


rrybarczyk:
# Background
This task is an outcome of the `protocols` Rust docs issues tracked in #845.

While documenting `protocols::v2::const-sv2` in #1016, areas of potential code debt were identified. This issue servers as a place to list out these items to then be addressed in an organized manner. The initial Rust documentation effort was an immediate action, while a refactoring (which implies breaking API changes) is not so urgent priority, so for now we should leave this in the backlog for an appropriate moment in the future.

# Identified Potential Code Debt

- `SV2_FRAME_HEADER_LEN_OFFSET`, `SV2_FRAME_HEADER_LEN_END`, and `NOISE_FRAME_MAX_SIZE` are not used anywhere --> we should just remove them 
- `ENCRYPTED_SV2_FRAME_HEADER_SIZE` and `NOISE_FRAME_HEADER_SIZE` are declared in `sv2-ffi`, and imported in `framing_sv2/src/header.rs`, which is then imported into `codec_sv2` just for this constant --> we should import them directly into `codec_sv2` (cc @rrybarczyk)
- `NOISE_FRAME_HEADER_LEN_OFFSET` is imported in `framing_sv2` in the same way [here](https://github.com/stratum-mining/stratum/blob/708e5e4ecb1f67d4e629b68dd4965271ced389c4/protocols/v2/framing-sv2/src/header.rs#L13), but then its alias is never used anywhere else --> we should remove it
- `RESPONDER_EXPECTED_HANDSHAKE_MESSAGE_SIZE` is an alias of `ELLSWIFT_ENCODING_SIZE` --> we should create the alias where it's used, not here
- `MAC` it's the same as `AEAD_MAC_LEN` --> we should remove it
- `NOISE_SUPPORTED_CIPHERS_MESSAGE` is used to signal AESG support but we're dropping support for AESG --> we should remove it when we will officially remove support from our implementation
- `SV2_TEMPLATE_DISTR_PROTOCOL_DISCRIMINANT` --> I would rename it in `SV2_TEMPLATE_DISTRIBUTION_PROTOCOL_DISCRIMINANT`
- `MESSAGE_TYPE_OPEN_EXTENDED_MINING_CHANNEL_SUCCES` --> fix typo with `MESSAGE_TYPE_OPEN_EXTENDED_MINING_CHANNEL_SUCCESS`
[...]

Issue #1129, opened 2024-08-21
Refactor: `codec-sv2` crate


rrybarczyk:
# Background
This task is an outcome of the `protocols` Rust docs issues tracked in #845.

While documenting `protocols::v2::codec-sv2`, areas of potential code debt were identified. This issue servers as a place to list out these items to then be addressed in an organized manner. The initial Rust documentation effort was an immediate action, while a refactoring (which implies breaking API changes) is not so urgent priority, so for now we should leave this in the backlog for an appropriate moment in the future.

> Note: This issue will replace the #873 discussion.

# Identified Potential Code Debt

## Error Variant Naming

Suggestions for `codec_sv2::error::Error` and `::CError` refactor:

* Instead of importing at the top of the file:
  * Use `noise_sv2` crate directly in the `Error` variants like `AeadError(noise_sv2::AeadError)` and `NoiseSv2Error(noise_sv2::Error)`.
  * Use `framing_sv2` crate directly in the `Error` variant like `FramingSv2Error(framing_sv2::Error)`.
* There are two definitions of the `framing_sv2::Error` variant: `FramingError` and `FramingSv2Error`, remove one of these variant and handle any downstream effects.
*  Rename (applies to both the `codec_sv2::Error` and `codec_sv2::CError` enums):
  * `AeadError` -> Either keep as is or do:`NoiseSv2Aead` or  `NoiseAead`
  * `BinarySv2Error` -> `BinarySv2`or `Binary`
  * `FramingSv2Error`/`FramingError` -> `FramingSv2` or `Framing`
  * `NoiseSv2Error` -> `Noise`
* Alphabetize the error variants, the match arms of the implementations for the `Error` enums, and the order of the `impl From<external-error> for Error`

## Uniform Spelling
* We are using both `HandShake` and `Handshake`. Everything should be `Handshake` with a lowercase "s" (this is how it is defined in [The Noise Protocol Framework](https://noiseprotocol.org/noise.html)) . This impacts:
  * `codec_sv2::`:
    * `lib::State::HandShake` -> `lib::State::Handshake`
    * `error::Error::NotInHandShake` -> `error::Error::NotInHandshake`
    * `error::CError::NotInHandShake` -> `error::CError::NotInHandshake`
  * `framing_sv2::framing2::` (there may be more in `framing_sv2`, but these are what I have noticed so far):
    * `HandShakeFrame` -> `HandshakeFrame` 
[...]

PR #1179, opened 2024-10-02
`framing-sv2` doc cmts to conform to standard

rrybarczyk:
Addresses #1178.

### DIFF
--- a/protocols/v2/framing-sv2/README.md
+++ b/protocols/v2/framing-sv2/README.md
@@ -1,3 +1,70 @@
-# framing_sv2
+# `framing_sv2`
 
-`framing_sv2` is a Rust `no_std` crate
+[![crates.io](https://img.shields.io/crates/v/framing_sv2.svg)](https://crates.io/crates/framing_sv2)
+[![docs.rs](https://docs.rs/framing_sv2/badge.svg)](https://docs.rs/framing_sv2)
+[![rustc+](https://img.shields.io/badge/rustc-1.75.0%2B-lightgrey.svg)](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html)
+[![license](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg)](https://github.com/stratum-mining/stratum/blob/main/LICENSE.md)
+[![codecov](https://codecov.io/gh/stratum-mining/stratum/branch/main/graph/badge.svg?flag=framing_sv2-coverage)](https://codecov.io/gh/stratum-mining/stratum)
+
+`framing_sv2` provides utilities for framing messages sent between Sv2 roles, handling both Sv2
+message and Noise handshake frames.
+
+The Sv2 protocol is binary, with fixed message framing. Each message begins with the extension
+type, message type, and message length (six bytes in total), followed by a variable length
+message. The message framing is outlined below ([according to Sv2 specs
+](https://github.com/stratum-mining/sv2-spec/blob/main/03-Protocol-Overview.md#32-framing)):
+
+| Field Name  | Byte Length | Description |
+|----------------|-------------|-------------|
+| `extension_type` | `U16` | Unique identifier of the extension associated with this protocol message. |
+| `msg_type` | `U8` | Unique identifier of this protocol message. |
+| `msg_length` | `U24` | Length of the protocol message, not including this header. |
+| `payload` | `BYTES` | Message-specific payload of length `msg_length`. If the MSB in `extension_type` (the `channel_msg` bit) is set the first four bytes are defined as a `U32` `"channel_id"`, though this definition is repeated in the message definitions below and these 4 bytes are included in `msg_length`. |
+
+Some bits of the `extension_type` field can also be repurposed for signaling on how the frame
+should be handled across channels.
+
+The least significant bit of `extension_type` (i.e.bit 15, 0-indexed, aka `channel_msg`) indicates
+a message which is specific to a channel, whereas if the most significant bit is unset, the message
+is to be interpreted by the immediate receiving device.
+
+Note that the `channel_msg` bit is ignored in the extension lookup, i.e.an `extension_type` of
+`0x8ABC` is for the same "extension" as `0x0ABC`.
+
+If the `channel_msg` bit is set, the first four bytes of the payload field is a `U32` representing
+the `channel_id` this message is destined for (these bytes are repeated in the message framing
+descriptions below).
+
+Note that for the Job Declaration and Template Distribution Protocols the `channel_msg` bit is
+always unset.
+
+## Main Components
+
+- **Header**: Defines the 6-byte Sv2 message header with information about the message payload,
+  including its extension type, if it is associated with a specific mining channel, the type of
+  message (e.g. `SetupConnection`, `NewMiningJob`, etc.) and the payload length.
+- **Sv2 Framing**: Use for serializing Sv2 messages.
+- **Noise Handshake Framing**: Use for serializing Noise protocol handshake messages.
+
+## Usage
+
+To include this crate in your project, run:
+
+```bash
+cargo add framing_sv2
+```
+
+This crate can be built with the following feature flags:
+
+- `with_buffer_pool`: Enables buffer pooling for more efficient memory management.
+- `with_serde`: builds [`binary_sv2`](https://crates.io/crates/binary_sv2) and
+  [`buffer_sv2`](https://crates.io/crates/buffer_sv2) crates with `serde`-based encoding and
+  decoding. Note that this feature flag is only used for the Message Generator, and deprecated
+  for any other kind of usage. It will likely be fully deprecated in the future.
+
+### Examples
+
+This crate provides an example demonstrating how to serialize and deserialize Sv2 message frames:
+
+1. **[Sv2 Frame](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/framing-sv2/examples/sv2_frame.rs)**:
+   Constructs, serializes, and deserialize a regular Sv2 message frame (`Sv2Frame`).
--- a/protocols/v2/framing-sv2/src/header.rs
+++ b/protocols/v2/framing-sv2/src/header.rs
@@ -1,3 +1,24 @@
+//! # Sv2 Frame Header
+//!
+//! Defines the [`crate::header::Header`] structure used in the framing of Sv2 messages.
+//!
+//! Each [`crate::framing::Sv2Frame`] starts with a 6-byte header with information about the
+//! message payload, including its extension type, if it is associated with a specific mining
+//! channel, the type of message (e.g. `SetupConnection`, `NewMiningJob`, etc.) and the payload
+//! length.
+//!
+//! ## Header Structure
+//!
+//! The Sv2 header includes the following fields:
+//!
+//! - `extension_type`: A `16`-bit field that describes the protocol extension associated with the
+//!   message. It also contains a special bit (the `channel_msg` bit) to indicate if the message is
+//!   tied to a specific channel.
+//! - `msg_type`: An `8`-bit field representing the specific message type within the given
+//!   extension.
+//! - `msg_length`: A `24`-bit field that indicates the length of the message payload, excluding the
+//!   header itself.
+
 use crate::Error;
 #[cfg(not(feature = "with_serde"))]
 use alloc::vec::Vec;
@@ -14,7 +35,7 @@ pub const NOISE_HEADER_LEN_OFFSET: usize = const_sv2::NOISE_FRAME_HEADER_LEN_OFF
 // Previously `NoiseHeader::HEADER_SIZE`
 pub const NOISE_HEADER_SIZE: usize = const_sv2::NOISE_FRAME_HEADER_SIZE;
 
-/// Abstraction for a SV2 Frame Header.
+/// Abstraction for a Sv2 Frame Header.
 #[derive(Debug, Serialize, Deserialize, Copy, Clone)]
 pub struct Header {
     // Unique identifier of the extension describing this protocol message.  Most significant bit
@@ -26,8 +47,10 @@ pub struct Header {
     // channel_id this message is destined for. Note that for the Job Declaration and Template
     // Distribution Protocols the channel_msg bit is always unset.
     extension_type: u16, // fix: use U16 type
+    //
     // Unique identifier of the extension describing this protocol message
     msg_type: u8, // fix: use specific type?
+
     // Length of the protocol message, not including this header
     msg_length: U24,
 }
@@ -35,7 +58,7 @@ pub struct Header {
 impl Header {
     pub const SIZE: usize = const_sv2::SV2_FRAME_HEADER_SIZE;
 
-    /// Construct a `Header` from raw bytes
+    /// Construct a [`Header`] from raw bytes
     #[inline]
     pub fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
         if bytes.len() < Self::SIZE {
@@ -59,7 +82,7 @@ impl Header {
         inner as usize
     }
 
-    // Construct a `Header` from payload length, type and extension type.
+    // Construct a [`Header`] from payload length, type and extension type.
     #[inline]
     pub(crate) fn from_len(msg_length: u32, msg_type: u8, extension_type: u16) -> Option<Header> {
         Some(Self {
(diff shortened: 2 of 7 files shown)


plebhash:
we should take this opportunity to add a `README.md`

rrybarczyk:
A couple notes of areas of improvement:

* I do not think I captured the need for `Header::encrypted_len()` properly. I understand this to be used when the message is too big to fit into a single frame, defined by the `SV2_FRAME_CHUNK_SIZE` constant. But I see this method is only used in the `codec_sv2::decoder::WithNoise` struct. Why don't we use this constant in the `Sv2Frame` decoder `WithoutNoise` struct? Do Sv2 messages ever exceed this max size that can fit into an `Sv2Frame`? Or do only the handshake messages (since they are so much bigger) ever exceed the size? To be clear, this technically does not have to do with Noise, but rather message chunking. I am just trying to understand why this method only is used on the Noise codec logic and not for `Sv2Frames`.
* I don't think my `handshake.rs` example does a great job of showcasing the handshake framing logic. Any feedback would be great, and I can always delete it if it is not helpful.
* I would like to make sure the reader/user understands that the `codec_sv2` crate is used directly after using this `framing_sv2` crate to encode the framed message for transmission and then decode it on receipt. However, I never could find the right wording or right spot to explain this. Not sure if it matters or not? 
* In the `README`, I used a lot of language in the doc comments to stay consistent. Happy to modify if needed.

protocols/v2/framing-sv2/README.md
```diff
@@ -0,0 +1,60 @@
+# `framing_sv2`
+
+[![crates.io](https://img.shields.io/crates/v/framing_sv2.svg)](https://crates.io/crates/framing_sv2)
+[![docs.rs](https://docs.rs/framing_sv2/badge.svg)](https://docs.rs/framing_sv2)
+[![rustc+](https://img.shields.io/badge/rustc-1.75.0%2B-lightgrey.svg)](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html)
+[![license](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg)](https://github.com/stratum-mining/stratum/blob/main/LICENSE.md)
+[![framingov](https://framingov.io/gh/stratum-mining/stratum/branch/main/graph/badge.svg?flag=framing_sv2-coverage)](https://framingov.io/gh/stratum-mining/stratum)
```
GitGab19, line 7:
```suggestion
[![codecov](https://codecov.io/gh/stratum-mining/stratum/branch/main/graph/badge.svg?flag=framing_sv2-coverage)](https://codecov.io/gh/stratum-mining/stratum)
```

protocols/v2/framing-sv2/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+`framing_sv2` provides utilities for framing messages sent between Sv2 roles, handling both Sv2
+message and Noise handshake frames.
+
+The Sv2 protocol is binary, with fixed message framing. Each message begins with the extension
+type, message type, and message length (six bytes in total), followed by a variable length
+message. The message framing is outlined below ([according to SV2 specs
+](https://stratumprotocol.org/specification/03-Protocol-Overview/#32-framing)):
+
```
GitGab19, line 16: I would always use the github repo link to specs. In this case it would be: https://github.com/stratum-mining/sv2-spec/blob/main/03-Protocol-Overview.md#32-framing

protocols/v2/framing-sv2/examples/handshake_frame.rs
```diff
@@ -0,0 +1,49 @@
...
+// - Define a handshake payload to be framed.
+// - Construct a handshake frame with the provided payload.
+// - Serialize the handshake frame into a byte array.
+// - Deserialize the handshake frame from the byte array back into a `HandShakeFrame`.
+// - Verify that the handshake frame is correctly processed.
+//
+// In the Sv2 protocol, these frames can then be encoded and transmitted between Sv2 roles that
+// have a connection secured with the Noise protocol.
+//
+// have established a secure Noise handshake.
```
GitGab19, line 15: Typo here. Duplicate sentence

Shourya742 (changes requested):

protocols/v2/framing-sv2/src/lib.rs
```diff
@@ -12,12 +19,38 @@
...
-//! # Features
+//!
+//! ## Usage
+//!
+//! Nearly all messages sent between Sv2 roles are serialized with the [`framing::Sv2Frame`]. The
+//! exception is when two Sv2 roles exchange Noise protocol handshake messages.
+//!
+//! Before Sv2 roles can communicate securely, they must perform a Noise handshake (note that Noise
+//! encryption is optional for communication between two local Sv2 roles (i.e. a local mining
+//! device and a local mining proxy), but required between two remote Sv2 roles (i.e. a local
```
plebhash, line 30: there's nothing really imposing encryption, even if the roles are remote.

```suggestion
//! device and a local mining proxy), but recommended between two remote Sv2 roles (i.e. a local
```

protocols/v2/framing-sv2/src/header.rs
```diff
@@ -51,15 +73,15 @@ impl Header {
         })
     }
 
-    /// Get the payload length
+    // Get the payload length
```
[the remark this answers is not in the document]
rrybarczyk: I chose private docs for `pub(crate)` functions as they are not rendered. @plebhash do you know which comment type we should use for types and methods marked public to the crate only?

protocols/v2/framing-sv2/src/lib.rs
rrybarczyk: I learned the concept of this being required from the [4 Protocol Security](https://github.com/stratum-mining/sv2-spec/blob/main/04-Protocol-Security.md#4-protocol-security) spec in the first section, third paragraph that says:

> Using the handshake protocol to establish secured communication is **optional** on the local network (e.g. local mining devices talking to a local mining proxy). However, it is **mandatory** for remote access to the upstream nodes, whether they be pool mining services, job declarating services or template distributors.

Do I misunderstand this?

plebhash: well if it's in the spec, I take it back

protocols/v2/framing-sv2/README.md
```diff
@@ -0,0 +1,60 @@
...
+  for any other kind of usage. It will likely be fully deprecated in the future.
+
+### Examples
+
+This crate provides two examples demonstrating how to serialize and deserialize Sv2 and Noise
+handshake message frames:
+
+1. **[Sv2 Frame](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/framing-sv2/examples/sv2_frame.rs)**:
+   Encode and decode standard Sv2 frames, detailing the message serialization and transmission
+   process for unencrypted communications.
```
plebhash, line 55: there is a lot of conceptual overlap between encode/decode vs serialize/deserialize

to be fully honest, I often get confused and also conflate the two

but for the sake of clarity, we should reserve the usage of encode/decode terminology for the `codec_sv2` crate.

plebhash: also, we should avoid bringing up "unencrypted" here... sure, the example is unencrypted, but from a conceptual standpoint, a `Sv2Frame` can be serialized into bytes just like on this example, and later be transmitted over an encrypted channel

so drawing attention to "unencrypted" here is misleading

protocols/v2/framing-sv2/src/header.rs
```diff
@@ -69,25 +92,37 @@ impl Header {
...
+    /// Calculates the total length of a chunked message, accounting for MAC overhead.
+    ///
+    /// Determines the total length of the message frame, including the overhead introduced by
+    /// MACs. If the message is split into multiple chunks (due to its size exceeding the maximum
+    /// frame chunk size), each chunk requires a MAC for integrity verification.
+    ///
+    /// This method is particularly relevant when the message is being encrypted using the Noise
+    /// protocol, where the payload is divided into encrypted chunks, and each chunk is appended
+    /// with a MAC. However, it can also be applied to non-encrypted chunked messages to calculate
+    /// their total length.
```
plebhash, line 122: > However, it can also be applied to non-encrypted chunked messages to calculate their total length.

how? that sounds rather counter intuitive, given that the method is called `encrypted_len`

plebhash: I mean, I guess I can imagine that this is possible, since the `Header` object doesn't really know whether the frame has been encrypted or not (since that happens at a higher layer of abstraction)... so in theory yes, I guess the method can be called even if there's no encryption

but if there's no encryption and this method is called, it feels like the API is just being misused?

or to put my confusion in different words:

what is the use-case where it's useful to know the MAC overhead, while there's no MAC because there's no encryption?

plebhash: it seems this was pointed out on https://github.com/stratum-mining/stratum/pull/1179#issuecomment-2412645998

> Why don't we use this constant in the `Sv2Frame` decoder `WithoutNoise` struct? 

> To be clear, this technically does not have to do with Noise, but rather message chunking.

I think those assumptions are not true. This is related to Noise. See this constant:

https://github.com/stratum-mining/stratum/blob/384df6cd40394508d9345c956005f31a32dc200d/protocols/v2/const-sv2/src/lib.rs#L56-L57

And on [Noise Protocol Framework spec](https://noiseprotocol.org/noise.html#message-format), we have:
> All Noise messages are less than or equal to 65535 bytes in length.

So encryption imposes that large messages will have to be broken down into chunks.

plebhash: I also noticed that the previous documentation I wrote for this method was actually misleading.

> /// Calculate the length of the encrypted `Header`

that is not the case, since this will give the length of the encrypted payload, **not including the header**.

so thanks for taking the initiative to clarify this

protocols/v2/framing-sv2/examples/handshake_frame.rs
plebhash, line 1: https://github.com/stratum-mining/stratum/pull/1179#issuecomment-2412645998

> I don't think my handshake.rs example does a great job of showcasing the handshake framing logic. Any feedback would be great, and I can always delete it if it is not helpful.

indeed, I don't feel we're really showing anything useful to the reader here

when you really look at it, `HandShakeFrame` is nothing but a simple wrapper around `payload: Slice`, and its methods don't really do anything fancy:
- `get_payload_when_handshaking` has a big name, while it's nothing but a getter for the inner `payload`
- `encoded_length` which just returns `len()` for the inner `payload`
-`from_bytes` and `from_bytes_unchecked`, which are actually redundant

so I guess there's nothing that really justifies a dedicated example around `HandShakeFrame`

plebhash: I remember we used to have a `Frame` trait on this crate, which has already been removed via https://github.com/stratum-mining/stratum/pull/976 because while it aimed to unify `Sv2Frame` and `HandShakeFrame` under the same interface, the trait was never really used as a trait bound anywhere

By trying to unify `Sv2Frame` and `HandShakeFrame` under the same interface, we ended up with the worst of both worlds

`Sv2Frame` is on its way to be improved via  via https://github.com/stratum-mining/stratum/pull/1033)

plebhash: but now this analysis is actually making me believe that there's really no point in having `HandShakeFrame` exist inside this crate, given that it's just a simple wrapper around `Slice` with some syntax sugar methods

if we ever move forward with merging `framing_sv2` (potentially with `binary_sv2`, as suggested via https://github.com/stratum-mining/stratum/issues/1129#issuecomment-2354934378), I would only take `Header` + `Sv2Frame` to `binary_sv2`, but I would not include `HandShakeFrame` in this migration

`HandShakeFrame` is only consumed in `codec_sv2`.

Wherever `HandShakeFrame` is being consumed in `codec_sv2`, we can simply replace that with a reference to `Slice` and make adaptations where the sugar methods were being called.

plebhash:
we should bump `framing_sv2` MINOR version on this PR

see https://github.com/stratum-mining/stratum/issues/1221 for more details

protocols/v2/framing-sv2/src/header.rs
rrybarczyk: Yeah, I was confused by this method as you saw in my [comment](https://github.com/stratum-mining/stratum/pull/1179#issuecomment-2412645998). I figured my comments were wrong, but it would at least get a conversation started.

My confusion was indeed coming from the name of `SV2_FRAME_CHUNK_SIZE`, which is why I incorrectly stated in that comment "this technically does not have to do with Noise, but rather message chunking". 

I will remove my comments here and change to:

```rs
/// Calculate the length of the encrypted payload, not including the header.
```

I should have a better understanding of how the AEAD MAC works before expanding further on that comment to avoid putting incorrect data in. Perhaps we just leave that single sentence. If I do add anything else, I will tag you to bring your attention to it. And if you have any suggestions, just drop them here.

I added a suggestion to change the name to `NOISE_CHUNK_SIZE` [here](https://github.com/stratum-mining/stratum/issues/1135#issuecomment-2419824853) in the `const-sv2` refactor issue #1135.

protocols/v2/framing-sv2/README.md
rrybarczyk: Good catch, I missed updating the explanation here after I reworked the examples to be Sv2 vs handshake frame focused.

protocols/v2/framing-sv2/examples/handshake_frame.rs
rrybarczyk: Thanks for that context, very helpful. Let me think on it a bit more.

plebhash (approved):
some general comments:
- love the new simplified example, feels tight and precise!
- love the new `README.md`

a few extra things to get this ready for merging:
- in contradiction to what I said [here](https://github.com/stratum-mining/stratum/pull/1179#issuecomment-2419430178), we already used #1300 to bump this crate (as a consequence of PR `no_std` PR #1230)... we have a tracker issue https://github.com/stratum-mining/stratum/issues/1285 with all crate bumping for v1.2.0... so we should consider dropping the commit bb27381b4fda8a646446c477b6b6f9d4c1b47c67 that bumps crate version here
- we found a some innacurate wording from the spec (reported via https://github.com/stratum-mining/sv2-spec/issues/117, potentially fixed via https://github.com/stratum-mining/sv2-spec/pull/118), which ended up on the docs here, so I we should consider cherry-picking https://github.com/plebhash/stratum/commit/0d99fe0c221e0b3de555ffae9a94ef88cadbdc89 here
- commit 288feb52627ea45c9422c14f52becea23cc89cfd broke `fmt` CI so it should be rebased to unblock this PR

rrybarczyk:
> some general comments:
> 
> * love the new simplified example, feels tight and precise!
> * love the new `README.md`
> 
> a few extra things to get this ready for merging:
> 
> * in contradiction to what I said [here](https://github.com/stratum-mining/stratum/pull/1179#issuecomment-2419430178), we already used [use `cargo semver-checks` v0.37 and bump crate versions as a follow-up to #1276 #1300](https://github.com/stratum-mining/stratum/pull/1300) to bump this crate (as a consequence of PR `no_std` PR [enforce `no_std` for all? protocol/v2 crates #1230](https://github.com/stratum-mining/stratum/pull/1230))... we have a tracker issue [version bumping as a consequence of Rust Docs on milestone 1.2.0 #1285](https://github.com/stratum-mining/stratum/issues/1285) with all crate bumping for v1.2.0... so we should consider dropping the commit [bb27381](https://github.com/stratum-mining/stratum/commit/bb27381b4fda8a646446c477b6b6f9d4c1b47c67) that bumps crate version here
> * we found a some innacurate wording from the spec (reported via [`Framing` section is not 100% accurate sv2-spec#117](https://github.com/stratum-mining/sv2-spec/issues/117), potentially fixed via [clarify framing section sv2-spec#118](https://github.com/stratum-mining/sv2-spec/pull/118)), which ended up on the docs here, so I we should consider cherry-picking [plebhash@0d99fe0](https://github.com/plebhash/stratum/commit/0d99fe0c221e0b3de555ffae9a94ef88cadbdc89) here to avoid propagating the confusion from the spec
> * commit [288feb5](https://github.com/stratum-mining/stratum/commit/288feb52627ea45c9422c14f52becea23cc89cfd) broke `fmt` CI so it should be rebased to unblock this PR

@plebhash fixed the above! thank you for the clear feedback, made it really easy to fix.

PR #1195, opened 2024-10-08
add `--nominal-hashrate-multiplier` to `mining-device`

### LINKED ISSUE #1188: `mining-device` cpu miner does not hit the target
When running the `mining-device` locally I was not able to hit the `target` set here https://github.com/stratum-mining/stratum/blob/main/roles/test-utils/mining-device/src/lib/mod.rs#L201

The test was not run for too long, probably for a couple of minutes. While in practice that might not be a problem, in test environment we want shares to be submitted in a matter of seconds in order to not run the test for too long.

plebhash:
close #1188

## first commit
- introduces a mechanism for advertising a custom nominal hashrate (by multiplying measured hashrate with a constant)
- fixes `measure_hashrate` so it also takes `handicap` into consideration (otherwise `measure_hashrate` will return something that doesn't reflect what will happen during the actual mining when `handicap` is used)

## second commit
- makes sure we use all available parallelism (there were some weird subtractions that didn't seem to make sense cc @Fi3)

### DIFF
--- a/roles/test-utils/mining-device/README.md
+++ b/roles/test-utils/mining-device/README.md
@@ -31,3 +31,30 @@ Usage example:
 ```
 cargo run --release -- --address-pool 127.0.0.1:20000 --id-device device_id::SOLO::bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh
 ```
+
+## handicap
+
+CPU mining could damage the system due to excessive heat.
+
+The `--handicap` parameter should be used as a safety mechanism to slow down the hashrate in order to preserve hardware.
+
+## nominal hashrate multiplier
+
+Let's imagine that:
+- the upstream wants to receive shares every ~100s (on average)
+- the CPU miner nominal hashrate is 1k H/s
+
+Maybe we want to do a test where we don't want to wait ~100s before a share is submitted by the CPU miner.
+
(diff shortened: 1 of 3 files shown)


GitGab19 (commented):
tACK.
Could you just add the new `nominal-hashrate-multiplier` param in the `README.md`?

roles/test-utils/mining-device/README.md
```diff
@@ -3,16 +3,28 @@
...
+          Pool pub key, when left empty the pool certificate is not checked
+  -i, --id-device <ID_DEVICE>
+          Sometimes used by the pool to identify the device
+  -a, --address-pool <ADDRESS_POOL>
+          Address of the pool in this format ip:port or domain:port
+      --handicap <HANDICAP>
+          This value is used to slow down the cpu miner, it represents the number of micro-seconds that are awaited between hashes [default: 0]
+      --id-user <ID_USER>
+          User id, used when a new channel is opened, it can be used by the pool to identify the miner
+      --nominal-hashrate-multiplier <NOMINAL_HASHRATE_MULTIPLIER>
```
rrybarczyk, line 19: Adding a explanation of why this was added, when to use which %, and give an example of a 1,000 H/s miner use case.

plebhash: done

rrybarczyk (approved):

GitGab19 (approved):

PR #1225, opened 2024-10-17
Dont provie TP pubkey in test env

### LINKED ISSUE #1224: Ignore TP pubkey when initializing `pool_sv2` in integration tests
Currently we try to pass pubkey here https://github.com/stratum-mining/stratum/blob/main/roles/tests-integration/tests/common/mod.rs#L251 but we should pass None as the pubkey of the TP is unknown to us and we can ignore this in a test scenario

jbesraa:
resolves #1224


GitGab19 (approved):

Shourya742 (approved):
ACK.
@jbesraa can you resolve the fmt error...

PR #1228, opened 2024-10-18
Allow altering messages exchanged by roles in test env

### LINKED ISSUE #1227: Add functionality to response to a role with alterd message in integration tests
In integration tests, we want to be able to tell the `Sniffer`(or potentially a new module) if role A is sending message X to role B, then change message X to Y so role B receives message Y.

jbesraa:
resolves #1227

### DIFF
--- a/roles/tests-integration/tests/common/sniffer.rs
+++ b/roles/tests-integration/tests/common/sniffer.rs
@@ -1,6 +1,6 @@
 use async_channel::{Receiver, Sender};
 use codec_sv2::{
-    framing_sv2::framing::Frame, HandshakeRole, Initiator, Responder, StandardEitherFrame,
+    framing_sv2::framing::Frame, HandshakeRole, Initiator, Responder, StandardEitherFrame, Sv2Frame,
 };
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
 use network_helpers_sv2::noise_connection_tokio::Connection;
@@ -13,8 +13,8 @@ use roles_logic_sv2::{
             IdentifyTransactionsSuccess, ProvideMissingTransactions,
             ProvideMissingTransactionsSuccess, SubmitSolution,
         },
-        TemplateDistribution,
-        TemplateDistribution::CoinbaseOutputDataSize,
+        PoolMessages,
+        TemplateDistribution::{self, CoinbaseOutputDataSize},
     },
     utils::Mutex,
 };
@@ -22,6 +22,7 @@ use std::{collections::VecDeque, convert::TryInto, net::SocketAddr, sync::Arc};
 use tokio::{
     net::{TcpListener, TcpStream},
     select,
+    time::sleep,
 };
 type MessageFrame = StandardEitherFrame<AnyMessage<'static>>;
 type MsgType = u8;
@@ -30,6 +31,7 @@ type MsgType = u8;
 enum SnifferError {
     DownstreamClosed,
     UpstreamClosed,
+    MessageInterrupted,
 }
 
 /// Allows to intercept messages sent between two roles.
@@ -50,17 +52,56 @@ pub struct Sniffer {
     upstream_address: SocketAddr,
     downstream_messages: MessagesAggregator,
     upstream_messages: MessagesAggregator,
+    interrupt_messages: Vec<InterruptMessage>,
+}
+
+#[derive(Debug, Clone)]
+pub struct InterruptMessage {
+    direction: MessageDirection,
+    expected_message_type: MsgType,
+    response_message: PoolMessages<'static>,
+    response_message_type: MsgType,
+    break_on: bool,
+}
+
+impl InterruptMessage {
+    pub fn new(
+        direction: MessageDirection,
+        expected_message_type: MsgType,
+        response_message: PoolMessages<'static>,
+        response_message_type: MsgType,
+        break_on: bool,
+    ) -> Self {
+        Self {
+            direction,
+            expected_message_type,
+            response_message,
+            response_message_type,
+            break_on,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MessageDirection {
+    ToDownstream,
+    ToUpstream,
 }
 
 impl Sniffer {
     /// Creates a new sniffer that listens on the given listening address and connects to the given
     /// upstream address.
-    pub async fn new(listening_address: SocketAddr, upstream_address: SocketAddr) -> Self {
+    pub async fn new(
+        listening_address: SocketAddr,
+        upstream_address: SocketAddr,
+        interrupt_messages: Option<Vec<InterruptMessage>>,
+    ) -> Self {
         Self {
             listening_address,
             upstream_address,
             downstream_messages: MessagesAggregator::new(),
             upstream_messages: MessagesAggregator::new(),
+            interrupt_messages: interrupt_messages.unwrap_or_default(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


roles/tests-integration/tests/common/sniffer.rs
```diff
@@ -175,13 +225,39 @@ impl Sniffer {
...
         send: Sender<MessageFrame>,
         upstream_messages: MessagesAggregator,
+        interrupt_messages: Vec<InterruptMessage>,
     ) -> Result<(), SnifferError> {
         while let Ok(mut frame) = recv.recv().await {
             let (msg_type, msg) = Self::message_from_frame(&mut frame);
-            upstream_messages.add_message(msg_type, msg);
+            for interrupt_message in interrupt_messages.iter() {
+                if interrupt_message.direction == MessageDirection::ToDownstream
+                    && interrupt_message.expected_message_type == msg_type
```
plebhash, line 234: I wonder if only matching against `direction` + `msg_type` isn't too generic?

shouldn't we consider matching against `msg` itself?

that way, there's no room for "accidentally" interrupting on a message simply because it happened before the expected trigger while having the same message type

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -44,3 +48,32 @@ async fn success_pool_template_provider_connection() {
...
+        MessageDirection::ToDownstream,
+        MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+        message,
+        MESSAGE_TYPE_SETUP_CONNECTION_ERROR,
+        true,
+    );
+    let sniffer = common::start_sniffer(sniffer_addr, tp_addr, Some(vec![interrupt_msgs])).await;
+    let _ = common::start_pool(Some(pool_addr), Some(sniffer_addr)).await;
+    assert_common_message!(&sniffer.next_downstream_message(), SetupConnection);
+    assert_common_message!(&sniffer.next_upstream_message(), SetupConnectionError);
```
plebhash, line 78: hmm I guess this test gives us confidence that we triggered the "interrupt" system on sniffer, but how do we know that pool actually received the expected modified message?

should we assert the pool state afterwards?

or maybe use an extra sniffer?

roles/tests-integration/tests/common/sniffer.rs
```diff
@@ -50,17 +52,56 @@ pub struct Sniffer {
...
+#[derive(Debug, Clone)]
+pub struct InterruptMessage {
+    direction: MessageDirection,
+    expected_message_type: MsgType,
+    response_message: PoolMessages<'static>,
+    response_message_type: MsgType,
+    break_on: bool,
+}
+
+impl InterruptMessage {
```
plebhash, line 67: this name feels a bit counterintuitive from the user perspective

"interrupt" implies that things will stop happening and this is the end of the test, which is not necessarily true

two alternative names that come to mind now are:
- `InterceptMessage`
- `ModifyMessage`

plebhash:
after discussing with @jbesraa on the call today, we will potentially pivot this approach towards adding a new primitive (aside to the `Sniffer`) that is simplified and just "mocks" a dummy role, essentially expecting predefined messages and also responding with predefined messages

plebhash:
as discussed in the call, here's a dummy test (pseudocode) that could be used to drive the development of this new primitive, at least from a conceptual perspective

(let's assume this new primitive will be called `Mock`, final name could be changed later)

`Mock` is created in a way that it has some predefined list of pair messages, where each pair has:
- some expected message
- response to the expected message

the scenario here is:

- `Mock` is created to mimic a TP
- `Sniffer` is placed between `Pool` and `Mock`
- `Pool` sends `SetupConnection` upstream
  - `Sniffer` logs `SetupConnection` into its `messages_from_downstream` queue
  - `Mock` is already expecting a `SetupConnection`
- `Mock` replies with `SetupConnection.Error`
  - `Sniffer` logs `SetupConnection.Error` into `messages_from_upstream`
  - `Pool` receives `SetupConnection.Error`

```rust
#[tokio::test]
async fn test_tp_mock_setup_connection_error() {
    let mock_addr = common::get_available_address();
    let sniffer_addr = common::get_available_address();
    let pool_addr = common::get_available_address();

    let mock_expected_setup_connection = ...
    let mock_response_setup_connection_error = ...

    let mock_message_pair = (mock_expected_setup_connection, mock_response_setup_connection_error);

    // this is a very simple test where we only have one pair, but the generic case could have multiple pairs
    let mock_message_pair_vec = vec![mock_message_pair];

    let mut mock = common::start_mock(mock_addr, mock_message_pair_vec).await;

    let sniffer = common::start_sniffer(sniffer_addr, tp_addr, Some(vec![interrupt_msgs])).await;
    let _ = common::start_pool(Some(pool_addr), Some(sniffer_addr)).await;

    assert_common_message!(&sniffer.next_downstream_message(), SetupConnection);

    // this specific assertion are essentially proving that mock is behaving as expected
    assert_common_message!(&sniffer.next_upstream_message(), SetupConnectionError);
}
```

jbesraa:
@plebhash This is ready for review

roles/tests-integration/tests/common/sniffer.rs
```diff
@@ -50,17 +52,56 @@ pub struct Sniffer {
     upstream_address: SocketAddr,
     messages_from_downstream: MessagesAggregator,
     messages_from_upstream: MessagesAggregator,
+    intercept_messages: Vec<InterceptMessage>,
```
plebhash, line 55: we have some comments above `Sniffer` declaration explaining its internal details

we should make sure those comments also include this new funcitonality

roles/tests-integration/tests/common/sniffer.rs
```diff
@@ -175,13 +246,39 @@ impl Sniffer {
...
+                    upstream_messages
+                        .add_message(msg_type, interrupt_message.response_message.clone());
+                    let _ = send.send(frame).await;
+                    if interrupt_message.break_on {
+                        return Err(SnifferError::MessageInterrupted);
+                    } else {
+                        continue;
+                    }
+                }
+            }
```
plebhash, line 277:
```suggestion
            for intercept_message in intercept_messages.iter() {
                if intercept_message.direction == MessageDirection::ToDownstream
                    && intercept_message.expected_message_type == msg_type
                {
                    let extension_type = 0;
                    let channel_msg = false;
                    let frame = StandardEitherFrame::<AnyMessage<'_>>::Sv2(
                        Sv2Frame::from_message(
                            intercept_message.response_message.clone(),
                            intercept_message.response_message_type,
                            extension_type,
                            channel_msg,
                        )
                        .expect("Failed to create the frame"),
                    );
                    upstream_messages
                        .add_message(msg_type, intercept_message.response_message.clone());
                    let _ = send.send(frame).await;
                    if interrupt_message.break_on {
                        return Err(SnifferError::MessageInterrupted);
                    } else {
                        continue;
                    }
                }
            }
```

roles/tests-integration/tests/common/sniffer.rs
```diff
@@ -50,17 +52,56 @@ pub struct Sniffer {
...
+    intercept_messages: Vec<InterceptMessage>,
+}
+
+#[derive(Debug, Clone)]
+pub struct InterceptMessage {
+    direction: MessageDirection,
+    expected_message_type: MsgType,
+    response_message: PoolMessages<'static>,
+    response_message_type: MsgType,
+    break_on: bool,
```
plebhash, line 64: what is the purpose of `break_on`?

under which circumstances can it be useful?

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -15,7 +15,7 @@ async fn success_pool_template_provider_connection() {
     let tp_addr = common::get_available_address();
     let pool_addr = common::get_available_address();
     let _tp = common::start_template_provider(tp_addr.port()).await;
-    let sniffer = common::start_sniffer(sniffer_addr, tp_addr).await;
+    let sniffer = common::start_sniffer(sniffer_addr, tp_addr, None).await;
```
plebhash, line 18: I think we should add a new `intercept_pool_template_provider_connection` to make sure the new `Sniffer` feature behaves as expected

potentially with 2 `Sniffer`s:
- one intercepts `SetupConnection.Success` and replaces with `SetupConnection.Error`
- one asserts `SetupConnection.Error`

roles/tests-integration/tests/common/sniffer.rs
jbesraa: when `break_on` is set to `true`, the sniffer will exist after sending the mutated message, which will result in closing the connection between downstream and upstream. Can be useful if we dont want the roles exchange messages after we mutate a message

plebhash: I assume you mean:

> the sniffer will EXIT after sending the mutated message

right?

plebhash: I still don't see how this could be useful. I am yet to encounter a scenario where it would make sense to break the connection and still continue the test.

Could you give some concrete example?

plebhash (approved):

Issue #1130, opened 2024-08-21
Refactor: `noise-sv2`


rrybarczyk:
# Background
This task is an outcome of the `protocols` Rust docs issues tracked in #845.

While documenting `protocols::v2::noise-sv2` in #1013, areas of potential code debt were identified. This issue servers as a place to list out these items to then be addressed in an organized manner. The initial Rust documentation effort was an immediate action, while a refactoring (which implies breaking API changes) is not so urgent priority, so for now we should leave this in the backlog for an appropriate moment in the future.

# Identified Potential Code Debt

## TODO

Shourya742:
I noticed a few `unwrap` calls in the `initiator`, `responder`, and `signature` modules. It's generally not advisable to use `unwrap` in a library crate since it can cause the program to panic if an error occurs. Below are the locations where `unwrap` is being used:

- **`initiator.rs`**
  ```rust
  let plaintext: [u8; SIGNATURE_NOISE_MESSAGE_SIZE] = to_decrypt.try_into().unwrap();
  let signature_message: SignatureNoiseMessage = plaintext.into();
  let rs_pub_key = PublicKey::from_ellswift(elligatorswift_theirs_static)
      .x_only_public_key()
      .0
      .serialize();
  let rs_pk_xonly = XOnlyPublicKey::from_slice(&rs_pub_key).unwrap();
  ```

- **`responder.rs`**
  ```rust
  // 7. appends `EncryptAndHash(SIGNATURE_NOISE_MESSAGE)` to the buffer
  let valid_from = std::time::SystemTime::now()
      .duration_since(std::time::UNIX_EPOCH)
      .unwrap()
      .as_secs();
  ```

- **`signature_message.rs`**
  ```rust
  let now = SystemTime::now()
      .duration_since(SystemTime::UNIX_EPOCH)
      .unwrap()
      .as_secs() as u32;
  ```

We should propagate these errors instead of using `unwrap` and allow the implementing binary to handle them appropriately. It's important to avoid using `unwrap` in a library crate to prevent unexpected panics.

Fi3:
they are inffalible, if you prefer you can use expect

Shourya742:
Based on the discussion about using the multi-cipher (referenced here: https://github.com/stratum-mining/stratum/discussions/916#discussioncomment-10416620), we need to remove AES_256_GCM from our codebase and update the specs accordingly.

Shourya742:
[...]

Issue #1200, opened 2024-10-10
custom logger


plebhash:
## background

`protocols` crates should use as little external dependencies as possible.

## problem

currently, the following crates use `tracing` as an external dependency:

- [`sv1_api`](https://github.com/stratum-mining/stratum/blob/8367a26a8ed552171563b18c941972196cb10c3d/protocols/v1/Cargo.toml#L24)
- [`binary_sv2`](https://github.com/stratum-mining/stratum/blob/8367a26a8ed552171563b18c941972196cb10c3d/protocols/v2/binary-sv2/binary-sv2/Cargo.toml#L21)
- [`codec_sv2`](https://github.com/stratum-mining/stratum/blob/8367a26a8ed552171563b18c941972196cb10c3d/protocols/v2/codec-sv2/Cargo.toml#L21)
- [`roles_logic_sv2`](https://github.com/stratum-mining/stratum/blob/8367a26a8ed552171563b18c941972196cb10c3d/protocols/v2/roles-logic-sv2/Cargo.toml#L26)

## solution

write a custom logger to replace `tracing`

reference from `rust-lightning`: https://github.com/lightningdevkit/rust-lightning/blob/f94bf9861289ce6a75fd7d5856d736a2f966a619/lightning/src/util/logger.rs#L165

Shourya742:
https://carbonated-machine-4e3.notion.site/Custom-Logger-878dbdf54d154ad68890c98814df707c

plebhash:
I created this issue to take note of some discussion point we had in a call last year

even though we try to use as little external dependencies as possible (to avoid introducing vulnerabilities etc), `tracing` is a widely used crate, and as @Fi3 pointed out it's extensively audited and therefore I don't see this dependency as a critical point of vulnerability

therefore I don't think this is high priority

plebhash:
@Georges760 please write an explanation of how embedded systems manage logging

Georges760:
> [@Georges760](https://github.com/Georges760) please write an explanation of how embedded systems manage logging

In embedded systems, logging is usually done using [`defmt`](https://crates.io/crates/defmt) crate (more info [here](https://defmt.ferrous-systems.com/)), it permit "deferred formatting", so the microcontroler does not have to manipulate utf8 strings and send them to the host debugger !

Most of the libraries that want to support embedded system have a feature `defmt` or often `defmt-03` that 
[...]

PR #1230, opened 2024-10-19
enforce `no_std` for all? protocol/v2 crates

Georges760:
Most of protocol/v2 crates are almost `no_std`, as listed [here](https://github.com/stratum-mining/stratum/issues/1130#issuecomment-2438642786).

Having a lib crate that can be `no_std` or `std` by the selection of a feature :
```rust
#![cfg_attr(not(feature = "std"), no_std)]
```
or the current implementation :
```rust
#![cfg_attr(feature = "no_std", no_std)]
```
is only usefull if :
- the crate could use some `std` related dependency (for instance `std::vec::Vec`)
- but we want the crate's user to be able to not have a dependancy on `std` so we disable the functionalities using `std`

In some case we can replace the `std` dep by its `alloc` equivalent if exists (`alloc::vec::Vec`) and we can keep the functionality in `no_std` environement.

Finally, a `std` binary can use some `no_std` dependencies, but an `no_std` binary cannot use any `std` deps. So making the low level crates all `no_std` will make them copamtible for every kind of binary (app).

We should use the `std` feature only if some functionality requires `std` and not have any `no_std` equivalent.

As far as I tried all protocol/v2/ crates (libs) can be 100% `no_std` a,d does not need the `std` (or `no_std`) feature to optionally add `std` dep.

This PR is targeting to get rid of `std` dep without loosing any functionality or efficiency.

Note: I will make 1 commit per crate to explain what and how I am doing it.
Note2: being `no_alloc` is another level of complexity for deep embedded system, which is not at all this PR concern.

### DIFF
--- a/protocols/v2/binary-sv2/binary-sv2/src/lib.rs
+++ b/protocols/v2/binary-sv2/binary-sv2/src/lib.rs
@@ -39,11 +39,10 @@ pub fn u256_from_int<V: Into<u64>>(value: V) -> U256<'static> {
 #[cfg(test)]
 mod test {
     use super::*;
+    use alloc::vec::Vec;
 
     mod test_struct {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
@@ -74,8 +73,6 @@ mod test {
 
     mod test_f32 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
@@ -106,8 +103,6 @@ mod test {
 
     mod test_b0255 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
@@ -153,8 +148,6 @@ mod test {
 
     mod test_u256 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
@@ -183,8 +176,6 @@ mod test {
 
     mod test_signature {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -213,8 +204,6 @@ mod test {
 
     mod test_b016m {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -263,8 +252,6 @@ mod test {
 
     mod test_b064k {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -296,8 +283,6 @@ mod test {
 
     mod test_seq0255_u256 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -338,8 +323,6 @@ mod test {
 
     mod test_0255_bool {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
@@ -366,8 +349,6 @@ mod test {
 
     mod test_seq0255_u16 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
@@ -394,8 +375,6 @@ mod test {
 
     mod test_seq_0255_u24 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -428,8 +407,6 @@ mod test {
 
     mod test_seqo255_u32 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
@@ -456,8 +433,6 @@ mod test {
 
     mod test_seq0255_signature {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -493,8 +468,6 @@ mod test {
 
     mod test_seq_064_u256 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -535,8 +508,6 @@ mod test {
 
     mod test_064_bool {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
@@ -571,8 +542,6 @@ mod test {
 
     mod test_se1o64k_u16 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
@@ -599,8 +568,6 @@ mod test {
 
     mod test_seq064k_u24 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -633,8 +600,6 @@ mod test {
 
     mod test_seq064k_u32 {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
@@ -660,8 +625,6 @@ mod test {
     }
     mod test_seq064k_signature {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -696,8 +659,6 @@ mod test {
     }
     mod test_seq064k_b016m {
         use super::*;
-        #[cfg(not(feature = "with_serde"))]
-        use alloc::vec::Vec;
         use core::convert::TryInto;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
@@ -732,7 +693,6 @@ mod test {
     }
     mod test_seq_0255_in_struct {
         use super::*;
-        use alloc::vec::Vec;
 
         #[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
         struct Test<'decoder> {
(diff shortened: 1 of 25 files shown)


protocols/DEPENDANCIES.md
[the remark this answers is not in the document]
Georges760: sure !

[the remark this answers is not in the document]
plebhash: thanks!

now that we have #1233 we should drop c658909741686ac6a9d9269057f534ae003f3ebb from this PR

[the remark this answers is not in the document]
Georges760: removed

jbesraa (commented):
While I think the general idea is very welcome, I think addressing https://github.com/stratum-mining/stratum/issues/1200 first would make more sense as it can potentially change how we approach this. I think in #1200 we should aim to provide a logger for std and no std env and then `impl Debug` manually, which would allow us to remove the debug feature IIUC whats going here

Georges760:
Current `binary_codec_sv2` :
- don't compile without `no_std` feature (this doesn't mean it is `no_std` as there is no enforcement => `#![cfg_attr(feature = "no_std", no_std)]`)
- has `no_std` feature by default (`default = ["no_std"]`)
- nowhere (not even in the CI), its default features are disable (`--no-default-features`)

So it's pseudo `std` support os broken and unused. We can :
- remove/clean it, and keep a real `no_std` crate only
- debug it, and have a `no_std` crate by default, with a `std` feature with the debugged `std` functionalities

Maybe this can be discussed in #1132 ?

Georges760:
> While I think the general idea is very welcome, I think addressing #1200 first would make more sense as it can potentially change how we approach this. I think in #1200 we should aim to provide a logger for std and no std env and then `impl Debug` manually, which would allow us to remove the debug feature IIUC whats going here

`tracing` has a `std` feature enabled by default, this means it actually add `std` dep to the crate that use it without disabling default features : `sv1_api`, `binary_sv2`, `codec_sv2` and `roles_logic_sv2` (the 4 crates listed in #1200). 2 of this 4 crates are targeted by this PR (`binary_sv2` and `codec_sv2`) and in order for them to become `no_std`, I will have to `no-default-features` tracing, and enable `std` in tracing using a local feature too. This can be done before or after #1200 without problem.

If I do it before #1200, when #1200 will remove `tracing` with the custom logger, the `std` feature will be available (and can be removed if not used after all).
If #1200 should be done before, then we can limit this PR to crates not depending on `binary_sv2` and `codec_sv2`, and wait to see what the new logger will propose for `no_std` support.

jbesraa:
> > While I think the general idea is very welcome, I think addressing #1200 first would make more sense as it can potentially change how we approach this. I think in #1200 we should aim to provide a logger for std and no std env and then `impl Debug` manually, which would allow us to remove the debug feature IIUC whats going here
> 
> `tracing` has a `std` feature enabled by default, this means it actually add `std` dep to the crate that use it without disabling default features : `sv1_api`, `binary_sv2`, `codec_sv2` and `roles_logic_sv2` (the 4 crates listed in #1200). 2 of this 4 crates are targeted by this PR (`binary_sv2` and `codec_sv2`) and in order for them to become `no_std`, I will have to `no-default-features` tracing, and enable `std` in tracing using a local feature too. This can be done before or after #1200 without problem.
> 
> If I do it before #1200, when #1200 will remove `tracing` with the custom logger, the `std` feature will be available (and can be removed if not used after all). If #1200 should be done before, then we can limit this PR to crates not depending on `binary_sv2` and `codec_sv2`, and wait to see what the new logger will propose for `no_std` support.

Yea I think #1200 shouldn't necessarily be about removing `tracing` crate, but generalizing how we use the logger 
and _mainly_ enabling a logger in a non-std env. Making everything non-std by default would make it a bit tricky to debug the crates generally.
With that said, If #1200 will require too much time and would block this for too long, I am happy for this to get in anyway.

Georges760:
Actual `binary_sv2` without default feature (`core`) and with `with_serde` feature, test is broken.

This PR is not fixing this.

CI does not see it because does not test it.

protocols/v2/binary-sv2/binary-sv2/src/lib.rs
```diff
@@ -37,6 +42,8 @@ mod test {
 
     mod test_struct {
         use super::*;
+        #[cfg(not(feature = "with_serde"))]
```
plebhash, line 45: why are these `with_serde` flags being added?

protocols/v2/binary-sv2/binary-sv2/src/lib.rs
```diff
@@ -1,5 +1,10 @@
 // TODO unify errors from serde_sv2 and no-serde-sv2
-//
+
+#![no_std]
```
plebhash, line 3: why aren't we following this pattern here?
```
#![cfg_attr(feature = "no_std", no_std)]
```

plebhash:
@Georges760 please note that we're aiming to deprecate `with_serde` feature soon.

I cannot give you a precise timeline but Q1 2025 would be my rough guess.

I wonder if that would make this process simpler?

protocols/v2/binary-sv2/no-serde-sv2/derive_codec/src/lib.rs
```diff
@@ -1,4 +1,13 @@
+#![no_std]
```
plebhash, line 1: why aren't we following this pattern here?
```
#![cfg_attr(feature = "no_std", no_std)]
```

protocols/v2/binary-sv2/serde-sv2/src/lib.rs
```diff
@@ -72,7 +72,7 @@
 //! [rkyv1]: https://docs.rs/rkyv/0.4.3/rkyv
 //! [rkyv2]: https://davidkoloski.me/blog/rkyv-is-faster-than/
 
-#![cfg_attr(feature = "no_std", no_std)]
+#![no_std]
```
plebhash, line 75: why aren't we following this pattern here?
```
#![cfg_attr(feature = "no_std", no_std)]
```

protocols/v2/binary-sv2/binary-sv2/src/lib.rs
Georges760: Enforcing #![no_std] at the top of lib.rs remove the implicitly use of std::vec::Vec everywhere in the crate. So here in one of the test that is run only with feature "with_serde" the Vec must be explicitly used from alloc instead of std. Because the test is depends on with_serde feature, I made the "use alloc::vec::Vec" also depending on with_serde feature to avoid a warning if we run the test without it.

protocols/v2/binary-sv2/binary-sv2/src/lib.rs
Georges760: Because this crate is no_std compliant no matter what. No need to use a feature for it. KISS

protocols/v2/binary-sv2/no-serde-sv2/derive_codec/src/lib.rs
Georges760: Same: KISS.

protocols/v2/binary-sv2/serde-sv2/src/lib.rs
Georges760: KISS.

Georges760:
> @Georges760 please note that we're aiming to deprecate `with_serde` feature soon.
> 
> I cannot give you a precise timeline but Q1 2025 would be my rough guess.
> 
> I wonder if that would make this process simpler?

It make sense, as it is already broken in binary_sv2 and framing_sv2 as stated above. 

So I understand the official/only codec way will be the "core" option ? Good

Georges760:
`noise_sv2` has its own PR #1238.
`codec_sv2` is dependant on both `noise_sv2` and all the crates of this PR.

So I stop here for this PR and open for reviews.
I think it is in good shape to fullfil the OP goal.

protocols/v2/binary-sv2/no-serde-sv2/codec/src/lib.rs
```diff
@@ -19,6 +19,9 @@
 //! Seq0255  <-> SEQ0_255[T]
 //! Seq064K  <-> SEQ0_64K[T]
 //! ```
+
+#![cfg_attr(feature = "no_std", no_std)]
```
Shourya742, line 23: We can now use `#![no_std]`, and I'm confident that we are not using any `std`-compliant methods in `binary_codec_sv2`  in any of the top-level crates that depend on it. Since almost all of our crates are `no_std`-compliant, makes sense to have no-serde-sv2 to no-std only. cc: @rrybarczyk

Shourya742 (changes requested):

protocols/v2/binary-sv2/no-serde-sv2/codec/src/lib.rs
Georges760: I asked about this in this [comment](https://github.com/stratum-mining/stratum/pull/1230#issuecomment-2441362158).

Current `binary_codec_sv2` has a broken `std` support, do we want to remove it or fix it ?

Look like you propose to remove it. It is also my preference.

protocols/v2/binary-sv2/binary-sv2/src/lib.rs
Georges760: done

plebhash:
> As far as I tried all protocol/v2/ crates (libs) can be 100% `no_std` a,d does not need the `std` (or `no_std`) feature to optionally add `std` dep.

if I understand correctly, we should expect this PR to leave all `protocols` crates (with exception of `noise_sv2` + `codec_sv2`, which will be covered in other PRs) without **ANY** `std`/`no_std` feature flag, since they will be `no_std` by default.

but I'm confused:

why are we still leaving some crates with the `no_std` feature flag? I inspected this branch and found the `no_std` feature flag on the following crates:
- `v2/codec-sv2/Cargo.toml` (expected, since this PR doesn't aim to touch this crate IIUC)
- `v2/subprotocols/job-declaration/Cargo.toml` 
- `v2/subprotocols/mining/Cargo.toml`
- `v2/binary-sv2/no-serde-sv2/codec/Cargo.toml`
- `v2/subprotocols/template-distribution/Cargo.toml`
- `v2/subprotocols/common-messages/Cargo.toml`

shouldn't `no_std` feature flag be completely removed from these crates?
shouldn't they simply have a `#![no_std]` on the top of the module, instead of `#![cfg_attr(feature = "no_std", no_std)]`?

plebhash:
also, we should bump the `PATCH` version of the crates that are being changed

Georges760:
> > As far as I tried all protocol/v2/ crates (libs) can be 100% `no_std` a,d does not need the `std` (or `no_std`) feature to optionally add `std` dep.
> 
> if I understand correctly, we should expect this PR to leave all `protocols` crates (with exception of `noise_sv2` + `codec_sv2`, which will be covered in other PRs) without **ANY** `std`/`no_std` feature flag, since they will be `no_std` by default.
> 
That is the end goal yes

> but I'm confused:
> 
> why are we still leaving some crates with the `no_std` feature flag? I inspected this branch and found the `no_std` feature flag on the following crates:
> 
> * `v2/codec-sv2/Cargo.toml` (expected, since this PR doesn't aim to touch this crate IIUC)
> * `v2/subprotocols/job-declaration/Cargo.toml`
> * `v2/subprotocols/mining/Cargo.toml`

I didn't consider subprotocols/* crates either in this PR

> * `v2/binary-sv2/no-serde-sv2/codec/Cargo.toml`

details [here](https://github.com/stratum-mining/stratum/pull/1230#issuecomment-2441362158) above
TL;DR: kept `no_std` feature for an existing yet broken implementation (don't want to remove any piece of code, even if doesn't compile, but yes I think we should also clean what is under the `#[cfg(not(feature = "no_std"))]` in this crate, maybe part of #1132 )

> * `v2/subprotocols/template-distribution/Cargo.toml`
> * `v2/subprotocols/common-messages/Cargo.toml`
> 
> shouldn't `no_std` feature flag be completely removed from these crates? shouldn't they KISS and only have a `#![no_std]` on the top of the module, instead of `#![cfg_attr(feature = "no_std", no_std)]`? e.g.: [422aee3](https://github.com/stratum-mining/stratum/commit/422aee3eecfd34877dc0363a92c5997b4cfb1ae7)

YES totally, but this PR being multi crate already, try to limit its action to not refactor any code, just remove `std` dependencies where possible *without* functional impact.

Georges760:
> following [this criteria](https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10973798), we should bump the `PATCH` version of the crates that are being changed on this PR

exact, but usually the version bump is done in a separate PR/commit to merge multiple improvement in a single version increment. Should I do it in this PR anyway ?

protocols/v2/binary-sv2/no-serde-sv2/codec/src/lib.rs
rrybarczyk: I also think we should enforce `no_std` on the `protocols` crates. So let's remove.

Shourya742: This PR looks good to me and is ready to go. We can enforce the `no-std` restriction during our protocol refactoring effort, once we have more concrete reasoning for removing `std`-related imports. However, I’m fine with either approach.

Shourya742 (approved):
ACK

protocols/v2/binary-sv2/no-serde-sv2/codec/src/lib.rs
rrybarczyk: Agree. Let's confirm in the dev meet before we approve.

plebhash:
> > following [this criteria](https://github.com/stratum-mining/stratum/discussions/970#discussioncomment-10973798), we should bump the `PATCH` version of the crates that are being changed on this PR
> 
> exact, but usually the version bump is done in a separate PR/commit to merge multiple improvement in a single version increment. Should I do it in this PR anyway ?

you're right, we can leave it for a new PR

rrybarczyk (approved):

plebhash:
@Georges760 could you please add a comment to the crates `README.md` mentioning that they are `no_std` by default?

some crates are missing a `README.md` (WIP under #845)

for the crates that are missing a `README.md`, feel free to create the file and only add this sentence

later, we're going to finish the `README.md` file with the rest of the info

rrybarczyk (approved):

plebhash (approved):

Issue #1184, opened 2024-10-03
Need Rust docs for `utils::buffer`


rrybarczyk:
While doing the #1160 issue documenting the `codec-sv2` crate and realized there are several core primitives that are dependent on primitives in the `utils::buffer`. This crate should be properly documented.

Unrelated to this issue, we may want to consider in the future bringing `utils::buffer` into the `protocols/` directory.

plebhash:
https://github.com/stratum-mining/stratum/issues/1009 is also closely related to this

PR #1232, opened 2024-10-21
[WIP] Utils buffer sv2 docs

rrybarczyk:
Addresses #1184.


utils/buffer/examples/buffer_pool_exhaustion.rs
```diff
@@ -0,0 +1,40 @@
...
+    let data_bytes = b"abcd"; // 4 bytes
+    println!("DATA: {:?}", data_bytes);
+    write_and_retrieve(&mut buffer_pool, data_bytes);
+    assert!(&buffer_pool.is_back_mode());
+    println!("");
+
+    let data_bytes = b"efg"; // 3 bytes
+    println!("DATA: {:?}", data_bytes);
+    write_and_retrieve(&mut buffer_pool, data_bytes);
+    assert!(&buffer_pool.is_front_mode());
```
rrybarczyk, line 33: Fails here as the pool is still in back mode.

```sh
DATA: [97, 98, 99, 100]
Buffer Pool: BufferPool { pool_back: PoolBack { back_start: 0, len: 0 }, mode: Back, shared_state: SharedState(0), inner_memory: InnerMemory { pool: [97, 98, 99, 100, 0, 0, 0, 0], raw_offset: 0, raw_len: 4, slots: [(0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0)], len: 0 }, system_memory: BufferFromSystemMemory { inner: [], cursor: 0, start: 0 }, start: 0 }
Slice from Buffer Pool: [97, 98, 99, 100]

DATA: [101, 102, 103]
Buffer Pool: BufferPool { pool_back: PoolBack { back_start: 0, len: 0 }, mode: Back, shared_state: SharedState(0), inner_memory: InnerMemory { pool: [97, 98, 99, 100, 101, 102, 103, 0], raw_offset: 0, raw_len: 7, slots: [(0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0)], len: 0 }, system_memory: BufferFromSystemMemory { inner: [], cursor: 0, start: 0 }, start: 0 }
Slice from Buffer Pool: [97, 98, 99, 100, 101, 102, 103]
thread 'main' panicked at buffer/examples/buffer_pool_exhaustion.rs:33:5:
assertion failed: &buffer_pool.is_front_mode()
```

@fi3, I set my buffer pool capacity to 8 bytes. I wrote 4 bytes to the back. Then I wrote 3 bytes, which I would expect to be written to the front of the pool, since we have the back half of the buffer filled with the first write. Instead, the pool is still in back mode during this second write. Then I write four more bytes, such that the 8 byte initialized capacity is exceeded by 4 bytes, and the pool mode switches to alloc mode, as expected.

Why is front mode never activated? 

In the example, [`PoolBuffer.get_writable`](https://github.com/stratum-mining/stratum/blob/b8e77156a8a368bcbfe37fe74b5a2a24b8474b04/utils/buffer/src/buffer_pool/mod.rs#L580) is called, which calls [`PoolBack.get_writable`](https://github.com/stratum-mining/stratum/blob/b8e77156a8a368bcbfe37fe74b5a2a24b8474b04/utils/buffer/src/buffer_pool/pool_back.rs#L146).  At the beginning of the `PoolBack.get_writable`, whether or not the pool has byte capacity is checked with a call to `InnerMemory::has_tail_capacity`(https://github.com/stratum-mining/stratum/blob/b8e77156a8a368bcbfe37fe74b5a2a24b8474b04/utils/buffer/src/buffer_pool/mod.rs#L265). 

If the data trying to be written to the buffer is less than the total buffer capacity the buffer pool was initialized with (8 bytes), then the `get_writable` method returns true, keeping it in back mode.

If the data exceeds the capacity, then we check if the tail is clearable. The tail is not clearable in this example because we have not removed any of the data we have written to the buffer yet. So the `get_writable` returns an [`Err(PoolMode::Alloc)`](https://github.com/stratum-mining/stratum/blob/b8e77156a8a368bcbfe37fe74b5a2a24b8474b04/utils/buffer/src/buffer_pool/pool_back.rs#L162).

Can you help me understand why we don't switch to front mode in this example? Is it the way I have set it up? What case would we proceed past the `tail_is_clearable` check in the `PoolBack::get_writable` and hit the `match` statement? Even in this match statement, the `PoolMode::Front` is ever set.

Shourya742: The Buffer Pool in the above example has a size of 8 bytes, with each slot containing 1 byte of memory capacity. In the example, we are not consuming the slices, so they persist and eventually fill the buffer.

This example clearly demonstrates the transition from back mode to front mode and then to alloc mode.
```
use std::collections::VecDeque;

use buffer_sv2::{Buffer, BufferPool};

fn main() {
    let mut buffer_pool = BufferPool::new(8); // 8 bytes capacity

    let mut slices = VecDeque::new();
    let data_bytes = b"a";
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);  

    let data_bytes = b"e";
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);  
 
    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    assert!(&buffer_pool.is_back_mode());

    slices.pop_front();

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);
    assert!(&buffer_pool.is_front_mode());

    let data_bytes = b"h"; // 1 bytes
    let writable = buffer_pool.get_writable(data_bytes.len());
    writable.copy_from_slice(data_bytes);
    let data_slice = buffer_pool.get_data_owned();  
    slices.push_back(data_slice);

    assert!(&buffer_pool.is_alloc_mode());
    
}

```

rrybarczyk: THANK YOU!

rrybarczyk:
The clippy lint CI is failing in `protocols/v2/subprotocols/template-distribution/src/new_template.rs` with:

```
warning: unused import: `std::convert::TryInto`
  --> /Users/runner/work/stratum/stratum/protocols/v2/subprotocols/template-distribution/src/new_template.rs:11:5
```

Anyone know why the clippy lint is failing in this PR that does not touch that file?

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -1,24 +1,60 @@
...
+// pre-allocated `BufferPool` memory; if not, it will remain in alloc mode.
+
 use crate::buffer_pool::{InnerMemory, PoolFront, PoolMode, POOL_CAPACITY};
 
+/// Manages the "back" section of the [`crate::BufferPool`].
+///
+/// Handles the allocation of memory slices at the back of the buffer pool. It tracks the number
+/// of slices in use and attempts to free unused slices when necessary to maximize available
+/// memory. The back of the buffer pool is used first, if it fills up, the front of the buffer pool
+/// is used.
```
Shourya742, line 32: /// -> //

Although the struct is `pub`, it is private to `mod.rs` and not visible to the crate user.

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -1,24 +1,60 @@
...
 pub struct PoolBack {
+    // Starting index of the back section of the buffer pool.
     back_start: usize,
+
+    // Number of allocated slices in the back section of the buffer pool.
     len: usize,
 }
 
 impl PoolBack {
+    /// Initializes a new [`PoolBack`] with no allocated slices.
```
Shourya742, line 43: /// -> //
Reasoning same as above

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -1,24 +1,60 @@
...
     pub fn new() -> Self {
         Self {
             back_start: 0,
             len: 0,
         }
     }
 
+    /// Returns the number of allocated slices in the back section of the buffer pool.
```
Shourya742, line 51:
```suggestion
    // Returns the number of allocated slices in the back section of the buffer pool.
```

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -1,24 +1,60 @@
...
     #[inline(always)]
     pub fn len(&self) -> usize {
         self.len
     }
 
+    /// Updates the length of the back section based on the state of the inner memory.
```
Shourya742, line 57:
```suggestion
    // Updates the length of the back section based on the state of the inner memory.
```

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -29,17 +65,27 @@ impl PoolBack {
         self.len = len;
     }
 
+    /// Returns the starting index of the back section in the buffer pool.
```
Shourya742, line 68:
```suggestion
    // Returns the starting index of the back section in the buffer pool.
```

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -29,17 +65,27 @@ impl PoolBack {
...
     #[inline(always)]
     pub fn back_start(&self) -> usize {
         self.back_start
     }
 
+    /// Resets the back section of the pool by clearing its start index and length.
```
Shourya742, line 74:
```suggestion
    // Resets the back section of the pool by clearing its start index and length.
```

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -29,17 +65,27 @@ impl PoolBack {
...
     }
 
+    /// From the back section, checks if there are any unset bits and adjusts the length if there
+    /// are.
+    ///
+    /// Assumes the caller has already ensured the safety of the operation (such as slice bounds
+    /// and memory validity). It skips internal safety checks, relying on the caller to manage the
+    /// state, making it faster but potentially unsafe if misused.
+    ///
+    /// Returns `true` if the tail was cleared successfully, otherwise `false`.
```
Shourya742, line 88:
```suggestion
    // From the back section, checks if there are any unset bits and adjusts the length if there
    // are.
    //
    // Assumes the caller has already ensured the safety of the operation (such as slice bounds
    // and memory validity). It skips internal safety checks, relying on the caller to manage the
    // state, making it faster but potentially unsafe if misused.
    //
    // Returns `true` if the tail was cleared successfully, otherwise `false`.
```

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -91,14 +137,24 @@ impl PoolBack {
         }
     }
 
+    /// Checks if the tail of the back section can be cleared.
+    ///
+    /// Should always be called before attempting to clear the tail. Returns `true` if the tail can
+    /// be cleared, otherwise `false`.
```
Shourya742, line 143:
```suggestion
    // Checks if the tail of the back section can be cleared.
    //
    // Should always be called before attempting to clear the tail. Returns `true` if the tail can
    // be cleared, otherwise `false`.
```

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -126,6 +182,16 @@ impl PoolBack {
...
+    /// Clears the tail or head of the back section, transitioning pool modes if necessary.
+    ///
+    /// Called when the state of the [`crate::BufferPool`], along with both `raw_offset` and
+    /// `raw_len`, reaches its maximum limits. It is used to clear any available space remaining in
+    /// the buffer's suffix or prefix. Based on that, the [`crate::BufferPool`]'s state is updated
+    /// by changing its pool mode.
+    ///
+    /// Returns `Ok` if the tail or head was cleared successfully. Otherwise an
+    /// `Err(PoolMode::Front)` if the buffer pool should switch to front mode, or an
+    /// `Err(PoolMode::Alloc)` if the buffer pool should switch to allocation mode.
```
Shourya742, line 194: Should we remove the explanation of return types, as discussed in rust docs standardization discussion,

utils/buffer/src/buffer_pool/pool_back.rs
```diff
@@ -142,6 +208,11 @@ impl PoolBack {
         self.try_clear_head(shared_state, memory)
     }
 
+    /// Returns a writable slice of memory from the back section or transitions to a new pool mode
+    /// if necessary.
+    ///
+    /// Returns `Ok(*mut u8)` if writable memory is available, otherwise an `Err(PoolMode)` if a
+    /// mode change is required.
```
Shourya742, line 215:
```suggestion
    // Returns a writable slice of memory from the back section or transitions to a new pool mode
    // if necessary.
    //
    // Returns `Ok(*mut u8)` if writable memory is available, otherwise an `Err(PoolMode)` if a
    // mode change is required.
```

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -1,3 +1,36 @@
...
+//
+// Provides primitives for reusing memory buffers, clearing old memory chunks, and switching
+// between different modes (back, front, and alloc) to manage memory effectively and reduce
+// expensive heap allocations. It uses atomic operations and shared state tracking to safely manage
+// memory across multiple threads.
+//
+// Supports different allocation modes to optimize memory usage:
+//
+// - **Back Mode**: Allocates from the back of the buffer pool (default).
+// - **Front Mode**: Allocates from the front when the back is full.
```
Shourya742, line 14:
```suggestion
// - **Front Mode**: Allocates from the front when the back is full and front has space.
```

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -110,25 +179,55 @@ impl PoolFront {
...
     pub(crate) raw_offset: usize,
+
+    // Length of the valid data within the contiguous block of memory, starting from `raw_offset`.
     pub(crate) raw_len: usize,
+
+    // Tracks individual chunks of memory being used in the buffer pool by tracking the start and
+    // length of each allocated memory slice.
     slots: [(usize, usize); POOL_CAPACITY],
+
+    // Number of active slices in use, represented by how many slots are currently occupied.
```
Shourya742, line 224: len is a pointer to current slot.

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -255,30 +377,47 @@ impl InnerMemory {
...
 
+    // Transfers ownership of the raw memory slice that has been written into the buffer, returning
+    // it as a `Slice`. This is the main mechanism for passing processed data from the buffer to
+    // the rest of the system.
+    //
+    // After this call, the buffer resets internally, allowing it to write new data into a fresh
+    // portion of memory. This prevents memory duplication and ensures efficient reuse of the
+    // buffer. This method is used when the data in the buffer is ready for processing, sending, or
+    // further manipulation. The returned `Slice` now owns the memory and the buffer no longer has
+    // responsibility for it, allowing the buffer to handle new incoming data.
```
Shourya742, line 420: Ownership is not transferred. After this operation the buffer points to a new slot.

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -682,14 +919,23 @@ impl<T: Buffer> Buffer for BufferPool<T> {
...
     }
 }
 
+// Allows `BufferPool` to be treated as a buffer.
 impl<T: Buffer> BufferPool<T> {
+    /// Determines if the [`BufferPool`] can be safely dropped.
+    ///
+    /// Returns `true` if all memory slices managed by the buffer pool have been released (i.e.,
+    /// the `shared_state` is zero), indicating that no other threads or components are using the
+    /// buffer pool's memory. This check helps prevent dropping the buffer pool while it's still in
```
Shourya742, line 937:
```suggestion
    /// the `shared_state` is zero), indicating that all the slices are dropped
    /// . This check helps prevent dropping the buffer pool while it's still in
```

Shourya742 (commented):
Super cool!!!!

utils/buffer/src/lib.rs
```diff
@@ -1,3 +1,42 @@
...
+//!
+//! For subsequent memory requests, it will continue to check whether the prefix or suffix of the
+//! shared state has been freed. If it has, [`BufferPool`] will switch modes and start consuming
+//! pre-allocated memory; if not, it will remain in alloc mode.
+//!
+//! For isolated operations, [`Slice`] should manages its own memory ([`Vec<u8>`]). In
+//! high-performance applications, [`Slice`] should references externally managed memory from
+//! [`BufferPool`], reducing dynamic memory allocation.
+//!
+//! **Note**: To prevent leaks or deadlocks, ensure that memory slices are properly released after
```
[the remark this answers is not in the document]
rrybarczyk: That is a good question. @fi3, do you know? Should we remove this note?

utils/buffer/src/lib.rs
```diff
@@ -1,3 +1,42 @@
...
+//! the head of shared state, [`BufferPool`] will switch to front mode and start allocating
+//! incoming memory requests from there. However, if the entire shared state is full, it will
+//! switch to alloc mode and begin allocating system memory to fulfill incoming requests at a
+//! performance reduction.
+//!
+//! For subsequent memory requests, it will continue to check whether the prefix or suffix of the
+//! shared state has been freed. If it has, [`BufferPool`] will switch modes and start consuming
+//! pre-allocated memory; if not, it will remain in alloc mode.
+//!
+//! For isolated operations, [`Slice`] should manages its own memory ([`Vec<u8>`]). In
```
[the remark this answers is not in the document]
rrybarczyk: I was trying to express when to use `Vec<u8>` vs `Slice`. My understanding is that`Slice` is for production SRI code that needs optimal performance. But `Vec<u8>` is there for non-critical code, like for examples. I don't know if there are places in the production SRI roles that use `Vec<u8>`. Do you know? @fi3?

utils/buffer/src/lib.rs
```diff
@@ -1,3 +1,59 @@
...
+//! high-performance applications, [`Slice`] should references externally managed memory from
+//! [`BufferPool`], reducing dynamic memory allocation.
```
rrybarczyk, line 47: @Shourya742, with the new "Memory Structure" section added, what do you think of this for the Usage section:

```suggestion
//! ## Usage
//!
//! When an incoming Sv2 message is received, it is buffered for processing. The [`BufferPool`]
//! attempts to allocate memory from its internal slots, starting in back mode. If the back slots
//! are full, it checks for available front slots to switch to front mode. If no internal slots are
//! free, it resorts to alloc mode, allocating memory from the system heap.
//!
//! For operations requiring dedicated buffers, the [`Slice`] type manages its own memory using
//! [`Vec<u8>`]. In high-performance scenarios, [`Slice`] can reference externally managed memory
//! from the [`BufferPool`], reducing dynamic memory allocations and increasing performance.
```

utils/buffer/src/buffer_pool/mod.rs
rrybarczyk: Can you write the changes you want to see here?

Shourya742:
```
/// Provides access to the raw memory slice containing the data written into the buffer, returning
/// it as a `Slice`. This method is the primary mechanism for making processed data available 
/// from the buffer to the rest of the system.
/// 
/// After this call, the buffer advances internally to a new slot, allowing new data to be written
/// into an unused portion of memory. This approach avoids memory duplication and ensures efficient 
/// reuse of the buffer without transferring ownership of the memory. 
///
/// This method is typically used when the data in the buffer is ready for processing, sending, or
/// further manipulation. The returned `Slice` contains the data, while the buffer itself remains 
/// ready to handle new incoming data by pointing to a fresh memory region.
```
Something on this line

utils/buffer/src/lib.rs
Shourya742: LGTM

utils/buffer/examples/multi_threaded_buffer_pool.rs
```diff
@@ -0,0 +1,58 @@
+// # Multi-Threaded Usage with Proper Memory Management
+//
+// This example shows how to use `BufferPool` in a multi-threaded context and ensures proper memory
+// slice release.
+//
+// - `Arc<Mutex<>>` is used to share `BufferPool` across threads safely.
+// - Each thread writes its own message into the buffer pool.
+// - The `data_slice` in each thread goes out of scope, releasing the memory slice.
+// - After joining all threads, the buffer pool is droppable.
+//
```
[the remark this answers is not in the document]
Shourya742: cc: @Fi3

utils/buffer/src/lib.rs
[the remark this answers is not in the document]
rrybarczyk: I will remove this note.

utils/buffer/README.md
```diff
@@ -1,105 +1,151 @@
...
+The `Buffer` trait is designed to work with the
+[`codec_sv2`](https://docs.rs/codec_sv2/1.3.0/codec_sv2/index.html) decoders, which operate by:
+
+1. Filling a buffer with the size of the protocol header being decoded.
+2. Parsing the filled bytes to compute the message length.
+3. Filling a buffer with the size of the message.
+4. Using the header and message to construct a
+   [`framing_sv2::framing::Frame`](https://docs.rs/framing_sv2/2.0.0/framing_sv2/framing/enum.Frame.html).
+
+To fill the buffer, the `codec_sv2` decoder must bass a reference of the buffer to a filler. To
```
GitGab19, line 76:
```suggestion
To fill the buffer, the `codec_sv2` decoder must pass a reference of the buffer to a filler. To
```

utils/buffer/README.md
```diff
@@ -1,105 +1,151 @@
...
+   the front.
+3. **Alloc Mode**: Falls back to system memory allocation (`BufferFromSystemMemory`) when both back
+   and front sections are full, providing additional capacity but with reduced performance.
 
-Usually, a message is decoded then a response is sent then a new message is decoded, etc.
-So `BufferPool` is optimized for use all the slots then check if the first slot has been dropped
-If so use it, then check if the second slot has been dropped, and so on.
-`BufferPool` is also optimized to drop all the slices.
-`BufferPool` is also optimized to drop the last slice.
+`BufferPool` can only be fragmented between the front and back and between back and end.
```
GitGab19, line 127: What does this mean? I'm having some problems in understanding how this back, front and end works tbh.

utils/buffer/README.md
```diff
@@ -1,105 +1,151 @@
...
 
-Below a graphical representation of the most optimized cases:
-```
-A number [0f] means that the slot is taken, the minus symbol (-) means the slot is free
-There are 8 slot
+#### Fragmentation, Overflow, and Optimization
+`BufferPool` can allocate a maximum of `8` `Slice`s (as it uses an `AtomicU8` to track used and
+freed slots) and up to the defined capacity in bytes. If all `8` slows are taken or there is no more
```
GitGab19, line 131:
```suggestion
freed slots) and up to the defined capacity in bytes. If all `8` slots are taken or there is no more
```

utils/buffer/README.md
```diff
@@ -147,40 +193,79 @@ CASE 3
...
 
 ```
+cargo bench --features criterion
+```
+
+## Benchmarks Comparisons
+
+`BufferPool` is benchmarked against `BufferFromSystemMemory` and two additional structure for
+reference: `PPool` (a hashmap-based pool) and `MaxEfficeincy` (a highly optimized but unrealistic
+control implementation writen such that the benchmarks do not panic and the compiler does not
```
GitGab19, line 210:
```suggestion
control implementation written such that the benchmarks do not panic and the compiler does not
```

utils/buffer/README.md
GitGab19, line 1: In CASE 1, we have:
```
12345678  BACK MODE
12345698  BACK MODE
```

Why don't we have:
```
12345678  BACK MODE
12345679  BACK MODE
```
?

utils/buffer/examples/basic_buffer_pool.rs
```diff
@@ -0,0 +1,41 @@
...
+    writable.copy_from_slice(data_to_write);
+    assert_eq!(buffer_pool.len(), 12);
+
+    // Retrieve the data as a referenced slice
+    let _data_slice = buffer_pool.get_data_by_ref(12);
+    assert_eq!(buffer_pool.len(), 12); // Buffer still owns data
+
+    // Retrieve the data as an owned slice
+    let data_slice = buffer_pool.get_data_owned();
+    assert_eq!(buffer_pool.len(), 0); // Buffer is now empty
```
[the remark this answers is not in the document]
GitGab19: So it is not empty after you retrieve data with `get_data_owned`?

utils/buffer/examples/buffer_pool_exhaustion.rs
```diff
@@ -0,0 +1,65 @@
...
+use buffer_sv2::{Buffer, BufferPool};
+use std::collections::VecDeque;
+
+fn main() {
+    // 8 byte capacity
+    let mut buffer_pool = BufferPool::new(8);
+    let mut slices = VecDeque::new();
+
+    // Write data to fill back slots
+    for _ in 0..4 {
```
GitGab19, line 28: Shouldn't this be `for _ in 0..8` ? Otherwise only 4 bytes are going to be written, but you defined a 8 bytes capacity. Maybe I'm missing something

utils/buffer/examples/buffer_pool_exhaustion.rs
```diff
@@ -0,0 +1,65 @@
...
+        let data_bytes = b"a"; // 1 byte
+        let writable = buffer_pool.get_writable(data_bytes.len()); // Mutable slice to internal
+                                                                   // buffer
+        writable.copy_from_slice(data_bytes);
+        let data_slice = buffer_pool.get_data_owned(); // Take ownership of allocated segment
+        slices.push_back(data_slice);
```
GitGab19, line 34: Or maybe you're adding the last 4 bytes here?

utils/buffer/src/buffer.rs
```diff
@@ -1,15 +1,41 @@
...
 #[derive(Debug)]
 pub struct BufferFromSystemMemory {
+    // Underlying buffer storing the data.
     inner: Vec<u8>,
+
+    // Current cursor indicating where the next byte should be written.
     cursor: usize,
+
+    // Starting index for the buffer. Useful for scenarios where part of the buffer is skipped or
+    // invalid.
```
GitGab19, line 33: Shouldn't these comments be public docs?

GitGab19 (commented):
Partial review, I will complete it in some hours, but some answers to my comments would make me easier to complete the review 🙏

utils/buffer/README.md
GitGab19: Can you expand more?

utils/buffer/README.md
rrybarczyk: @Fi3 can we get some clarity on `CASE 1`. It is:

```
CASE 1
--------  BACK MODE
1-------  BACK MODE
12------  BACK MODE
123-----  BACK MODE
1234----  BACK MODE
12345---  BACK MODE
123456--  BACK MODE
1234567-  BACK MODE
12345678  BACK MODE
12345698  BACK MODE
1234569a  BACK MODE
123456ba  BACK MODE
123456ca  BACK MODE
..... and so on
```

We are unclear on the line `12345698  BACK MODE`. I would expect it to be something like:

```
...
1234567-  BACK MODE
12345678  BACK MODE
12345678  ALLOC MODE (new bytes being allocated in a new space in the heap)
12345678  ALLOC MODE
--------  RESET
9a------  BACK MODE
9ab-----  BACK MODE
```

utils/buffer/README.md
```diff
@@ -1,105 +1,151 @@
...
+considering further allocation. It is also optimized to drop all the slices and to drop the last
+slice. It also efficiently handles scenarios where all slices are dropped or when the last slice is
+released, reducing memory fragmentation.
+
+The following cases illustrate typical memory usage patterns within `BufferPool`:
+1. Slots fill from back to front, switching as each area reaches capacity.
+2. Pool resets upon full usage, then reuses back slots.
+3. After filling the back, front slots are used when they become available.
+
+Below is a graphical representation of the most optimized cases. A number (`0f`) means that the
```
plebhash, line 145: > A number (`0f`)

is `0f` meant to be a random or specific example?

utils/buffer/README.md
```diff
@@ -1,105 +1,151 @@
...
 
-## Write
-Waiting for `Write` in `core` a compatible trait is used so that it can be replaced.
+2. **[Buffer Pool Exhaustion Example](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/codec-sv2/examples/buffer_pool_exhaustion.rs)**:
+   Demonstrates how data is added to a buffer pool and dynamically allocates directly to the heap
+   once the buffer pool's capacity has been exhausted.
 
-## Buffer
-The `Buffer` trait has been written to work with `codec_sv2::Decoder`.
+3. **[Variable Sized Messages Example](https://github.com/stratum-mining/stratum/blob/main/protocols/v2/codec-sv2/examples/variable_sized_messages.rs)**:
```
plebhash, line 62: these links are all pointing as if the examples lived under `protocols/v2/codec-sv2`

utils/buffer/examples/basic_buffer_pool.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    let expect_data = [67, 105, 97, 111, 44, 32, 109, 117, 110, 100, 111, 33];
```
plebhash, line 39: what are these numbers? are they somewhat related to `b"Ciao, mundo!"`? if so, shouldn't we make that explicitly clear?

right now, I don't really understand the point that is being illustrated here

utils/buffer/examples/basic_buffer_pool.rs
```diff
@@ -0,0 +1,41 @@
...
+    assert_eq!(data_slice.as_ref(), expect_data);
```
plebhash, line 40: I switched this and the test still works... I feel `expect_data` is actually causing more confusion with no real benefit
```suggestion
    assert_eq!(data_slice.as_ref(), data_to_write);
```

utils/buffer/src/buffer.rs
```diff
@@ -1,15 +1,41 @@
+// # Buffer
```
plebhash, line 1: aren't we documenting `BufferFromSystemMemory` here?

`Buffer` is on a different file

utils/buffer/README.md
plebhash: btw @Fi3 told me via DM he's busy and won't be able to look into this, so we shouldn't wait for his input before fixing this

utils/buffer/src/lib.rs
```diff
@@ -15,23 +54,39 @@ pub use aes_gcm::aead::Buffer as AeadBuffer;
...
     fn write(&mut self, buf: &[u8]) -> Result<usize, WriteError>;
 
+    /// Attempts to write the entire byte slice (`buf`) into the buffer. If the buffer cannot
+    /// accept the full length of the data, an error is returned.
     fn write_all(&mut self, buf: &[u8]) -> Result<(), WriteError>;
 }
 
 impl Write for Vec<u8> {
+    /// Writes data from a byte slice into a [`Vec<u8>`] buffer by extending the vector with the
+    /// contents of the provided slice.
```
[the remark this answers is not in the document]
rrybarczyk: These are generated to the docs, I know we already have the comments on the `Write` trait def, but I like adding these to explicitly discuss the `Vec<u8>` impl.

Screenshot 2024-11-27 at 7 27 47 AM

utils/buffer/README.md
rrybarczyk: You can only add data at the beginning or end of the buffer array. You cannot add data in the middle.

Like if our buffer is 6 bytes:
`[0, 0, 0, 0, 0, 0]`

You can add data like:
`[0, 0, 0, 0, 0, 1]` 
or:
`[1, 0, 0, 0, 0, 0]` 

but NOT:
`[0, 0, 0, 1, 0, 0]`

utils/buffer/examples/buffer_pool_exhaustion.rs
plebhash: @rrybarczyk did you miss this one?

utils/buffer/examples/basic_buffer_pool.rs
```diff
@@ -0,0 +1,41 @@
...
+// # Run
+//
+// ```
+// cargo run --example basic_buffer_pool
+// ```
+
+use buffer_sv2::{Buffer, BufferPool};
+
+fn main() {
+    // Create a new BufferPool with a capacity of 32 bytes
```
plebhash, line 19: this comment feels a bit confusing/misleading

by reading it with naive eyes, my first thought is: "this buffer pool can only fit 32 bytes"

which is not true!

AFAIU, every buffer pool has exactly 8 slots of variable size (which is 32 bytes here)

utils/buffer/examples/buffer_pool_exhaustion.rs
```diff
@@ -0,0 +1,65 @@
...
+//
+// ```
+// cargo run --example buffer_pool_exhaustion
+// ```
+
+use buffer_sv2::{Buffer, BufferPool};
+use std::collections::VecDeque;
+
+fn main() {
+    // 8 byte capacity
```
plebhash, line 23: ditto https://github.com/stratum-mining/stratum/pull/1232/files#r1865225440

utils/buffer/examples/variable_sized_messages.rs
```diff
@@ -0,0 +1,60 @@
...
+//
+// ```
+// cargo run --example variable_sized_messages
+// ```
+
+use buffer_sv2::{Buffer, BufferPool};
+use std::collections::VecDeque;
+
+fn main() {
+    // Initialize a BufferPool with a capacity of 32 bytes
```
plebhash, line 15: ditto https://github.com/stratum-mining/stratum/pull/1232/files#r1865225440

utils/buffer/examples/buffer_pool_exhaustion.rs
plebhash: this is how the buffer pool evolves over as the example progresses:

```
--------  BACK MODE
a-------  BACK MODE (add "a" via loop)
aa------  BACK MODE (add "a" via loop)
aaa-----  BACK MODE (add "a" via loop)
aaaa----  BACK MODE (add "a" via loop)
-aaa----  BACK MODE (pop front)
-aaab---  BACK MODE (add "b")
-aaabc--  BACK MODE (add "c" via loop)
-aaabcc-  BACK MODE (add "c" via loop)
-aaabccc  BACK MODE (add "c" via loop)
caaabccc  BACK MODE (add "c" via loop, which gets added via front mode)
caaabccc  ALLOC MODE (add "d", allocated in a new space in the heap)
```

utils/buffer/examples/buffer_pool_exhaustion.rs
plebhash: see https://github.com/stratum-mining/stratum/pull/1232/files#r1865236367

utils/buffer/examples/buffer_pool_exhaustion.rs
plebhash: @rrybarczyk it could be interesting to add this explanation above into the header comments

I'll add a suggestion

utils/buffer/examples/buffer_pool_exhaustion.rs
```diff
@@ -0,0 +1,65 @@
...
+// This example demonstrates how a buffer pool is filled. The back slots of the buffer pool are
+// exhausted first, followed by the front of the buffer pool. Once both the back and front are
+// exhausted, data is allocated on the heap at a performance decrease.
+//
+// 1. Fills up the back slots of the buffer pool until they’re exhausted.
+// 2. Releases one slot to allow the buffer pool to switch to front mode.
+// 3. Fully fills the front slots of the buffer pool.
+// 4. Switches to alloc mode for direct heap allocation when both the buffer pool's back and front
+//    slots are at capacity.
+//
```
plebhash, line 12:
```suggestion
// Below is a visual representation of how the buffer pool evolves evolves as the example progresses:
//
// --------  BACK MODE
// a-------  BACK MODE (add "a" via loop)
// aa------  BACK MODE (add "a" via loop)
// aaa-----  BACK MODE (add "a" via loop)
// aaaa----  BACK MODE (add "a" via loop)
// -aaa----  BACK MODE (pop front)
// -aaab---  BACK MODE (add "b")
// -aaabc--  BACK MODE (add "c" via loop)
// -aaabcc-  BACK MODE (add "c" via loop)
// -aaabccc  BACK MODE (add "c" via loop)
// caaabccc  BACK MODE (add "c" via loop, which gets added via front mode)
// caaabccc  ALLOC MODE (add "d", allocated in a new space in the heap)
```

utils/buffer/examples/variable_sized_messages.rs
```diff
@@ -0,0 +1,60 @@
...
+    let large_message = b"This message is larger than the remaining buffer pool capacity.";
+    write_data(&mut buffer_pool, large_message, &mut slices);
+    assert!(buffer_pool.is_alloc_mode());
+    assert_eq!(slices.back().unwrap().as_ref(), large_message);
+
+    // Clean up: Drop slices to free up buffer pool space
+    slices.pop_front();
+    while let Some(slice) = slices.pop_front() {
+        drop(slice);
+    }
```
plebhash, line 51: why do we do one `pop_front` individually, and then do multiple over the while loop?

is that intentional?

if not, why not just doing everything over the while loop?

plebhash:
@rrybarczyk I guess I said on the last call that I felt this PR was ready, but after another look I left a few extra comments

utils/buffer/src/slice.rs
```diff
@@ -1,21 +1,99 @@
...
 #[cfg(feature = "debug")]
 use std::time::SystemTime;
 
+// A special index value used to mark `Slice` as ignored in certain operations, such as memory pool
+// tracking or state management.
+//
+// It can be used as a sentinel value for slices that should not be processed or tracked, helping
+// differentiate valid slices from those that need to be skipped. When a `Slice`'s `index` is set
+// to `INGORE_INDEX`, it is flagged to be ignored and by any logic that processes or tracks slice
+// indices.
```
GitGab19, line 35: Why did you use normal comments and not public docs ones here?

utils/buffer/src/slice.rs
```diff
@@ -1,21 +1,99 @@
...
+    //
+    // Allows for efficient access to the underlying memory. Care should be taken when working with
+    // raw pointers to avoid memory safety issues. The pointer must be valid and must point to a
+    // properly allocated and initialized memory region.
     pub(crate) offset: *mut u8,
+
+    // Length of the memory block in bytes.
+    //
+    // Represents how much memory is being used. This is critical for safe memory access, as it
+    // prevents reading or writing outside the bounds of the buffer.
```
GitGab19, line 62: Why did you use normal comments and not public docs ones here?

utils/buffer/src/slice.rs
```diff
@@ -49,48 +135,82 @@ impl Slice {
...
     fn index(&self, index: core::ops::Range<usize>) -> &Self::Output {
         self.as_ref().index(index)
     }
 }
+
 impl core::ops::IndexMut<core::ops::Range<usize>> for Slice {
+    /// Provides mutable slicing access to the specified range within the `Slice`.
+    ///
+    /// Uses `as_mut` to get a mutable reference to the underlying buffer and returns the specified
+    /// range.
```
GitGab19, line 192: Why did you use public docs comments and not normal ones here?

utils/buffer/src/slice.rs
```diff
@@ -49,48 +135,82 @@ impl Slice {
...
         self.as_mut().index_mut(index)
     }
 }
+
 impl core::ops::Index<core::ops::RangeFull> for Slice {
     type Output = [u8];
 
+    /// Provides immutable access to the entire range of the [`Slice`].
+    ///
+    /// Uses `as_ref` to get a reference to the entire underlying buffer.
```
GitGab19, line 203: Why did you use public docs comments and not normal ones here?

utils/buffer/src/slice.rs
```diff
@@ -49,48 +135,82 @@ impl Slice {
...
     fn index(&self, index: core::ops::RangeFull) -> &Self::Output {
         self.as_ref().index(index)
     }
 }
 
 impl AsMut<[u8]> for Slice {
+    /// Converts the [`Slice`] into a mutable slice of bytes (`&mut [u8]`).
+    ///
+    /// Returns the owned buffer if present, otherwise converts the raw pointer and length into a
+    /// mutable slice.
```
GitGab19, line 213: Why did you use public docs comments and not normal ones here?

utils/buffer/src/slice.rs
```diff
@@ -99,7 +219,12 @@ impl AsMut<[u8]> for Slice {
         }
     }
 }
+
 impl AsRef<[u8]> for Slice {
+    /// Converts the [`Slice`] into an immutable slice of bytes (`&[u8]`).
+    ///
+    /// Returns the owned buffer if present, otherwise converts the raw pointer and length into an
+    /// immutable slice.
```
GitGab19, line 227: Why did you use public docs comments and not normal ones here?

utils/buffer/src/slice.rs
```diff
@@ -110,20 +235,29 @@ impl AsRef<[u8]> for Slice {
 }
 
 impl Drop for Slice {
+    /// Toggles the shared state when the slice is dropped, allowing the memory to be reused.
+    ///
+    /// In debug mode, it also tracks the `mode` of the slice when it is dropped.
```
GitGab19, line 240: Why did you use public docs comments and not normal ones here?

utils/buffer/src/slice.rs
```diff
@@ -110,20 +235,29 @@ impl AsRef<[u8]> for Slice {
...
+
         #[cfg(not(feature = "debug"))]
         self.shared_state.toogle(self.index);
     }
 }
 
 impl From<Vec<u8>> for Slice {
+    /// Creates a [`Slice`] from a [`Vec<u8>`], taking ownership of the vector.
+    ///
+    /// Initializes the [`Slice`] with the vector's pointer and sets the length to `0`.
```
GitGab19, line 253: Why did you use public docs comments and not normal ones here?

utils/buffer/src/slice.rs
```diff
@@ -135,25 +269,66 @@ impl From<Vec<u8>> for Slice {
...
+    // Creates a new `SharedState` with an internal `AtomicU8` initialized to `0`, indicating no
+    // memory slots are in use.
     fn default() -> Self {
         Self::new()
     }
 }
 
 impl SharedState {
+    // Creates a new `SharedState` with an internal `AtomicU8` initialized to `0`, indicating no
+    // memory slots are in use.
```
GitGab19, line 304: Why did you use  normal comments and not public docs ones here?

utils/buffer/src/slice.rs
```diff
@@ -135,25 +269,66 @@ impl From<Vec<u8>> for Slice {
...
     pub fn new() -> Self {
         Self(Arc::new(AtomicU8::new(0)))
     }
 
+    // Atomically loads and returns the current value of the `SharedState` using the specified
+    // memory ordering.
+    //
+    // Returns the current state of the memory slots as an 8-bit value.
```
GitGab19, line 312: Why did you use  normal comments and not public docs ones here?

utils/buffer/src/slice.rs
```diff
@@ -135,25 +269,66 @@ impl From<Vec<u8>> for Slice {
...
+    // After a message held by a buffer slice has been processed, the corresponding bit in the
+    // shared state is toggled (flipped). When the shared state for a given region reaches zero
+    // (i.e., all bits are cleared), the buffer pool knows it can safely reclaim or reuse that
+    // memory slice.
+    //
+    // Uses atomic bitwise operations to ensure thread-safe toggling without locks. It manipulates
+    // the shared state in-place using the `AtomicU8::fetch_update` method, which atomically
+    // applies a bitwise XOR (`^`) to toggle the bit at the specified `position`.
+    //
+    // Panics if the `position` is outside the range of 1-8, as this refers to an invalid bit.
```
GitGab19, line 331: Why did you use  normal comments and not public docs ones here?

utils/buffer/src/slice.rs
```diff
@@ -182,6 +357,18 @@ impl SharedState {
...
```
GitGab19, line 371: Why did you use  normal comments and not public docs ones here?

utils/buffer/src/slice.rs
rrybarczyk: It's private.

utils/buffer/src/slice.rs
rrybarczyk: It's public.

utils/buffer/src/slice.rs
rrybarczyk: It's private.

utils/buffer/src/slice.rs
rrybarczyk: It's private.

utils/buffer/src/slice.rs
rrybarczyk: It's private.

utils/buffer/examples/variable_sized_messages.rs
rrybarczyk: Yes, it is intentional. You only switch modes if you have popped off the first one in the front.

utils/buffer/examples/basic_buffer_pool.rs
rrybarczyk: I think it is 32 bytes, with each slot holding 4 bytes.

```
 BufferPool { pool_back: PoolBack { back_start: 0, len: 0 }, mode: Back, shared_state: SharedState(0), inner_memory: InnerMemory { pool: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], raw_offset: 0, raw_len: 0, slots: [(0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0)], len: 0 }, system_memory: BufferFromSystemMemory { inner: [], cursor: 0, start: 0 }, start: 0 }
```

utils/buffer/examples/buffer_pool_exhaustion.rs
rrybarczyk: I think it is overkill since we have these cases in the `README.md`.

utils/buffer/examples/basic_buffer_pool.rs
Shourya742: We have 8 slots with a total capacity of 32 bytes. The slots do not need to have equal capacity, such as 32/8 = 4 bytes each. They can vary in length, as long as the total size remains 32 bytes.

plebhash: I see, my original comment was wrong then.

utils/buffer/examples/buffer_pool_exhaustion.rs
plebhash: the `README.md`has a similar table but it's not identical to this

the motivation was the fact that @GitGab19 was confused about the loops, so the goal of this one is to help the reader understand the flow of the example, especially what's happening around each loop

utils/buffer/examples/variable_sized_messages.rs
plebhash: my question is: how is that different from this?

```diff
     // Clean up: Drop slices to free up buffer pool space
-    slices.pop_front();
     while let Some(slice) = slices.pop_front() {
         drop(slice);
     }
```

I applied this change and running `cargo run --example variable_sized_messages` has the exact same output.

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -13,19 +46,39 @@ use std::time::SystemTime;
...
-pub const POOL_CAPACITY: usize = 8;
 mod pool_back;
 pub use pool_back::PoolBack;
 
+// Maximum number of memory slices the buffer pool can concurrently manage.
+//
+// This value limits the number of slices the `BufferPool` can track and manage at once. Once the
+// pool reaches its capacity of `8` slices, it may need to free memory or switch modes (e.g., to
+// system memory). The choice of `8` ensures a balance between performance and memory management.
+// The use of `usize` allows for compatibility with platform-dependent memory operations.
```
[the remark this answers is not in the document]
rrybarczyk: It is private. Before commenting on comment slashes, can you execute the docs and check it out to confirm?

utils/buffer/README.md
```diff
@@ -1,106 +1,152 @@
...
+The `Buffer` trait is implemented for `BufferFromSystemMemory` and `BufferPool`. It includes a
+`Write` trait to replace `std::io::Write` in `no_std` environments.
 
-`get_data_owned(..)` returns the inner vector.
+## `BufferPoolFromSystemMemory`
+`BufferFromSystemMemory` is a simple implementation of the `Buffer` trait. Each time a new buffer is
+needed, it creates a new `Vec<u8>`.
 
+- `get_writable(..)` returns mutable references to the inner vector.
+- `get_data_owned(..)` returns the inner vector.
```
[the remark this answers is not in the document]
GitGab19: @rrybarczyk I saw you resolved this, but I don't see any answers here, nor changes

utils/buffer/README.md
```diff
@@ -1,106 +1,152 @@
...
-get_data_owned(&mut self) -> Slice {
+get_data_owned(&mut self) -> Slice;
 ```
-It returns `Slice`: something that implements `AsMut[u8]` and `Send`, and sets the buffer len to 0.
 
-## BufferFromSystemMemory
-Is the simplest implementation of a `Buffer`: each time that a new buffer is needed it create a new
-`Vec<u8>`.
+This `get_data_owned` method returns a `Slice` that implements `AsMut<[u8]>` and `Send`, and resets
+the buffer length to `0`.
```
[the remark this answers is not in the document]
GitGab19: What's the answer to this question?

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -110,25 +178,54 @@ impl PoolFront {
     }
 }
 
+// Current mode of operation for the `BufferPool`.
+//
+// The pool operates in three modes based on memory availability: it first allocates from the back,
+// then from the front if the back is full, and finally from system memory (with reduced
+// performance) if both sections are exhausted.
```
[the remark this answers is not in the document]
GitGab19: Since this `pub enum PoolMode` is a field of the public struct `BufferPool`, shouldn't it have public doc comments? 

I see it when I render docs locally:
Screenshot 2024-12-10 at 09 26 07

utils/buffer/src/buffer_pool/mod.rs
```diff
@@ -110,25 +178,54 @@ impl PoolFront {
...
+    // The buffer pool is operating in "back" mode, where memory is allocated from the back of the
+    // buffer pool.
     Back,
+
+    // The buffer pool is operating in "front" mode, where memory is allocated from the front of
+    // the buffer pool. Used when the back is full.
     Front(PoolFront),
+
+    // The pool has exhausted its internal memory, and it is now allocating directly from the
+    // system memory (heap).
```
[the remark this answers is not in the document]
GitGab19: Depending on your answer on [this comment](https://github.com/stratum-mining/stratum/pull/1232#discussion_r1877107753), this could change

GitGab19:
Alright, I just cleaned up many old comments from my last review. What I left there is something which is worth resolving/answering before proceeding @rrybarczyk

utils/buffer/README.md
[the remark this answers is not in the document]
rrybarczyk: It does not. You can check yourself by running the `variable_sized_message` example.

utils/buffer/README.md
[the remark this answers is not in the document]
rrybarczyk: https://github.com/stratum-mining/stratum/pull/1232#discussion_r1877171461

GitGab19 (approved):

plebhash (approved):

PR #1233, opened 2024-10-25
draw a handy dependancy graph

Georges760:
to better undersatnding of the current protocol/v2/dependancies

### DIFF
--- a/protocols/DEPENDANCIES.md
+++ b/protocols/DEPENDANCIES.md
@@ -1,0 +1,40 @@
+```mermaid
+stateDiagram-v2
+    secp256k1 --> noise_sv2
+    rand --> noise_sv2
+    aes_gcm --> noise_sv2
+    const_sv2 --> noise_sv2
+    rand_chacha --> noise_sv2
+    chacha20poly1305 --> noise_sv2
+
+    serde --> buffer_sv2: with_serde
+    criterion --> buffer_sv2: criterion
+    aes_gcm --> buffer_sv2
+
+    buffer_sv2 --> binary_codec_sv2: with_buffer_pool
+    quickcheck --> binary_codec_sv2: prop_test
+
+    binary_codec_sv2 --> derive_codec_sv2
+
+    serde --> serde_sv2
+    buffer_sv2 --> serde_sv2
+
+    serde_sv2 --> binary_sv2: with_serde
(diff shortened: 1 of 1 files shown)


protocols/DEPENDANCIES.md
plebhash, line 1: the correct spelling should be `DEPENDENCIES.md`

jbesraa:
Thanks for working on this.
Why is this good? in which scenarios we can use it?

Georges760:
> Thanks for working on this. Why is this good? in which scenarios we can use it?

It draws a dependencies graph with features propagation, for better understanding of the relations between crates : https://github.com/GPTechinno/stratum/blob/dep-graph/protocols/DEPENDENCIES.md

Shourya742:
@rrybarczyk, in this context, the dependency curve considers the individual flavors of our crates. For someone new to our protocol crates, they might appear as separate entities. For example, the `binary_sv2` crate has two variations: one with `serde` support and one without. Encapsulating both the `serde` and non-`serde` versions within a single `binary_sv2` block could make it less confusing, instead of representing them as separate crates. Let me know your thoughts on this. Based on your feedback, we can ask @Georges760 to make the necessary adjustments, and then we’ll be ready to proceed with this PR.

Shourya742:
> > @rrybarczyk, in this context, the dependency curve considers the individual flavors of our crates. For someone new to our protocol crates, they might appear as separate entities. For example, the `binary_sv2` crate has two variations: one with `serde` support and one without. Encapsulating both the `serde` and non-`serde` versions within a single `binary_sv2` block could make it less confusing, instead of representing them as separate crates. Let me know your thoughts on this. Based on your feedback, we can ask @Georges760 to make the necessary adjustments, and then we’ll be ready to proceed with this PR.
> 
> nack since we are going to drop the one with serde eventually

Does it make sense to have this dependency graph after we complete the refactor?

GitGab19:
Moving this PR to draft status, as it makes sense to have a final graph after protocols crates refactoring.

Shourya742:
> > @Georges760 Any plans to update this PR with new repo structure?
> 
> I would have to study the changes since my first shot. But why not !
> 
> Is the architecture stabilized now ?

No drastic changes in future comes to my mind, atleast regarding the dependencies.

PR #1236, opened 2024-10-25
Fix ambiguous naming on Integration Tests Sniffer API (`next_upstream_message` vs `next_downstream_message`)

### LINKED ISSUE #1235: Ambiguous naming on Integration Tests Sniffer API (`next_upstream_message` vs `next_downstream_message`)
While doing some experimentation with the Integration Tests, I found myself confused with regards to the usage of `sniffer.next_downstream_message()` vs `sniffer.next_upstream_message()`.

For example, imagine the following setup:
```
Role_A
            ^
            |
Sniffer  Message X
            |
            |
Role_B
```

As a user, my first intuition was to call `sniffer.next_upstream_message()` so I could fetch `Message X`.

That stemmed from the semantic interpretation of the term upstream as a **DIRECTION**.

Then, I realized I actually had to call `sniffer.next_downstream_message()`, because this method was written with the mindset that the message was coming FROM the downstream role. Here, downstream actually means **RELATIVE POSITION**, not direction.

When we're writing complex tests, having to constantly fight the cognitive dissonance coming from this ambiguity becomes a potential footgun, and also slows productivity.

The fix should be very straightfoward.

We can simply rename:
- `next_downstream_message()` into `next_message_from_downstream()`
- `next_upstream_message()` into `next_message_from_upstream()`

and
- `Sniffer.downstream_messages` into `Sniffer.messages_from_downstream`
- `Sniffer.upstream_messages` into `Sniffer.messages_from_upstream`
 
The semantic interpretation of "downstream" and "upstream" words would remain the same as it is now (relative position), but introducing the word **FROM** would make that interpretation explicitly clear and resolve any kind of ambiguity.

plebhash:
close #1235

### DIFF
--- a/roles/tests-integration/tests/common/sniffer.rs
+++ b/roles/tests-integration/tests/common/sniffer.rs
@@ -37,9 +37,9 @@ enum SnifferError {
 /// The downstream (or client) role connects to the [`Sniffer`] `listening_address` and the
 /// [`Sniffer`] connects to the `upstream` server. This way, the Sniffer can intercept messages sent
 /// between the downstream and upstream roles. The downstream will send its messages to the
-/// [`Sniffer`] which will save those in the `downstream_messages` aggregator and forward them to
-/// the upstream role. When a response is received it is saved in `upstream_messages` and
-/// forwarded to the downstream role. Both `downstream_messages` and `upstream_messages` can be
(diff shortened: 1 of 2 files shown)


jbesraa (changes requested):
LGTM

Could please update the docs above the `pub struct Sniffer` accordingly?

jbesraa (approved):

GitGab19 (approved):

PR #1237, opened 2024-10-26
const_sv2: remove unused dep (secp256k1)

Georges760:
const_sv2 does depend on secp256k1 anymore

### DIFF
--- a/protocols/v2/const-sv2/Cargo.toml
+++ b/protocols/v2/const-sv2/Cargo.toml
@@ -13,12 +13,6 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[dependencies]
-secp256k1 = { version = "0.28.2", default-features = false, features =["hashes", "alloc","rand","rand-std"] }
-
-#[dev-dependencies]
-#cbindgen = "0.16.0"
-
 [features]
 no_std = []
 
--- a/roles/Cargo.lock
+++ b/roles/Cargo.lock
@@ -750,9 +750,6 @@ dependencies = [
 [[package]]
 name = "const_sv2"
 version = "2.1.0"
-dependencies = [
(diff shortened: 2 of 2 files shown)


Shourya742 (approved):
ACK 0743f82839911f445f9e22b5e684c6f62d52108f

plebhash:
hey @Georges760 thanks for this contribution

can you elaborate on the rationale that motivated this contribution

Georges760:
#1130 ask to no_std noise_sv2, so I asked in Discord dev channel if I can give it a try. Doing so, I first had to understand the dependencies graph so I wrote a mermaid graph that ended in #12 and used #1230 to unify the no_std nature of protocol/sv2/* crates. 
The top level crate (the one all other depends on) is const_sv2, which is just a bunch of 'pub const' declaration. And it has a dependency use anywhere, so I asked about it on the Discord and @Fi3 reply me I could remove it. It was first part of #1230 (like the dependencies graph) then I made single PR for both of them. And here we are.

plebhash:
> #1130 ask to no_std noise_sv2, so I asked in Discord dev channel if I can give it a try. Doing so, I first had to understand the dependencies graph so I wrote a mermaid graph that ended in #1233 and used #1230 to unify the no_std nature of protocol/sv2/* crates. The top level crate (the one all other depends on) is const_sv2, which is just a bunch of 'pub const' declaration. And it has a dependency not used anywhere, so I asked about it on the Discord and @Fi3 reply me I could remove it. It was first part of #1230 (like the dependencies graph) then I made single PR for both of them. And here we are.

ah ok! sorry I missed that discussion on Discord while I was away for a few days

thank you for the context, I just caught up on that discussion

plebhash (approved):

GitGab19 (approved):
Since we are going to refactor a bit this crate (look at https://github.com/stratum-mining/stratum/issues/1135), it's ok to merge this PR now, but I want to underline that we are not going to bump the version of this crate for this tiny change. We will do it after the refactoring.

PR #1238, opened 2024-10-29
Propose an implementation of `noise_sv2` with optional `no_std`

Georges760:
Following @rrybarczyk [comment](https://github.com/stratum-mining/stratum/issues/1130#issuecomment-2422975674) on #1130.

Some equivalent conversion have been done (does not change any functionality, just using the `no_std` equivalents) :
- std::ptr -> core::ptr
- std::boxed::Box -> alloc::boxed::Box
- std::vec::Vec -> alloc::vec::Vec
- std::string::{String, ToString} -> alloc::string::{String, ToString}
- std::convert::TryInto -> core::convert::TryInto
- std::fmt::{Debug, Formatter, Result} -> core::fmt::{Debug, Formatter, Result}
- std::time::Duration -> core::time::Duration

To have a `no-std` version, the `--no-default-features` must be used.

Current public API (`std` dependant) is unchanged.
Additional public API for `no_std` compliance is available using the `*_with_rng` and `*_with_now` suffix, and the corresponding arguments. This delegate the choice of the Ramdom Number Generator and the current System Time to the caller, instead of assuming using the ones from `std`.
- `Initiator::new_with_rng()`, `Initiator::from_raw_k_with_rng()`, `Initiator::without_pk_with_rng()`, `Responder::new_with_rng()`, `Responder::from_authority_kp_with_rng()` and `Responder::generate_key_with_rng()` take an additional argument: `rng` implementing rand::Rng + ?Sized
- `SignatureNoiseMessage::sign_with_rng()` take an additional argument: `rng` implementing rand::Rng + rand::CryptoRng
- `Initiator::step_2_with_now()`  and `SignatureNoiseMessage::verify_with_now()` take an additional argument: `now` for the current system time epoch
- `Responder::step_1()_with_now_rng` take two additional arguments: `rng` implementing rand::Rng + rand::CryptoRng and `now` for the current system time epoch

### DIFF
--- a/protocols/v2/noise-sv2/Cargo.toml
+++ b/protocols/v2/noise-sv2/Cargo.toml
@@ -12,13 +12,24 @@ homepage = "https://stratumprotocol.org"
 keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 [dependencies]
-secp256k1 = { version = "0.28.2", default-features = false, features =["hashes", "alloc","rand","rand-std"] }
-rand = {version = "0.8.5", default-features = false, features = ["std","std_rng"] }
+secp256k1 = { version = "0.28.2", default-features = false, features = ["hashes", "alloc", "rand"] }
+rand = {version = "0.8.5", default-features = false }
 aes-gcm = "0.10.2"
 chacha20poly1305 = "0.10.1"
-rand_chacha = "0.3.1"
+rand_chacha = { version = "0.3.1", default-features = false }
 const_sv2 = { version = "^3.0.0", path = "../../../protocols/v2/const-sv2"}
+
+[features]
+default = ["std"]
+std = ["rand/std", "rand/std_rng", "rand_chacha/std", "secp256k1/rand-std"]
 
 [dev-dependencies]
 quickcheck = "1.0.3"
 quickcheck_macros = "1"
+rand = {version = "0.8.5", default-features = false, features = ["std", "std_rng"] }
+
+[profile.dev]
+panic = "unwind"
+
+[profile.release]
+panic = "abort"
--- a/protocols/v2/noise-sv2/src/handshake.rs
+++ b/protocols/v2/noise-sv2/src/handshake.rs
@@ -30,6 +30,8 @@
 // mandatory for communication across external networks (e.g., between a local mining proxy and a
 // remote pool).
 
+use alloc::{string::String, vec::Vec};
+
 use crate::{aed_cipher::AeadCipher, cipher_state::CipherState, NOISE_HASHED_PROTOCOL_NAME_CHACHA};
 use chacha20poly1305::ChaCha20Poly1305;
 use secp256k1::{
@@ -99,14 +101,19 @@ pub trait HandshakeOp<Cipher: AeadCipher>: CipherState<Cipher> {
     // Generates a fresh key pair, consisting of a secret key and a corresponding public key,
     // using the [`Secp256k1`] elliptic curve. If the generated public key does not match the
     // expected parity, a new key pair is generated to ensure consistency.
+    #[cfg(feature = "std")]
     fn generate_key() -> Keypair {
+        Self::generate_key_with_rng(&mut rand::thread_rng())
+    }
+    #[inline]
+    fn generate_key_with_rng<R: rand::Rng + ?Sized>(rng: &mut R) -> Keypair {
         let secp = Secp256k1::new();
-        let (secret_key, _) = secp.generate_keypair(&mut rand::thread_rng());
+        let (secret_key, _) = secp.generate_keypair(rng);
         let kp = Keypair::from_secret_key(&secp, &secret_key);
         if kp.x_only_public_key().1 == crate::PARITY {
             kp
         } else {
-            Self::generate_key()
+            Self::generate_key_with_rng(rng)
         }
     }
 
@@ -273,10 +280,11 @@ pub trait HandshakeOp<Cipher: AeadCipher>: CipherState<Cipher> {
 #[cfg(test)]
 mod test {
     use super::*;
+    use alloc::string::ToString;
+    use core::convert::TryInto;
     use quickcheck::{Arbitrary, TestResult};
     use quickcheck_macros;
     use secp256k1::{ecdh::SharedSecret, SecretKey, XOnlyPublicKey};
-    use std::convert::TryInto;
 
     struct TestHandShake {
         k: Option<[u8; 32]>,
@@ -464,9 +472,27 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "std")]
     fn test_ecdh() {
         let key_pair_1 = TestHandShake::generate_key();
         let key_pair_2 = TestHandShake::generate_key();
+
+        let secret_1 = key_pair_1.secret_bytes();
+        let secret_2 = key_pair_2.secret_bytes();
+
+        let pub_1 = key_pair_1.x_only_public_key();
+        let pub_2 = key_pair_2.x_only_public_key();
+
+        let ecdh_1 = TestHandShake::ecdh(&secret_1, &pub_2.0.serialize());
+        let ecdh_2 = TestHandShake::ecdh(&secret_2, &pub_1.0.serialize());
+
+        assert!(ecdh_1 == ecdh_2);
+    }
+
+    #[test]
+    fn test_ecdh_with_rng() {
+        let key_pair_1 = TestHandShake::generate_key_with_rng(&mut rand::thread_rng());
+        let key_pair_2 = TestHandShake::generate_key_with_rng(&mut rand::thread_rng());
 
         let secret_1 = key_pair_1.secret_bytes();
         let secret_2 = key_pair_2.secret_bytes();
--- a/protocols/v2/noise-sv2/src/initiator.rs
+++ b/protocols/v2/noise-sv2/src/initiator.rs
@@ -34,7 +34,11 @@
 // The [`Drop`] trait is implemented to automatically trigger secure erasure when the [`Initiator`]
 // instance goes out of scope, preventing potential misuse or leakage of cryptographic material.
 
-use std::{convert::TryInto, ptr};
+use alloc::{
+    boxed::Box,
+    string::{String, ToString},
+};
+use core::{convert::TryInto, ptr};
 
 use crate::{
     cipher_state::{Cipher, CipherState, GenericCipher},
@@ -93,8 +97,8 @@ pub struct Initiator {
     c2: Option<GenericCipher>,
 }
 
-impl std::fmt::Debug for Initiator {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl core::fmt::Debug for Initiator {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.debug_struct("Initiator").finish()
     }
 }
@@ -164,14 +168,31 @@ impl Initiator {
     /// If the responder public key is provided, the initiator uses this key to authenticate the
     /// responder during the handshake. The initial initiator state is instantiated with the
     /// ephemeral key pair and handshake hash.
+    #[cfg(feature = "std")]
     pub fn new(pk: Option<XOnlyPublicKey>) -> Box<Self> {
+        Self::new_with_rng(pk, &mut rand::thread_rng())
+    }
+
+    /// Creates a new [`Initiator`] instance with an optional responder public key and a custom
+    /// random number generator.
+    ///
+    /// See [`Self::new`] for more details.
+    ///
+    /// The custom random number generator is used to generate the ephemeral key pair. It should be
+    /// provided in order to not implicitely rely on `std` and allow `no_std` environments to
+    /// provide a hardware random number generator for example.
+    #[inline]
+    pub fn new_with_rng<R: rand::Rng + ?Sized>(
+        pk: Option<XOnlyPublicKey>,
+        rng: &mut R,
+    ) -> Box<Self> {
         let mut self_ = Self {
             handshake_cipher: None,
             k: None,
             n: 0,
             ck: [0; 32],
             h: [0; 32],
-            e: Self::generate_key(),
+            e: Self::generate_key_with_rng(rng),
             responder_authority_pk: pk,
             c1: None,
             c2: None,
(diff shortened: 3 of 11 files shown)


Sjors:
Is it possible to run tests on both the regular and `no_std` build?

Sjors:
I see. But I wonder if it's possible & safer to keep two versions, because IIUC `no_std` misses various safety features.

https://docs.rust-embedded.org/book/intro/no-std.html

Georges760:
> ble & saf

a double API conditioned by the `std` feature ? yeah why not ! I will push that this weekend.

Georges760:
> I see. But I wonder if it's possible & safer to keep two versions, because IIUC `no_std` misses various safety features.
> 
> https://docs.rust-embedded.org/book/intro/no-std.html

a double API conditioned by the std feature ? yeah why not ! I will push that this weekend.

Georges760:
After reflexion, I think that a `#![no_std]` lib crate still does not enforce the main crate (the one choosing to link against std or not) so `stack overflow protection` is only lost if and only if the main crate choose to be `#![no_std]` (real embedded system).

If the main crate is `std` aware, even if it use some `#![no_std]` lib as dep, it will profit from the `std` security features.

[handshake example](https://github.com/GPTechinno/stratum/blob/noise_sv2_no_std/protocols/v2/noise-sv2/examples/handshake.rs) is a good exemple of a 'main' std crate (using `std::rand` and `std::time`) using a `#![no_std]` lib dep (noise_sv2) and providing the Random Number Generator and System Time (of std) through the lib API so the lib does require std itself.

Breaking the API is a problem for `std` enviroement, and I fully agree it should be broken. So I will push this dual API (std unchanged, no_std new) according to a `std` feature enabled by default, in order to keep backward compat.

Georges760:
Thanks to @Sjors good comment, the current API is unchanged and an addition one allow `no_std`.

This is the way `secp256k1` crate is doing it too, see [sign_schnorr_with_rng](https://github.com/rust-bitcoin/rust-secp256k1/blob/master/src/schnorr.rs#L165C12-L165C33).

protocols/v2/noise-sv2/src/handshake.rs
```diff
@@ -99,14 +101,19 @@ pub trait HandshakeOp<Cipher: AeadCipher>: CipherState<Cipher> {
...
+    fn generate_key_with_rng<R: rand::Rng + ?Sized>(rng: &mut R) -> Keypair {
         let secp = Secp256k1::new();
-        let (secret_key, _) = secp.generate_keypair(&mut rand::thread_rng());
+        let (secret_key, _) = secp.generate_keypair(rng);
         let kp = Keypair::from_secret_key(&secp, &secret_key);
         if kp.x_only_public_key().1 == crate::PARITY {
             kp
         } else {
-            Self::generate_key()
+            Self::generate_key_with_rng(rng)
```
Shourya742, line 116: Can you negate the keys here, rather than recursive frames otherwise it gonna bloat the stack space.

protocols/v2/noise-sv2/src/signature_message.rs
```diff
@@ -70,12 +71,21 @@ impl SignatureNoiseMessage {
...
     // is already valid without further verification.
+    #[cfg(feature = "std")]
     pub fn verify(self, pk: &XOnlyPublicKey, authority_pk: &Option<XOnlyPublicKey>) -> bool {
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_secs() as u32;
+        self.verify_with_now(pk, authority_pk, now)
+    }
+    pub fn verify_with_now(
```
Shourya742, line 82: making this method #[inline] here, will help to reduce the stack space.

protocols/v2/noise-sv2/src/responder.rs
```diff
@@ -331,7 +367,13 @@ impl Responder {
     // certificate validity period, and a cryptographic signature. The signature is created using
     // the responder's static public key and authority keypair, ensuring that the responder's
     // identity and certificate validity are cryptographically verifiable.
-    fn get_signature(&self, version: u16, valid_from: u32, not_valid_after: u32) -> [u8; 74] {
+    fn get_signature<R: rand::Rng + rand::CryptoRng>(
```
Shourya742, line 370: making this method #[inline] here, will help to reduce the stack space.

protocols/v2/noise-sv2/src/responder.rs
```diff
@@ -226,9 +248,27 @@ impl Responder {
...
+            .unwrap()
+            .as_secs() as u32;
+
+        self.step_1_with_now_rng(
+            elligatorswift_theirs_ephemeral_serialized,
+            now,
+            &mut rand::thread_rng(),
+        )
+    }
+    pub fn step_1_with_now_rng<R: rand::Rng + rand::CryptoRng>(
```
Shourya742, line 267: making this method #[inline] here, will help to reduce the stack space.

protocols/v2/noise-sv2/src/handshake.rs
```diff
@@ -30,6 +30,8 @@
 // mandatory for communication across external networks (e.g., between a local mining proxy and a
 // remote pool).
 
+use alloc::{string::String, vec::Vec};
+
```
Shourya742, line 34: As we are now having support for both env, we should consider to have environment specific imports.

Shourya742 (changes requested):
utack 9f2a908f229fa2079fe3ab81849c3368f08a1c8e. Will test the PR once with the whole setup up and running. Rest looks ok on first glance with minor nits

protocols/v2/noise-sv2/src/signature_message.rs
Georges760: done

protocols/v2/noise-sv2/src/responder.rs
Georges760: done

protocols/v2/noise-sv2/src/responder.rs
Georges760: done

protocols/v2/noise-sv2/src/handshake.rs
Georges760: std::string::String is falling into alloc::string::String anyway

same for Vec

protocols/v2/noise-sv2/src/handshake.rs
Georges760: it is strickly not a `no_std` related job, I would like to keep this PR not refactoring any existing functionality.

Georges760:
> utack [9f2a908](https://github.com/stratum-mining/stratum/commit/9f2a908f229fa2079fe3ab81849c3368f08a1c8e). Will test the PR once with the whole setup up and running. Rest looks ok on first glance with minor nits

PR good to merge (for me) nothing more to add, you can test it

protocols/v2/noise-sv2/src/initiator.rs
```diff
@@ -164,14 +168,21 @@ impl Initiator {
     /// If the responder public key is provided, the initiator uses this key to authenticate the
     /// responder during the handshake. The initial initiator state is instantiated with the
     /// ephemeral key pair and handshake hash.
+    #[cfg(feature = "std")]
     pub fn new(pk: Option<XOnlyPublicKey>) -> Box<Self> {
+        Self::new_with_rng(pk, &mut rand::thread_rng())
+    }
+    pub fn new_with_rng<R: rand::Rng + ?Sized>(
```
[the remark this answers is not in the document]
Georges760: done

protocols/v2/noise-sv2/src/initiator.rs
```diff
@@ -187,19 +198,30 @@ impl Initiator {
     /// valid [`XOnlyPublicKey`], an [`Error::InvalidRawPublicKey`] error is returned.
     ///
     /// Typically used when the initiator is aware of the responder's public key in advance.
+    #[cfg(feature = "std")]
     pub fn from_raw_k(key: [u8; 32]) -> Result<Box<Self>, Error> {
+        Self::from_raw_k_with_rng(key, &mut rand::thread_rng())
+    }
+    pub fn from_raw_k_with_rng<R: rand::Rng + ?Sized>(
```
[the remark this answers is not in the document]
Georges760: done

protocols/v2/noise-sv2/src/initiator.rs
```diff
@@ -187,19 +198,30 @@ impl Initiator {
...
     /// This function initializes the [`Initiator`] with a default empty state and is intended
     /// for use when both the initiator and responder are within the same network. In this case,
     /// the initiator does not validate the responder's static key from a certificate. However,
     /// the connection remains encrypted.
+    #[cfg(feature = "std")]
     pub fn without_pk() -> Result<Box<Self>, Error> {
-        Ok(Self::new(None))
+        Self::without_pk_with_rng(&mut rand::thread_rng())
+    }
+    pub fn without_pk_with_rng<R: rand::Rng + ?Sized>(rng: &mut R) -> Result<Box<Self>, Error> {
```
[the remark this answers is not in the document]
Georges760: done

protocols/v2/noise-sv2/src/responder.rs
```diff
@@ -195,17 +208,26 @@ impl Responder {
...
     /// certificate validity duration is also set here. Fails if the key pair is mismatched.
+    #[cfg(feature = "std")]
     pub fn from_authority_kp(
         public: &[u8; 32],
         private: &[u8; 32],
         cert_validity: Duration,
+    ) -> Result<Box<Self>, Error> {
+        Self::from_authority_kp_with_rng(public, private, cert_validity, &mut rand::thread_rng())
+    }
+    pub fn from_authority_kp_with_rng<R: rand::Rng + ?Sized>(
```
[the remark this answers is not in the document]
Georges760: done

protocols/v2/noise-sv2/src/signature_message.rs
```diff
@@ -100,11 +111,20 @@ impl SignatureNoiseMessage {
     // Creates a Schnorr signature for the message, combining the version, validity period, and
     // the static public key of the server (`static_pk`). The resulting signature is then written
     // into the provided message buffer (`msg`).
+    #[cfg(feature = "std")]
     pub fn sign(msg: &mut [u8; 74], static_pk: &XOnlyPublicKey, kp: &Keypair) {
+        Self::sign_with_rng(msg, static_pk, kp, &mut rand::thread_rng());
+    }
+    pub fn sign_with_rng<R: rand::Rng + rand::CryptoRng>(
```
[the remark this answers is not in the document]
Georges760: done

protocols/v2/noise-sv2/src/responder.rs
Georges760: done

protocols/v2/noise-sv2/src/signature_message.rs
Georges760: with rust-doc

Georges760:
> @Georges760 You’ll need to bump the major version to make CI happy.

do you want me to do it in this PR ?

also should I rebase on main ?

Georges760:
> > The CI fails because of infra, need to be re-run when CI depending server/files are accessible again
> 
> Ya I noticed it now... I have opened a PR to address this, #1293

CI continue to fail, I am rebasing on every PR merge.

Georges760:
CI is still randomly failing :
- on last dev call, the re-run made it pass
- after my last reabse, it PASS
- after my README last push, it FAIL again (i am unable to re-run the job)

Georges760:
AFAIK @Shourya742 already TACK this PR, waiting for @rrybarczyk review.

rrybarczyk (approved):
Looks good to me. Thanks @Georges760.

plebhash:
@Georges760 I think this is almost ready for merging, but commit history looks a bit confusing.

For example, commit title of 771832b84781c453325a2b7e56e3592f7f77883d is:
> utils/key-utils: revert changes, doesn't have to be changed in this PR

also some commits could be squashed

plebhash:
> CI is still randomly failing :
> 
> * on last dev call, the re-run made it pass
> * after my last reabse, it PASS
> * after my README last push, it FAIL again (i am unable to re-run the job)

we're looking into a way to make it deterministic, apologies for that

but doesn't look like anything serious

Georges760:
> @Georges760 I think this is almost ready for merging, but commit history looks a bit confusing.
> 
> For example, commit title of 771832b84781c453325a2b7e56e3592f7f77883d is:
> > utils/key-utils: revert changes, doesn't have to be changed in this PR
> 
> I'm not sure how to interpret this... is this commit intended to be on this PR or not?
> 
> also some commits could be squashed (or dropped, in case they're not meant to be on this PR)
> 

I will squash everything tomorrow;)

protocols/v2/noise-sv2/Cargo.toml
```diff
@@ -12,13 +12,24 @@ homepage = "https://stratumprotocol.org"
...
+rand = {version = "0.8.5", default-features = false }
 aes-gcm = "0.10.2"
 chacha20poly1305 = "0.10.1"
-rand_chacha = "0.3.1"
+rand_chacha = { version = "0.3.1", default-features = false }
 const_sv2 = { version = "^3.0.0", path = "../../../protocols/v2/const-sv2"}
 
+[features]
+default = ["std"]
+std = ["rand/std", "rand/std_rng", "rand_chacha/std", "secp256k1/rand-std"]
```
GitGab19, line 24: Why are we keeping `std` for this crate, but we haven't done the same for the others in the previous PRs?

protocols/v2/noise-sv2/src/handshake.rs
```diff
@@ -480,6 +489,23 @@ mod test {
...
+        let key_pair_2 = TestHandShake::generate_key_with_rng(&mut rand::thread_rng());
+
+        let secret_1 = key_pair_1.secret_bytes();
+        let secret_2 = key_pair_2.secret_bytes();
+
+        let pub_1 = key_pair_1.x_only_public_key();
+        let pub_2 = key_pair_2.x_only_public_key();
+
+        let ecdh_1 = TestHandShake::ecdh(&secret_1, &pub_2.0.serialize());
+        let ecdh_2 = TestHandShake::ecdh(&secret_2, &pub_1.0.serialize());
```
GitGab19, line 504: How can these two be equal?

protocols/v2/noise-sv2/Cargo.toml
Georges760: to not break current API relying on `std`.

I basically added an intermediary API for `no_std` (so it is only a semver MINOR upgrade) and keep the `std` dependant API under the `std` feature

protocols/v2/noise-sv2/src/handshake.rs
Georges760: I don't know, I didn't mess with any functional part of the crate.

This specific test `test_ecdh_with_rng` is a copy/past of the existing `test_ecdh` above, but with the new `no_std` API...

plebhash: by definition ECDH takes two independently different (and without mutual trust) keypairs and generates a shared secret from them

that's what `TestHandShake::ecdh` does, and therefore we expect `ecdh_1` and `ecdh_2` to be identical, because both were generated from the same keypairs (although switching secret vs public on each), but they must both arrive to the same identical shared secret

Issue #1077, opened 2024-07-31
issues lacking CI


plebhash:
# context

there's been a lot of discussion around the need for a sustainable strategy to enforce SRI is not only compliant to specs, but also that every bugfix is accompanied by some CI that enforces that this bug doesn't manifest again

Discussion https://github.com/stratum-mining/stratum/discussions/944 started as a proposal for a new policy where every bugfix is accompanied with a Message Generator test integrated into the CI workflows

We started noticing some bottlenecks on this approach, namely:
- lack of deterministic behavior on some MG tests
- high complexity and steep learning curve for writing MG tests
- missing features on MG preventing full coverage (e.g.: https://github.com/stratum-mining/stratum/issues/1059)

@jbesraa is working on a proposal for a new integration testing framework via #1066, which could potentially present an alternative strategy for CI

the best approach for CI remains to be discussed on the SRI community

# 

during the SRI community dev call of 30/07/24, it was agreed to de-prioritize some issues to unlock the next milestone

these issues were originally conceived as adding MG tests to CI, but with the current ongoing discussions, it was agreed it's best to set them aside for a while

however, it remains critical that those issues are not forgotten, and once there's consensus on the best long-term approach for CI, they should be revisited

so this issue is actually a compilation of all issues that need to be addressed in the future in CI

# issue list

| bug   | CI tracker |
|-------|---------|
| #840  | #955    |
| #901  | #923    |
| #844  | -       |
| #1004 | -       |

PR #1239, opened 2024-10-29
Fix missing transaction in light mempool bug

Shourya742:
Closes: #840

### DIFF
--- a/roles/jd-server/src/lib/job_declarator/mod.rs
+++ b/roles/jd-server/src/lib/job_declarator/mod.rs
@@ -145,7 +145,7 @@ impl JobDeclaratorDownstream {
                     .ok_or(Box::new(JdsError::ImpossibleToReconstructBlock(
                         "Txid found in jds mempool but transactions not present".to_string(),
                     )))?;
-                transactions_list.push(tx);
+                transactions_list.push(tx.0);
             } else {
                 return Err(Box::new(JdsError::ImpossibleToReconstructBlock(
                     "Unknown transaction".to_string(),
--- a/roles/jd-server/config-examples/jds-config-local-example.toml
+++ b/roles/jd-server/config-examples/jds-config-local-example.toml
@@ -28,4 +28,4 @@ core_rpc_pass =  "password"
 # Time interval used for JDS mempool update 
 [mempool_update_interval]
 unit = "secs"
-value = 1
+value = 0.1
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -117,7 +117,6 @@ impl ParseClientJobDeclarationMessages for JobDeclaratorDownstream {
                 .known_transactions
                 .append(&mut known_transactions);
 
-
             if missing_txs.is_empty() {
                 let message_success = DeclareMiningJobSuccess {
                     request_id: message.request_id,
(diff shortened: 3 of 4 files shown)


GitGab19:
@Shourya742 I had a quick look at this solution and I have two major points to discuss:
- You remove all the txs contained in an old job from the mempool --> probably the majority of the txs contained in the new job will be the same --> it doesn't make too much sense to remove and add them again 
- To properly remove old txs, you transform the mempool from txid to short_id (which doesn't scale well as approach and it's already a bottleneck we encounter when we receive a `DeclareMiningJob` message.

Have you explored the idea about not removing the old txs, but simply updating the mempool (as it's already done) and adding the txs contained in the last declared jobs (if they are not already there in the mempool after the update)? 
Does it make sense to you? If not, which could be the issues?

Shourya742:
> @Shourya742 I had a quick look at this solution and I have two major points to discuss:
> 
> * You remove all the txs contained in an old job from the mempool --> probably the majority of the txs contained in the new job will be the same --> it doesn't make too much sense to remove and add them again
> * To properly remove old txs, you transform the mempool from `txid` to `short_id` (which doesn't scale well as approach and it's already a bottleneck we encounter when we receive a `DeclareMiningJob` message.
> 
> Have you explored the idea about not removing the old txs, but simply updating the mempool (as it's already done) and adding the txs contained in the last declared jobs (if they are not already there in the mempool after the update)? Does it make sense to you? If not, which could be the issues?

Here's a refined and grammatically improved version of your message:

---
**Point 1**: We are checking if the `txid` is not present in the new job. Based on this, we decrement the counter value rather than deleting everything from the old job.

**Point 2**: I didn’t know that—let me check on it.

I initially thought keeping the JDS mempool as lean as possible was a requirement. However, it does make sense to let transactions persist in the mempool without a specific logic to remove them once a new DMJ is anticipated. In my testing, while we have logic to remove transactions from the old job, it didn’t result in a significant reduction in the number of transactions deleted from the mempool. But if we’re okay with not keeping the JDS mempool lean, then we can follow the approach you suggested.

GitGab19 (commented):
LGTM.
I would change the `[mempool_update_interval]` parameter to 0.1 secs as we tested together.

roles/jd-server/src/lib/job_declarator/mod.rs
```diff
@@ -145,7 +145,7 @@ impl JobDeclaratorDownstream {
                     .ok_or(Box::new(JdsError::ImpossibleToReconstructBlock(
                         "Txid found in jds mempool but transactions not present".to_string(),
                     )))?;
-                transactions_list.push(tx);
+                transactions_list.push(tx.0);
```
GitGab19, line 148: Why did you change this?

GitGab19 (commented):
LGTM. I just left a minor comment. I would say that we could add an integration test for this scenario in the future. I remember we were tracking these issues to be tested by the new integration tests framework somewhere, but I can't find it anymore (cc @plebhash)

roles/jd-server/src/lib/job_declarator/mod.rs
Shourya742: Since the hashmap structure is now `(txid, (tx, count))` instead of `(txid, tx)`, we used the first element of the inner tuple to reference `tx`.

GitGab19 (approved):

plebhash:
> LGTM. I just left a minor comment. I would say that we could add an integration test for this scenario in the future. I remember we were tracking these issues to be tested by the new integration tests framework somewhere, but I can't find it anymore (cc @plebhash)

it's https://github.com/stratum-mining/stratum/issues/1077

Issue #1234, opened 2024-10-25
Need to fix fragmented patterns on Integration Test starter APIs


plebhash:
Role starter APIs on Integration Test framework is starting to slip into some fragmented patterns, which is resulting in a suboptimal user experience while writing tests.

I noticed this while doing the experimentation described here: https://github.com/stratum-mining/stratum/pull/1226#issuecomment-2438910821

Given that #1226 hasn't yet been merged, I believe it can be used to address `start_jdc` and `start_jds`.

But `start_sniffer`, `start_pool` and `start_template_provider` probably deserve a separate PR (ideally in sync with #1226).

Here's a detailed description of each fragmented pattern:

## arg types

All starter functions take parameters with Socket Address information.
However, each role is following a different standard:
- `start_sniffer` takes `SocketAddr`
- `start_pool` takes `Option<SocketAddr>`
- `start_template_provider` takes `u16` (representing a port)
- `start_jdc` takes `SocketAddr` (not yet merged, WIP under #1226)
- `start_jds` takes `SocketAddr` (not yet merged, WIP under #1226)

## assumptions on roles socket addresses
Not all starter functions are taking Socket Address information for the actual role.

- `start_sniffer` assumes `sniffer_addr` will be provided by the user
- `start_pool`  assumes `pool_addr` will be provided by the user
- `start_template_provider` assumes `tp_addr` will be provided by the user
- `start_jdc` assumes nothing is provided, but calls `common::get_available_address()` inside the function body (not yet merged, WIP under #1226)
- `start_jds` assumes nothing is provided, but calls `common::get_available_address()` inside the function body (not yet merged, WIP under #1226)

To be honest, I feel it makes more sense to always call `common::get_available_address()` inside the function bodies and simply abstract that away from the user (potentially even making `common::get_available_address()` into a private function under `common`).

In other words, I think the direction `start_jdc` and `start_jds` are taking is better than the other starters.

## return types

Starter functions are returning types under different patterns:
- `start_sniffer` returns `Sniffer`
- `start_pool` returns `PoolSv2`
- `start_template_provider` returns `TemplateProvider`
- `start_jdc` returns `SocketAddr` (not yet merged, WIP under #1226)
- `start_jds` returns `SocketAddr` (not yet merged, WIP under #1226)

jbesraa:
Totally agree with the stance here. I was about to address this after completing #1226 as it made it more clear how we should approach this.

PR #1249, opened 2024-11-14
`sv1-mining-device` lib

### LINKED ISSUE #1245: Create library for `mining-device-sv1`
`mining-device-sv1` is currently a binary crate. We should move the code around and create a library as well, same as all the other roles here https://github.com/stratum-mining/stratum/issues/1093.

This is needed in order to be able to migrate some of the tests from MG to integration.

jbesraa:
Resolves #1245


roles/test-utils/sv1-mining-device/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "sv1-mining-device"
+name = "sv1_mining_device"
```
Shourya742, line 2: It makes sense to keep this consistent with other crates in similar roles. I'm not sure if we've published this crate, but if we have, we should deprecate the previous version.

roles/test-utils/sv1-mining-device/Cargo.toml
```diff
@@ -13,6 +13,10 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[lib]
+name = "sv1_mining_device"
```
Shourya742, line 17: The `name` field can be removed since it's already defined in the `[package]` section. However, if we want to keep the original package name and use a different library name, including the `name` field in `[lib]` makes sense.

roles/test-utils/sv1-mining-device/src/client.rs
```diff
@@ -69,8 +67,8 @@ impl Client {
     ///    the information from `sender_share`, it is formatted as a `v1::client_to_server::Submit`
     ///    and then serialized into a json message that is sent to the Upstream via
     ///    `sender_outgoing`.
-    pub(crate) async fn connect(client_id: u32) {
-        let stream = std::sync::Arc::new(TcpStream::connect(ADDR).await.unwrap());
+    pub async fn connect(client_id: u32, upstream_addr: std::net::SocketAddr) {
```
Shourya742, line 70:
```suggestion
    pub async fn connect(client_id: u32, upstream_addr: SocketAddr) {
```
`std::net::SocketAddr` can be added to the import statement above.

roles/test-utils/sv1-mining-device/src/main.rs
```diff
@@ -1,9 +1,16 @@
...
+
 pub(crate) use client::Client;
 
 #[async_std::main]
 async fn main() {
-    Client::connect(80).await
+    const ADDR: &str = "127.0.0.1:34255";
+    Client::connect(
+        80,
+        std::net::SocketAddr::from_str(ADDR).expect("Invalid upstream address"),
```
Shourya742, line 13:
```suggestion
       SocketAddr::from_str(ADDR).expect("Invalid upstream address"),
```
`std::net::SocketAddr` can be added to the import statement above.

Shourya742 (changes requested):
Minor nits and a few follow-ups. Everything else looks good.

roles/Cargo.lock
```diff
@@ -1,12 +1,12 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
plebhash, line 3: this breaks MSRV

plebhash: I highly recommend locking SRI development setup to Rust 1.75.0

that's what I did with my setup when the same thing happened to me

roles/test-utils/sv1-mining-device/Cargo.toml
plebhash: it has not yet been published to `crates.io`, so it's still not too late to change things

plebhash: +1 on consistency for naming patterns

roles/test-utils/sv1-mining-device/Cargo.toml
plebhash: this is true, however we have been following this same pattern with every roles

in case we decide to change this, IMO it should be a separate PR doing this in a consistent way across all roles

arguably cosmetic and therefore low prio IMHO

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +282,28 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
...
+            ),
+            None,
+            None,
+            0,
+            Some(0.1),
+        )
+        .await;
+    });
+    tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+}
```
plebhash, line 309: is this PR about Sv1 CPU miner or [Sv1 + Sv2] CPU miners?

roles/tests-integration/Cargo.toml
```diff
@@ -23,6 +23,8 @@ once_cell = "1.19.0"
 network_helpers_sv2 = { path = "../roles-utils/network-helpers", features =["with_tokio","with_buffer_pool"] }
 pool_sv2 = { path = "../pool" }
 roles_logic_sv2 = { path = "../../protocols/v2/roles-logic-sv2" }
+sv1_mining_device = { path = "../test-utils/sv1-mining-device" }
+mining_device = { path = "../test-utils/mining-device" }
```
plebhash, line 27: is this PR about Sv1 CPU miner or [Sv1 + Sv2] CPU miners?

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +282,28 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
     template_provider.generate_blocks(16);
     template_provider
 }
+
+pub async fn start_sv1_mining_device(pool_address: SocketAddr) {
```
plebhash, line 286: let's try to be consistent with https://github.com/stratum-mining/stratum/issues/1234

plebhash: also, `pool_address` is a legacy naming convention that's a bit misleading IMO

I would call this `upstream_address`

roles/test-utils/sv1-mining-device/Cargo.toml
jbesraa: whats the consistency? I don't think we can use `-` in library names, that's why I changed this..

roles/test-utils/sv1-mining-device/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
-name = "sv1-mining-device"
+name = "sv1_mining_device"
 version = "0.1.0"
 authors = ["The Stratum V2 Developers"]
-edition = "2021"
+edition = "2018"
```
jbesraa, line 5: Changed this to keep it consistent with all of the other roles, we should upgrade all of them to 2021 at some point

roles/tests-integration/tests/common/mod.rs
jbesraa: > let's try to be consistent with #1234

No need to return anything here as the mining device is the lowest in the hierarchy,  it never acts as a server

roles/test-utils/sv1-mining-device/Cargo.toml
plebhash: `_` instead of `-` is in the right direction

I would switch `sv1` to suffix instead of prefix: `mining_device_sv1`

that's what we have on other roles, e.g.: `pool_sv2`, `mining_proxy_sv2`

roles/Cargo.lock
plebhash: @jbesraa you missed this.

plebhash:
overall LGTM

let's fix MSRV and move forward for merging

jbesraa:
Addressed comments and fixed MSRV

plebhash (approved):

jbesraa:
Thanks @plebhash

Issue #1229, opened 2024-10-18
Vardiff currently unsupported pool side


rrybarczyk:
Supporting variable difficulty in the mining proxy and pool is a needed feature for potential users. Previously, #251 was opened but was subsequently closed without addressing it. I do not remember why.

pavlenex:
According to @GitGab19 variable difficulty is included with tProxy. Since this feedback comes from early adopter, we should prfioritize this feature when tackling the pool role.

plebhash:
I remember that @rrybarczyk had previously mentioned the idea of modularizing the vardiff logic, so that it is not statically baked into tProxy

this makes a lot of sense, as ideally we need to be able to easily integrate vardiff as a feature to applications as needed

I wonder whether it makes sense to isolate vardiff into its own crate?

---

as we tackle this, we should be very strategic about test-driven development

the main challenge around this kind of feature is its statistical nature, which makes it very prone to non-deterministic behavior, especially as different platforms (with different computational capacity) are used to emulate the incoming hashrate

we have some valuable lessons learned from MG CI of tProxy, and more insight can be found on https://github.com/stratum-mining/stratum/issues/1208

plebhash:
not directly but somewhat correlated to this topic is the need for modularizing translation logic, which could fit into a potential approach for `cgminer` integration, as it was laid out here: https://github.com/plebhash/cgminer-sv2/issues/1#issuecomment-2185278906

AkulovOleg:
Hi, guys!
We need vardiff too
It will be nice if we can prioritize this feature

pavlenex:
@rrybarczyk Since we have this feature request by two adopters, we should prioritize it as agreed on the last dev call. I'll leave it up to you to decide when you'll have capacity to tackle this one. Not sure if we can get it into 1.2.0 milestone (cc @plebhash @GitGab19)?

plebhash:
> @rrybarczyk Since we have this feature request by two adopters, we should prioritize it as agreed on the last dev call. I'll leave it up to you to decide when you'll have capacity to tackle this one. Not sure if we can get it into 1.2.0 milestone (cc @plebhash @GitGab19)?

I don't have any objections to the idea of starting to work on this right away.

My only remark is that I really feel that we should make sure this functionality is properly tested for the long run.

We still have some moving parts under Integration Test Framework.

So in case Vardiff is finished before ITF, I would push for later having a new dev cycle around Vardiff that adds Integration Tests around it. 

[...]

PR #1254, opened 2024-11-14
Add `tproxy` initializer for Integration Tests

### LINKED ISSUE #1250: Add tproxy initializer in integration tests

jbesraa:
resolves #1250

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -5,10 +5,11 @@ use flate2::read::GzDecoder;
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
 use once_cell::sync::Lazy;
 use pool_sv2::PoolSv2;
+use rand::{thread_rng, Rng};
 use sniffer::Sniffer;
 use std::{
     collections::HashSet,
-    convert::TryFrom,
+    convert::{TryFrom, TryInto},
     env,
     fs::{create_dir_all, File},
     io::{BufReader, Read},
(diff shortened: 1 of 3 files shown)


roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +286,86 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
...
+        translator_sv2::proxy_config::ProxyConfig::new(upstream_conf, downstream_conf, 2, 2, 8);
+    let translator_v2 = translator_sv2::TranslatorSv2::new(config);
+    tokio::spawn(async move {
+        translator_v2.start().await;
+    });
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    listening_address
+}
+
+fn measure_hashrate(duration_secs: u64) -> f64 {
```
plebhash, line 339: interesting approach of dynamically measuring local CPU hashrate

this strategy will save us A LOT of headache and it's quite a game changer in comparison to the static nature of MG tests

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +286,86 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
...
+pub async fn start_sv2_translator(upstream: SocketAddr) -> SocketAddr {
+    let upstream_address = upstream.ip().to_string();
+    let upstream_port = upstream.port();
+    let upstream_authority_pubkey = Secp256k1PublicKey::try_from(
+        "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72".to_string(),
+    )
+    .expect("failed");
+    let listening_address = get_available_address();
+    let listening_port = listening_address.port();
+    let hashrate = measure_hashrate(3) as f32 / 100.0;
```
plebhash, line 299: why divide by 100?

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +286,86 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
...
+        let mut rng = thread_rng();
+        let mut arr = [0u8; 80];
+        rng.fill(&mut arr[..]);
+        arr
+    };
+    let start_time = std::time::Instant::now();
+    let mut hashes: u64 = 0;
+    let duration = std::time::Duration::from_secs(duration_secs);
+
+    let hash = |share: &mut [u8; 80]| -> Target {
```
plebhash, line 350: why return `Target`, it seems to be used nowhere?

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +286,86 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
...
+fn measure_hashrate(duration_secs: u64) -> f64 {
+    let mut share = {
+        let mut rng = thread_rng();
+        let mut arr = [0u8; 80];
+        rng.fill(&mut arr[..]);
+        arr
+    };
+    let start_time = std::time::Instant::now();
+    let mut hashes: u64 = 0;
+    let duration = std::time::Duration::from_secs(duration_secs);
```
plebhash, line 348: I'm fine with this approach but it seems a bit convoluted and confusing to read/understand

I'm not convinced we really need to replicate the bitcoin protocol so closely if all we want is to know the capacity of the system to calculate sha256d (attention to the `d`, it's 2x sha256)

why not simply hash some simple input?

if you REALLY want to closely replicate what happens on bitcoin mining, I would suggest leveraging rust-bitcoin

you can initialize a [`Header`](https://docs.rs/bitcoin/latest/bitcoin/block/struct.Header.html) struct and then call `header.block_hash` to "mine" it.

that's how both `mining-device` and `sv1-mining-device` do it

roles/tests-integration/tests/common/mod.rs
jbesraa: Credit to @GitGab19

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +286,86 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
     template_provider.generate_blocks(16);
     template_provider
 }
+
+pub async fn start_sv2_translator(upstream: SocketAddr) -> SocketAddr {
```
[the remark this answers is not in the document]
jbesraa: I think the approach here is mostly what we want. No need for the end user to call `get_available_address` but the role itself does that and returns it

roles/tests-integration/tests/common/mod.rs
jbesraa: This is done just part of the testing, I think this will change also until we figure out the best numbers here and in `measure_hashrate`

jbesraa:
@plebhash the `measure_hashrate` function and any function it uses is just a copy-pasta from a code @GitGab19 previously wrote. This pull request only uses this functionality without making changes to it yet. I think we should leave those change to be examined in the actual testing pull requests that will follow up.

plebhash:
> @plebhash the `measure_hashrate` function and any function it uses is just a copy-pasta from a code @GitGab19 previously wrote. This pull request only uses this functionality without making changes to it yet. I think we should leave those change to be examined in the actual testing pull requests that will follow up.

I must point out that my comments above are not purely aesthetic.

in it's current form, this `measure_hashrate` function is going to provide a wrong hashrate estimation because it's only calculating `sha256`, not `sha256d`.

https://github.com/stratum-mining/stratum/pull/1254#discussion_r1843847431

roles/tests-integration/tests/common/mod.rs
```diff
@@ -282,3 +286,86 @@ pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
...
+    let start_time = std::time::Instant::now();
+    let mut hashes: u64 = 0;
+    let duration = std::time::Duration::from_secs(duration_secs);
+
+    let hash = |share: &mut [u8; 80]| -> Target {
+        let nonce: [u8; 8] = share[0..8].try_into().unwrap();
+        let mut nonce = u64::from_le_bytes(nonce);
+        nonce += 1;
+        share[0..8].copy_from_slice(&nonce.to_le_bytes());
+        let hash = Sha256::digest(&share).to_vec();
```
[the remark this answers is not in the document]
plebhash: here's an easy way to calculate `sha256d` hashes with `stratum_common`:

```rust

use stratum_common::bitcoin::hashes::{sha256d, Hash, HashEngine};

let input = 0;
let mut engine = sha256d::Hash::engine();
engine.input(&input);
let hashed = sha256d::Hash::from_engine(engine).into_inner();
```

plebhash:
after discussing with @GitGab19 (he did not write this code) we tracked down the root cause of the confusion around `measure_hashrate` function:

https://github.com/stratum-mining/stratum/blob/8c986472b82b61e1b2dd1aacd3279cf680f81efc/roles/translator/src/lib/downstream_sv1/diff_management.rs#L374-L400

if this is going to be used as inspiration for any hashrate measurement algos, it should be noted that this hashrate estimation algo is wrong (should be `sha256d`, not `sha256`)

I'm taking note of this on https://github.com/stratum-mining/stratum/issues/1229

PR #1255, opened 2024-11-14
Update CI benchmarks workflows

GitGab19:
This PR updates the workflows used to run benchmarks on our CI. According to @epompeii, we need to add `permissions:
  checks: write` to solve errors we are getting when merging PRs to `main`. 

This PR sets a fixed version for `bencher` in the other related workflows, while adding a new workflow to automatically archive benches from forks after PRs are merged/closed.


.github/workflows/run-and-track-benchmarks-on-main.yaml
```diff
@@ -5,69 +5,68 @@ on:
...
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: 1.75.0
           override: true
-      - name: Checkout repository
-        uses: actions/checkout@v2
-            
+
       - uses: actions/checkout@v4
```
Shourya742, line 53:
```suggestion
      - name: Checkout repository
        uses: actions/checkout@v4
```
Keeping it like this would keep it consistent to other actions.

.github/workflows/run-and-track-benchmarks-on-main.yaml
```diff
@@ -5,69 +5,68 @@ on:
...
```
Shourya742, line 28:
```suggestion
      - name: Checkout repository
        uses: actions/checkout@v4
```
Keeping it this way would maintain consistency with other actions.

.github/workflows/run-and-track-benchmarks-on-main.yaml
```diff
@@ -5,69 +5,68 @@ on:
     branches:
       - main
 
+permissions:
+  checks: write
+
```
Shourya742, line 10: Why do we need to grant write permissions?

.github/workflows/fork_pr_benchmark_closed.yaml
```diff
@@ -0,0 +1,17 @@
+on:
+  pull_request_target:
+    types: [closed]
+
+jobs:
+  archive_fork_pr_branch:
+    name: Archive closed fork PR branch with Bencher
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
Shourya742, line 10:
```suggestion
      - name: Checkout repository
        uses: actions/checkout@v4
```
Keeping it this way would maintain consistency with other actions.

Shourya742 (changes requested):
Few nits.

.github/workflows/run-and-track-benchmarks-on-main.yaml
GitGab19: https://bencher.dev/docs/how-to/github-actions/#base-branch

It is needed by the github action as indicated here

Shourya742 (approved):
ACK

plebhash (approved):

PR #1258, opened 2024-11-18
`Sniffer` Improvements

### LINKED ISSUE #1253: Sniffer improvements
After working with the `Sniffer` a bit, I think we should allow more flexibility when working with it.
For example, sometimes we don't really want to assert all the messages exchanged between two roles, because we have some messages that can be sent twice or messages that are irrelevant for a test. Example: we want to test that a tproxy can start successfully and exchange all messages until the mining channel is ready, without caring about the `SubmitShare`. 

I suggest we do two more functionalities to the `Sniffer`:
1) A flag to decide how the `drop` function should behave(panic or not on unpopulated messages)
2) Introduce a function to check if a message was exchanged regardless of the order(this is good because sometimes in a `SubmitShare` scenario, upstream next message can be `SubmitShare[STATUS]` or it can be something totally different, while it is handling the received share

jbesraa:
Partially resolves #1253

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -194,8 +194,12 @@ pub fn get_available_address() -> SocketAddr {
     SocketAddr::from(([127, 0, 0, 1], port))
 }
 
-pub async fn start_sniffer(listening_address: SocketAddr, upstream: SocketAddr) -> Sniffer {
-    let sniffer = Sniffer::new(listening_address, upstream).await;
+pub async fn start_sniffer(
+    identifier: String,
+    listening_address: SocketAddr,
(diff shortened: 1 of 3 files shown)


plebhash:
from description of https://github.com/stratum-mining/stratum/issues/1253:

> 2) Introduce a function to check if a message was exchanged regardless of the order(this is good because sometimes in a `SubmitShare` scenario, upstream next message can be `SubmitShare[STATUS]` or it can be something totally different, while it is handling the received share

I don't see this being implemented here.

Is that the reason why this PR says it only partially solves #1253?

will that be done in a follow up PR?

other than that, LGTM

plebhash (approved):

PR #1262, opened 2024-11-21
Duplicate tproxy MG test in integration tests

### LINKED ISSUE #1208: CI migration: `translation-proxy` MG to Integration Test
The [`translation-proxy`](https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/translation-proxy/translation-proxy.json) MG test displays undeterministic behavior on CI.

The MG test does the following:

- launch a tProxy
- launch a SV1 CPU miner
- injects messages to tProxy while mocking an upstream

tProxy will establish a channel with the mocked upstream with some predefined difficulty target (via `OpenExtendedMiningChannel`)

once tProxy receives a `NewExtendedMiningJob`, it forwards it to the CPU miner (as Sv1 messages)

the lack of determinism comes from the fact that two different things could happen after `NewExtendedMiningJob` is sent:

- maybe CPU miner finds a share that is valid under the Channel difficulty target, therefore tProxy sends `SubmitSharesExtended` to its upstream
- maybe tProxy finished its difficulty adjustment (by looking at SV1 shares submitted by the CPU miner), and it needs to change the difficulty target of the Channel via `UpdateChannel`

it's impossible for us to know beforehand the `nominal_hash_rate` of the CPU miner (since it depends on the available system resources)

as a consequence, we often see [the following error on the current MG CI](https://github.com/stratum-mining/stratum/actions/runs/11279363283/job/31369908317?pr=1179#step:4:2263):

```
WRONG MESSAGE TYPE expected: 27 received: 22
```

- 27 is decimal for `0x1b`, which means `SubmitSharesExtended`
- 22 is decimal for `0x16`, which means `UpdateChannel`

(ref: [Message Types Spec](https://github.com/stratum-mining/sv2-spec/blob/main/08-Message-Types.md))

that is because the MG test is fixed in a way that after sending `NewExtendedMiningJob`, it expects ONLY `SubmitSharesExtended`, so whenever the difficulty adjustment finishes BEFORE a valid share is found, CI breaks

https://github.com/stratum-mining/stratum/issues/1059 was meant to add this kind of functionality to MG, which would allow it to assert that EITHER one specific message OR another arrived. Unfortunately, the trade-offs on the engineering effort for implementing this feature on MG didn't seem appealing.

hopefully the Integration Test framework will give us enough flexibility to avoid this kind of problem

jbesraa:
resolves #1208

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -395,7 +395,7 @@ pub async fn start_sv2_translator(upstream: SocketAddr) -> SocketAddr {
     .expect("failed");
     let listening_address = get_available_address();
     let listening_port = listening_address.port();
-    let hashrate = measure_hashrate(3) as f32 / 100.0;
+    let hashrate = measure_hashrate(1) as f32 / 100.0;
     let min_individual_miner_hashrate = hashrate;
     let shares_per_minute = 60.0;
     let channel_diff_update_interval = 60;
--- a/roles/tests-integration/tests/translator_integration.rs
+++ b/roles/tests-integration/tests/translator_integration.rs
@@ -1,0 +1,57 @@
+mod common;
+
+use common::MessageDirection;
+use const_sv2::{MESSAGE_TYPE_SETUP_CONNECTION, MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED};
+use roles_logic_sv2::parsers::{CommonMessages, Mining, PoolMessages};
+
+// This test runs an sv2 translator between an sv1 mining device and a pool. the connection between
+// the translator and the pool is intercepted by a sniffer. The test checks if the translator and
+// the pool exchange the correct messages upon connection. And that the miner is able to submit
+// shares.
+#[tokio::test]
+async fn translation_proxy() {
+    let pool_translator_sniffer_addr = common::get_available_address();
+    let tp_addr = common::get_available_address();
+    let pool_addr = common::get_available_address();
+    let pool_translator_sniffer = common::start_sniffer(
+        "0".to_string(),
+        pool_translator_sniffer_addr,
+        pool_addr,
+        false,
+        None,
+    )
+    .await;
+    let _tp = common::start_template_provider(tp_addr.port()).await;
+    let _pool = common::start_pool(Some(pool_addr), Some(tp_addr)).await;
+    let tproxy_addr = common::start_sv2_translator(pool_translator_sniffer_addr).await;
+    let _mining_device = common::start_mining_device_sv1(tproxy_addr).await;
+    pool_translator_sniffer
+        .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
+        .await;
+    assert_common_message!(
+        &pool_translator_sniffer.next_message_from_downstream(),
+        SetupConnection
+    );
+    assert_common_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        SetupConnectionSuccess
+    );
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_downstream(),
+        OpenExtendedMiningChannel
+    );
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        OpenExtendedMiningChannelSuccess
+    );
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        NewExtendedMiningJob
+    );
+    pool_translator_sniffer
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED,
+        )
+        .await;
+}


roles/tests-integration/tests/pool_integration.rs
```diff
@@ -53,3 +56,55 @@ async fn success_pool_template_provider_connection() {
     assert_tp_message!(&sniffer.next_message_from_upstream(), NewTemplate);
     assert_tp_message!(sniffer.next_message_from_upstream(), SetNewPrevHash);
 }
+
+// covers
+// https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/translation-proxy/translation-proxy.json
```
GitGab19, line 61: Can you add the explanation here like you did for the other test?

roles/tests-integration/tests/common/mod.rs
```diff
@@ -396,9 +396,9 @@ pub async fn start_sv2_translator(upstream: SocketAddr) -> SocketAddr {
     .expect("failed");
     let listening_address = get_available_address();
     let listening_port = listening_address.port();
-    let hashrate = measure_hashrate(3) as f32 / 100.0;
+    let hashrate = measure_hashrate(3) as f32 / 20.0;
```
GitGab19, line 399: I did test many times, and some of them it was failing. So I tried to change this in:
```suggestion
    let hashrate = measure_hashrate(1) as f32 / 20.0;
```
With that, it never failed again. Don't know if it's random or not btw :)

GitGab19 (commented):
I tested it many times, and some of them I got this error, I don't know if you already noticed it:
```
running 2 tests
thread 'translation_proxy' panicked at tests-integration/tests/common/mod.rs:162:14:
called `Result::unwrap()` on an `Err` value: JsonRpc(Transport(SocketError(Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" })))
test translation_proxy ... FAILED
thread 'success_pool_template_provider_connection' panicked at tests-integration/tests/common/mod.rs:162:14:
called `Result::unwrap()` on an `Err` value: JsonRpc(Transport(SocketError(Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" })))
test success_pool_template_provider_connection ... FAILED

failures:

failures:
    success_pool_template_provider_connection
    translation_proxy

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 37.93s

error: test failed, to rerun pass `-p integration-test --test pool_integration`
```

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -94,3 +95,61 @@ async fn test_sniffer_interrupter() {
     assert_common_message!(&sniffer.next_message_from_downstream(), SetupConnection);
     assert_common_message!(&sniffer.next_message_from_upstream(), SetupConnectionError);
 }
+
+// covers
+// https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/translation-proxy/translation-proxy.json
+#[tokio::test]
+async fn translation_proxy() {
```
plebhash, line 102: I don't think it makes sense to add this to a file called `pool_integration.rs`

also, as we already discussed in the past with @Shourya742 , by having multiple tests in the same file, we end up executing them in parallel

this can cause problems with CPU miners

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -94,3 +95,61 @@ async fn test_sniffer_interrupter() {
...
+    let pool_jdc_sniffer = common::start_sniffer(
+        "0".to_string(),
+        pool_jdc_sniffer_addr,
+        pool_addr,
+        false,
+        None,
+    )
+    .await;
+    let _tp = common::start_template_provider(tp_addr.port()).await;
+    let _pool_1 = common::start_pool(Some(pool_addr), Some(tp_addr)).await;
```
plebhash, line 116: why is this called `_pool_1`?

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -94,3 +95,61 @@ async fn test_sniffer_interrupter() {
...
+    let jds_addr = common::start_jds(tp_addr).await;
+    let jdc_addr = common::start_jdc(pool_jdc_sniffer_addr, tp_addr, jds_addr).await;
```
plebhash, line 118: why do we have JDC + JDS on this test?

this should be JD-less, tProxy should simply connect to Pool

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -94,3 +95,61 @@ async fn test_sniffer_interrupter() {
...
+        SetCustomMiningJob
+    );
+    assert_mining_message!(
+        &pool_jdc_sniffer.next_message_from_upstream(),
+        SetNewPrevHash
+    );
+    assert_mining_message!(
+        &pool_jdc_sniffer.next_message_from_downstream(),
+        SubmitSharesExtended
+    );
```
plebhash, line 154: I don't understand these assertions

what's the point of asserting messages between JDC and Pool?

the original MG test doesn't even have JDC + JDS

roles/tests-integration/tests/pool_integration.rs
plebhash: most importantly, as we already discussed in the past with @Shourya742 , by having multiple tests in the same file, we end up executing all of them in parallel

this can cause problems with CPU miners

roles/tests-integration/tests/common/mod.rs
plebhash: tbh I still don't understand why we are dividing the measured hashrate

that will make tProxy set a very low difficulty and make CPU miner submit way more shares than necessary

plebhash:
currently blocked by:
- https://github.com/stratum-mining/stratum/issues/1288
- https://github.com/stratum-mining/stratum/issues/1287

jbesraa:
@plebhash @GitGab19 This is ready for another review. Thanks for the help here.

roles/tests-integration/tests/common/mod.rs
```diff
@@ -395,7 +395,7 @@ pub async fn start_sv2_translator(upstream: SocketAddr) -> SocketAddr {
     .expect("failed");
     let listening_address = get_available_address();
     let listening_port = listening_address.port();
-    let hashrate = measure_hashrate(3) as f32 / 100.0;
+    let hashrate = measure_hashrate(1) as f32 / 100.0;
```
jbesraa, line 398: This change here makes the test run faster

GitGab19 (approved):

plebhash (approved):

Issue #1014, opened 2024-06-26
Need Rust docs on `protocols::v2::roles-logic-sv2`


rrybarczyk:
# Background
Currently, the `protocols::v2::roles-logic-sv2` logic does not contain any in-line rust documentation, resulting in no documentation on its [crates.io](https://crates.io/crates/roles_logic_sv2) site. 

> Note: While this issue addresses the need for doc comments, it will also serve as a means to identify code debt that needs to be cleaned up and addressed in future issues.

# Problem
When a user navigates to this crate crates.io, there is no useful information for the user. This impedes understanding and adoption.

# Solution
Navigate through each crate in `protocols::v2::roles-logic-sv2` and add:
- Rust Docs comments (following the formatting structure in [this on-going discussion](https://github.com/stratum-mining/stratum/discussions/970)), to be displayed in `docs.rs`
- a descriptive `README.md` file to be displayed in `crates.io`

While this work is being performed, specific code debt should be looked for and recorded to turn into future issues

PR #1263, opened 2024-11-22
docs `roles_logic_sv2`

plebhash:
close https://github.com/stratum-mining/stratum/issues/1014

### DIFF
--- a/protocols/v2/roles-logic-sv2/README.md
+++ b/protocols/v2/roles-logic-sv2/README.md
@@ -1,0 +1,34 @@
+# `roles_logic_sv2`
+
+[![crates.io](https://img.shields.io/crates/v/roles_logic_sv2.svg)](https://crates.io/crates/roles_logic_sv2)
+[![docs.rs](https://docs.rs/roles_logic_sv2/badge.svg)](https://docs.rs/roles_logic_sv2)
+[![rustc+](https://img.shields.io/badge/rustc-1.75.0%2B-lightgrey.svg)](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html)
+[![license](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg)](https://github.com/stratum-mining/stratum/blob/main/LICENSE.md)
+[![codecov](https://codecov.io/gh/stratum-mining/stratum/branch/main/graph/badge.svg?flag=roles_logic_sv2-coverage)](https://codecov.io/gh/stratum-mining/stratum)
+
+`roles_logic_sv2` provides the core logic and utilities for implementing roles in the Stratum V2 (Sv2) protocol, such as miners, pools, and proxies. It abstracts message handling, channel management, job creation, and routing logic, enabling efficient and secure communication across upstream and downstream connections.
+
+## Main Components
+
+- **Channel Logic**: Manages the lifecycle and settings of communication channels (standard, extended, and group ones) between roles.
+- **Handlers**: Provides traits for handling logic of Sv2 protocol messages.
+- **Job Management**: Facilitates the creation, validation, and dispatching of mining jobs.
+- **Parsers**: Handles serialization and deserialization of Sv2 messages via [`binary_sv2`](https://docs.rs/binary_sv2/latest/binary_sv2/index.html).
+- **Routing Logic**: Implements message routing and downstream/upstream selector utilities. Useful for advanced proxy implementations with multiplexing of Standard Channels across different upstreams.
+- **Utilities**: Provides helpers for safe mutex locking, mining-specific calculations, and more.
+
+## Usage
+
+To include this crate in your project, run:
+
+```bash
+cargo add roles_logic_sv2
+```
+
+This crate can be built with the following feature flags:
+
+- `with_serde`: Enables serialization and deserialization support using the serde library. This feature flag also activates the with_serde feature for dependent crates such as `binary_sv2`, `common_messages_sv2`, `template_distribution_sv2`, `job_declaration_sv2`, and `mining_sv2`.
+  Note that this feature flag is only used for the Message Generator, and deprecated
+  for any other kind of usage. It will likely be fully deprecated in the future.
+- `prop_test`: Enables property-based testing features for template distribution logic, leveraging dependencies' testing capabilities such as `template_distribution_sv2` crate.
+- `disable_nopanic`: Disables the nopanic logic in scenarios where code coverage tools might conflict with it.
--- a/protocols/v2/roles-logic-sv2/src/common_properties.rs
+++ b/protocols/v2/roles-logic-sv2/src/common_properties.rs
@@ -18,32 +18,28 @@ use std::{collections::HashMap, fmt::Debug as D};
 pub struct CommonDownstreamData {
     /// Header-only mining flag.
     ///
-    /// Enables the processing of block headers only, leaving transaction management and template
-    /// construction to the upstream node. It reduces the amount of data processed and transmitted
-    /// by the downstream node, making it useful when bandwidth is limited and transmitting full
-    /// block templates is costly.
+    /// Enables the processing of standard jobs only, leaving merkle root manipulation to the
+    /// upstream node.
     ///
-    /// - `true`: The downstream node only processes block headers, relying on the upstream for
-    ///   transaction management.
-    /// - `false`: The downstream node handles full blocks.
+    /// - `true`: The downstream node only processes standard jobs, relying on the upstream for
+    ///   merkle root manipulation.
+    /// - `false`: The downstream node handles extended jobs and merkle root manipulation.
     pub header_only: bool,
 
     /// Work selection flag.
     ///
-    /// Enables the selection of which transactions or templates the node will work on. It allows
-    /// for more autonomy for downstream nodes to define specific version bits to roll, adjust the
-    /// difficulty target, apply `timestamp` range constraints, etc.
+    /// Enables the selection of which transactions or templates the node will work on.
     ///
-    /// - `true`: The downstream node can modify or customize the mining job, such as choosing
-    ///   specific transactions to include in a block template.
-    /// - `false`: The downstream node strictly follows the work provided by the upstream, such as
-    ///   pre-constructed templates from a pool.
+    /// - `true`: The downstream node works on a custom block template, using the Job Declaration
+    ///   Protocol.
+    /// - `false`: The downstream node strictly follows the work provided by the upstream, based on
+    ///   pre-constructed templates from the upstream (e.g. Pool).
     pub work_selection: bool,
 
     /// Version rolling flag.
     ///
     /// Enables rolling the block header version bits which allows for more unique hash generation
-    /// on the same block template by expanding the nonce-space. Used when other fields (e.g.
+    /// on the same mining job by expanding the nonce-space. Used when other fields (e.g.
     /// `nonce` or `extranonce`) are fully exhausted.
     ///
     /// - `true`: The downstream supports version rolling and can modify specific bits in the
@@ -60,7 +56,7 @@ pub struct CommonDownstreamData {
 /// range, and flag settings.
 #[derive(Debug, Copy, Clone)]
 pub struct PairSettings {
-    /// Role the settings apply to.
+    /// Protocol the settings apply to.
     pub protocol: Protocol,
 
     /// Minimum protocol version the node supports.
@@ -70,7 +66,7 @@ pub struct PairSettings {
     pub max_v: u16,
 
     /// Flags indicating optional protocol features the node supports (e.g. header-only mining,
-    /// Noise protocol support, job configuration parameters, etc.). Each protocol field as its own
+    /// work selection, version-rolling, etc.). Each protocol field as its own
     /// flags.
     pub flags: u32,
 }
@@ -122,8 +118,9 @@ pub enum UpstreamChannel {
 
     /// A grouped channel for aggregated mining.
     ///
-    /// Aggregates mining work for multiple devices or clients under a single channel, enabling the
-    /// upstream to manage work distribution and result aggregation for an entire group of miners.
+    /// Aggregates mining work for multiple standard channels under a single group channel,
+    /// enabling the upstream to manage work distribution and result aggregation for an entire
+    /// group of channels.
     ///
     /// Typically used by a mining proxy managing multiple downstream miners.
     Group,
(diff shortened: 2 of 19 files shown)


plebhash:
we need to add a `README.md`

protocols/v2/roles-logic-sv2/README.md
```diff
@@ -0,0 +1,34 @@
...
+[![rustc+](https://img.shields.io/badge/rustc-1.75.0%2B-lightgrey.svg)](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html)
+[![license](https://img.shields.io/badge/license-MIT%2FApache--2.0-blue.svg)](https://github.com/stratum-mining/stratum/blob/main/LICENSE.md)
+[![codecov](https://codecov.io/gh/stratum-mining/stratum/branch/main/graph/badge.svg?flag=roles_logic_sv2-coverage)](https://codecov.io/gh/stratum-mining/stratum)
+
+`roles_logic_sv2` provides the core logic and utilities for implementing roles in the Stratum V2 (Sv2) protocol, such as miners, pools, and proxies. It abstracts message handling, channel management, job creation, and routing logic, enabling efficient and secure communication across upstream and downstream connections.
+
+## Main Components
+
+- **Channel Logic**: Manages the lifecycle and settings of communication channels (standard, extended, and group ones) between roles.
+- **Handlers**: Centralizes the processing and routing logic of Sv2 protocol messages.
```
plebhash, line 14:
```suggestion
- **Handlers**: Provides traits for handling logic of Sv2 protocol messages.
```

protocols/v2/roles-logic-sv2/README.md
```diff
@@ -0,0 +1,34 @@
...
+- **Job Management**: Facilitates the creation, validation, and dispatching of mining jobs.
+- **Parsers**: Handles serialization and deserialization of Sv2 messages.
```
plebhash, line 16:
```suggestion
- **Parsers**: Handles serialization and deserialization of Sv2 messages via [`binary_sv2`](https://docs.rs/binary_sv2/latest/binary_sv2/index.html).
```

protocols/v2/roles-logic-sv2/README.md
```diff
@@ -0,0 +1,34 @@
...
+- **Routing Logic**: Implements message routing and downstream/upstream selector utilities.
```
plebhash, line 17:
```suggestion
- **Routing Logic**: Implements message routing and downstream/upstream selector utilities. Useful for advanced proxy implementations with multiplexing of Standard Channels across different upstreams.
```

plebhash:
> ACK. We're not adding documentation for some structs, enums, and methods right now because they're likely to change in the near future. This crate needs significant refactoring, which will involve architectural changes. While the current documentation isn't complete, it provides enough context to understand the crate. We'll add examples once the refactoring is finished.

thanks!

I made one last round of modifications (based on `#![deny(missing_docs)]` 🙈) and I made sure to cover every single pub function/method.

The ones still missing are:
- struct members
- enum variants

all of them are obvious and meaning is self-evident... writing them would be extremely laborious and chatGPT proved a bit error-prone, so IMO it's fine to leave them like this

protocols/v2/roles-logic-sv2/src/parsers.rs
```diff
@@ -140,6 +148,7 @@ pub enum JobDeclaration<'a> {
     SubmitSolution(SubmitSolutionJd<'a>),
 }
 
+/// A parser of messages of Mining subprotocol, to be used for parsing raw messages
```
rrybarczyk, line 151:
```suggestion
  /// Mining subprotocol messages: categorization, encapsulation, and parsing.
  ///
  /// Encapsulates mining-related Sv2 protocol messages, providing both a structured representation
  /// of parsed messages and an abstraction for communication between mining-related roles. These
  /// messages are essential for managing mining channels, distributing jobs, and processing shares.
  ///
  /// ## Purpose
  /// - **Parsing Raw Messages**:
  ///   - Converts raw binary Sv2 mining subprotocol messages into strongly-typed Rust
  ///     representations.
  ///   - Simplifies deserialization by mapping raw data directly to the appropriate enum variant.
  ///   - Once parsed, the [`Mining`] enum provides a structured interface that can be passed through
  ///     routing and processing layers in roles like proxies or pools.
  /// - **Encapsulation**:
  ///   - Groups mining-related messages into a unified type, abstracting away low-level subprotocol
  ///     details and making it easier to interact with Sv2 protocol messages.
  /// - **Facilitating Modular Handling**:
  ///   - Categorizes mining messages under a single enum, enabling roles (e.g., proxies or pools) to
  ///     route and process messages more efficiently using pattern matching and centralized logic.
  /// - **Bridging Parsed Messages and Role Logic**:
  ///   - Acts as a bridge between parsed subprotocol messages and role-specific logic, providing a
  ///     unified interface for handling mining-related communication. This reduces complexity and
  ///     ensures consistency across roles.
```

protocols/v2/roles-logic-sv2/src/parsers.rs
rrybarczyk, line 1: After reviewing this module, I think its responsibilities have evolved beyond just parsing, and the current documentation does not fully reflect this expanded role (which is understandable given the module name).

### Parsing Aspect
Parsing is certainly a key function of this module:
- It converts raw/serialized Sv2 protocol data into structured types via `TryFrom` implementations on enums like `Mining` and `PoolMessages`.
- These enums encapsulate parsed protocol messages, providing strongly-typed representations of subprotocol messages like `SubmitSharesExtended`.

### Expanded Role Beyond Parsing
However, the module also takes on broader responsibilities:
1. Encapsulation of parsed data:
   - Enums like `Mining` and `PoolMessages` abstract away low-level subprotocol details, making it easier for roles to interact with protocol messages.
   - For example, `Mining::SubmitSharesExtended` wraps the `subprotocols::mining_sv2::SubmitSharesExtended` message, clearly marking it as mining-related.
2. Categorization for routing:
   - The `Mining` enum is categorized under `PoolMessages`, which represents all messages a pool could send or receive (theoretically).
   - This categorization simplifies routing and modularizes handling by grouping related messages into enums.
3. Message preparation:
   - In a role application, the encapsulated message is serialized and wrapped into transportable frames (`StdFrame`, `EitherFrame`) for transmission.
   - The module ensures messages are properly typed and structured, facilitating seamless integration into the message flow.

### Looking at its usage in the translator role
In the translator's [`handle_submit` method](https://github.com/stratum-mining/stratum/blob/e8d76d68642ea28aa48a2da7e41fb4470bbe2681/roles/translator/src/lib/upstream_sv2/upstream.rs#L480), we can see how this module’s types enable streamlined handling of a `SubmitSharesExtended` message:

1. A `SubmitSharesExtended` message is received from a downstream via the `receiver` channel:
   ```rust
   let mut sv2_submit: SubmitSharesExtended = handle_result!(tx_status, receiver.recv().await);
   ```

2. The `channel_id` and `job_id` fields are updated to reflect the Translator’s context:
   ```rust
   sv2_submit.channel_id = ...; // Retrieves channel ID
   sv2_submit.job_id = ...;     // Retrieves job ID
   ```

3. The updated message is wrapped in the `Mining::SubmitSharesExtended` enum:
   ```rust
   let message = Message::Mining(
       roles_logic_sv2::parsers::Mining::SubmitSharesExtended(sv2_submit),
   );
   ```

4. The `Mining` enum is further categorized under the `PoolMessages` enum (aliased as `Message`):
   ```rust
   pub type Message = PoolMessages<'static>;
   ```

5. The `Message::Mining` instance is serialized into a `StdFrame` and then wrapped in an `EitherFrame` for transmission:
   ```rust
   let frame: StdFrame = handle_result!(tx_status, message.try_into());
   let frame: EitherFrame = frame.into();
   tx_frame.send(frame).await;
   ```

This process demonstrates that `parsers.rs` plays a role in:
1. Abstracting subprotocol messages.
2. Categorizing messages for structured routing.
3. Preparing messages for transmission.

### Proposed doc updates
The current documentation should be updated to reflect the module’s full range of responsibilities:

1. Data abstraction: Wraps parsed subprotocol messages (e.g., `SubmitSharesExtended`) in meaningful enums (`Mining`, `PoolMessages`), abstracting low-level details.
2. Routing and categorization: Groups related messages into broader categories (e.g., `Mining`, `CommonMessages`), facilitating modular handling and routing.
3. Message flow integration: Ensures messages are properly typed and serialized, making them ready for processing or transmission.

While parsing remains a key aspect of this module, its expanded role suggests that a name like `message_types.rs` or `protocol_messages.rs` might better reflect its responsibilities. This could be considered during the refactor phase, depending on how the module evolves.

Check out the review comment I left on the `Mining` enum for an example of what I think should be included for the enum documentation and lmk what you think.

rrybarczyk (changes requested):
- `common_properties` - added comments
- `selectors` 
  - We decided not to do "Arguments" and "Returns" sections as it is redundant to the function definition. And if more info is needed on those things, we can express. So I modified to conform to those standards.
  - I moved the top level module doc comments under the "Traits" and "Details" sections to belong with their traits/struct definitions.
- `parsers.rs` - left a detailed comment on the mod and an example for the `Mining` enum on what I think the docs should look like, given my understanding of the module
- For each module, we need to make sure we have a `# Header ` name at the top.
- For the` lib.rs`, we need to follow the doc convention in [`codec-sv2/src/lib.rs`](https://github.com/stratum-mining/stratum/blob/e8d76d68642ea28aa48a2da7e41fb4470bbe2681/protocols/v2/codec-sv2/src/lib.rs#L1)
- We need to be sure that all sentences start with a capital letter and end in a period.
- I have not gotten to the modules in the `handlers` module. It looks like there is doc coverage there, but we need to remove the `# Parameters` and `# Returns` as we agreed not to have these as they are redundant to the function declaration. However, if there are specific important things to point out about these parameters/return args, they can be explicitly stated in the doc comments.
- In the top level modes, we don't need to re-list out the traits, structs, types, etc (see `routing_logic.rs` as an example). Those should speak on their own with their own documentation. Rather the purpose/usage of the module should be expressed in the top level comments.
- I still think examples would be important to have here.
- I will keep going with the review, but wanted to give the feedback I currently have back to you.

GitGab19:
@rrybarczyk I added a commit to complete this PR with all your suggestions, fixing some inaccuracies here and there.

To summarize, I have:
- added a proper header in every module
- modified the `lib.rs` following the one on `codec_sv2`
- removed useless components descriptions on some modules (traits, enums, etc)

Everything is ready for a final review now.

protocols/v2/roles-logic-sv2/src/common_properties.rs
```diff
@@ -8,32 +13,82 @@ use mining_sv2::{Extranonce, Target};
...
+    /// Header-only mining flag.
+    ///
+    /// Enables the processing of block headers only, leaving transaction management and template
+    /// construction to the upstream node. It reduces the amount of data processed and transmitted
+    /// by the downstream node, making it useful when bandwidth is limited and transmitting full
+    /// block templates is costly.
+    ///
+    /// - `true`: The downstream node only processes block headers, relying on the upstream for
+    ///   transaction management.
+    /// - `false`: The downstream node handles full blocks.
```
plebhash: @rrybarczyk this wording seems to be based on some assumptions that should be clarified:

header only mining (aka HOM) is only related to not using the extranonce, while this description seems to imply it's something related to JD

the selection of the transaction set always happens on the upstream node (pretty much the entire merkle path minus coinbase)

what happens depending on this boolean is one of the following:
- standard job: extranonce is not rolled, and coinbase has a unique extranonce_prefix (unit of search space), which results in a unique (but fixed) merkle root
- extendend job: extranonce is rolled, and coinbase also has unique extranonce_prefix

when shares are submitted:
- standard shares: no info about extranonce is reported, as merkle root is already cached upstream
- extended shares: extranonce is reported so upstream can reconstruct merkle root and validate job

just taking note as I feel that having clarity on this distinction is an important conceptual foundation of this crate

plebhash: this description was re-written on https://github.com/stratum-mining/stratum/commit/63b2c2be30a7686e915727a3b96c68104646488c

protocols/v2/roles-logic-sv2/src/common_properties.rs
```diff
@@ -8,32 +13,82 @@ use mining_sv2::{Extranonce, Target};
...
+    /// Work selection flag.
+    ///
+    /// Enables the selection of which transactions or templates the node will work on. It allows
+    /// for more autonomy for downstream nodes to define specific version bits to roll, adjust the
+    /// difficulty target, apply `timestamp` range constraints, etc.
+    ///
+    /// - `true`: The downstream node can modify or customize the mining job, such as choosing
+    ///   specific transactions to include in a block template.
+    /// - `false`: The downstream node strictly follows the work provided by the upstream, such as
+    ///   pre-constructed templates from a pool.
```
plebhash: this one also should be clarified:

this flag is kind of a "feature gate" for whether `SetCustomMiningJob` will be allowed to flow within this connection or not (set via `REQUIRES_WORK_SELECTION` bit flag when the connection is established)

the description is conflating it with version rolling and other extra features have no relationship

plebhash: also re-written and simplified in https://github.com/stratum-mining/stratum/commit/63b2c2be30a7686e915727a3b96c68104646488c

GitGab19:
> @plebhash rn roles-logic-sv2 is not even spec complaints I would merge my PRs before documenting it.

What is not spec-compliant?

This PR is the last step of our efforts in documenting all crates (see #845) before refactoring them. And it's pretty much ready to be merged since a while now. We are just waiting for a final ACK from @rrybarczyk.

plebhash:
> 2. we removed shortId from the spec, [this pr](https://github.com/stratum-mining/stratum/pull/1301/files)

this PR is only changing SRI, we haven't enacted this change in the spec repo yet

I'm about to start working on a PR that will introduce the different JD spec changes, but ideally we should merge this first: https://github.com/stratum-mining/sv2-spec/pull/104

@Fi3 can you ACK it?

plebhash:
> @plebhash rn roles-logic-sv2 is not even spec complaints I would merge my PRs before documenting it.

NACK

spec changes have not yet been enacted on `sv2-spec` repo

we only have social consensus, so let's not block this PR because of that

rrybarczyk (approved):

PR #1271, opened 2024-12-05
Add Rust Docs CI

### LINKED ISSUE #1270: Rust Docs CI
as @Shourya742 pointed out, there's a very interesting rustdoc lint macro available: [`missing_docs`](https://doc.rust-lang.org/rustdoc/lints.html#missing_docs)

by combining it with the `deny` filter, we have a powerful filter that if integrated into SRI CI, will make sure we never let any public API go without docs ever again.

this will give us confidence to move forward from #845 knowing that we didn't let anything go without notice

some files already contain this, but we need a PR that adds this macro to every file that still doesn't have it.

```
#![deny(missing_docs)]
```

this PR should also add a new CI workflow that enforces running `cargo doc --all-features` recursively across all crates in the repo

plebhash:
close #1270

this PR does two things:
- enforce `#![deny(missing_docs)]` on every `*.rs` file that still doesn't have it
- enforce CI does `cargo doc --all-features` on every crate of the repo (via shell recursion)

we should aim to merge this PR by the end of Milestone 1.2.0

this will be the final toothcomb that will give us confidence to move forward from #845 knowing that we didn't let anything go without notice

### DIFF
--- a/.github/workflows/rust-docs.yaml
+++ b/.github/workflows/rust-docs.yaml
@@ -1,0 +1,35 @@
+name: Validate Rust Docs
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  generate-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+
+      - name: Find and Document All Crates
+        run: |
+          # Fail on any error
+          set -e
+          
+          # Find all Cargo.toml files (crates) and run `cargo doc` for each
+          find . -name "Cargo.toml" | while read -r crate; do
+            crate_dir=$(dirname "$crate")
+            echo "Generating documentation for $crate_dir..."
+            (
+              cd "$crate_dir" || exit 1
+              cargo doc --all-features --no-deps
+            )
+          done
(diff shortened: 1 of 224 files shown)


jbesraa (commented):
A couple of general notes:

1) I think at this point we should just have a single `rust.yml` file in the CI and do all the rust-related stuff, like building/testing/docs/msrv.. It is getting crowded here because we add a new file for each action
2) I think it should be enough to add the `deny` macro to `main`/`lib`, it should be reflected across the crate?

plebhash:
> A couple of general notes:
> 
> 1) I think at this point we should just have a single `rust.yml` file in the CI and do all the rust-related stuff, like building/testing/docs/msrv.. It is getting crowded here because we add a new file for each action
> 2) I think it should be enough to add the `deny` macro to `main`/`lib`, it should be reflected across the crate? 


1 would be good on a separate PR

2 is interesting feedback, but the engineering effort is higher, because we would need human eyes going with a fine comb over each module

do you feel strongly about this?

the current approach is naive and not elegant, but it gets the job done without significant negative consequences

jbesraa:
for (1) https://github.com/stratum-mining/stratum/issues/1272

with (2), what is the difference when you include the flag in every files?

plebhash:
closing this

after trying to implement all the changes on `roles_logic_sv2` I quickly realized that the engineering effort here is not worth it

introducing the `deny(missing_docs)` macro, A LOT of new errors come up

to make the CI happy, we would need to spend too much human hours adding redundant docs

---

I'm not against having this CI... if someone is opinionated and wants to spend time on making this PR ready, feel free to spend time on it (e.g.: @Shourya742 )

I just don't want to spend my time here, I feel there's higher prio things to be done right now

PR #1290, opened 2024-12-11
Handle test TemplateProvider drop error

### LINKED ISSUE #1288: need to fix TP dropping logic on ITF
when TP is dropped, sometimes we get a panic which causes Integration Tests to randomly fail:

```
thread 'translation_proxy' panicked at tests-integration/tests/common/mod.rs:149:45:
called `Result::unwrap()` on an `Err` value: JsonRpc(Transport(SocketError(Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })))
test translation_proxy ... FAILED

```

apparently it's coming from here:
https://github.com/stratum-mining/stratum/blob/5bcd54d6eee28a55fcc20c2aaaa707e659cf2136/roles/tests-integration/tests/common/mod.rs#L148-L150

jbesraa:
Resolves #1288 

As the current error we are seeing is `ConnectionRefused`, it might be that bitcoind is already down. Lets fail gracefully and keep an eye on this if it keeps poping

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -146,7 +146,7 @@ impl TemplateProvider {
     }
 
     fn stop(&self) {
-        let _ = self.bitcoind.client.stop().unwrap();
+        self.bitcoind.stop().expect("Failed to stop bitcoind");
     }
 
     fn generate_blocks(&self, n: u64) {


plebhash:
hey @jbesraa thanks for this

I was about to submit a PR with a slightly different approach to fix this

can you check the updated description of #1288?

roles/tests-integration/tests/common/mod.rs
```diff
@@ -146,7 +146,10 @@ impl TemplateProvider {
     }
 
     fn stop(&self) {
-        let _ = self.bitcoind.client.stop().unwrap();
+        match self.bitcoind.client.stop() {
+            Ok(_) => (),
+            Err(e) => eprintln!("Failed to stop bitcoind: {:?}", e),
```
plebhash, line 151: this is a bit too YOLO and doesn't fully fix the root of the problem

we need to figure out how to properly stop `bitcoind` via the crate APIs

please check the updated description of https://github.com/stratum-mining/stratum/issues/1288

roles/tests-integration/tests/common/mod.rs
```diff
@@ -145,8 +145,11 @@ impl TemplateProvider {
         TemplateProvider { bitcoind }
     }
 
-    fn stop(&self) {
-        let _ = self.bitcoind.client.stop().unwrap();
+    fn stop(&mut self) {
+        match self.bitcoind.stop() {
+            Ok(_) => (),
+            Err(e) => eprintln!("Failed to stop bitcoind: {:?}", e),
```
plebhash, line 151: we should still `panic!` here in case of `Err`

the difference now is that we don't expect this `Err` case to be triggered anymore

but in case it does, we don't want to mask that with a `eprintln!`

plebhash:
I had to make one small modification to get it to compile:
```diff
-    fn stop(&self) {
+    fn stop(&mut self) {
         self.bitcoind.stop().expect("Failed to stop bitcoind");
     }
```

 I cherry picked this into `2024-11-14-rewrite-tproxy-test` of #1262 and unfortunately the error did not go away:
```
thread 'translation_proxy' panicked at tests-integration/tests/common/mod.rs:149:30:
Failed to stop bitcoind: JSON-RPC error: transport error: Couldn't connect to host: Connection refused (os error 61)
Reader task terminated.
test translation_proxy ... FAILED

failures:

failures:
    translation_proxy

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 9.61s
```

GitGab19:
Closing this in favor of #1291

PR #1293, opened 2024-12-12
add unix specific execution permission to tp binary executable

### LINKED ISSUE #1278: Test TemplateProvider OS E13

Shourya742:
closes #1278 
The error Permission denied (os error 13) indicates that your process does not have the necessary permissions to execute the bitcoind binary located at /tmp/.template-provider/bitcoin-sv2-tp-0.1.9/bin/bitcoind
Solution: changing macos runner version worked

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,10 +14,10 @@ jobs:
     strategy:
       matrix:
         os:
-          - macos-latest
+          - macos-14
           - ubuntu-latest
         include:
-          - os: macos-latest
+          - os: macos-14
             target: x86_64-apple-darwin
           - os: ubuntu-latest
             target: x86_64-unknown-linux-musl


plebhash:
@Shourya742 it seems we changed the approach here?

can we get a description of this new approach? (overall let's try to keep PRs descriptive, otherwise the review process becomes guesswork, which is not really ideal)

plebhash:
tbh I'm not a fan of slapping `sudo` everywhere... as a rule-of-thumb, this is usually a very dirty approach to fixing errors and it makes me worried that we are going in this direction here

therefore I feel it's important to have a detailed understanding on why we are doing this... and most importantly: why did we pivot away from the original approach?

plebhash:
even tests that have nothing to do with the original problem (`/tmp` permissions for Integration Tests) are being changed for priviledged execution

`sudo` = big powers, big responsabilities

slapping `sudo` as a umbrella solution to solve problems generically is lazy, dirty and dangerous

plebhash:
the original solution description was:

> The Permission denied error occurs because the downloaded bitcoind binary lacks execute permissions, and on macOS, the com.apple.quarantine attribute blocks execution. This fix ensures the binary is executable and removes the quarantine attribute with `xattr`.

our solution needs to encompass the scope of this problem without overescalating execution to `root`

we can achieve that with:
```yaml
- name: Roles Integration Tests (MAC)
        if: matrix.os == 'macos-latest'
        run: |
        xattr -d com.apple.quarantine /tmp/.template-provider/bitcoin-sv2-tp-*/bin/bitcoin*
        cargo test --manifest-path=roles/Cargo.toml --verbose --test '*' -- --nocapture
```

the `xattr` could potentially require `sudo` execution, and that is ok, because we are limiting `root` execution to this specific action, instead of the entire execution.

the point here is that we are fixing the problem with a specific solution, instead of an umbrella with potential unintended consequences.

plebhash:
> > ```yaml
> > xattr -d com.apple.quarantine /tmp/.template-provider/bitcoin-sv2-tp-*/bin/bitcoin*
> > ```
> 
> Won't this fail, as the directory `/tmp/.template-provider` will be created during execution and won't be present before the action starts? The reason I opted to remove the quarantine attribute via code was because of this limitation.

I see your point.

Perhaps instead of calling `xattr` we could do:

```yaml
- name: Roles Integration Tests (MAC)
        if: matrix.os == 'macos-latest'
        run: |
        sudo mkdir -p /tmp/.template-provider
        sudo chmod 777 /tmp/.template-provider
        cargo test --manifest-path=roles/Cargo.toml --verbose --test '*' -- --nocapture
```

this approach tweaks permissions from a different perspective:

we just create a `/tmp/.template-provider` beforehand and make sure every user has write permission

then shouldn't even need to tweak write permissions for the `bitcoind` executable, so `xattr` execution is no longer needed

plebhash:
I think the current approach is much better.

a few remarks:
- let's not forget to update the PR description so if we need to check this in the future, it is accurately describing what the PR did
- let's squash commit history

plebhash (approved):

Issue #1244, opened 2024-11-11
Improve panic output in `Sniffer`


jbesraa:
Currently the panic output modification with 'set_hook' is not consistent, sometimes it would print the full trace although it's not suppose to. 
I suspect it is the fact that 'set_hook' is called within the 'drop' function.

Moreover, the 'drop' function only prints 'downstream' messages even if also 'upstream' have uncovered messages. We should change the 'if' in 'drop' to 'match' and print both values in case of '(true, true)'.

roles/tests-integration/tests/common/sniffer.rs L430

jbesraa:
Maybe something like this

```rust
 impl Drop for Sniffer {
     fn drop(&mut self) {
-        // Don't print backtrace on panic
-        std::panic::set_hook(Box::new(|_| {
-            println!();
-        }));
-        if !self.downstream_messages.is_empty() {
-            println!(
-                "You didn't handle all downstream messages: {:?}",
-                self.downstream_messages
-            );
-            panic!();
-        }
-        if !self.upstream_messages.is_empty() {
-            println!(
-                "You didn't handle all upstream messages: {:?}",
-                self.upstream_messages
-            );
-            panic!();
+        let downstream_messages = self.downstream_messages.clone();
+        let upstream_messages = self.upstream_messages.clone();
+
+        match (downstream_messages.is_empty(), upstream_messages.is_empty()) {
+            (true, true) => {
+                println!("Test Passed, all messages were handled");
+            }
+            (true, false) => {
+                println!("Test Failed, not all upstream messages were handled:\n");
+                let formatted = format!("{:?}", upstream_messages);
+                println!("messages from upstream:\n {:?}", formatted);
+                panic!("");
+                // std::process::exit(1);
+            }
+            (false, true) => {
+                println!("Test Failed, not all messages were handled:\n");
+                let formatted = format!("{:?}", downstream_messages);
+                println!("messages from downstream:\n {:?}", formatted);
+                panic!("");
+                // std::process::exit(1);
+            }
+            (false, false) => {
+                println!("Test Failed, not all messages were handled:\n");
+                let downstream_messages = format!("{:?}", downstream_messages);
+                println!("messages from downstream:\n {:?}", downstream_messages);
+                let upstream_messages = format!("{:?}", upstream_messages);
+                println!("messages from upstream:\n {:?}", upstream_messages);
+                panic!("");
+                // std::process::exit(1);
+            }
```

And move the `set_hook` to the new function

```rust
[...]

PR #1297, opened 2024-12-13
Integration tests cleanup

### LINKED ISSUE #1296: Integration test `test_sniffer_interrupter()` should be under `/roles/tests-integration/tests/common` module
We currently have a test called `test_sniffer_interrupter()` in `/roles/tests-integration/tests/pool_integration.rs` ([here](https://github.com/stratum-mining/stratum/blob/2675ef949e3c7aeebdfec98e6abb0d06fc59d0ed/roles/tests-integration/tests/pool_integration.rs#L63)).

The test is testing the `sniffer` functionality, so it should live under `/roles/tests-integration/tests/common` module.

jbesraa:
Resolves #1296  #1244 #1181 #1234 #1317


roles/tests-integration/README.md
```diff
@@ -0,0 +1,21 @@
+# SV2 Integration Tests
+
+This is a test crate and it can be used in order to test the behavior of different roles when
+working together. Each role should have a `start_[role_name]` function under `common` folder that
+can be called in order to run the role. In order to assert the behavior of the role or the messages
+it exchanges with other roles, you can use the `Sniffer` helper in order to listen to the messages
+exchanged between the roles, and assert those messages using the `assert_message_[message_type]`
+function. For examples on how to use the `Sniffer` helper, you can check the
+`sniffer_integration.rs` module or other tests in the `tests` folder.
+
```
plebhash, line 10: should we add some words about `regtest` TP here?

roles/jd-client/src/lib/mod.rs
```diff
@@ -54,6 +54,7 @@ pub static IS_NEW_TEMPLATE_HANDLED: AtomicBool = AtomicBool::new(true);
 /// switching to backup Pools in case of declared custom jobs refused by JDS (which is Pool side).
 /// As a solution of last-resort, it is able to switch to Solo Mining until new safe Pools appear
 /// in the market.
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
jbesraa: You are right that none of them is currently used beside the `Sniffer`. I think in the future some of those roles will have some functionality that we might need while testing, thats the main motivation for including them

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  ci:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
```
plebhash, line 20: why are we using `musl`?

plebhash:
@jbesraa please see these commits and if they make sense, cherry-pick them into this PR:

---

with #1317 in mind, I forked this branch and added the following commits:
- 38dd0b885878b9ce5f8ee9836d61fcbbfc4e7102 renaming the crate (now snake_case with `*_sv2` suffix, following the most common pattern on SRI)
- 0ec0951dfb37148aa37a10038efd88d2e0588180 making `sniffer` module externally public, so that other projects can use this crate as a dependency and import `sniffer::Sniffer`

if we merge this PR including these commits we can close #1317
---

the following commit is a simple cleanup of variable naming:
- 2539393e5faf5a25b10b40af715de1aac5868dfa

plebhash (approved):

plebhash:
oops, looks like I forgot to fix one import path on deb9747b2a2d10ff8013c7d79e95685b8d58cb8f and now it's breaking CI

should be fixed on https://github.com/plebhash/stratum/commit/3328f36df127408cd48981db1e36318ebd8d6c2b

@jbesraa after this I think this is ready for merging

jbesraa:
> oops, looks like I forgot to fix one import path on [deb9747](https://github.com/stratum-mining/stratum/commit/deb9747b2a2d10ff8013c7d79e95685b8d58cb8f) and now it's breaking CI
> 
> should be fixed on [plebhash@3328f36](https://github.com/plebhash/stratum/commit/3328f36df127408cd48981db1e36318ebd8d6c2b)
> 
> @jbesraa after fixing this I think this is ready for merging

Squashed into deb9747b2a2d10ff8013c7d79e95685b8d58cb8f

plebhash (approved):

plebhash (changes requested):
as stated on the PR review call earlier today, I found some issues while running `cargo doc` which should be addressed before we merge this PR

they came from the fact that some public Rust Doc comments have links to private items, for example:
```
warning: public documentation for `Sniffer` links to private item `InterceptMessage::response_message`
  --> tests-integration/lib/sniffer.rs:50:7
   |
50 | /// [`InterceptMessage::response_message`].
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this item is private
```

I started addressing this, but I quickly noticed some other issues with `InterceptMessage`

although I previously approved this PR for merging, this rabbit hole made me realize we still have some things to think more deeply about

## `InterceptMessage::response_message` field name

this name is a bit misleading, as this is not really a response, it's just the new message that will replace the old one

so I propose we rename it to `replace_message` via 3b387ac76db4f03fbe5efcd8befe920a3173019f

## optionality of `replace_message` field

currently the `break_on` functionality is a bit limited, because it forces the user to replace the message every time they want to kill the `Sniffer`

imagine a scenario where the user wants to kill `Sniffer` as soon as some specific message is intercepted, but there's no need to actually modify it with a new one... currently, that is impossible, and the user will be forced to create some dummy new message just to be able to use the `break_on` flag

so I made some adaptations to make it optional on d5a60c4c9abc4b648aeb448a9ae307c701600160

BUT I'm not sure this is actually useful and I'm not really asking for this commit to be cherry-picked into this PR, please check the item below

## `test_sniffer_interrupter` is not comprehensive, and motivation for `break_on` is not clear

the name of this test implies that we are testing that the "interrupt" functionality (`break_on: true`) works as expected, but we never actually assert that the connection was terminated

we only assert that `SetupConnectionSuccess` was replaced with `SetupConnectionError`, so in the very least this test has a bad name

but I believe the problem goes a bit deeper

tbh looking at this test gives me the impression that we're kind of mixing "interrupt" + "intercept" terminology, which IMHO is something we need to be careful moving forward in order to further propagate this confusion

for clarity, here's my understanding of these terms:
- "intercept": this is a very generic term... it simply means `Sniffer` took some message and it's about to do something with it (maybe just log it, maybe replace it with a different one)
- "interrupt": it means that after some interception, the `Sniffer` is supposed to be killed

during this analysis, I also noticed that if `break_on` is `true`, `recv_from_down_send_to_up` and `recv_from_up_send_to_down` return a `SnifferError::MessageInterrupted`

this error is never checked nor used for anything, so what is supposed to happen after this error is returned? does `Sniffer` terminate its TCP connections? is it dropped from memory? currently, absolutely nothing happens, and we also don't have any test for that (despite the misleading name of `test_sniffer_interrupter`)

in other words: there's no real "interrupt" functionality, only a superficial appearance of one

which then makes me wonder: what is the point of this "interrupt" functionality in the first place? do we have some real demand for this?

it was introduced on https://github.com/stratum-mining/stratum/pull/1228 where the goal was simply to allow for messages to be modified, but I cannot find any reference in our discussions to really justify this extra functionality

in fact, I found [this question](https://github.com/stratum-mining/stratum/pull/1228/files#r1850376798) that I resolved without a proper answer

so my impression is that this was introduced as a potential "nice-to-have", but it was never properly designed and it feels it's only polluting and adding complexity to the code without any real benefit

so IMHO we should make things simpler and drop this "interrupt" functionality, until we realize that we actually need it (and then properly design and test it)

this is why I believe d5a60c4c9abc4b648aeb448a9ae307c701600160 is not really useful, but I'm still referencing here to provide the full context of my investigation

however, I still believe 3b387ac76db4f03fbe5efcd8befe920a3173019f is important and I would encourage to be included in this PR before the "interrupt" issue is addressed

---

after all the investigation above (which took a few hours), I didn't spend any time actually fixing the original Rust Docs issues, as I believe these items should be addressed first

plebhash:
on top of [this ignored rewiew point](https://github.com/stratum-mining/stratum/pull/1228/files#r1850376798), another review point of #1228 that was not properly addressed:

https://github.com/stratum-mining/stratum/pull/1228#discussion_r1848953490

currently, `test_sniffer_interrupter` only has one single sniffer, and we only assert that it holds `SetupConnectionError` internally

but we do not assert that the downstream actually received this `SetupConnectionError` , which was the original motivation for my request for 2 sniffers on this test

so we're only partially testing things and this test does not really give us full confidence that this functionality

jbesraa:
> as stated on the PR review call earlier today, I found some issues while running `cargo doc` which should be addressed before we merge this PR
> 
> they came from the fact that some public Rust Doc comments have links to private items, for example:
> 
> ```
> warning: public documentation for `Sniffer` links to private item `InterceptMessage::response_message`
>   --> tests-integration/lib/sniffer.rs:50:7
>    |
> 50 | /// [`InterceptMessage::response_message`].
>    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this item is private
> ```
> 
> I started addressing this, but I quickly noticed some other issues with `InterceptMessage`
> 
> although I previously approved this PR for merging, this rabbit hole made me realize we still have some things to think more deeply about
> ## `InterceptMessage::response_message` field name
> 
> this name is a bit misleading, as this is not really a response, it's just the new message that will replace the old one
> 
> so I propose we rename it to `replace_message` via [3b387ac](https://github.com/stratum-mining/stratum/commit/3b387ac76db4f03fbe5efcd8befe920a3173019f)
> ## optionality of `replace_message` field
> 
> currently the `break_on` functionality is a bit limited, because it forces the user to replace the message every time they want to kill the `Sniffer`
> 
> imagine a scenario where the user wants to kill `Sniffer` as soon as some specific message is intercepted, but there's no need to actually modify it with a new one... currently, that is impossible, and the user will be forced to create some dummy new message just to be able to use the `break_on` flag
> 
> so I made some adaptations to make it optional on [d5a60c4](https://github.com/stratum-mining/stratum/commit/d5a60c4c9abc4b648aeb448a9ae307c701600160)
> 
> BUT I'm not sure this is actually useful and I'm not really asking for this commit to be cherry-picked into this PR, please check the item below
> ## `test_sniffer_interrupter` is not comprehensive, and motivation for `break_on` is not clear
> 
> the name of this test implies that we are testing that the "interrupt" functionality (`break_on: true`) works as expected, but we never actually assert that the connection was terminated
> 
> we only assert that `SetupConnectionSuccess` was replaced with `SetupConnectionError`, so in the very least this test has a bad name
> 
> but I believe the problem goes a bit deeper
> 
> tbh looking at this test gives me the impression that we're kind of mixing "interrupt" + "intercept" terminology, which IMHO is something we need to be careful moving forward in order to further propagate this confusion
> 
> for clarity, here's my understanding of these terms:
> 
>     * "intercept": this is a very generic term... it simply means `Sniffer` took some message and it's about to do something with it (maybe just log it, maybe replace it with a different one)
> 
>     * "interrupt": it means that after some interception, the `Sniffer` is supposed to be killed
> 
> 
> during this analysis, I also noticed that if `break_on` is `true`, `recv_from_down_send_to_up` and `recv_from_up_send_to_down` return a `SnifferError::MessageInterrupted`
> 
> this error is never checked nor used for anything, so what is supposed to happen after this error is returned? does `Sniffer` terminate its TCP connections? is it dropped from memory? currently, absolutely nothing happens, and we also don't have any test for that (despite the misleading name of `test_sniffer_interrupter`)
> 
> in other words: there's no real "interrupt" functionality, only a superficial appearance of one
> 
> which then makes me wonder: what is the point of this "interrupt" functionality in the first place? do we have some real demand for this?
> 
> it was introduced on #1228 where the goal was simply to allow for messages to be modified, but I cannot find any reference in our discussions to really justify this extra functionality
> 
> in fact, I found [this question](https://github.com/stratum-mining/stratum/pull/1228/files#r1850376798) that I resolved without a proper answer (maybe we discussed it on Discord, or some call?)
> 
> so my impression is that this was introduced as a potential "nice-to-have", but it was never properly designed and it feels it's only polluting and adding complexity to the code without any real benefit
> 
> so IMHO we should make things simpler and drop this "interrupt" functionality, until we realize that we actually need it (and then properly design and test it)
> 
> this is why I believe [d5a60c4](https://github.com/stratum-mining/stratum/commit/d5a60c4c9abc4b648aeb448a9ae307c701600160) is not really useful, but I'm still referencing here to provide the full context of my investigation
> 
> however, I still believe [3b387ac](https://github.com/stratum-mining/stratum/commit/3b387ac76db4f03fbe5efcd8befe920a3173019f) is important and I would encourage to be included in this PR before the "interrupt" issue is addressed
> 
> after all the investigation above (which took a few hours), I didn't spend any time actually fixing the original Rust Docs issues, as I believe these items should be addressed first

`cargo doc` bug is here because we tried to add  https://github.com/stratum-mining/stratum/issues/1317 to this PR although it was not planned initially. I would suggest to remove the commits regarding #1317 and not block this.

Regarding the other concerns, I also recommend to open an issue and discuss things and then open new PR. Your notes are general about the sniffer and integration tests design and I am not sure this pull request should be blocked by that.

My main concern is that this PR is already doing too much, I initially resolved 4-5 issues but you added another 3-4 and I don't think it is a good idea to try and more stuff here, it is gonna be too annoying to review/improve/track.

jbesraa:
Rebased without adding changes

plebhash (approved):

Issue #1275, opened 2024-12-07
Crates were not bumped in PR #1230


GitGab19:
PR #1230 changed some crates under `protocols` to be `no_std`, without bumping their versions. 

Now this is creating some errors in the CI execution of `semver-checks`, such as here: https://github.com/stratum-mining/stratum/actions/runs/12211524808/job/34069218188?pr=1242.

We need to bump versions of crates which were modified there to fix this and unblock all the docs PRs.

Specifically, we need to bump the major version number of:
- `serde_sv2`
- `const_sv2`
- `framing_sv2`
- `buffer_sv2`

jbesraa:
I would remove the GH from ci and move the checks into a script that we run manually before creating a new release

plebhash:
> I would remove the semver check from ci and move it into a script that we run manually before creating a new release

@GitGab19 feels we should catch semver deviations as soon as possible

but it's also important to highlight that something is not working as expected

we merged #1230 without CI blocks, and now we are getting this behavior

ideally #1230 should have been blocked by this semver check

plebhash:
> we merged https://github.com/stratum-mining/stratum/pull/1230 without CI blocks, and now we are getting this behavior
> 
> ideally https://github.com/stratum-mining/stratum/pull/1230 should have been blocked by this semver check

this happened because our CI was using whatever was the latest release of `cargo semver-checks`, and https://github.com/obi1kenobi/cargo-semver-checks/pull/1015 introduced a change of behavior into v0.37.0

the solution was https://github.com/stratum-mining/stratum/pull/1276 where we are locking CI to always use `cargo semver-checks` on release v0.33.0 in order to always have the same behavior of the tool

plebhash:
todo: follow up to https://github.com/stratum-mining/stratum/pull/1276 that locks `cargo semver-checks` on 0.37.0

this CI will have to run on rust stable

PR #1300, opened 2024-12-16
use `cargo semver-checks` v0.37 and bump crate versions as a follow-up to #1276

plebhash:
follow up to #1276 according to this comment: https://github.com/stratum-mining/stratum/issues/1275#issuecomment-2532472582

---

# rationale

we are planning to remove more cargo build features in the future, and therefore it's desirable to enforce semver around this via CI

as a consequence, we need to use `cargo semver-check` v0.37 (or greater, but always locked)

this creates incompatibilities with MSRV, therefore we need to run `cargo semver-check` on `stable`

finally, switching `cargo semver-check` to v0.37 forced us to bump many crate versions

---

PR changes:
- use `cargo semver-check` v0.37
- use `stable` toolchain instead of MSRV 1.75
- bump MAJOR for crates that dropped `no_std` features on #1276 
- bump PATCH for crates that had dependencies on the crates that were bumped because of #1276

### DIFF
--- a/protocols/v2/subprotocols/common-messages/Cargo.toml
+++ b/protocols/v2/subprotocols/common-messages/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "common_messages_sv2"
-version = "2.0.0"
+version = "2.0.1"
 authors = ["The Stratum V2 Developers"]
 edition = "2018"
 readme = "README.md"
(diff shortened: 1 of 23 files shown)


protocols/v2/subprotocols/common-messages/Cargo.toml
GitGab19, line 1: Here you missed to bump the patch version number

protocols/v2/subprotocols/job-declaration/Cargo.toml
GitGab19, line 1: Same here

protocols/v2/subprotocols/mining/Cargo.toml
GitGab19, line 1: Same here

protocols/v2/sv2-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sv2_ffi"
-version = "1.0.0"
+version = "1.0.2"
```
GitGab19, line 3:
```suggestion
version = "1.0.1"
```

PR #1248, opened 2024-11-13
Improve coinbase

Fi3:
Better management of the coinbase's input script:

- [X] add a method that the pool can use in order to change coinbase_script_additional_data of an already opened channel and send downstream the new extranonce_prefix
- [X] rename the pool signature field of the channel factory into coinbase_script_additional_data, and change the type from `String` to `Vec<u8>`. Sv2 do not say anything about pool's signature, but we still want to let the user add some arbitrary data and the beginning of the extranonce. We do not assume that this data is a pool signature. We do not assume that this data is valid utf8.
- [X] send downstream coinbase_script_additional_data as part of the extranonce_prefix and not as part of the coinbase_prefix. This make the implementation of a JDC more straightforward, as it do not need anymore to observe the jobs sent and created by the pool and reconstruct and parse the coinbase to know the coinbase_script_additional_data.

PR #1301, opened 2024-12-17
Remove short ids

Fi3:
As per https://github.com/stratum-mining/sv2-spec/issues/99 this PR change the implementation so that in the JD protocol full tx ids are used instead of short tx ids. It also remove the IdentifyTransaction and IdentifyTransactionSuccess messages since are not needed anymore.

it need https://github.com/stratum-mining/stratum/pull/1248


GitGab19:
As agreed with @Fi3, we are going to adapt this PR to current `main`. 

This will be part of `roles_logic_sv2` refactoring. 

Opened an issue for it: https://github.com/stratum-mining/stratum/issues/1440.

Moving this PR to draft so that it can still be used by @plebhash while working on #1440.

GitGab19:
Closing since this has been done in PR https://github.com/stratum-mining/stratum/pull/1650.

Issue #1286, opened 2024-12-11
need to migrate `sv1-mining-device` logs from `println! / eprintln!` to `tracing`


plebhash:
[`sv1-mining-device`](https://github.com/stratum-mining/stratum/tree/main/roles/test-utils/mining-device-sv1) currently has a logging system based on `println!` and `eprintln!`

we should migrate this to `tracing`

use `mining-device` as a reference

devworlds:
Hey @plebhash i uptade the printlns to info, but i cant't push a new branch to create pr, its my first make something to open-source, i need to so something before?

plebhash:
closed via #1309

PR #1309, opened 2024-12-19
feat: update println! and eprintln! to info using tracing cargo.

devworlds:
Update mining-device-sv1 from println! to info! using tracing following mining-device as example.

the issue: https://github.com/stratum-mining/stratum/issues/1286


plebhash (approved):

plebhash:
thanks @devworlds 

I rebased the commit history to make it a bit cleaner, and also switched some `info!` for `error!` and `warn!`

just waiting for CI so I can merge it

PR #1310, opened 2024-12-19
Bump TP to 0.1.13 for CI

Sjors:
It's had enough manual testing that I feel good about this release. Let's make sure there's breaking changes I missed though.

### DIFF
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -24,7 +24,7 @@ use tar::Archive;
 // prevents get_available_port from ever returning the same port twice
 static UNIQUE_PORTS: Lazy<Mutex<HashSet<u16>>> = Lazy::new(|| Mutex::new(HashSet::new()));
 
-const VERSION_TP: &str = "0.1.9";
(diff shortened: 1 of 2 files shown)


Sjors:
The pool_integration test fails mysteriously here:

```rs
assert_tp_message!(&sniffer.next_message_from_upstream(), NewTemplate);
```

Looking at the Bitcoin Core log the message _does_ arrive.

It does take a bit longer, about little under half a second. So I tried adding:

```rs
std::thread::sleep(std::time::Duration::from_millis(1000));
```

But that didn't help. I can reproduce the issue locally as well.

Any ideas?

cc @jbesraa

jbesraa:
Here is a fix:
https://github.com/stratum-mining/stratum/commit/f332727c09d3feffecfd656d9ea3b9ca772aa39d 

Instead of checking immediately for `NewTemplate` message, we wait for it to arrive first

Sjors:
@jbesraa thanks! I took your fix.

Also improved the TP logging a bit in the tests.

jbesraa (approved):

plebhash (approved):

Issue #1260, opened 2024-11-19
Fixing config naming across roles


jbesraa:
Config object across different roles is not well named. For example JDC config is called `ProxyConfig` we should change it to `JDCConfig` and so on for all other roles `[RoleName]Config`.

devworlds:
Basically, need to change the ProxyConfig Object to JDCConfig, and update all references that use it?
![image](https://github.com/user-attachments/assets/687545ac-4223-4590-85c7-752b57d69899)

jbesraa:
Updated top comment - We only have translator and mining proxy left.

Sansh2356:
@plebhash  @jbesraa is this still pending ?

jbesraa:
Yes, for the translator and mining-proxy. See top comment for more info or refer to #1493 and #1507

PR #1316, opened 2024-12-21
feat: update ProxyConfig in roles to JDCConfig, and update the references

devworlds:
Hi, i not found another object with Config, so i changed only ProxyConfig, to check if is what was  wanted.

Issue: https://github.com/stratum-mining/stratum/issues/1260

### DIFF
--- a/INTEROPERABILITY-TESTS.md
+++ b/INTEROPERABILITY-TESTS.md
@@ -61,7 +61,7 @@ Once the pool role is running you can start up the Translation Proxy. The transl
    
 If this successfully starts up you'll see these messages as part of the output:
 ```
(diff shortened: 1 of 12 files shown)


jbesraa (changes requested):
Thanks for working on this!

Please go to all of the `main.rs` files inside `roles/[role_name]` and there you can that each different role have its own config. We should change the name in each role to reflect the name of the role. For example, in translator role we should change to `TranslatorConfig`, in jd-server role we should change to `JDSConfig` and so on.

devworlds:
> Thanks for working on this!
> 
> Please go to all of the `main.rs` files inside `roles/[role_name]` and there you can that each different role have its own config. We should change the name in each role to reflect the name of the role. For example, in translator role we should change to `TranslatorConfig`, in jd-server role we should change to `JDSConfig` and so on.

got it!

jbesraa (changes requested):
Let me just explain a bit about the project structure. You have the `roles` folder which is a cargo workspace, i.e. it contains multiple crates. Each one of the crates have it is own configuration object. For example, the `translator` role under `roles/translator` have the `roles/translator/src/lib/proxy_config.rs` config. This config should be changed to `TranslatorConfig`.
Whats correct for the `translator` is correct for all the other roles. We have `jd-client` and its config should be `JDCConfig`. We also have `jd-server` and its config should be `JDSConfig` and so on.

Let me know if that is not clear.

GitGab19:
Closing this as stale

Issue #1311, opened 2024-12-20
Need `cargo doc --all-features` test in CI/CD


rrybarczyk:
Currently there is no checks to ensure the cargo documentation builds. To ensure no code changes break documentation, we need to ensure that `cargo doc --all-features` passes for every crate.

plebhash:
it should be sufficient to run `cargo doc --all-features` under the `protocols` workspace

this is not a comprehensive analysis (maybe more crates are broken), but I tried it locally and it seems `template_distribution_sv2` is broken:

```
    Checking template_distribution_sv2 v1.0.3 (/Users/plebhash/develop/stratum/protocols/v2/subprotocols/template-distribution)
error[E0433]: failed to resolve: use of undeclared type `Vec`
   --> v2/subprotocols/template-distribution/src/new_template.rs:209:28
    |
209 |             let mut path = Vec::new();
    |                            ^^^ use of undeclared type `Vec`
    |
help: consider importing one of these items
    |
7   + use alloc::vec::Vec;
    |
7   + use crate::vec::Vec;
    |

error[E0433]: failed to resolve: use of undeclared type `Vec`
   --> v2/subprotocols/template-distribution/src/new_template.rs:230:34
    |
230 |             coinbase_tx_outputs: Vec::new().try_into().unwrap(),
    |                                  ^^^ use of undeclared type `Vec`
    |
help: consider importing one of these items
    |
7   + use alloc::vec::Vec;
    |
7   + use crate::vec::Vec;
    |

error[E0599]: no method named `try_into` found for struct `Vec<u8>` in the current scope
   --> v2/subprotocols/template-distribution/src/new_template.rs:195:66
    |
195 |         let coinbase_prefix: binary_sv2::B0255 = coinbase_prefix.try_into().unwrap();
    |                                                                  ^^^^^^^^
    |
   ::: /Users/plebhash/.rustup/toolchains/1.78-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/convert/mod.rs:611:8
    |
611 |     fn try_into(self) -> Result<T, Self::Error>;
    |        -------- the method is available for `Vec<u8>` here
    |
    = help: items from traits can only be used if the trait is in scope
    = note: 'core::convert::TryInto' is included in the prelude starting in Edition 2021
help: trait `TryInto` which provides `try_into` is implemented but not in scope; perhaps you want to import it
    |
1   + use core::convert::TryInto;
    |
help: there is a method `into` with a similar name
[...]

Issue #1210, opened 2024-10-14
Enable all features in for published Rust Docs of `protocols` crates


plebhash:
Currently, the public Rust Docs for `protocols` crates are rendered without enabling any of the feature flags available on the crates.

That prevents many important elements from being displayed (e.g.: `State` in `codec_sv2`, which depends on the `noise_sv2` feature flag being enabled).

In order to change that, we need to add the following to their `Cargo.toml`:
```toml
[package.metadata.docs.rs]
all-features = true
```

ref: https://docs.rs/about/metadata

PR #1318, opened 2024-12-23
simplify `[package.metadata.docs.rs]` on `protocols` crates

plebhash:
currently we are making sure all published `protocols` crates cover all of the crates features on Rust Docs

we enforce that by having the following on `Cargo.toml` of each crate:
```toml
[package.metadata.docs.rs]
all-features = true
```

this strategy was suggested on #1210 and implemented via #1211

during some discussions in #1311 @jbesraa pointed out that this strategy is a bit overengineered

I agree `all-features` is more generalistic than necessary, and it is causing trouble (e.g.: trying to build `cargo b --all-features` on `template_distribution_sv2` breaks due to `with_serde` and `prop_test` being built together)

also, features like `no_std` and `with_serde` are being deprecated, so there's no point in wasting engineering effort on them

---

this PR is simplifying the `[package.metadata.docs.rs]` parameters

now, for each crate, we only enable publishing Rust Docs with features that are relevant for API usage, namely:

- `binary_sv2`: `features = ["core", "with_buffer_pool"]`
- `binary_codec_sv2`: `features = ["with_buffer_pool"]`
- `codec_sv2`: `features = ["with_buffer_pool", "noise_sv2"]`
- `framing_sv2`: `features = ["with_buffer_pool"]`

the following crates have no features that are relevant for API usage, and therefore we don't enforce anything on their `Cargo.toml`:
- `derive_codec_sv2`
- `serde_sv2`
- `const_sv2`
- `noise_sv2`
- `roles_logic_sv2`
- `common_messages_sv2`
- `job_declaration_sv2`
- `mining_sv2`
- `template_distribution_sv2`
- `sv2_ffi`
- `buffer_sv2`

### DIFF
--- a/protocols/v2/binary-sv2/binary-sv2/Cargo.toml
+++ b/protocols/v2/binary-sv2/binary-sv2/Cargo.toml
@@ -28,4 +28,4 @@ prop_test = ["binary_codec_sv2/prop_test", "derive_codec_sv2"]
 with_buffer_pool = ["binary_codec_sv2/with_buffer_pool", "derive_codec_sv2"]
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["core", "with_buffer_pool"]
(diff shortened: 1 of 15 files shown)


jbesraa (changes requested):
Please make this a single commit. This makes ALOT of noise in the `main` branch.

jbesraa (approved):

Issue #1266, opened 2024-11-28
Investigate ghost processes in tproxy and JDC


jbesraa:
While working on #1207 it was noticed that some of the spawned tasks are not exiting upon dropping the main task.  This results in hanging test as the future of those spawned tasks is never resolved and the test never ends.

@Shourya742  @plebhash  tagging you as you have done some work here already

edit by @plebhash: it's important to get to the bottom of this, because it seems it's become a blocker to Integration Test issues such as https://github.com/stratum-mining/stratum/issues/1207

plebhash:
one of the main findings around this is the fact that this usually happens with JDC and tProxy, and both are using a `task_collector` pattern

```
task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>
```

plebhash:
here's some findings from manual tests

they were observed without any Integration Tests or `tokio-console`

the text below might seem very dense, but we can actually infer many patterns from these observations

---

The first setup had the following:
- TP on testnet4
- Pool
- JDS
- JDC
- tProxy (connected to JDC)

I tried **killing JDC** and observing tProxy in two different ways:
- **wait** for tProxy to establish the connection with JDC before killing JDC: **tProxy becomes unresponsive to ctrl+C and needs to be killed manually**
- **don't wait** for tProxy to open an extended channel with JDC before killing JDC: tProxy is still responsive to ctrl+C

then I tried **killing tProxy** and observing JDC, and the same patterns was observed:
- **wait** for tProxy to establish the connection with JDC before killing tProxy: **JDC becomes unresponsive to ctrl+C and needs to be killed manually**
- **don't wait** for tProxy to establish the connection with JDC before killing tProxy: tProxy is still responsive to ctrl+C

then I tried **killing Pool** and observing JDC, but the same pattern was **not** observed:
- **wait** for JDC to establish connection with Pool before killing Pool: JDC is still responsive to ctrl+C
- **don't wait** for JDC to establish connection with Pool before killing Pool: JDC is still responsive to ctrl+C

then I tried **killing JDS** and observing JDC, but the same pattern was **not** observed:
- **wait** for JDC to establish connection with JDS before killing JDS: JDC is still responsive to ctrl+C
- **don't wait** for JDC to establish connection with JDS before killing JDS: JDC is still responsive to ctrl+C

---

The second setup was the following:
- TP on testnet4
- Pool
- tProxy (connected to Pool)

[...]

PR #1295, opened 2024-12-13
fix: clarify logging

average-gary:

Issue #1207, opened 2024-10-11
CI migration: `interop-jdc-change-upstream` MG to Integration Test


plebhash:
as part of #1121 a good first MG test to migrate is [`interop-jdc-change-upstream`](https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/interop-jdc-change-upstream/interop-jdc-change-upstream.json) because:
- it covers many roles (including JD)
- it requires the usage of a mock, which we still don't have in the integration test framework

this will help us continue shaping up the integration test framework, as a continuation of #1122

plebhash:
to elaborate a bit more on the requirement of a mock:

this test is essentially making sure that JDC has a sane implementation for the fallback mechanism, which is one of the major selling points of JD:

assuming a JDS+Pool have already acknowledged some specific job, Pool rejecting shares must trigger JDC to switch upstreams

now, how do we get a Pool to reject shares? it wouldn't really make sense to add that kind of functionality to SRI Pool implementaiton

therefore, we need a way to mock a role

plebhash:
I did some experimentation where I tried modifying the `Sniffer` struct to give it the ability to inject messages, but this modification proved not to be trivial

I wonder if an easier approach could consist of the following:

we add a new `Mock` struct that does not relay messages... in fact, it doesn't even need to have both `upstream`+`downstream` connections, as one single connection would suffice

`Mock` would keep a record of all messages it receives form its connection, and allow the user to assert against them (much like `Sniffer`)

but it would also have a method that would allow the user to inject some custom message into the connection

jbesraa:
> I did some experimentation where I tried modifying the `Sniffer` struct to give it the ability to inject messages, but this modification proved not to be trivial
> 
> I wonder if an easier approach could consist of the following:
> 
> we add a new `Mock` struct that does not relay messages... in fact, it doesn't even need to have both `upstream`+`downstream` connections, as one single connection would suffice
> 
> `Mock` would keep a queue record with all messages it receives form its single connection, and allow the user to assert against them (much like `Sniffer`)
> 
> but it would also have a method that would allow the user to inject some custom message into the connection

Yea I also think it need to be something a bit different than the Sniffer. It would need to be a "role" that can act as either upstream or downstream. As mentioned, we would also need to define some functionality to tell it "when you receive message X return message Y".  
[...]

PR #1319, opened 2024-12-23
fix sigterm issue with translator sv2

Shourya742:
Partially closes: #1266

The issue lies with the blocking nature of `self.internal_start().await` inside the loop. Since the .await suspends the current task, it prevents the tokio::select! from responding promptly to other events, such as the ctrl_c() signal. This PR tends to spawn to internal_start as a separate task, while still keeping the select! macro active to listen for other events.

### DIFF
--- a/roles/translator/src/lib/mod.rs
+++ b/roles/translator/src/lib/mod.rs
@@ -10,7 +10,8 @@ use std::{
 };
 
 use tokio::{
-    sync::broadcast,
+    select,
+    sync::{broadcast, Notify},
     task::{self, AbortHandle},
 };
 use tracing::{debug, error, info, warn};
@@ -32,6 +33,7 @@ pub mod utils;
 pub struct TranslatorSv2 {
     config: ProxyConfig,
     reconnect_wait_time: u64,
+    shutdown: Arc<Notify>,
 }
 
 impl TranslatorSv2 {
@@ -41,6 +43,7 @@ impl TranslatorSv2 {
         Self {
             config,
             reconnect_wait_time: wait_time,
+            shutdown: Arc::new(Notify::new()),
         }
     }
 
@@ -58,7 +61,8 @@ impl TranslatorSv2 {
         let task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>> =
             Arc::new(Mutex::new(Vec::new()));
 
-        self.internal_start(
+        Self::internal_start(
+            self.config.clone(),
             tx_sv1_notify.clone(),
             target.clone(),
             tx_status.clone(),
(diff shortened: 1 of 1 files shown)


jbesraa (changes requested):
Please make this PR easier to review by undoing the formatting/indentation changes. If you are moving a function outside the `TranslatorSV2` struct, do it in a separate commit and then commit the changes to make the fix

roles/translator/src/lib/mod.rs
```diff
@@ -72,74 +75,78 @@ impl TranslatorSv2 {
...
+                    info!("Shutting down gracefully...");
+                    break;
                 }
             }
         }
     }
 
     async fn internal_start(
-        &self,
+        proxy_config: ProxyConfig,
```
plebhash, line 144: are we changing the signature for this function for any specific purpose?

roles/translator/src/lib/mod.rs
```diff
@@ -72,74 +75,78 @@ impl TranslatorSv2 {
...
+            tokio::select! {
+                task_status = rx_status.recv().fuse() => {
+                    if let Ok(task_status_) = task_status {
+                        match task_status_.state {
+                            State::DownstreamShutdown(err) | State::BridgeShutdown(err) | State::UpstreamShutdown(err) => {
+                                error!("SHUTDOWN from: {}", err);
+                                self.shutdown.notify_one();
+                            }
+                            State::UpstreamTryReconnect(err) => {
+                                error!("SHUTDOWN from: {}", err);
```
plebhash, line 99: let's bring this change here: https://github.com/stratum-mining/stratum/pull/1295

however I think it's better to make it into a `warn!` instead of `error!`

roles/translator/src/lib/mod.rs
```diff
@@ -10,7 +10,7 @@ use std::{
 };
 
 use tokio::{
-    sync::broadcast,
+    sync::{broadcast, Notify as notification},
```
plebhash, line 13: we're not using this alias on #1321 

is that intentional?

if not, we should try to keep the patterns consistent

plebhash (approved):

plebhash:
tACK

roles/translator/src/lib/mod.rs
```diff
@@ -72,74 +75,79 @@ impl TranslatorSv2 {
...
+                                tokio::spawn (async move {
+                                    // wait a random amount of time between 0 and 3000ms
+                                    // if all the downstreams try to reconnect at the same time, the upstream may
+                                    // fail
+                                    tokio::time::sleep(std::time::Duration::from_millis(wait_time)).await;
 
-                    // kill al the tasks
-                    let task_collector_aborting = task_collector_.clone();
-                    kill_tasks(task_collector_aborting.clone());
+                                    // kill al the tasks
```
jbesraa, line 111: Indentation is a bit off here and in L115

jbesraa (changes requested):
Looks good. Just one small nit.

IMO before we merge this or #1321, a test to cover the feature/bug fix should be implemented.

jbesraa:
This PR is resolving(partially, the translator part) https://github.com/stratum-mining/stratum/issues/1266. 
And #1266 was opened because we were not able to do #1207. IMO in order to make sure this is the solution we want(it might be a solution to a different problem), #1207 need to be resolved.

jbesraa:
@Shourya742 https://github.com/stratum-mining/stratum/pull/1343 this branch have your code and the test I wrote back then(I did some changes to align it with current API design). Feel free to pick it up, otherwise Ill look into it tomorrow.

jbesraa (approved):

plebhash (approved):

PR #1323, opened 2024-12-27
Add sv2-serde-json

Shourya742:
This PR introduces a sv2-serde-json as a utility module. The goal is to eliminate the dependency on `serde` from the message generator, ultimately phasing out `binary_sv2_serde`. Currently, `sv2_serde_json` handles serialization and deserialization by converting data into a `Value` construct, which must then be mapped onto Rust's structs and enums. 

By adopting this approach, we aim to remove all dependencies on `serde`, paving the way for the eventual deprecation of `binary_sv2_serde` entirely. Alternatively, if we decide not to proceed with this new utility, we will extend the existing `binary_sv2` to support JSON serialization and deserialization directly.

This PR is marked as a draft while we finalize the chosen approach. In the meantime, this branch will be maintained for developing JSON features using the external utility, rather than extending `binary_sv2`.


Georges760:
What crates will depend on this new `sv2_serde_json` ? Because it brings many hard dep on `std`.

Georges760:
> > What crates will depend on this new `sv2_serde_json` ? Because it brings many hard dep on `std`.
> 
> This change would only impact `protocol/sv1` and the message generator within the test module. An alternative approach could involve extending `binary_sv2` with JSON serializer and deserializer, but this would be gated under a feature flag. As a result, it wouldn’t compromise our `no-std` compliance for the protocol crates.
> 

For reference only, I did a Stratum v1 no_std client implementation (https://github.com/Foundation-Devices/foundation-rs/tree/main/stratum-v1) and the json parsing was a nightmare...

Georges760:
> > What crates will depend on this new `sv2_serde_json` ? Because it brings many hard dep on `std`.
> 
> This change would only impact `protocol/sv1` and the message generator within the test module. An alternative approach could involve extending `binary_sv2` with JSON serializer and deserializer, but this would be gated under a feature flag. As a result, it wouldn’t compromise our `no-std` compliance for the protocol crates.
> 

Also I was confused because this new crate have the "sv2" prefix, if it is sv1 oriented (JSON data) or at least not specific to sv2, maybe we should use this prefix?

jbesraa (commented):
I am fairly confused about this PR. Can you elaborate why we need this? and why we need to implement a mini serde as utility instead of using serde(if it is really needed..?)

jbesraa:
> > I am fairly confused about this PR. Can you elaborate why we need this? and why we need to implement a mini serde as utility instead of using serde(if it is really needed..?)
> 
> As you may know, we have two versions of `binary_sv2`:
> 
>     1. **Serde Variant:** Utilizes serialization and deserialization mechanisms from the `serde` crate. This version is primarily used by the `message_generator` and the `sv1` crate due to their reliance on JSON serialization and deserialization. Our current `binary_sv2` implementation lacks this functionality.
> 
>     2. **Non-Serde Variant:** Specifically designed for our custom datatypes defined in the protocol specifications.
> 
> 
> The coexistence of these two implementations is challenging due to conflicting APIs for byte serialization and deserialization. In one of our previous meeting, we decided to develop our own custom JSON serializer and deserializer to replace `serde` within the `message_generator` and `sv1` crates. This approach aims to address the current limitations. We may consider extending the `binary_sv2` to incorporate these custom JSON serialization and deserialization features. Still open to suggestions and reviews from the other team members.

I am still kinda confused. Why actually the message generator need ser/de if its working with sv2 and we already know how to handle sv2 message? Also, in case MG need to do this ser/de, why not add `serde` to MG and call it a day?

jbesraa:
I see. Why then not add serde to MG and thats it? Why adding serde to MG trickledown to many different crates in the past?

GitGab19:
@Shourya742 after reading again the discussion here, I think the best approach is the one you started: developing a separate specific crate that can be then imported by the `sv1` crate and by the `Message Generator`.

Are you facing some particular issues somewhere? Is there anything else you would like to get feedback about?

utils/sv2_serde_json/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "sv2_serde_json"
```
plebhash, line 2: should we use a more generic name here? this implies it's sv2 only (not sv1)

jbesraa:
@Shourya742 should we close this for now and reopen in MG repo?

Issue #1202, opened 2024-10-11
Refactor: `subprotocols` crates


GitGab19:
This task is an outcome of the `protocols` Rust docs issues tracked in #845.

While documenting `protocols::v2::subprotocols` in #1201, areas of potential code debt were identified. This issue servers as a place to list out these items to then be addressed in an organized manner. The initial Rust documentation effort was an immediate action, while a refactoring (which implies breaking API changes) is not so urgent priority, so for now we should leave this in the backlog for an appropriate moment in the future.

# Identified Potential Code Debt

## TODO

Shourya742:
common_message_sv2: https://github.com/stratum-mining/stratum/pull/1242#discussion_r1851827737

Shourya742:
common_message_sv2: https://github.com/stratum-mining/stratum/pull/1242#discussion_r1851851553

Shourya742:
template-distribution: https://github.com/stratum-mining/stratum/pull/1257#discussion_r1852013899

Shourya742:
template-distribution: https://github.com/stratum-mining/stratum/pull/1257#discussion_r1852014658

Shourya742:
template-distribution: https://github.com/stratum-mining/stratum/pull/1257#discussion_r1852331612

Shourya742:
template-distribution: https://github.com/stratum-mining/stratum/pull/1257#discussion_r1852416791

jbesraa:
One of the key things I think we are missing in the subprotocols is how to actually use them. 

For example, there is no explanation how to connect to a pool and open SV2 compatible channel. Or simpler, how do I connect to a template provider. 

For us this stuff is intuitive and we have roles crates as reference but I actually think the main reference should be in the subprotocols crates.

plebhash:
> One of the key things I think we are missing in the subprotocols is how to actually use them.
> 
> For example, there is no explanation how to connect to a pool and open SV2 compatible channel. Or simpler, how do I connect to a template provider.
> 
> For us this stuff is intuitive and we have roles crates as reference but I actually think the main reference should be in the subprotocols crates.

I see your point, but is that technical debt or lack of docs?

GitGab19:
We should rename `min_ntime` field (inside `SetCustomMiningJob`, `NewMiningJob`, `NewExtendedMiningJob` messages) into `ntime` or `header_timestamp`: https://github.com/stratum-mining/stratum/pull/1325#issuecomment-2569369954

[...]

PR #1325, opened 2025-01-02
Fix timestamp bug - `min_ntime` becomes always the one sent by TP

### LINKED ISSUE #1324: Pool and JDC update `nTime` field with current timestamp - they should use the one sent by TP
Our current Pool and JDC update the `nTime` field of every job using the current system time, instead of using the value sent by TP. This leads to mine invalid blocks when the value is too old with respect to the network one. We were able to notice this bug on `testnet4`, since its MTP is moved in the future.

Here some logs which shows TP logs when this bug is in place:

timestamp-bug

cc @Sjors

GitGab19:
This PR introduces a new `last_ntime` field into `JobsCreators` struct to save the last `nTime` value sent by TP. 

In this way, the correct value is always used when creating new jobs (future and not), instead of using the current system time.

Closes #1324

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/job_creator.rs
+++ b/protocols/v2/roles-logic-sv2/src/job_creator.rs
@@ -27,6 +27,7 @@ pub struct JobsCreators {
     templte_to_job_id: HashMap<u64, u32, BuildNoHashHasher<u64>>,
     ids: Id,
     last_target: mining_sv2::Target,
+    last_ntime: Option<u32>,
     extranonce_len: u8,
 }
 
@@ -56,6 +57,7 @@ impl JobsCreators {
             templte_to_job_id: HashMap::with_hasher(BuildNoHashHasher::default()),
             ids: Id::new(),
             last_target: mining_sv2::Target::new(0, 0),
+            last_ntime: None,
             extranonce_len,
         }
     }
@@ -91,6 +93,7 @@ impl JobsCreators {
             next_job_id,
             version_rolling_allowed,
             self.extranonce_len,
+            self.last_ntime,
         )
     }
 
@@ -106,6 +109,7 @@ impl JobsCreators {
     /// we clear all the saved templates.
     pub fn on_new_prev_hash(&mut self, prev_hash: &SetNewPrevHash<'static>) -> Option<u32> {
         self.last_target = prev_hash.target.clone().into();
+        self.last_ntime = prev_hash.header_timestamp.into(); // set correct ntime
         let template: Vec<NewTemplate<'static>> = self
             .lasts_new_template
             .clone()
@@ -162,6 +166,7 @@ pub fn extended_job_from_custom_job(
         0,
         true,
         extranonce_len,
+        Some(referenced_job.min_ntime),
     )
 }
 
@@ -181,6 +186,7 @@ fn new_extended_job(
     job_id: u32,
     version_rolling_allowed: bool,
     extranonce_len: u8,
+    ntime: Option<u32>,
 ) -> Result<NewExtendedMiningJob<'static>, Error> {
     coinbase_outputs[0].value = match new_template.coinbase_tx_value_remaining.checked_mul(1) {
         //check that value_remaining is updated by TP
@@ -207,13 +213,7 @@ fn new_extended_job(
 
     let min_ntime = match new_template.future_template {
         true => binary_sv2::Sv2Option::new(None),
-        false => {
-            let now = std::time::SystemTime::now()
-                .duration_since(std::time::UNIX_EPOCH)
-                .unwrap()
-                .as_secs() as u32;
-            binary_sv2::Sv2Option::new(Some(now))
-        }
+        false => binary_sv2::Sv2Option::new(ntime),
     };
 
     let new_extended_mining_job: NewExtendedMiningJob<'static> = NewExtendedMiningJob {
@@ -432,11 +432,11 @@ impl StrippedCoinbaseTx {
         let mut inputs = self.inputs.clone();
         let last_input = inputs.last_mut().ok_or(Error::BadPayloadSize)?;
         let new_last_input_len =
-        32 // outpoint
-        + 4 // vout
-        + 1 // script length byte -> TODO can be also 3 (based on TODO in `coinbase_tx_prefix()`)
-        + self.bip141_bytes_len // space for bip34 bytes
-        ;
+            32 // outpoint
+                + 4 // vout
+                + 1 // script length byte -> TODO can be also 3 (based on TODO in `coinbase_tx_prefix()`)
+                + self.bip141_bytes_len // space for bip34 bytes
+            ;
         last_input.truncate(new_last_input_len);
         let mut prefix: Vec<u8> = vec![];
         prefix.extend_from_slice(&self.version.to_le_bytes());
--- a/protocols/v2/roles-logic-sv2/src/job_creator.rs
+++ b/protocols/v2/roles-logic-sv2/src/job_creator.rs
@@ -211,10 +211,11 @@ fn new_extended_job(
         extranonce_len,
     );
 
-    let min_ntime = match new_template.future_template {
-        true => binary_sv2::Sv2Option::new(None),
-        false => binary_sv2::Sv2Option::new(ntime),
-    };
+    let min_ntime = binary_sv2::Sv2Option::new(if new_template.future_template {
+        None
+    } else {
+        ntime
+    });
 
     let new_extended_mining_job: NewExtendedMiningJob<'static> = NewExtendedMiningJob {
         channel_id: 0,
--- a/roles/tests-integration/tests/common/mod.rs
+++ b/roles/tests-integration/tests/common/mod.rs
@@ -81,20 +81,21 @@ pub struct TemplateProvider {
 }
 
 impl TemplateProvider {
-    pub fn start(port: u16) -> Self {
+    pub fn start(port: u16, sv2_interval: u32) -> Self {
         let temp_dir = PathBuf::from("/tmp/.template-provider");
         let mut conf = Conf::default();
         let staticdir = format!(".bitcoin-{}", port);
         conf.staticdir = Some(temp_dir.join(staticdir));
-        let port = format!("-sv2port={}", port);
+        let port_arg = format!("-sv2port={}", port);
+        let sv2_interval_arg = format!("-sv2interval={}", sv2_interval);
         conf.args.extend(vec![
             "-txindex=1",
             "-sv2",
-            &port,
+            &port_arg,
             "-debug=rpc",
             "-debug=sv2",
-            "-sv2interval=20",
-            "-sv2feedelta=1000",
+            &sv2_interval_arg,
+            "-sv2feedelta=0",
             "-loglevel=sv2:trace",
             "-logtimemicros=1",
         ]);
@@ -284,8 +285,9 @@ pub async fn start_pool(
     pool
 }
 
-pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
-    let template_provider = TemplateProvider::start(tp_port);
+pub async fn start_template_provider(tp_port: u16, sv2_interval: Option<u32>) -> TemplateProvider {
+    let sv2_interval = sv2_interval.unwrap_or(20);
+    let template_provider = TemplateProvider::start(tp_port, sv2_interval);
     template_provider.generate_blocks(16);
     template_provider
 }
--- a/roles/tests-integration/tests/pool_integration.rs
+++ b/roles/tests-integration/tests/pool_integration.rs
@@ -1,15 +1,15 @@
 mod common;
 
-use std::convert::TryInto;
-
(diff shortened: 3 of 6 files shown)


GitGab19:
@plebhash @jbesraa do you think we need to add a new integration test for this? We are able to assert specific fields value inside specific messages, right?

Also, let me know if you want me to bump `roles_logic_sv2` patch version here or not.

plebhash:
> do you think we need to add a new integration test for this? 

IMO more tests are always desirable

the real question is how complex it is to write a comprehensive test for this bug

if it is too complex and we decide it's not worth spending our time on it now, we could always take note of this on https://github.com/stratum-mining/stratum/issues/1077 and come back to it later

> We are able to assert specific fields value inside specific messages, right?

yes, see:

https://github.com/stratum-mining/stratum/blob/a58246ab04534f4a665afce9166a5004675118b0/roles/tests-integration/tests/common/sniffer.rs#L465-L470

or a more concrete example:
https://github.com/stratum-mining/stratum/blob/a58246ab04534f4a665afce9166a5004675118b0/roles/tests-integration/tests/pool_integration.rs#L40-L51

> Also, let me know if you want me to bump roles_logic_sv2 patch version here or not.

are we doing a patch release for this bugfix?

if yes, then we should bump and release right away

if not, then we should take note of this change and then bump before v1.3.0 (maybe there will be a MAJOR or MINOR bump that would overwrite this)

Sjors:
You should the `curtime` field, not `mintime`. For the timewarp scenario that this PR fixes, they're the same. But normally `mintime` is equal to MTP + 1, which on mainnet is typically an hour in the past. Whereas `curtime` is the node clock time, adjusted for MTP + 1 and for timewarp where needed.

GitGab19:
> You should the `curtime` field, not `mintime`. For the timewarp scenario that this PR fixes, they're the same. But normally `mintime` is equal to MTP + 1, which on mainnet is typically an hour in the past. Whereas `curtime` is the node clock time, adjusted for MTP + 1 and for timewarp where needed.

I'm not following you here, weren't we saying that we should always use the timestamp communicated by the TP?

Sjors:
Maybe I'm just confused by the name. The `getblocktemplate` RPC has both a `curtime` and `mintime` field. But the TP doesn't use that field and just gives you the `nTime` value in the block. That's _not_ the minimum time, so the variable name `mintime` is confusing. But it should work correctly.

GitGab19:
What I'm using in this PR is the `header_timestamp` field contained in the `SetNewPrevHash` message, as [specified here](https://github.com/stratum-mining/sv2-spec/blob/main/07-Template-Distribution-Protocol.md#73-setnewprevhash-server---client). It's correct, right?

Regarding the naming, do you think something could be optimized in the specs?

Sjors:
Yes, it's indeed `header_timestamp`.

The spec is clear:

> The nTime field in the block header at which the client should start (usually current time). This is NOT the minimum valid nTime value.

Some part of the SRI codebase is calling it `Some(referenced_job.min_ntime)`. I think that variable should be renamed from `min_ntime` to either `ntime` or `header_timestamp`.

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -162,6 +166,7 @@ pub fn extended_job_from_custom_job(
         0,
         true,
         extranonce_len,
+        Some(referenced_job.min_ntime),
```
Sjors, line 169: ^

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -207,13 +213,7 @@ fn new_extended_job(
 
     let min_ntime = match new_template.future_template {
```
Sjors, line 214: ^

GitGab19:
> Yes, it's indeed `header_timestamp`.
> 
> The spec is clear:
> 
> > The nTime field in the block header at which the client should start (usually current time). This is NOT the minimum valid nTime value.
> 
> Some part of the SRI codebase is calling it `Some(referenced_job.min_ntime)`. I think that variable should be renamed from `min_ntime` to either `ntime` or `header_timestamp`.

I think that's because on the mining protocol specs is specified as min_ntime: https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md#5315-newminingjob-server---client

Don't you think we should enhance that part of the specs? I can push changes on this PR, but I think that if we agree on it, we should also clarify specs.

Sjors:
Yes it makes sense to improve the spec there.

> Smallest nTime value available for hashing for the new mining job. 

There's currently no way to know this, unless we add a new field to the NewTemplate and/or SetNewPrevHash message.

GitGab19:
@Sjors took note of this improvements on both:
- specs (https://github.com/stratum-mining/sv2-spec/issues/119)
- SRI (https://github.com/stratum-mining/stratum/issues/1202#issuecomment-2569489461)

GitGab19:
@plebhash integration test added in e4e5e61a6c1058347c0e9ee12bda8a73b438c4c7.

Wrt version bumping, I would discuss about it during next `dev call`. So that we can better agree on the need about doing a patch release or not.

roles/tests-integration/tests/common/mod.rs
```diff
@@ -284,8 +285,8 @@ pub async fn start_pool(
     pool
 }
 
-pub async fn start_template_provider(tp_port: u16) -> TemplateProvider {
-    let template_provider = TemplateProvider::start(tp_port);
+pub async fn start_template_provider(tp_port: u16, sv2_interval: u32) -> TemplateProvider {
```
jbesraa, line 288: Instead of repeating `20` value for tests that don't need to set this, you can set a default `sv2_interval` value here.

`let sv2_interval = sv2_interval.unwrap_or(20)`

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -64,12 +66,128 @@ async fn success_pool_template_provider_connection() {
...
+async fn header_timestamp_value_assertion_in_new_extended_mining_job() {
+    let tp_pool_sniffer_addr = common::get_available_address();
+    let pool_translator_sniffer_addr = common::get_available_address();
+    let tp_addr = common::get_available_address();
+    let pool_addr = common::get_available_address();
+    let sv2_interval = 5;
+    let _tp = common::start_template_provider(tp_addr.port(), sv2_interval).await;
+
+    let tp_pool_sniffer =
+        common::start_sniffer("0".to_string(), tp_pool_sniffer_addr, tp_addr, false, None).await;
```
jbesraa, line 93: Both sniffers have `"0"` as the first argument which is the ID of the sniffer. The ID makes it easier to debug errors in the test, if they occur. 
It would be better to give unique names here.

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -64,12 +66,128 @@ async fn success_pool_template_provider_connection() {
...
+
+    let _ = common::start_pool(Some(pool_addr), Some(tp_pool_sniffer_addr)).await;
+
+    let pool_translator_sniffer = common::start_sniffer(
+        "0".to_string(),
```
jbesraa, line 98: ^

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -64,12 +66,128 @@ async fn success_pool_template_provider_connection() {
...
+        OpenExtendedMiningChannelSuccess
+    );
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        NewExtendedMiningJob
+    );
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        SetNewPrevHash
+    );
```
jbesraa, line 156: Do we actually need to assert all of this?

jbesraa (changes requested):
It is really great to see you using the integration tests!

Added a few nits mainly in that regard.

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -432,11 +432,11 @@ impl StrippedCoinbaseTx {
...
-        32 // outpoint
-        + 4 // vout
-        + 1 // script length byte -> TODO can be also 3 (based on TODO in `coinbase_tx_prefix()`)
-        + self.bip141_bytes_len // space for bip34 bytes
-        ;
+            32 // outpoint
+                + 4 // vout
+                + 1 // script length byte -> TODO can be also 3 (based on TODO in `coinbase_tx_prefix()`)
+                + self.bip141_bytes_len // space for bip34 bytes
+            ;
```
plebhash, line 439: what's the motivation behind this change?

did it come from `cargo fmt`?

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -64,12 +66,128 @@ async fn success_pool_template_provider_connection() {
...
+//   occurred with non-future jobs.
+//
+// Related issue: https://github.com/stratum-mining/stratum/issues/1324
+#[tokio::test]
+async fn header_timestamp_value_assertion_in_new_extended_mining_job() {
+    let tp_pool_sniffer_addr = common::get_available_address();
+    let pool_translator_sniffer_addr = common::get_available_address();
+    let tp_addr = common::get_available_address();
+    let pool_addr = common::get_available_address();
+    let sv2_interval = 5;
```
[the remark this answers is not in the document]
GitGab19: I set it to 5s so that a second template (and so a second job) is created in 5s. In this way we avoid to wait for too much time for this test.

roles/tests-integration/tests/pool_integration.rs
```diff
@@ -64,12 +66,128 @@ async fn success_pool_template_provider_connection() {
...
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        NewExtendedMiningJob
+    );
+    assert_mining_message!(
+        &pool_translator_sniffer.next_message_from_upstream(),
+        SetNewPrevHash
+    );
+
+    // Wait for a second NewExtendedMiningJob message
```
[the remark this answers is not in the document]
GitGab19: The focus of the test is on the second job, because the bug described by issue #1324 is manifesting on NON-future jobs. But when you connect a client to TP, TP sends a first future template (followed by its corresponding `SetNewPrevHash`), which becomes a future job sent downstream. After `sv2interval` seconds, a NON-future template is sent from the TP, which becomes a NON-future job sent downstream.

I hope this helps you clarify also the reason behind this comment: https://github.com/stratum-mining/stratum/pull/1325#discussion_r1905214342

roles/tests-integration/tests/pool_integration.rs
GitGab19: No, but I was not able to make it work without asserting (and so popping from the sniffer) all those messages. Any suggestion to avoid them?

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -207,13 +213,7 @@ fn new_extended_job(
...
     let min_ntime = match new_template.future_template {
         true => binary_sv2::Sv2Option::new(None),
-        false => {
-            let now = std::time::SystemTime::now()
-                .duration_since(std::time::UNIX_EPOCH)
-                .unwrap()
-                .as_secs() as u32;
-            binary_sv2::Sv2Option::new(Some(now))
-        }
+        false => binary_sv2::Sv2Option::new(ntime),
```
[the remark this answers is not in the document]
GitGab19: Done in b53152c0548cf849ecd4feb7910f32dca9d37164

roles/tests-integration/tests/common/mod.rs
GitGab19: Done in b53152c0548cf849ecd4feb7910f32dca9d37164

roles/tests-integration/tests/pool_integration.rs
GitGab19: Done in b53152c0548cf849ecd4feb7910f32dca9d37164

roles/tests-integration/tests/pool_integration.rs
GitGab19: I removed some of them in b53152c0548cf849ecd4feb7910f32dca9d37164

plebhash:
I did some local manual testing for the semver issue:
- make sure `cargo semver-checks` is 0.37.0 (same as CI)
- `cd protocols/v2/roles-logic-sv2; cargo +stable semver-checks --default-features`

no errors were reported

it's very good that @Shourya742 had the eagle eyes to detect this potential versioning issue, but we should definitely investigate deeper to know if:
- @Shourya742 was mistaken and a PATCH bump is sufficient
- OR we need to fix the CI, as it's not reliable

plebhash:
for future reference:

after some investigation, it was confirmed that the semver issue was a false alarm

although the `JobsCreators` is a public struct, the added field is private, which therefore does not categorize a MAJOR version bump

plebhash (approved):

PR #1328, opened 2025-01-06
remove `readme` field from `Cargo.toml` of crates that dont have a `README.md`

### LINKED ISSUE #1326: crates are missing reference to `README.md` on `Cargo.toml`
last week I did the v1.2.0 release

everything seemed fine, but now I just realized some problems on the `Release Libs` Workflow

https://github.com/stratum-mining/stratum/actions/runs/12519062802/job/34922524754

the publishing of some crates failed because the `Cargo.toml` has a `readme` field, but the crate doesn't have a `README.md` file

this happened to the following crates:
- `roles_logic_sv2`
- `sv2_ffi`
- `network_helpers_sv2`

plebhash:
close #1326

### DIFF
--- a/protocols/v2/roles-logic-sv2/Cargo.toml
+++ b/protocols/v2/roles-logic-sv2/Cargo.toml
@@ -3,7 +3,6 @@ name = "roles_logic_sv2"
 version = "1.2.2"
 authors = ["The Stratum V2 Developers"]
 edition = "2018"
-readme = "README.md"
 description = "Common handlers for use within SV2 roles"
 documentation = "https://docs.rs/roles_logic_sv2"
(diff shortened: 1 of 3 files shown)


jbesraa:
Why not adding readme file where it is missing?

plebhash:
> Why not adding readme file where it is missing?

because that will require considerable effort, some of which is part of other ongoing tasks (e.g.: #1263)

this was a miss from v1.2.0, which means we'll have to re-do the release

therefore it's best to just get it out of the way in the simplest possible way... we can always properly add these files later

GitGab19 (approved):
ACK

jbesraa (approved):

PR #1329, opened 2025-01-06
add `scripts/release-libs.sh`

### LINKED ISSUE #1327: `release-libs.yaml` has a footgun `continue-on-error: true`
during christmas week I did the v1.2.0 release

everything seemed fine, but now I just realized some problems on the `Release Libs` Workflow

https://github.com/stratum-mining/stratum/actions/runs/12519062802/job/34922524754#step:21:7

this specific problem is described in more detail on #1326

but we also need to avoid that similar problems happen again in the future

the root cause was the fact that every job on `release-libs.yaml` has `continue-on-error: true`

if we try to publish some crate version that already exists on crates.io, the `cargo publish` command fails

so `continue-on-error: true` was introduced (along with [some header comments as instructions](https://github.com/stratum-mining/stratum/blob/1f2c5e81b7c3a5f10bd4e82e2fd1a8d81a00edf9/.github/workflows/release-libs.yaml#L2)) to allow for all jobs to run, even if some other crate has already been published

the problem with this strategy is that it overly relies on human monitoring of those silent failures, as it's very easy for a human to miss some failed publishing (like I did on v1.2.0)

ideally we need the publishing workflow to be fully automated and explicitly fail in case something goes wrong, without relying on human monitoring of those silent failures

plebhash:
close #1327 

this PR adds a new `scripts/release-libs.sh`, which is used to publish each crate to `crates.io`

the script returns `0` on success of `cargo publish`, and `1` on failure
the only exception is when `cargo publish` fails because the crate is already published
in that case, the script also returns `0`

### DIFF
--- a/.github/workflows/release-libs.yaml
+++ b/.github/workflows/release-libs.yaml
@@ -1,12 +1,8 @@
 # This workflow is used to publish SV2 crates to crates.io
-# The workflow tries to update all the library crates, so if a crate is not to updated, the step will fail
-# for that each step have continue-on-error set to true.
-# Since each step can fail, the output ot the action must be manually checked to make sure that all
-# the library intended to be published are published.
-# Running cargo release in the various workspace help to prepare the version number and everything.
-# ATTENTION
-# Is very important to check the output manually cause when too many crates are updated crates.io could fail
-# and ask to rerun the action later
+# the workflow tries to publish all the library crates by running scripts/release-libs.sh
+# in case the `cargo publish` command fails, the script returns 1 and the entire workflow fails
+# the only exception is when the `cargo publish` command fails because the crate has already
+# been published, in which case the workflow continues
 
 name: Release Libs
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release-libs.yaml
```diff
@@ -1,12 +1,8 @@
...
-# for that each step have continue-on-error set to true.
-# Since each step can fail, the output ot the action must be manually checked to make sure that all
-# the library intended to be published are published.
-# Running cargo release in the various workspace help to prepare the version number and everything.
-# ATTENTION
-# Is very important to check the output manually cause when too many crates are updated crates.io could fail
-# and ask to rerun the action later
+# the workflow tries to publish all the library crates by running scripts/release-libs.sh
+# in case the `cargo publish` command fails, the script returns 1 and the entire workflow fails
+# the only exception is when the `cargo publish` command fails due to the crate already
```
jbesraa, line 4: what do you mean "already being published"? how could that happen?

.github/workflows/release-libs.yaml
```diff
@@ -32,113 +28,91 @@ jobs:
             override: true
       - name: Login
         run: cargo login ${{ secrets.CRATES_IO_DEPLOY_KEY }}
+
       - name: Publish crate common
-        continue-on-error: true
         run: |
-          cd common
-          cargo publish
+          ./scripts/release-libs.sh common
```
jbesraa, line 34: I suggest to move all the `./scripts/release-libs.sh [crate-name]` calls to a shell script with a for loop instead of the repeated code here.

scripts/release-libs.sh
```diff
@@ -0,0 +1,55 @@
...
+#
+# if --dry-run is provided, the script runs cargo publish --dry-run,
+# otherwise it performs a real publish.
+#
+# the script returns 0 on success of cargo publish, and 1 on failure
+# the only exception is when cargo publish fails because the crate is already published
+# in that case, the script also returns 0
+
+CRATE_DIR="$1"
+DRY_FLAG="${2:-}"
```
jbesraa, line 14: Is this being used somewhere? I think we should run a dry-run release on each pull request that gets merged into main branch.

jbesraa (changes requested):

Issue #1333, opened 2025-01-07
Add rust-toolchain.toml 


Shourya742:
We have observed instances where pull requests include unexpected formatting changes and Clippy fixes that were not present initially. This issue occurs because contributors are using different Rust versions, leading to inconsistencies caused by differences in toolchain behavior. To resolve this, we should add a `rust-toolchain.toml` file to the repository to enforce a consistent Rust version across all contributors. This will ensure that formatting and linting remain uniform, preventing unrelated changes from appearing in pull requests.

Few recent examples:
1. https://github.com/stratum-mining/stratum/pull/1319#pullrequestreview-2525455878

GitGab19:
With regards to the 2nd example noted here, I can tell that my local rust version is `1.75`. So it couldn't be a proper example of the issue you're describing here..

plebhash:
@Fi3 we were discussing this in a call earlier today

one of the questions that came up was why we're using `nightly` on the `fmt` workflow:

https://github.com/stratum-mining/stratum/blob/74830903e340698f8cc4e2d6f0247eab808e8796/.github/workflows/fmt.yaml#L30

@jbesraa mentioned that you had explained this a while ago over Discord but we couldn't find the messages

PR #1335, opened 2025-01-08
Add rust toolchain

Shourya742:
closes: #1333

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel = "1.75.0"
+profile = "default" 
+components = ["clippy", "rustfmt"]


jbesraa:
nack, our formatting works with nightly and always was. Also, you didnt provide any example where your statement is correct, your pr was not formatted with nightly and thats why you had so many changes.

We had this discussion a couple of times already with adding/not the rust-toolchain file, not sure why we are back at it again.

jbesraa:
The workflow uses always latest nightly https://github.com/stratum-mining/stratum/blob/main/.github/workflows/fmt.yaml

plebhash:
let's investigate on the need for using `nightly` on `fmt` CI and see how we can make progress on this

Georges760 (approved):
I made myself almost the exact same file in order to have my shell.nix synced with toolchain version

GitGab19:
Is this still needed? 

I haven't followed the discussion opened after this.

I would say it's not needed and we can close this PR, but I just want to be sure before proceeding.

jbesraa:
still big nack from me, I dont see any reason for this. whats described in #1333 happened (maybe) once and I think that was a problem with the authors text editor

GitGab19:
Closing this since we are using `stable` for clippy and `nightly` for fmt on our CI

PR #1337, opened 2025-01-09
Handle `cargo build` warnings

### LINKED ISSUE #1096: Cleanup cargo build warnings
Handle warnings generated when running cargo build, preferably in an incremental way with separate commit for each set of warnings

jbesraa:
resolves #1096  and #1048 

None of the dead code is removed and only marked with `dead_code` macro.

### DIFF
--- a/protocols/v2/binary-sv2/no-serde-sv2/codec/src/codec/mod.rs
+++ b/protocols/v2/binary-sv2/no-serde-sv2/codec/src/codec/mod.rs
@@ -112,6 +112,7 @@ pub trait Fixed {
 }
 
 // Not used and will be removed during refactoring
+#[allow(dead_code)]
 pub trait Variable {
     const HEADER_SIZE: usize;
     //const ELEMENT_SIZE: usize;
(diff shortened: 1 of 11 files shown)


GitGab19:
The ones generated in `protocols` crates should be tackled in the corresponding refactoring

jbesraa:
> The ones generated in `protocols` crates should be tackled in the corresponding refactoring

hmm.. it might take a while until we get there. Any reason why not to remove dead code before that?

GitGab19:
> > The ones generated in `protocols` crates should be tackled in the corresponding refactoring
> 
> hmm.. it might take a while until we get there. Any reason why not to remove dead code before that?

I would say just to remember to properly fix the corresponding warnings during refactoring PRs.. but I'm not totally against disabling dead_code as you are doing here.

jbesraa:
> > > The ones generated in `protocols` crates should be tackled in the corresponding refactoring
> > 
> > 
> > hmm.. it might take a while until we get there. Any reason why not to remove dead code before that?
> 
> I would say just to remember to properly fix the corresponding warnings during refactoring PRs.. but I'm not totally against disabling dead_code as you are doing here.

going with `dead_code` macro sounds like a good middle ground :+1:

GitGab19 (approved):
tACK

PR #1340, opened 2025-01-09
`integration_tests_sv2` docs cleanup

### LINKED ISSUE #1317: `integration-test` crate as a development tool for SRI ecosystem
imagine a future where the SRI-ecosystem uses Integration Testing for developing SRI-based apps

we would probably publish the ITF crate to `crates.io`

there's a few things that need to be addressed before this is possible:
- currently, this crate is called `integration-test`, which is obvious if the reader is inside the SRI repo and already has full context... but for a wider audience, it would be better to give this crate a more meaningful name with some kind of reference to SRI or Sv2
- the crate doesn't export any `lib`, which means the `common` module is not accessible, even if we add the crate as a dependency for another project

plebhash:
builds on top of #1339 

closes #1317

### DIFF
--- a/roles/tests-integration/lib/sniffer.rs
+++ b/roles/tests-integration/lib/sniffer.rs
@@ -35,21 +35,22 @@ enum SnifferError {
 
 /// Allows to intercept messages sent between two roles.
 ///
-/// The downstream (or client) role connects to the [`Sniffer`] `listening_address` and the
-/// [`Sniffer`] connects to the `upstream` server. This way, the Sniffer can intercept messages sent
-/// between the downstream and upstream roles. The downstream will send its messages to the
-/// [`Sniffer`] which will save those in the `messages_from_downstream` aggregator and forward them
-/// to the upstream role. When a response is received it is saved in `messages_from_upstream` and
-/// forwarded to the downstream role. Both `messages_from_downstream` and `messages_from_upstream`
-/// can be accessed as FIFO queues.
-///
-/// In order to alter the messages sent between the roles, the [`Sniffer::intercept_messages`]
-/// field can be used. It will look for the [`InterceptMessage::expected_message_type`] in the
-/// specified [`InterceptMessage::direction`] and replace it with
-/// [`InterceptMessage::replacement_message`].
-///
 /// Can be useful for testing purposes, as it allows to assert that the roles have sent specific
 /// messages in a specific order and to inspect the messages details.
+///
+/// The downstream (or client) role connects to the [`Sniffer`] `listening_address` and the
+/// [`Sniffer`] connects to the `upstream` server. This way, the Sniffer can intercept messages sent
+/// between the downstream and upstream roles.
+///
+/// Messages received from downstream are stored in the `messages_from_downstream` aggregator and
+/// forwarded to the upstream role. Alternatively, messages received from upstream are stored in
+/// the `messages_from_upstream` and forwarded to the downstream role. Both
+/// `messages_from_downstream` and `messages_from_upstream` aggregators can be accessed as FIFO
+/// queues via [`Sniffer::next_message_from_downstream`] and
+/// [`Sniffer::next_message_from_upstream`], respectively.
+///
+/// In order to replace the messages sent between the roles, a set of [`InterceptMessage`] can be
+/// used in [`Sniffer::new`].
 #[derive(Debug, Clone)]
 pub struct Sniffer {
     identifier: String,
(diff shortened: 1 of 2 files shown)


jbesraa (approved):

roles/tests-integration/lib/sniffer.rs
```diff
@@ -35,21 +35,22 @@ enum SnifferError {
 
 /// Allows to intercept messages sent between two roles.
 ///
+/// Can be useful for testing purposes, as it allows to assert that the roles have sent specific
+/// messages in a specific order and to inspect the messages details.
+///
```
plebhash, line 40: this paragraph was just moved from the bottom to the top, as it feels more like an introduction about `Sniffer`

it was a bit confusing to read it on the bottom, as it wasn't clear what it was referring to (`Sniffer` in general or some internal detail)

roles/tests-integration/lib/sniffer.rs
```diff
@@ -35,21 +35,22 @@ enum SnifferError {
...
 /// The downstream (or client) role connects to the [`Sniffer`] `listening_address` and the
 /// [`Sniffer`] connects to the `upstream` server. This way, the Sniffer can intercept messages sent
-/// between the downstream and upstream roles. The downstream will send its messages to the
-/// [`Sniffer`] which will save those in the `messages_from_downstream` aggregator and forward them
-/// to the upstream role. When a response is received it is saved in `messages_from_upstream` and
-/// forwarded to the downstream role. Both `messages_from_downstream` and `messages_from_upstream`
-/// can be accessed as FIFO queues.
```
plebhash, line 45: this was re-rewitten below, as it implies that messages from upstream are always a response, which is not necessarily true

roles/tests-integration/lib/sniffer.rs
```diff
@@ -35,21 +35,22 @@ enum SnifferError {
...
+/// between the downstream and upstream roles.
 ///
-/// In order to alter the messages sent between the roles, the [`Sniffer::intercept_messages`]
-/// field can be used. It will look for the [`InterceptMessage::expected_message_type`] in the
-/// specified [`InterceptMessage::direction`] and replace it with
-/// [`InterceptMessage::replacement_message`].
```
plebhash: this was replaced with a new paragraph about `InterceptMessage` in the bottom in a more summarized way

roles/tests-integration/lib/sniffer.rs
```diff
@@ -31,29 +31,26 @@ type MsgType = u8;
...
-/// In order to alter the messages sent between the roles, the [`Sniffer::intercept_messages`]
-/// field can be used. It will look for the [`InterceptMessage::expected_message_type`] in the
-/// specified [`InterceptMessage::direction`] and replace it with
-/// [`InterceptMessage::response_message`].
+/// between the downstream and upstream roles.
 ///
-/// If `break_on` is set to `true`, the [`Sniffer`] will stop the communication after sending the
-/// response message.
+/// Messages received from downstream are stored in the `messages_from_downstream` aggregator and
+/// forward them to the upstream role. Alternatively, messages received from upstream are stored in
```
GitGab19, line 46:
```suggestion
/// forwarded to the upstream role. Alternatively, messages received from upstream are stored in
```

roles/tests-integration/lib/sniffer.rs
```diff
@@ -65,29 +62,27 @@ pub struct Sniffer {
     intercept_messages: Vec<InterceptMessage>,
 }
 
+/// Allows [`Sniffer`] to replace some intercepted message before forwarding it.
```
GitGab19, line 65: I would expand this comment by explaining the meaning of some parameters (e.g. `break_on`)

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -10,10 +10,14 @@ use roles_logic_sv2::{
 use sniffer::{InterceptMessage, MessageDirection};
 use std::convert::TryInto;
 
+// this test aims to assert that Sniffer is able to intercept and replace some message
```
GitGab19, line 13:
```suggestion
// This test aims to assert that Sniffer is able to intercept and replace some messages.
```

roles/tests-integration/lib/sniffer.rs
plebhash: `break_on` is being removed via https://github.com/stratum-mining/stratum/pull/1339, which this PR builds on top of

while I was working on this PR, for a moment I had the same idea as you, but then I felt that all the other parameter names feel reasonably self-explanatory... I agree that `break_on` wasn't too obvious but now that it's gone I feel the user can probably understand everything from this short line

plebhash: also, please keep in mind that those struct members are private, this is how it would render:

image

the comments explaining each one should be added on the constructor `InteceptMessage::new`

plebhash: ok ended up adding these comment lines to the constructor:

image

PR #1341, opened 2025-01-09
`Sniffer` messages cleaner

### LINKED ISSUE #1336: need  new method `Sniffer::has_message_type_with_remove`
while reviewing #1325 we detected the need for a new `Sniffer` method for Integration Tests

this should help us avoid repeating assertions across multiple tests with the sole purpose of popping from the `Sniffer` queue

@jbesraa wrote this draft on Discord:
```rust
    fn has_message_type_with_remove(&self, message_type: u8) -> bool {
        let has_message: bool = self
            .messages
            .safe_lock(|messages| {
                for (t, _) in messages.iter() {
                    if *t == message_type {
                        messages.retain(|(_, _)| false);
                        return true; // Exit early with `true`
                    }
                }
                false // Default value if no match is found
            })
            .unwrap();
        has_message
    }
```

jbesraa:
resolves #1336

### DIFF
--- a/roles/tests-integration/lib/sniffer.rs
+++ b/roles/tests-integration/lib/sniffer.rs
@@ -423,8 +423,8 @@ impl Sniffer {
         }
     }
 
-    /// used to block the test runtime
-    /// while we wait until Sniffer has received a message of some specific type
+    /// Waits until a message of the specified type is received into the `message_direction`
+    /// corresponding queue.
     pub async fn wait_for_message_type(
         &self,
         message_direction: MessageDirection,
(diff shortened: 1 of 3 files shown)


roles/tests-integration/tests/pool_integration.rs
```diff
@@ -92,24 +95,12 @@ async fn header_timestamp_value_assertion_in_new_extended_mining_job() {
...
-    );
-    assert_mining_message!(
-        &pool_translator_sniffer.next_message_from_upstream(),
-        NewExtendedMiningJob
-    );
-    assert_mining_message!(
-        &pool_translator_sniffer.next_message_from_upstream(),
-        SetNewPrevHash
-    );
+    pool_translator_sniffer
```
jbesraa, line 98: @GitGab19

plebhash: should we take the opportunity on this PR and also apply this new function to the test introduced on https://github.com/stratum-mining/stratum/pull/1325 ?

since #1325 was where this feature request was born, seems like a good way to validate the new feature

plebhash: oh duh nevermind, that's exactly what's already being done here

plebhash (approved):

roles/tests-integration/lib/sniffer.rs
```diff
@@ -456,6 +456,53 @@ impl Sniffer {
             sleep(Duration::from_secs(1)).await;
         }
     }
+
+    pub async fn wait_for_message_type_with_remove(
```
GitGab19, line 460: Couldn't we call this something like `wait_for_message_type_and_clean_queue`? It seems more clear to understand what it does. Otherwise I would suggest to put a comment over this function to explain and describe what it does.

GitGab19: Actually I would add a comment over the function in any case :)

roles/tests-integration/lib/sniffer.rs
```diff
@@ -456,6 +456,56 @@ impl Sniffer {
             sleep(Duration::from_secs(1)).await;
         }
     }
+
+    /// Similar to `[Sniffer::wait_for_message_type]` but also removes the message from the queue
```
GitGab19, line 460:
```suggestion
    /// Similar to `[Sniffer::wait_for_message_type]` but also removes the messages from the queue
```

GitGab19:
Can you fix fmt errors?

GitGab19 (approved):

PR #1342, opened 2025-01-09
use `target` dir for integration tests TP

### LINKED ISSUE #1278: Test TemplateProvider OS E13

plebhash:
close #1278

unfortunately #1293 didn't fix it

this PR changes the directory where TP is downloaded into, replacing `/tmp` (which was giving permission errors in an undeterministic way) to the `target` directory crated by `cargo` (which shouldn't ever have permission issues)

### DIFF
--- a/roles/tests-integration/lib/template_provider.rs
+++ b/roles/tests-integration/lib/template_provider.rs
@@ -67,10 +67,13 @@ pub struct TemplateProvider {
 
 impl TemplateProvider {
     pub fn start(port: u16, sv2_interval: u32) -> Self {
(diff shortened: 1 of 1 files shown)


jbesraa:
Ha, it is happening all over the place suddenly.

jbesraa (changes requested):
nack. writing to the same folder cargo writes to can lead to unexpected behavior. it would also make the local size of the project much bigger. 
maybe create a specific folder under `tests-integration` and clean it after the `TemplateProvider` is dropped.

plebhash:
closing in favor of #1359

PR #1354, opened 2025-01-16
Test Translator, JD and miner setup

### LINKED ISSUE #1353: Move interop-jd-translator from mg to integration test
https://github.com/stratum-mining/stratum/tree/main/test/message-generator/test/interop-jd-translator

jbesraa:
resolves #1353, #1474 and #1475

~~This is currently not working as expected. Same problem faced previously with #1207 where ghost processes are not exiting. Probably something need to be fixed at #1319/#1321~~


plebhash:
perhaps this is also suffering from the issues reported on https://github.com/stratum-mining/stratum/issues/1420

jbesraa:
This is now unblocked after making [this](https://github.com/stratum-mining/stratum/pull/1354/commits/99729554198873290f48d69b66f50d52f3fd476b) modification to the miner

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -4,14 +4,18 @@
...
+    MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED,
+};
 use integration_tests_sv2::{sniffer::*, *};
 use roles_logic_sv2::parsers::{CommonMessages, Mining, PoolMessages};
 
 // This test runs an sv2 translator between an sv1 mining device and a pool. the connection between
 // the translator and the pool is intercepted by a sniffer. The test checks if the translator and
 // the pool exchange the correct messages upon connection. And that the miner is able to submit
 // shares.
+#[ignore]
```
plebhash, line 18: as pointed out during PR review club today, this was left-over from testing and needs to be removed

test/message-generator/test/sv1-test/sv1-test.json
```diff
@@ -92,7 +92,7 @@
                         "pool_sv2",
                         "--",
                         "-c",
-                        "../test/config/interop-jd-translator/pool-config.toml"
```
plebhash, line 95: why change this?

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -51,3 +55,59 @@ async fn translate_sv1_to_sv2_successfully() {
...
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED,
+        )
+        .await;
+    jdc_pool_sniffer
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_MINING_SET_NEW_PREV_HASH,
+        )
+        .await;
```
plebhash, line 112: why are we waiting for a new `SetNewPrevHash` here?

jbesraa: To fully cover this test https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/pool-sri-test-1-standard/pool-sri-test-1-standard.json

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -51,3 +55,59 @@ async fn translate_sv1_to_sv2_successfully() {
...
+    jdc_pool_sniffer
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED,
+        )
+        .await;
```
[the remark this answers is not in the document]
jbesraa: I dont think it is necessary because we have the second SetNewPrevHash which indicates a previous submission was successful. But added it anyway.

test/message-generator/test/sv1-test/sv1-test.json
jbesraa: This is is using the pool-config from `interop-jd-translator` test which is deleted in this PR. I left the `pool-config.toml` in so sv1-test keep working

jbesraa:
Addressed comments and rebased

GitGab19:
@jbesraa you pointed this PR to issue #1476 in the description.

But #1476 is related to standard channels, and SV2 mining device.

This PR is not covering that, can you confirm? In case, update the PR description please

roles/tests-integration/tests/translator_integration.rs
plebhash: I don't think we should keep it, but the reasons go a bit deeper than extended vs standard channels.

Theoretically speaking, a Pool should never send a `SetNewPrevHash` to a JDC. A mining connection between JDC and Pool is used for Custom Jobs. If you check `SetCustomMiningJob`, you will see that it contains a `prev_hash` field, which JDC got from a `SetNewPrevHash` sent by TP.

The only reason that we can see Pool sending `SetNewPrevHash` to JDC is because our dummy Pool doesn't distinguish between its downstream connections. Still, JDC simply ignores this message.

But we definitely shouldn't assert for this, because that would be kind of enforcing behavior that shouldn't even exist in the first place.

roles/tests-integration/tests/translator_integration.rs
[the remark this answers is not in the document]
plebhash: on the `pool-sri-test-1-standard.json` MG test, this second `SetNewPrevHash` only happened because the mocked TP send it 2x.

It has nothing to do with an acknowledgement for share submission.

I think it's important to clarify this distinction, because this phrase below makes me wonder whether you're holding some misguided assumptions:

> SetNewPrevHash which indicates a previous submission was successful.

The only reason we **happen** to see a `SetNewPrevHash` on this integration test is because the difficulty set on tProxy is high enough to qualify for a valid block on regtest. So as soon as Sv1 CPU miner finds a share, that is already a block.

And as stated here https://github.com/stratum-mining/stratum/pull/1354#discussion_r1973787357, `SetNewPrevHash` should never happen in a connection between JDC and Pool, which is another reason we shouldn't assert this.

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -51,3 +54,53 @@ async fn translate_sv1_to_sv2_successfully() {
         )
         .await;
 }
+
+// Test full flow with Translator and Job Declarator. An SV1 mining device is connected to
+// Translator and is successfully submitting shares to the pool. It also asserts that Pool role
+// sends a subsequent `SET_NEW_PREV_HASH`.
```
plebhash, line 60: these comments are outdated

test/message-generator/test/sv1-test/sv1-test.json
plebhash: this is breaking CI https://github.com/stratum-mining/stratum/actions/runs/13586446905/job/37986127705#step:3:592

PR #1359, opened 2025-01-18
Integration tests avoid tmp dir

### LINKED ISSUE #1278: Test TemplateProvider OS E13

plebhash:
closes #1278, alternative to #1342

unfortunately #1293 didn't fix #1278

this PR changes the directory where TP is downloaded into, replacing `/tmp` (which was giving permission errors in an undeterministic way) to a new `template-provider` directory crated at the Integration Test execution (which shouldn't ever have permission issues)

this new directory is listed on a new `.gitignore` so that we don't commit binary executables to the SRI repo, which would be bad practice

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ cobertura.xml
 /roles/*/*-config.toml
 /examples/*/Cargo.lock
 /scripts/sv2.h
+/roles/tests-integration/template-provider
--- a/roles/tests-integration/README.md
+++ b/roles/tests-integration/README.md
@@ -22,5 +22,9 @@ $ git clone git@github.com:stratum-mining/stratum.git
 $ cargo test --manifest-path=roles/Cargo.toml --verbose --test '*' -- --nocapture
 ```
 
+Note: during the execution of the tests, a new directory called `template-provider` is created.
+This directory holds the executable for Template Provider node, as well as the different data 
+directories created for each execution.
+
 ## License
 MIT OR Apache-2.0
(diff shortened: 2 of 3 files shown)


plebhash:
> ACK on the current changes. However, I am not entirely sure about the implementation of removing the template provider directory in the `Drop` trait. If this is done, all the tests might take significantly longer since they would need to download the binary every time or might fail due to download errors. [#1342 (review)](https://github.com/stratum-mining/stratum/pull/1342#pullrequestreview-2555031362).

can you elaborate? where are we removing the TP directory? there's no `Drop` trait implementation for the `template_provider` module, and the `Drop` trait implementation from underlying `bitcoind` crate simply stops the process.

plebhash:
> > > ACK on the current changes. However, I am not entirely sure about the implementation of removing the template provider directory in the `Drop` trait. If this is done, all the tests might take significantly longer since they would need to download the binary every time or might fail due to download errors. [#1342 (review)](https://github.com/stratum-mining/stratum/pull/1342#pullrequestreview-2555031362).
> > 
> > 
> > can you elaborate? where are we removing the TP directory? there's no `Drop` trait implementation for the `template_provider` module, and the `Drop` trait implementation from underlying `bitcoind` crate simply stops the process.
> 
> I was mentioning about Esraa's suggestion...

ah ok sorry I missed the link on the end of the comment

yeah I didn't go in that direction... instead, marked the new directory on `.gitignore`

roles/tests-integration/.gitignore
```diff
@@ -0,0 +1 @@
+template-provider
```
jbesraa, line 1: There is a single `.gitignore` in the project root used for all workspace crates

jbesraa (commented):
LGTM. Do you think this is something worth mentioning in the readme?  we are writing to a user disk and we might end up taking a good chunk of space.

GitGab19 (approved):

jbesraa (approved):

Issue #1348, opened 2025-01-16
Test JDC shutdown behavior


jbesraa:
https://github.com/stratum-mining/stratum/pull/1321
Introduced a `shutdown` object to `JobDeclaratorClient` but add a unit test to cover its functionality. 

Ideally something like the following should be added as a unit test:

```rust
#[cfg(test)]
mod tests {
    use ext_config::{Config, File, FileFormat};

    use crate::*;

    #[test]
    fn test_shutdown() {
        let config_path = "config-examples/jdc-config-hosted-example.toml";
        let config: ProxyConfig = match Config::builder()
            .add_source(File::new(config_path, FileFormat::Toml))
            .build()
        {
            Ok(settings) => match settings.try_deserialize::<ProxyConfig>() {
                Ok(c) => c,
                Err(e) => {
                    dbg!(&e);
                    return;
                }
            },
            Err(e) => {
                dbg!(&e);
                return;
            }
        };
        let jdc = JobDeclaratorClient::new(config.clone());
        let cloned = jdc.clone();
        tokio::spawn(async move {
            cloned.start().await;
        });
        jdc.shutdown();
        let ip = config.downstream_address.clone();
        let port = config.downstream_port;
        let jdc_addr = format!("{}:{}", ip, port);
        assert!(std::net::TcpListener::bind(jdc_addr).is_ok());
    }
}
```

jbesraa:
This is mainly raised because of this test https://github.com/stratum-mining/stratum/blob/main/test/message-generator/test/jds-do-not-panic-if-jdc-close-connection/jds-do-not-panic-if-jdc-close-connection.json

We can migrate this test easily to ITF with the following:

```rust
#[tokio::test]
async fn jds_should_not_panic_if_jdc_shutsdown() {
    let (_tp, tp_addr) = start_template_provider(None).await;
    let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
    let (_jds, jds_addr) = start_jds(tp_addr).await;
    let (jdc, jdc_addr) = start_jdc(vec![pool_addr], tp_addr, jds_addr).await;
    // This shutdown behavior is not working as expected as `jdc_addr` is not available
    jdc.shutdown();
    let ret = tokio::net::TcpListener::bind(jdc_addr).await;
    dbg!(&ret);
    assert!(ret.is_ok());
    let (sniffer, sniffer_addr) = start_sniffer("0".to_string(), jds_addr, false, None).await;
[...]

Issue #1349, opened 2025-01-16
Change JDC behavior on shutdown


jbesraa:
after #1321, JDC would shutdown with `process::exit(0)` https://github.com/stratum-mining/stratum/pull/1321/files#diff-b344c0ca590735a830cbfb638690b53af5889deeef68717d4586b5da8af62479R173

This behavior can be problematic in functions where the JDC is initialized with other roles as JDC shutdown will result in shutting down all other roles.

plebhash:
closed via #1360

Issue #1320, opened 2024-12-23
Killing a running process(role) in integration tests


jbesraa:
Sometimes we need the ability to stop a role in order to perform testing on the behaviour of other roles. For example if jd-client connects to jd-server and then jd-client goes offline, we want to test how jd-server behaves, is it still able to accept connections from new jd-client(s). the problem is that we don't have access to the async communication channels started within the roles to send a shutdown signal. I wonder if someone can think of a neat way to achieve this without doing a hell of refactor in the code..

GitGab19:
Closing this since we were able to migrate all tests to the ITF.

Let's reopen more specific issues when we are gonna need them in the future

PR #1360, opened 2025-01-20
Add jds-do-not-fail-on-jdc-shutdown integration test

Shourya742:
Closes: #1348 #1320 #1349 

1. On side this PR changes the shutdown api on tproxy and jdc: 
```rust
pub fn shutdown(&self) 
```
2. Add unit test for these
3. Add IT for jds-do-not-fail-on-jdc-shutdown
4. Remove jds-do-not-fail-on-jdc-shutdown MG test


.github/workflows/mg.yaml
```diff
@@ -43,14 +43,6 @@ jobs:
...
         run: sh ./test/message-generator/test/jds-do-not-fail-on-wrong-tsdatasucc/jds-do-not-fail-on-wrong-tsdatasucc.sh
 
-  jds-do-not-panic-if-jdc-close-connection:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-      - name: Run jds-do-not-panic-if-jdc-close-connection
-        run: sh ./test/message-generator/test/jds-do-not-panic-if-jdc-close-connection/jds-do-not-panic-if-jdc-close-connection.sh
-
```
plebhash, line 53: we should also delete `test/message-generator/test/jds-do-not-panic-if-jdc-close-connection/`

roles/jd-client/src/lib/mod.rs
```diff
@@ -368,8 +373,10 @@ impl JobDeclaratorClient {
         .await;
     }
 
-    pub fn shutdown(&self) -> Arc<Notify> {
-        self.shutdown.clone()
+    #[allow(dead_code)]
+    pub fn shutdown(&self) {
+        dbg!("I am here");
```
plebhash, line 378: was this left here on purpose? seems a left-over from debugging or development?

roles/translator/src/lib/mod.rs
```diff
@@ -301,3 +301,42 @@ fn kill_tasks(task_collector: Arc<Mutex<Vec<(AbortHandle, String)>>>) {
         }
     });
 }
+
+#[cfg(test)]
```
plebhash, line 305: why are we adding this unit test to tProxy? isn't the issue/PR about JDC?

we should either:
- add integration + unit test to both JDC and tProxy (and make it clear on the PR description)
- add integration + unit test to JDC only (maybe tProxy on a different PR)

the current approach feels a bit weird, as we're doing integration test for JDC and unit test for tProxy, and the motivation for that is not clear to me

plebhash: ah ok, now I see that we're adding the unit test on both JDC and tProxy

plebhash (approved):

roles/jd-client/src/lib/mod.rs
```diff
@@ -368,8 +368,9 @@ impl JobDeclaratorClient {
         .await;
     }
 
-    pub fn shutdown(&self) -> Arc<Notify> {
-        self.shutdown.clone()
+    #[allow(dead_code)]
+    pub fn shutdown(&self) {
```
jbesraa, line 377: Could you please document this function? it would be a good practice to start using the `roles` as well

roles/jd-client/src/lib/mod.rs
```diff
@@ -170,7 +170,7 @@ impl JobDeclaratorClient {
                     },
                     _ = self.shutdown.notified().fuse() => {
                         info!("Shutting down gracefully...");
-                        std::process::exit(0);
+                        break 'outer;
```
jbesraa, line 178: In `inner` loop, `break` is used to exit current(`inner`) loop. I would use `break` here as well to keep it consistent.

roles/jd-client/src/lib/mod.rs
```diff
@@ -165,11 +166,15 @@ impl JobDeclaratorClient {
                             }
                         } else {
                             info!("Received unknown task. Shutting down.");
+                            handler.abort();
+                            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
jbesraa, line 170: why the sleep?

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -44,3 +44,18 @@ async fn translation_proxy() {
...
 }
+
+#[tokio::test]
+async fn jds_should_not_panic_if_jdc_shutsdown() {
+    let (_tp, tp_addr) = start_template_provider(None).await;
+    let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
+    let (_jds, jds_addr) = start_jds(tp_addr).await;
+    let (jdc, jdc_addr) = start_jdc(pool_addr, tp_addr, jds_addr).await;
+    jdc.shutdown();
+    tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```
jbesraa, line 55: this should not be needed anymore, if so, it is probably 1-2 secs

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -44,3 +44,18 @@ async fn translation_proxy() {
...
+    let ret = tokio::net::TcpListener::bind(jdc_addr).await;
+    assert!(ret.is_ok());
```
jbesraa, line 57: this assert can be moved to the line above it 
`assert!(tokio::net::TcpListener::bind(jdc_addr).await.is_ok());`

roles/tests-integration/tests/translator_integration.rs
```diff
@@ -44,3 +44,18 @@ async fn translation_proxy() {
         )
         .await;
 }
+
+#[tokio::test]
+async fn jds_should_not_panic_if_jdc_shutsdown() {
```
jbesraa, line 49: this test should be added in a new test file `tests/jd_integration`

roles/jd-client/src/lib/mod.rs
```diff
@@ -375,7 +375,6 @@ impl JobDeclaratorClient {
 
     #[allow(dead_code)]
     pub fn shutdown(&self) {
-        dbg!("I am here");
```
jbesraa: This should be removed in the commit it was introduced in

test/config/jds-do-not-panic-if-jdc-close-connection/jds-config.toml
```diff
@@ -1,20 +0,0 @@
-# SRI Pool config
-authority_public_key = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72"
-authority_secret_key = "mkDLTBBRxdBv998612qipDYoTK3YUrqLe8uWw7gu3iXbSrn2n"
```
jbesraa, line 3: IMO deleting the MG files and CI appearances of the same test should be in the same commit, so I would combine this with the one before it

jbesraa (changes requested):
Looks super with the tests!

roles/jd-client/src/lib/mod.rs
jbesraa: aint this match arm in the `outer` loop?

roles/jd-client/src/lib/mod.rs
```diff
@@ -415,3 +415,41 @@ impl PoolChangerTrigger {
         }
     }
 }
+
+#[cfg(test)]
```
jbesraa, line 419: Can you please move this test to https://github.com/stratum-mining/stratum/pull/1360/commits/de393492d4cf3dbb8ff8efc8d9d216a338629c62

roles/translator/src/lib/mod.rs
```diff
@@ -288,8 +288,13 @@ impl TranslatorSv2 {
             task_collector.safe_lock(|t| t.push((task.abort_handle(), "init task".to_string())));
     }
 
-    pub fn shutdown(&self) -> Arc<Notify> {
-        self.shutdown.clone()
+    /// Notifies the translator to shut down gracefully.
```
jbesraa, line 291:
```
Close Translator role and any open connection associated with it.

Note that this method will result in a full exit of the  running Translator and any open connection most be re-initiated upon new start.
```

roles/translator/src/lib/mod.rs
```diff
@@ -288,8 +288,13 @@ impl TranslatorSv2 {
...
+    ///
+    /// This method triggers the shutdown process by sending a notification.
+    /// It ensures that any ongoing operations are properly handled before
+    /// the translator stops functioning.
```
jbesraa, line 295: I think this part is redundant

jbesraa (changes requested):
LGTM. Added few nits. Could you also please apply the docs changes to JDC::shutdown?

.github/workflows/mg.yaml
plebhash: hey @Shourya742 check out this error: https://github.com/stratum-mining/stratum/actions/runs/13083583326

we forgot to remove the following lines:

https://github.com/stratum-mining/stratum/blob/4637a824decaec51a6efdcf64e9ce06c9d008d4d/.github/workflows/mg.yaml#L151

https://github.com/stratum-mining/stratum/blob/4637a824decaec51a6efdcf64e9ce06c9d008d4d/.github/workflows/mg.yaml#L170

they are used to aggregate the MG tests... the fact that we're leaving them behing is the last thing preventing this PR from being merged

PR #1352, opened 2025-01-16
Handle unwrap when shutting down JDC server

jbesraa:
1st commit(9a35baae51adf3d64ebdd222fd003db86e818f0c) resolves #1189 

2nd and 3rd resolves #1351

PR #1373, opened 2025-01-21
avoid running workflows AFTER merging PRs

### LINKED ISSUE #1372: consider avoid running workflows AFTER merging PRs
most of our CI workflows are set to that we run them before and after PRs are merged

```yml
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

I'm not sure about the motivation for running them after the PRs have already been merged... personally speaking, I never really check that, I only care about CI before the PRs are merged (as a safeguard against introducing bugs etc)

given that our repo can only run a limited number of workflows at the same time, this is actually introducing a significant bottleneck that severely slows down the process of merging PRs, especially when there's multiple PRs already approved and buffered for merging

unless I'm missing some important detail for the motivation for this, IMO we should stop running workflows after the PRs have been merged

plebhash:
close #1372


jbesraa (commented):
what if you push again to this pull request, would that rerun ci?

plebhash:
> what if you push again to this pull request, would that rerun ci?

just rebased against `main`, CI is re-running from scratch

jbesraa:
> > But it would have been ideal if we can have merge policies a bit sorted, like unrelated change need not to be rebased or stuff.
> 
> Github has a new UI for merging PRs. It's currently under "preview" mode but I opted it and apparently we can merge PRs even if they are "out-of-date" with the base branch.
> 
> Here's what I currently see on #1352
> 
> image
> 
> So maybe we won't need the `ready-to-be-marged` label so often anymore. Let's see.

Well, if we are going to use this button, we should definitely keep the checks on main as well. I would vote for this option.

plebhash:
marked the two comments above as outdated

the new Github UI seems a bit buggy, and IMO we should just keep rebasing PRs against `main` and only run CI before the PR is merged

GitGab19 (approved):

PR #1376, opened 2025-01-22
Add shutdown API to JDS along with task handler

Shourya742:
Partially Closes: #1320  

This PR introduces a shutdown API, similar to the implementation in JDC and TProxy. Additionally, to facilitate a graceful shutdown, we have added a task handler that manages the lifecycle of all tasks spawned within the `start` method.


jbesraa:
Dunno really. I understand the urgency but in the same time we lived with this for a bit now. Lets see what others think.

GitGab19:
Do we still need this PR?

Don't we have a way to shutdown JDS?

GitGab19:
But it seems to me that we were able to migrate all tests without needing this.

Shouldn't we close this for now and open specific issues when we are gonna need them?

Issue #1377, opened 2025-01-22
Make the pool wait for the template provider to start


Shourya742:
Currently, the pool exits if the template provider is not running. This behavior is incorrect; the pool should wait and attempt to connect to the template provider instead.

PR #1378, opened 2025-01-22
Add retry logic to ensure the pool waits and connect to TP

Shourya742:
closes: #1377

### DIFF
--- a/roles/pool/src/lib/template_receiver/mod.rs
+++ b/roles/pool/src/lib/template_receiver/mod.rs
@@ -18,7 +18,7 @@ use roles_logic_sv2::{
 };
 use std::{convert::TryInto, net::SocketAddr, sync::Arc};
 use tokio::{net::TcpStream, task};
-use tracing::info;
+use tracing::{info, warn};
 
 mod message_handler;
 mod setup_connection;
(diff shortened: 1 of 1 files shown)


roles/pool/src/lib/mod.rs
```diff
@@ -32,17 +32,32 @@ impl PoolSv2 {
...
-            s_prev_hash,
-            r_solution,
-            r_message_recv_signal,
-            status::Sender::Upstream(status_tx.clone()),
-            coinbase_output_len,
-            tp_authority_public_key,
-        )
-        .await?;
+
+        loop {
```
jbesraa, line 36: Here you loop through the template provider connection regardless of the error, not only if TP is not up. Instead the loop should be added here https://github.com/stratum-mining/stratum/blob/main/roles/pool/src/lib/template_receiver/mod.rs#L48

jbesraa (changes requested):

plebhash (approved):

jbesraa (approved):
I would change the sleep to 1 second tbh, but wont block this - I can change it later

plebhash:
> > I would change the sleep to 1 second tbh, but wont block this - I can change it later
> 
> Let me change and push,

let's squash the commits into one

PR #1380, opened 2025-01-22
add shutdown api to pool

Shourya742:
closes: #1320


jbesraa (commented):
Same as with #1376, I think we should wait with this for a bit

Issue #1385, opened 2025-01-23
Replace bitcoind Crate with corpc-node in Integration Tests


Shourya742:
**Description**  
The `bitcoind` crate is currently being used in our integration tests. However, this crate is no longer maintained, and all development has been transitioned to `corpc-node`.  

**Task**  
Replace the `bitcoind` crate in the integration tests with `corpc-node` to ensure compatibility and maintainability moving forward.

jbesraa:
Related to https://github.com/stratum-mining/stratum/issues/1278

PR #1386, opened 2025-01-23
replace bitcoind with corpc-node crate

Shourya742:
closes: #1385

### DIFF
--- a/roles/tests-integration/lib/template_provider.rs
+++ b/roles/tests-integration/lib/template_provider.rs
@@ -1,4 +1,4 @@
-use bitcoind::{bitcoincore_rpc::RpcApi, BitcoinD, Conf};
+use corepc_node::{Conf, Node};
 use flate2::read::GzDecoder;
 use std::{
     env,
@@ -62,7 +62,7 @@ fn get_bitcoind_filename(os: &str, arch: &str) -> String {
 
 #[derive(Debug)]
 pub struct TemplateProvider {
-    bitcoind: BitcoinD,
+    bitcoind: Node,
 }
 
 impl TemplateProvider {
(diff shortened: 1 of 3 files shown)


jbesraa (approved):
LGTM

roles/tests-integration/lib/template_provider.rs
```diff
@@ -127,24 +127,18 @@ impl TemplateProvider {
         }
 
         env::set_var("BITCOIND_EXE", bitcoin_exe_home.join("bitcoind"));
-        let exe_path = bitcoind::exe_path().unwrap();
+        let exe_path = corepc_node::exe_path().unwrap();
```
plebhash, line 130: should we take this opportunity to clean up these `unwrap`s?

if this is unfailable, we should at least replace them with `expect`.

plebhash: please note that this code is not a test, technically speaking

it's a library that enables testing

roles/tests-integration/lib/template_provider.rs
```diff
@@ -127,24 +127,18 @@ impl TemplateProvider {
...
 
-        let bitcoind = BitcoinD::with_conf(exe_path, &conf).unwrap();
+        let bitcoind = Node::with_conf(exe_path, &conf).unwrap();
```
plebhash, line 132: same here

roles/tests-integration/lib/template_provider.rs
```diff
@@ -127,24 +127,18 @@ impl TemplateProvider {
...
 
     pub fn generate_blocks(&self, n: u64) {
-        let mining_address = self
-            .bitcoind
-            .client
-            .get_new_address(None, None)
-            .unwrap()
-            .require_network(bitcoind::bitcoincore_rpc::bitcoin::Network::Regtest)
-            .unwrap();
+        let mining_address = self.bitcoind.client.new_address().unwrap();
```
plebhash, line 138: same here

roles/tests-integration/lib/template_provider.rs
```diff
@@ -127,24 +127,18 @@ impl TemplateProvider {
...
         self.bitcoind
             .client
-            .generate_to_address(n, &mining_address)
+            .generate_to_address(n as usize, &mining_address)
             .unwrap();
```
plebhash, line 142: same here

roles/tests-integration/lib/template_provider.rs
plebhash: > Since this is not a user-facing library

why not? I fundamentally disagree with this assumption, I believe we should always see this crate as something that could eventually be used externally

see https://github.com/stratum-mining/stratum/issues/1317

and that is precisely why I said this:
> please note that this code is not a test, technically speaking
>
> it's a library that enables testing

---

anyways, I'm not arguing for adding a new error enum, the proposal is simply to replace `unwrap` with `expect`

plebhash: > We have MG for interoperability which can be used by SRI ecosystem for compatibility, like thats why we have MG?

if you're writing some Sv2 app in a different language, then yeah MG is your only option

but if you're writing some Sv2 app in Rust, you can choose between ITF and MG, and I would def not recommend MG when there's a way better option available

plebhash (approved):
tACK

Issue #1395, opened 2025-01-24
Ambiguous NewMiningJob stratum-message format


jakubtrnka:
NewMiningJob is defined here:
https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md#5315-newminingjob-server---client

the merkle-root is supposed to be 32-bytes array.
It is, however, declared as B32, which can easily be mistaken for a generic byte array Bytes0_32 which has a one-byte length prefix

"B32" is not defined in the section of types.
https://github.com/stratum-mining/sv2-spec/blob/main/03-Protocol-Overview.md#31-data-types-mapping

It's a question how to exactly reword the specification - Bytes type of length 32 or the U256? But in any case B0_32 is incorrect and is the reason why
SRI is not compatible with Braiins SV2 miners

Sjors:
Nice catch!

PR #1396, opened 2025-01-24
Braiins/cross testing

jakubtrnka:
Fix for [#1395](https://github.com/stratum-mining/stratum/issues/1395)

There has been some incompatibility between braiins miner and SRI pool. I was looking into it and found the root cause of the issue. It was in Merkle Root.
1. I added an independent test-vector for the merkle-root calculation. It worked correctly.
2. Then I figured out that SRI pool sends a 33 bytes merkle-root array, while braiins miner expects only 32 byte array. I wrote a unit test for the expected message format
3. I fixed the message format to pass the test
4. Tested on Braiins Miniminer BMM100 with nightly firmware release. - it successfully mined on on the SRI pool and template provider https://github.com/Sjors/bitcoin/releases/tag/sv2-tp-0.1.13 Submits produced by miner were accepted

Please note that the unit tests are probably not in the correct place. I didn't have time to try to find the best place for them.


plebhash:
overall I think this is in a good direction, and almost done

a few remarks:
- we're breaking the `Rustfmt` CI... ideally we should fix it by running `cargo +nightly fmt` for each commit (as opposed to adding a new "fmt" commit)
- I would place the `new_mining_job_serialization` test inside `roles_logic_sv2::parsers` module, not inside `pool_sv2` crate.
- I would squash the two last commits, since they are:
  - fixing the problem on `NewMiningJob` serialization
  - adding a test to enforce this fix

protocols/v2/subprotocols/mining/src/new_mining_job.rs
```diff
@@ -2,7 +2,7 @@
 use alloc::vec::Vec;
 #[cfg(not(feature = "with_serde"))]
 use binary_sv2::binary_codec_sv2;
-use binary_sv2::{Deserialize, Seq0255, Serialize, Sv2Option, B032, B064K, U256};
+use binary_sv2::{Deserialize, Seq0255, Serialize, Sv2Option, B064K, U256};
```
Sjors, line 5: If B032 doesn't exist in the spec, maybe `binary_sv2` shouldn't define it?

More rigorously, should it wiped from the codebase?

Sjors: Oh I see, it's still used for `extranonce` and `extranonce_prefix`.

Out of scope for this PR, but is it really worth having a separate `B0_32` datatype instead of just using `B0_255` and then having the spec limit the extranonce size to 31?

It would be less confusing, because `extranonce_size` is U16 which allows for much bigger values. (U8 would have made more sense, but not worth changing now).

Adding an explanation as to why that limit was picked would be good too.

plebhash (approved):

PR #1401, opened 2025-01-27
Sniffer fixes

### LINKED ISSUE #1384: Look into `Sniffer::recv_from_up_send_to_down` as it seems broken

jbesraa:
resolves #1384

### DIFF
--- a/roles/tests-integration/lib/sniffer.rs
+++ b/roles/tests-integration/lib/sniffer.rs
@@ -234,8 +234,10 @@ impl Sniffer {
                         )
                         .expect("Failed to create the frame"),
                     );
-                    downstream_messages
-                        .add_message(msg_type, intercept_message.replacement_message.clone());
+                    downstream_messages.add_message(
+                        intercept_message.replacement_message.message_type(),
+                        intercept_message.replacement_message.clone(),
+                    );
                     let _ = send.send(frame).await;
                 }
             }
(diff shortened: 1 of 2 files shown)


roles/tests-integration/lib/sniffer.rs
```diff
@@ -219,35 +219,35 @@ impl Sniffer {
...
-                            channel_msg,
-                        )
-                        .expect("Failed to create the frame"),
-                    );
-                    downstream_messages.add_message(
-                        intercept_message.replacement_message.message_type(),
+            let intercept_message = intercept_messages.iter().find(|im| {
+                im.direction == MessageDirection::ToDownstream
+                    && im.expected_message_type == msg_type
+            });
```
plebhash, line 225: I assume this was copy-pasted for both functions?

the direction here has to be `ToUpstream`

roles/tests-integration/lib/sniffer.rs
```diff
@@ -256,30 +258,33 @@ impl Sniffer {
...
+                send.send(frame)
+                    .await
+                    .map_err(|_| SnifferError::DownstreamClosed)?;
             }
-            if send.send(frame).await.is_err() {
-                return Err(SnifferError::DownstreamClosed);
-            };
-            upstream_messages.add_message(msg_type, msg);
         }
         Err(SnifferError::UpstreamClosed)
```
plebhash, line 289: same here... shouldn't this be `DownstreamClosed`?

I'm not sure about the direction, but it definitely feels wrong to return the same error type for both functions

jbesraa: This is `recv_from_up_send_to_down`, if something fails while sending, it should return` DownstreamClosed` and if it fails when receiving it should return `UpstreamClosed`.  The incorrect appearance was at `recv_from_down_send_to_up`, which is fixed now.

plebhash (approved):

Issue #1402, opened 2025-01-27
remove `async_std` support


plebhash:
maintaining support for `async_std` in SRI substantially increases the engineering efforts

meanwhile, the SRI application layer is built with `tokio`, which is already quite challenging to maintain

therefore, we should drop the `async_std` support to make things simpler and cleaner, and most importantly: something that allows our limited engineering resources to sustainably achieve production readiness

SRI application layer (`roles`, `network_helpers_sv2`, `integration_tests_sv2`) should only officially support `tokio`

plebhash:
related:

issues:
- https://github.com/stratum-mining/stratum/issues/1389

PRs:
- https://github.com/stratum-mining/stratum/pull/1007
- https://github.com/stratum-mining/stratum/pull/977

plebhash:
cc @Fi3 , just pinging you so you're aware

does Demand heavily rely on anything `async_std`?

plebhash:
re-opening, as we still have some left-overs:

- `network_helpers_sv2` has `async_std` modules
- `benches` makes use of `network_helpers_sv2` on `async_std` mode

I'll send a follow-up PR soon

plebhash:
> > re-opening, as we still have some left-overs:
> > 
> > * `network_helpers_sv2` has `async_std` modules
> > * `benches` makes use of `network_helpers_sv2` on `async_std` mode
> > 
> > I'll send a follow-up PR soon
> 
> It's ready on my fork... I will open the PR..

sorry only saw this now

PR #1404, opened 2025-01-27
move Sv1 example to crate

plebhash:
while investigating #1402, I noticed that the Sv1 example currently lives on the `examples` directory without any meaningful motivation

this PR moves that example code to the `sv1_api` crate, while also removing `async_std` and `async_channel` from its dependencies

---

note for reviewers:

this example code actually feels widely overengineered... the commit making the necessary adaptations for removal of `async_std` and `async_channel` could prove challenging to review, and IMHO it's not really worth the effort of spending too much time doing that

maybe someday we can re-write a simplified example... but for now, I'd point out that the code compiles and execution remains the same as before... that should be sufficient in terms of reviewing the changes here


GitGab19:
Why did you remove the `async_std` and `async_channel` stuff in this PR?
It was just to avoid adding those dependencies to the `sv1` crate, right? 

In that case, I would leave the review of that part to @Shourya742 or @jbesraa since they are working on channels

Moving this to the `sv1` crate makes sense to me btw

plebhash:
> Why did you remove the async_std and async_channel stuff in this PR?
It was just to avoid adding those dependencies to the sv1 crate, right?

correct

> In that case, I would leave the review of that part to @Shourya742 or @jbesraa since they are working on channels.

please note what I mentioned in the PR description about this:

> this example code actually feels widely overengineered... the commit making the necessary adaptations for removal of async_std and async_channel could prove challenging to review, and IMHO it's not really worth the effort of spending too much time doing that

> maybe someday we can re-write a simplified example... but for now, I'd point out that the code compiles and execution remains the same as before... that should be sufficient in terms of reviewing the changes here

tbh if the review process here becomes too complex (as there's obviously better ways to write examples for this crate), I would actually prefer one of these options:
- simply delete this example from the codebase
- not do anything about this example

as I mentioned above, this example feels widely overengineered anyways...

GitGab19 (approved):

PR #1407, opened 2025-01-27
move `interop-cpp` examples to `sv2-ffi` crate

plebhash:
similar to https://github.com/stratum-mining/stratum/pull/1404

this PR moves the `interop-cpp` examples to `sv2-ffi` crate

builds on top of #1405


GitGab19 (approved):

plebhash:
turns out that adapting the scripts is not as simple as I initially thought

not sure this is worth spending time now, so I'll just close this PR

PR #1411, opened 2025-01-29
Remove async channel from Pool

Shourya742:
Partially closes: stratum-mining/sv2-apps#17

Notes for reviewers:

Run pool using this `RUSTFLAGS="--cfg tokio_unstable" RUST_LOG=debug cargo run -- -c config-examples/pool-config-local-tp-example.toml` as I have added console_subscriber for observing the task.


GitGab19:
Should we close this for now @Shourya742 ?

PR #1417, opened 2025-01-29
remove `sv2-ffi` crate

### LINKED ISSUE #1416: consider removing `sv2-ffi` crate from the repo
As discussed [here](https://github.com/stratum-mining/stratum/discussions/895#discussioncomment-9314350), for historical reasons, the FFIs were started with the intention to be imported into Bitcoin Core (TP).

As a consequence, functionality of `sv2-ffi` is currently limited to only a few SRI crates, and expanding functionality to fully cover Sv2 is a considerable engineering effort (originally tracked by https://github.com/stratum-mining/stratum/issues/919, which has since been closed).

The Template Provider strategy has since evolved, and it no longer is a meaningful motivation for maintaining FFIs.

While one could argue potential firmware integration as a motivation for maintaining FFIs (e.g.: https://github.com/plebhash/cgminer-sv2), the Embedded Rust ecosystem is very mature in 2025, and we even have projects in the community that are already implementing full-Rust firmware (e.g.: https://github.com/bitaxeorg/esp-miner-rs).

Meanwhile, keeping the `sv2-ffi` crate has been a meaningful burden. Whenever we change APIs on some of the crates that are covered by the FFIs, we have to do extra steps to keep `sv2-ffi` up-to-date (e.g.: run `scripts/sv2-header-check.sh` and make sure `sv2.h` is not broken).

Recently I also tried to clean up the `examples/interop-cpp*` examples via https://github.com/stratum-mining/stratum/pull/1407, but eventually gave up as the effort didn't really seem to justify.

So my proposal here is: IMHO we should delete `sv2-ffi` from the repository.

If we ever get feedback from some adopter that they do need this, we can either:
- start a FFI coverage task-force from scratch (which would be my preferred approach)
- ressurect `sv2-ffi` from the git history

plebhash:
close #1416


GitGab19 (commented):
Still two things to remove:
- `.github/workflows/sv2-header-check.yaml `
- step called `interop-test` in `ci` job of `.github/workflows/test.yaml`

GitGab19:
Since we didn't find any consensus on this, maybe it's better to close this PR for now.

We can always reopen it in the future.

What do you think @plebhash?

Issue #1421, opened 2025-01-30
Point crates in Message Generator to point at crates.io and not relative workspace path.


Shourya742:

plebhash:
> this will render the MG useless to test PRs so maybe we should do that only that we have all the new test framework in place.

I disagree.

where the low-level crates came from (local or crates.io) when compiling the MG executable is not the relevant for those tests.

those tests leverage MG to assert the correct message flow between roles.

as long as:
- MG executable builds (which will happen if we use dependencies from crates.io)
- the JSON files are still doing `cargo run` with the local roles implementation

then the MG tests are still doing what they are supposed to do.

if we accidentally introduce a bug on a low-level crate, it will break roles, which will break the MG test.

PR #1422, opened 2025-01-30
point MG crate to crate.io, than using protocol relative path

Shourya742:
closes: #1421


plebhash (approved):

utils/message-generator/Cargo.toml
GitGab19, line 1: I think you missed to change `key-utils`'s path

GitGab19 (approved):

PR #1430, opened 2025-01-31
Add `tracing-subscriber` to itf

### LINKED ISSUE #1370: Consider adding a logger to `tests-integration`

jbesraa:
This should allow us to examine the logs from the roles that are using `tracing` crate for logging. As a default only ERROR logs will be shown, in order to examine a different level of logging, `RUST_LOG` env variable should be modified.

For example `RUST_LOG=info cargo test --test '*'`

resolves #1370

### DIFF
--- a/roles/tests-integration/lib/mod.rs
+++ b/roles/tests-integration/lib/mod.rs
@@ -3,19 +3,30 @@ use jd_client::JobDeclaratorClient;
 use jd_server::JobDeclaratorServer;
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
 use pool_sv2::PoolSv2;
-use translator_sv2::TranslatorSv2;
-
 use rand::{thread_rng, Rng};
 use std::{
     convert::{TryFrom, TryInto},
     net::SocketAddr,
     str::FromStr,
+    sync::Once,
 };
+use tracing_subscriber::EnvFilter;
+use translator_sv2::TranslatorSv2;
 use utils::get_available_address;
 
 pub mod sniffer;
 pub mod template_provider;
 mod utils;
+
+static LOGGER: Once = Once::new();
+
+pub fn start_tracing() {
+    LOGGER.call_once(|| {
+        tracing_subscriber::fmt()
+            .with_env_filter(EnvFilter::from_default_env())
+            .init();
+    });
+}
 
 pub async fn start_sniffer(
     identifier: String,
--- a/roles/tests-integration/tests/jd_integration.rs
+++ b/roles/tests-integration/tests/jd_integration.rs
@@ -1,9 +1,15 @@
+// This file contains integration tests for the `JDC/S` module.
+//
+// `JDC/S` are modules that implements the Job Decleration roles in the Stratum V2 protocol.
+//
+// Note that it is enough to call `start_tracing()` once in the test suite to enable tracing for
+// all tests. This is because tracing is a global setting.
 use integration_tests_sv2::*;
 
 use roles_logic_sv2::parsers::{CommonMessages, PoolMessages};
 
-// This test verifies that the `jds` (Job Distributor Server) does not panic when the `jdc`
-// (Job Distributor Client) shuts down.
+// This test verifies that the `jds` (Job Decleration Server) does not panic when the `jdc`
+// (Job Decleration Client) shuts down.
 //
 // The test follows these steps:
 // 1. Start a Template Provider (`tp`) and a Pool.
--- a/roles/tests-integration/tests/pool_integration.rs
+++ b/roles/tests-integration/tests/pool_integration.rs
@@ -1,10 +1,9 @@
-use integration_tests_sv2::*;
-
 use crate::sniffer::MessageDirection;
 use const_sv2::{
     MESSAGE_TYPE_MINING_SET_NEW_PREV_HASH, MESSAGE_TYPE_NEW_EXTENDED_MINING_JOB,
     MESSAGE_TYPE_NEW_TEMPLATE,
 };
+use integration_tests_sv2::*;
 use roles_logic_sv2::{
     common_messages_sv2::{Protocol, SetupConnection},
     parsers::{AnyMessage, CommonMessages, Mining, PoolMessages, TemplateDistribution},
@@ -16,6 +15,7 @@ use roles_logic_sv2::{
 // Pool will connect to the Sniffer, and the Sniffer will connect to the Template Provider.
 #[tokio::test]
 async fn success_pool_template_provider_connection() {
+    start_tracing();
     let (_tp, tp_addr) = start_template_provider(None).await;
     let (sniffer, sniffer_addr) = start_sniffer("".to_string(), tp_addr, true, None).await;
     let _ = start_pool(Some(sniffer_addr)).await;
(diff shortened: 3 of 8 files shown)


roles/tests-integration/tests/pool_integration.rs
```diff
@@ -16,6 +17,9 @@ use roles_logic_sv2::{
 // Pool will connect to the Sniffer, and the Sniffer will connect to the Template Provider.
 #[tokio::test]
 async fn success_pool_template_provider_connection() {
+    tracing_subscriber::fmt()
```
jbesraa, line 20: This is initialized in a single test in each file. This way it works for all tests in the file.. it is repetitive but could not think about a better way..

plebhash:
> As a default only ERROR logs will be shown, in order to examine a different level of logging, RUST_LOG env variable should be modified.

I think for local execution it's fine to leave this default, but we should consider doing `RUST_LOG=debug` on CI for max verbosity, hopefully it could help us debug un-deterministic behavior

roles/tests-integration/tests/pool_integration.rs
plebhash: I tried to follow the suggestion from @Shourya742 above (with a few adaptations to get it to compile)

this was added to `lib/mod.rs`:
```rust
pub static LOGGER: once_cell::sync::Lazy<()> = once_cell::sync::Lazy::new(|| {
    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
        .init();
});
```

that way I could replace:
```rust
#[tokio::test]
async fn test_name() {
  tracing_subscriber::fmt()
    .with_env_filter(EnvFilter::from_default_env())
    .init();
  ...
}
```

with this one-liner:
```rust

#[tokio::test]
async fn test_name() {
  *LOGGER;
  ...
}
```

plebhash: while the experiment above did remove the amount of lines, it can also be achieved in a much simpler way

we can simply add this to `lib/mod.rs`:
```rust
pub fn tracing_subscribe() {
    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
        .init();
}
```

and then also use a one-liner on the test:
```rust

#[tokio::test]
async fn test_name() {
  tracing_subscribe();
  ...
}
```

therefore using `once_cell::sync::Lazy` feels like over-engineering IMHO

plebhash: I also saw this error when trying to use the `once_cell::sync::Lazy` approach:

```
failures:

---- test_sniffer_intercept_to_downstream stdout ----
thread 'test_sniffer_intercept_to_downstream' panicked at /Users/plebhash/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-subscriber-0.3.19/src/fmt/mod.rs:517:14:
Unable to install global subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

I only added the `*LOGGER` line to one single test

plebhash: deleted my previous comments here as I just realized I was writing nonsense... apologies for the noise!

please ignore them in case you're reading notifications on Discord

anyways, I could not get the suggestion from @Shourya742 above to work, as there seems to be no `call_once` API on the `once_cell` crate:

```
error[E0423]: expected value, found crate `once_cell`
  --> tests-integration/lib/mod.rs:24:5
   |
24 |     once_cell.call_once(|| {
   |     ^^^^^^^^^- help: use the path separator to refer to an item: `::`

```

I suspect @Shourya742 mixed this with `std::sync::Once`.

---

but indeed, the problem with the current approach is:
- it's repetitive, as we're forced to initialize the subscriber on each file
- it's a bit counter-intuitive, because if we initialize the subscriber more than 1x on each file, it breaks with:

```
Unable to install global subscriber: SetGlobalDefaultError("a global default trace dispatcher has already been set")
```

not the end of the world... just something we would have to be careful with...

but one middle-ground approach (that I validated locally) could be to add this to `lib/mod.rs`:
```rust
use tracing_subscriber::EnvFilter;
use std::sync::Once;

static INIT: Once = Once::new();

pub fn init_tracing() {
    INIT.call_once(|| {
        tracing_subscriber::fmt()
            .with_env_filter(EnvFilter::from_default_env())
            .init();
    });
}
```

then call `init_tracing()` on the test.

it should be sufficient to call `init_tracing()` only 1x per file and it will enable logging on all tests of such file, just like it's currently done on this PR.

however, the advantage here is:

if for some reason `init_tracing()` is called on multiple tests on the same file, it will not break with the error mentioned above.

jbesraa:
Made changes based on the discussion above. Thank you.

roles/tests-integration/tests/jd_integration.rs
```diff
@@ -1,3 +1,9 @@
+// This file contains integration tests for the `JDC/S` module.
+//
+// `JDC/S` is a module that implements the Job Distributor role in the Stratum V2 protocol.
```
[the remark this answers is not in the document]
plebhash: ref https://github.com/stratum-mining/stratum/issues/1065

roles/tests-integration/tests/pool_integration.rs
plebhash: @jbesraa the `ctor` approach above seems like an interesting alternative to the current approach of initializing the logger 1x per file

roles/tests-integration/lib/mod.rs
```diff
@@ -3,20 +3,31 @@ use jd_client::JobDeclaratorClient;
...
 pub mod template_provider;
 mod utils;
 
+static LOGGER: Once = Once::new();
+
+pub fn start_tracing() {
+    LOGGER.call_once(|| {
+        tracing_subscriber::fmt()
+            .with_env_filter(EnvFilter::from_default_env())
+            .init();
```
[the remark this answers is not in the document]
jbesraa: Hmm current approach seems more versatile - allowing us to have different level in CI and locally. I prefer to keep it simple as is.

roles/tests-integration/tests/pool_integration.rs
jbesraa: not entirely sold on this. adding new dep just for this feels too much, and current approach is not too bad IMO. If you both think otherwise, happy to add it though.

roles/tests-integration/lib/mod.rs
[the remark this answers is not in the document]
jbesraa: current approach does not require setting RUST_LOG, if it is not set it will fallback to ERROR

jbesraa:

roles/tests-integration/tests/pool_integration.rs
jbesraa: I added docs at the top of the file regarding docs. I would appreciate if we unblock this.

plebhash: I agree with @Shourya742 in the sense that the code could look a bit more clean, cohesive and intuitive with regards to this issue of tracer subscription

but I also agree with @jbesraa that we shouldn't spend too much time refining this, as the context of this PR is more in the "nice-to-have" category and we have more important things to focus on

as time goes on we can create new issues and submit new PRs to continue refining things on ITF

Issue #1420, opened 2025-01-29
JDC does not exchange messages with TP in Integration Tests


plebhash:
while working on #1419 I noticed that JDC is not connecting to TP

here's a little snippet that illustrates the problem:

```rust
#[tokio::test]
async fn jdc_test() {
    let (_tp, tp_addr) = start_template_provider(None).await;
    let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
    let (_jds, jds_addr) = start_jds(tp_addr).await;

    let (sniffer_pool, sniffer_pool_addr) =
        start_sniffer("0".to_string(), pool_addr, false, None).await;

    let (sniffer_jds, sniffer_jds_addr) =
        start_sniffer("1".to_string(), jds_addr, false, None).await;

    let (sniffer_tp, sniffer_tp_addr) =
        start_sniffer("2".to_string(), tp_addr, false, None).await;

    let (_jdc, jdc_addr) = start_jdc(sniffer_pool_addr, sniffer_tp_addr, sniffer_jds_addr).await;
    let (_translator, tproxy_addr) = start_sv2_translator(jdc_addr).await;

    println!("sniffer_pool {:?}", sniffer_pool.next_message_from_downstream());
    println!("sniffer_pool {:?}", sniffer_pool.next_message_from_upstream());

    println!("sniffer_jds {:?}", sniffer_jds.next_message_from_downstream());
    println!("sniffer_jds {:?}", sniffer_jds.next_message_from_upstream());
    
    println!("sniffer_tp {:?}", sniffer_tp.next_message_from_downstream());
    println!("sniffer_tp {:?}", sniffer_tp.next_message_from_upstream());
}
```

the behavior:
```
    Finished test [optimized + debuginfo] target(s) in 0.91s
     Running tests/jds_integration.rs (/Users/plebhash/develop/stratum/roles/target/debug/deps/jds_integration-caf345f2043f64aa)

running 1 test
sniffer_pool Some((0, Common(SetupConnection(SetupConnection { protocol: MiningProtocol, min_version: 2, max_version: 2, flags: 6, endpoint_host: Owned([48, 46, 48, 46, 48, 46, 48]), endpoint_port: 50, vendor: Owned([]), hardware_version: Owned([]), firmware: Owned([]), device_id: Owned([]) }))))
sniffer_pool Some((1, Common(SetupConnectionSuccess(SetupConnectionSuccess { used_version: 2, flags: 6 }))))
[...]

Issue #1418, opened 2025-01-29
JDS connects to wrong RPC port on Integration Tests


plebhash:
Currently, `start_jds` function takes `tp_address` (which contains Sv2 port for Template Distribution Protocol).

Then, it tries to connect to the Sv2 port as if it was RPC.

https://github.com/stratum-mining/stratum/blob/c30fd6dc6bb0e7f261f441ca3c387847d02472f6/roles/tests-integration/lib/mod.rs#L158-L179

I'm not sure why this is not breaking anything (yet), but this is definitely not right.

plebhash:
an important note related to this:

thinking that JDS connects to a Sv2 TP is a common misconception

the truth is that JDS doesn't even take part in the Template Distribution Protocol, and therefore it does not even need a Sv2-patched bitcoin node

the only roles that really connect to a TP (and take part in TDP) are:
- Pool
- JDC

while working on https://github.com/stratum-mining/sv2-spec/pull/121, more specifically the diagrams, I made sure to make a clear distinction (mostly in colors and legends) about which node is a TP and which is just a regular node

![](https://github.com/plebhash/sv2-spec/raw/upgrade-jdp/img/jd_template_mode.png)

jbesraa:
IIUC JDS is all about managing a local mempool, how would that work without a connection to TP?

GitGab19:
JDS maintains a light local version of the mempool, but it uses the RPC port of the node to get those information, not the same used to get templates (`8442`).

plebhash:
> IIUC JDS is all about managing a local mempool, how would that work without a connection to TP?

please look at the blue connections on the diagram above

plebhash:
closed via https://github.com/stratum-mining/stratum/pull/1429

plebhash:
> > IIUC JDS is all about managing a local mempool, how would that work without a connection to TP?
> 
> please look at the blue connections on the diagram above

rectifying what I said here, as this diagram was updated and now there's an orange TP connected to JDS+Pool

https://github.com/plebhash/sv2-spec/blob/upgrade-jdp/img/jd_template_mode.png

technically speaking, JDS still and Pool don't **need** a TP... both can get all the on-chain stateful info via RPC, ZMQ or whatever

but Pool does have the possibility of receiving the on-chain stateful info via TDP, and it's better to showcase that in the diagram

PR #1434, opened 2025-02-01
[WIP] Integration Test for disconnecting downstream connections on all roles

### LINKED ISSUE #1432: Integration Test for disconnecting downstream connections on all roles
https://github.com/stratum-mining/stratum/issues/1426 made us notice an edge case that ideally we should cover in CI for all roles that accept downstream connections.

the tests should look something like this:

- launch role under test (together with all upstream roles it needs as a pre-requisite)
- connect at least two downstreams to the role under test
- kill one downstream

the test should assert that the role stays alive (and doesn't die like in #1426)

plebhash:
close #1432

WIP, blocked by #1420 and #1418

- adds a new `is_listening` API to `pool_sv2`, `jd_server`, `jd_client`, and `translator_sv2` crates
- adds integration tests described in #1432

note: `mining_proxy_sv2`crate does not follow the same pattern of providing a struct representing the role (where `is_listening` API was added)... as a consequence, its listening port is not accessible in the test environment and therefore this role is not being covered on this PR

### DIFF
--- a/roles/jd-client/src/lib/job_declarator/mod.rs
+++ b/roles/jd-client/src/lib/job_declarator/mod.rs
@@ -11,7 +11,7 @@ use roles_logic_sv2::{
     template_distribution_sv2::SetNewPrevHash,
     utils::{hash_lists_tuple, Mutex},
 };
-use std::{collections::HashMap, convert::TryInto, str::FromStr};
+use std::{collections::HashMap, convert::TryInto};
 use stratum_common::bitcoin::{util::psbt::serialize::Deserialize, Transaction};
 use tokio::task::AbortHandle;
 use tracing::{error, info};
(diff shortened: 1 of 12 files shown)


jbesraa (commented):
hmm.. this "is_listening" is nice to use in testing env from the outside but I dont see how this is related to any of the roles. All of the is_listening functions can be "is_port_open" in test env and give it the role port(thats why the roles in test env return their SocketAddr). 
If is_listening had access to the actual socket and checked if it can receive messages, that would be great - but checking with TcpStream::connect and calling the function "is_listening" gives the wrong impression

plebhash:
> hmm.. this "is_listening" is nice to use in testing env from the outside but I dont see how this is related to any of the roles. All of the is_listening functions can be "is_port_open" in test env and give it the role port(thats why the roles in test env return their SocketAddr). If is_listening had access to the actual socket and checked if it can receive messages, that would be great - but checking with TcpStream::connect and calling the function "is_listening" gives the wrong impression

so are you suggesting that:
- this API should be renamed to more accurately reflect its functionality
OR
- this API is not a sufficient criteria for what we're aiming to test

?

plebhash:
this was a bad idea

Issue #1387, opened 2025-01-23
Decoupling Message Generator and Deprecating Serde Variant in Binary_SV2


Shourya742:
Currently, we maintain two variants of `binary_sv2`: the serde variant and the no-serde variant. The serde variant was originally retained to support the Message Generator. However, since we are now working to decouple the Message Generator into a separate repository, maintaining the serde variant of `binary_sv2` is no longer necessary. This issue will serve as a tracker for all required changes to complete the migration and deprecate the serde variant.

### Action Items:

- [x] 1. **Decouple the Message Generator**: Move the Message Generator to its own repository. The new repository will depend on protocol crates that reference a previously released version of the protocol where serde was still supported as a feature flag. This task is currently led by @plebhash, and the ongoing development can be tracked here:
 GitHub Repository - Stratum Message Generator: https://github.com/stratum-mining/stratum-message-generator

- [x] 2. **Deprecate the serde variant of `binary_sv2`**: Designate the no-serde variant as the primary and sole variant of `binary_sv2`. Additionally, remove all `with_serde` feature flags throughout the roles and protocol.
     - [x] https://github.com/stratum-mining/stratum/pull/1454
     - [x] https://github.com/stratum-mining/stratum/pull/1455
     - [x] https://github.com/stratum-mining/stratum/pull/1459

- [x] 3. **Refining the current MG Test [workflow](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/mg.yaml) to checkout the MG from its future repo**
     - [x] #1421
     - [x] https://github.com/stratum-mining/stratum/pull/1428

Both task 1 and 2 are disjoint, and share no dependence.

jbesraa:
for (2) we should include roles workspace as well

GitGab19:
We should also add another point in this list which is:
- refining the current `MG Test` [workflow](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/mg.yaml) to checkout the MG from its future repo

In this way we could move it out and still run those tests in our CI, while migrating every test.

GitGab19:
Closing after last task has been addressed with https://github.com/stratum-mining/stratum/pull/1428

PR #1455, opened 2025-02-05
remove serde feature from protocol crate

Shourya742:
Part of: #1387

### DIFF
--- a/protocols/v2/framing-sv2/examples/sv2_frame.rs
+++ b/protocols/v2/framing-sv2/examples/sv2_frame.rs
@@ -14,36 +14,22 @@
 // ## Run
 //
 // ```
-// cargo run --example sv2_frame --features with_serde
+// cargo run --example sv2_frame
 // ```
 
-#[cfg(feature = "with_serde")]
-use binary_sv2::{GetSize, Serialize};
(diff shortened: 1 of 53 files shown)


protocols/v2/framing-sv2/examples/sv2_frame.rs
plebhash, line 1: do we really need to delete the entire example?

isn't there a way to keep it, just not using `serde`?

plebhash: writing this example was an outcome of the documentation effort

it would be a pity to lose it

plebhash:
looks like we're going to have to bump MAJOR for all the crates where the `with_serde` feature is being removed

https://github.com/stratum-mining/stratum/actions/runs/13152202385/job/36701547815?pr=1455

protocols/v2/framing-sv2/examples/sv2_frame.rs
plebhash: thanks, but please note that this should have been fixed on the original commit, not adding a new commit 

doing this way effectively erases the history of the original file

I'm rebasing the branch to fix it

plebhash (approved):

PR #1442, opened 2025-02-03
Rename `pool_signature` field of `channel_factory` into `additional_coinbase_script_data`, and change the type from `String` to `Vec<u8>`

plebhash:
close #1436

loosely inspired by https://github.com/stratum-mining/stratum/pull/1248/commits/3f2815b5a10992158f4b04529eeb71c3c902302b from @Fi3 at #1248, but with a few corrections (and split into two different commits)

note: this renaming only happens on `roles_logic_sv2`, but not on SRI application layer (`roles`), where the field remains called `pool_signature`

PR #1461, opened 2025-02-06
remove assumption that coinbase scriptSig prefix contains only BIP34

### LINKED ISSUE #1460: SRI should not assume that coinbase `scriptSig` prefix contains only BIP34
in this discussion https://github.com/stratum-mining/stratum/discussions/1256#discussioncomment-12060362 it was noted that SRI (more specifically `job_creator` module of `roles_logic_sv2`) seems to be based on the assumption that the coinbase `scriptSig` prefix contains only BIP34

while this is true now, that's not a safe assumption

[to quote](https://github.com/stratum-mining/stratum/discussions/1256#discussioncomment-12065541) @Sjors :

> ... In any case, one day Bitcoin Core might add more stuff to the scriptSig after the BIP34 commitment. Or it might stop adding the useless 0 for heights > 15. So it's best not to make any assumptions here.

plebhash:
close #1460

builds on top of #1442

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/job_creator.rs
+++ b/protocols/v2/roles-logic-sv2/src/job_creator.rs
@@ -204,18 +204,18 @@ fn new_extended_job(
         .try_into()
         .map_err(|_| Error::TxVersionTooBig)?;
 
-    let bip34_bytes = get_bip_34_bytes(new_template, tx_version)?;
-    let script_prefix_len = bip34_bytes.len() + additional_coinbase_script_data.len();
+    let script_sig_prefix = new_template.coinbase_prefix.to_vec();
+    let script_sig_prefix_len = script_sig_prefix.len() + additional_coinbase_script_data.len();
 
     let coinbase = coinbase(
-        bip34_bytes,
+        script_sig_prefix,
         tx_version,
         new_template.coinbase_tx_locktime,
         new_template.coinbase_tx_input_sequence,
         coinbase_outputs,
         additional_coinbase_script_data,
         extranonce_len,
-    );
+    )?;
 
     let min_ntime = binary_sv2::Sv2Option::new(if new_template.future_template {
         None
@@ -230,8 +230,8 @@ fn new_extended_job(
         version: new_template.version,
         version_rolling_allowed,
         merkle_path: new_template.merkle_path.clone().into_static(),
-        coinbase_tx_prefix: coinbase_tx_prefix(&coinbase, script_prefix_len)?,
-        coinbase_tx_suffix: coinbase_tx_suffix(&coinbase, extranonce_len, script_prefix_len)?,
+        coinbase_tx_prefix: coinbase_tx_prefix(&coinbase, script_sig_prefix_len)?,
+        coinbase_tx_suffix: coinbase_tx_suffix(&coinbase, extranonce_len, script_sig_prefix_len)?,
     };
 
     debug!(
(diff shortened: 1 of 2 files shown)


protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -284,78 +284,80 @@ fn coinbase_tx_suffix(
...
     sequence: u32,
     coinbase_outputs: &[TxOut],
     additional_coinbase_script_data: Vec<u8>,
     extranonce_len: u8,
-) -> Transaction {
-    // If script_prefix_len is not 0 we are not in a test environment and the coinbase have the 0
-    // witness
-    let witness = match bip34_bytes.len() {
+) -> Result<Transaction, Error> {
+    // outside of test environments, we need to make sure the BIP34 data is sane
```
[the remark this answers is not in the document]
plebhash: this is to make sure TP sent the correct BIP34 bytes

otherwise, what's the purpose of the `Error::InvalidBip34Bytes` variant?

and most importantly: what happens if TP sends a `scriptSig` with bad BIP34 bytes?

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -203,18 +203,18 @@ fn new_extended_job(
         .try_into()
         .map_err(|_| Error::TxVersionTooBig)?;
 
-    let bip34_bytes = get_bip_34_bytes(new_template, tx_version)?;
-    let script_prefix_len = bip34_bytes.len() + additional_coinbase_script_data.len();
+    let script_sig_prefix = new_template.coinbase_prefix.to_vec();
+    let script_sig_prefix_len = script_sig_prefix.len() + additional_coinbase_script_data.len();
```
GitGab19, line 207: Isn't this part of the `script_sig` field? Why is it called "prefix"?

protocols/v2/roles-logic-sv2/src/job_creator.rs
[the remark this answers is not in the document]
Sjors: It's indirectly covered by the fact that the Bitcoin Core test suite creates blocks and then validates them. But not (yet) by the Template Provider codebase, maybe that can be added once it becomes an independent project.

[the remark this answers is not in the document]
GitGab19: This confirms that there's no need for the test here, since we can assume the template which comes out is always valid

protocols/v2/roles-logic-sv2/src/job_creator.rs
plebhash: `script_sig` contains a lot of things, namely:
- bip34
- [maybe something else in the future](https://github.com/stratum-mining/stratum/discussions/1256#discussioncomment-12069191)
- extranonce prefix
- extranonce

the goal of this PR is not to assume that `script_sig` can only hold bip34 bytes before the extranonce prefix

but it's also important to remember that extranonce prefix and extranonce are still part of the `script_sig`

therefore, the two items from the list above are being generically called `script_sig_prefix`

GitGab19: I see. My confusion was because of the naming, since `bip34` and `maybe something else in the future` are part of the `script_sig`.

But I see your point 👍

plebhash: image

jbesraa (commented):
LGTM
I think a test should be added to examine the behavior when there is additional data in the scriptsig prefix beside bip34 data

Issue #1462, opened 2025-02-08
need to deprecate `binary_codec_sv2` and restructure `binary_sv2` + `derive_codec_sv2`


Shourya742:
The current structure of `binary_sv2` does not follow best practices. Typically, a crate defining a `proc_macro` only includes an additional crate for the macro’s implementation. However, in `binary_sv2`, we have two separate crates—one for the `proc_macro` implementation and another for the primitives required for macro scaffolding. Due to the tightly coupled nature of the `proc_macro` implementation with the module structure, restructuring it requires rewriting some parts of the macro to align with the proposed design.  

Kudos to @plebhash for pointing this out and suggesting a better structure: https://github.com/stratum-mining/stratum/pull/1459#issuecomment-2644185529

plebhash:
one detail coming from #1459 that we should clean up here is some leftover lifetime parameters that were originally motivated by `serde` support:

reference: https://github.com/stratum-mining/stratum/pull/1459/commits/3f807180571f5858d57e846107a3b445c393d83b#r1947844273

plebhash:
this will introduce breaking changes to `binary_sv2` APIs

plebhash:
> this will introduce breaking changes to `binary_sv2` APIs

as a consequence, we will have to refactor `framing_sv2`, which could also be a good opportunity to improve APIs there

Levi0804:
> Reimplement or redesign parts of `proc_macro` in `derive_codec` to work around bugs like this: [#1468 (comment)](https://github.com/stratum-mining/stratum/pull/1468#discussion_r1949507770).

Hey, Would it be a good idea to start with introducing quote and syn in code? Also, regarding [#1468 (comment)](https://github.com/stratum-mining/stratum/pull/1468#discussion_r1949507770) I think we can return a more appropriate error message, something like 'the name of the field cannot be data.' or we could prefix data with an _ or rename it appropriately. What are your thoughts on this?

plebhash:
> > Reimplement or redesign parts of `proc_macro` in `derive_codec` to work around bugs like this: [#1468 (comment)](https://github.com/stratum-mining/stratum/pull/1468#discussion_r1949507770).
> 
[...]

PR #1459, opened 2025-02-06
remove binary sv2 serde

Shourya742:
part of: #1387

PR #1468, opened 2025-02-09
[WIP] refactor `framing_sv2` example

### LINKED ISSUE #1467: `framing_sv2` example is incomplete
while working on the presentation for `framing_sv2`, I noticed that the example is slightly incomplete.

more specifically, the last two assertions are kinda arbitrary and don't make much sense.

https://github.com/stratum-mining/stratum/blob/19f26e3519a53a4e31d26bc57488dda929aa8793/protocols/v2/framing-sv2/examples/sv2_frame.rs#L33-L58

ideally, we should assert that the deserialized message has the same contents from the original, not assert for some arbitrary frame size.

plebhash:
close #1467


protocols/v2/framing-sv2/examples/sv2_frame.rs
```diff
@@ -17,15 +17,16 @@
...
 use framing_sv2::framing::Sv2Frame;
+use std::convert::TryInto;
 
 // Example message type (e.g., SetupConnection)
 const MSG_TYPE: u8 = 1;
 // Example extension type (e.g., a standard Sv2 message)
 const EXT_TYPE: u16 = 0x0001;
 
-#[derive(Serialize, Debug)]
+#[derive(Serialize, Deserialize, Debug)]
```
plebhash, line 29: this line is causing the following error:

```
error[E0308]: mismatched types
  --> v2/framing-sv2/examples/sv2_frame.rs:29:21
   |
29 | #[derive(Serialize, Deserialize, Debug)]
   |                     ^^^^^^^^^^^
   |                     |
   |                     expected `&[u8]`, found `&Vec<FieldMarker>`
   |                     arguments to this method are incorrect
   |
   = note: expected reference `&[u8]`
              found reference `&Vec<FieldMarker>`
note: method defined here
  --> /Users/plebhash/develop/stratum/protocols/v2/binary-sv2/no-serde-sv2/codec/src/codec/mod.rs:85:8
   |
85 |     fn size_hint_(&self, data: &[u8], offset: usize) -> Result<usize, Error>;
   |        ^^^^^^^^^^
   = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
error: could not compile `framing_sv2` (example "sv2_frame") due to previous error
```

this error is the main reason why this PR is still draft

I wonder if this is related to #1459, and if it requires anything from #1462?

cc @Shourya742

plebhash: I rebased to bring in the changes from #1455 and the error did not go away.

plebhash: that did the trick!

but now I'm quite puzzled! is this somehow conflicting with something from `binary_sv2`?

jbesraa (approved):
Looks good. do we actually run this in the ci?

plebhash:
> Looks good. do we actually run this in the ci?

nice catch, we didn't but I just added a commit that runs it from `test.yml`

protocols/v2/framing-sv2/examples/sv2_frame.rs
```diff
@@ -17,33 +17,29 @@
...
+use std::convert::TryInto;
 
 // Example message type (e.g., SetupConnection)
 const MSG_TYPE: u8 = 1;
 // Example extension type (e.g., a standard Sv2 message)
 const EXT_TYPE: u16 = 0x0001;
 
-#[derive(Serialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct CustomMessage {
```
jbesraa, line 30: thinking out loud here, wouldn't it be more beneficial to actually have an SV2 message here? for the MSG_TYPE, SETUP_CONNECTION is used, maybe here as well?

GitGab19:
I think this PR is ready to go since it received 2 approvals.

Could you try to rebase it @plebhash ?

protocols/v2/framing-sv2/examples/sv2_frame.rs
plebhash: rationale here is to limit this example to the scope of `framing_sv2` while highlighting the assumption that the crate is agnostic to which messages actually exist in the Sv2 protocol

it's meant to showcase that as long as a message is defined this way, `framing_sv2` will be able to transmit it over the wire (without any kind of assertion of whether this message is part of the spec or not)

plebhash:
> I think this PR is ready to go since it received 2 approvals.
> 
> Could you try to rebase it @plebhash ?

pav bot called to action (which btw reminds me of #1488)

PR #1482, opened 2025-02-13
Remove reject auth test from MG and its footprint

### LINKED ISSUE #1481: Remove `pool-sri-reject-auth` MG test
.. no auth functionality is in place and this is only testing the feature flag(!)

jbesraa:
close #1481

.. Auth rejection is not implemented, this is just testing the feature flag.


plebhash:
> Damn... What's the purpose of this test then? Is it just to check whether `OpenMiningChannel.Error` is received? That seems pretty vague.

it was introduced via https://github.com/stratum-mining/stratum/commit/84a3ddb34eba41adec44ac234a8b38b3aa6e90df

perhaps a deep inspection of this commit could give us some insights?

GitGab19:
> > Damn... What's the purpose of this test then? Is it just to check whether `OpenMiningChannel.Error` is received? That seems pretty vague.
> 
> it was introduced via [84a3ddb](https://github.com/stratum-mining/stratum/commit/84a3ddb34eba41adec44ac234a8b38b3aa6e90df)
> 
> perhaps a deep inspection of this commit could give us some insights?

Probably this was a test related to some initial work related to authentication mechanisms which were never really implemented..
I think we should add a proper test in the future, as soon as we will have some kind of auth system in place on the Pool.

ACK to remove it for now.

PR #1486, opened 2025-02-17
JDC Fallback test

### LINKED ISSUE #1207: CI migration: `interop-jdc-change-upstream` MG to Integration Test

jbesraa:
blocked by #1485 
superseded #1343 

Closes #1207


roles/tests-integration/tests/jd_integration.rs
```diff
@@ -33,11 +38,46 @@ async fn jds_should_not_panic_if_jdc_shutsdown() {
...
-    let (_jdc, _jdc_addr) = start_jdc(pool_addr, tp_addr, sniffer_addr).await;
+    let (_jdc, _jdc_addr) = start_jdc(vec![pool_addr], tp_addr, sniffer_addr).await;
     assert_common_message!(sniffer.next_message_from_downstream(), SetupConnection);
 }
+
+// Start JDClient with two pools in the pool list config. The test forces the first pool to send a
+// share submission rejection by altering a message(MINING_SET_NEW_PREV_HASH) sent by the pool. The
+// test verifies that the second pool will be used as a fallback.
+#[tokio::test]
+async fn test_jdc_pool_fallback_after_submit_rejection() {
```
jbesraa, line 54: For me sometimes this run forever but mostly passes. It might be a bug in how JDC is handling the fallback

jbesraa:
Super nice to see CI passing here!

jbesraa:
This is ready for review

GitGab19:
Have we officially decided to add a file for every test?

I see that the tracker issue [here](https://github.com/stratum-mining/stratum/issues/1523) is still opened without a final conclusion.

Maybe it's worth discussing this during the next PR Review Club call.

jbesraa:
> Have we officially decided to add a file for every test?
> 
> I see that the tracker issue [here](https://github.com/stratum-mining/stratum/issues/1523) is still opened without a final conclusion.
> 
> Maybe it's worth discussing this during the next PR Review Club call.

Yea in one of the last calls we agreed that at least tests with a miner should be in a separate file.

roles/tests-integration/tests/tproxy_jd_miner.rs
plebhash, line 1: why is this file being added on this commit?

I understand the idea of separating tests with CPU miners, but:
- this commit is titled "Allow `start_jdc` to accept multiple pool addresses"
- we create this file but the test remains on `translator_integration.rs`?

roles/tests-integration/tests/jdc_fallback.rs
```diff
@@ -0,0 +1,47 @@
...
+use roles_logic_sv2::{
+    mining_sv2::SubmitSharesError,
+    parsers::{AnyMessage, CommonMessages, Mining},
+};
+use sniffer::{MessageDirection, ReplaceMessage};
+use std::convert::TryInto;
+
+// Start JDC with two pools in the pool list config. The test forces the first pool to send a share
+// submission rejection by altering a message(MINING_SET_NEW_PREV_HASH) sent by the pool. The test
+// verifies that the second pool will be used as a fallback.
```
plebhash, line 12: This description seems wrong?

roles/tests-integration/lib/mod.rs
```diff
@@ -137,12 +136,17 @@ pub async fn start_jdc(
...
-        jds_address.to_string(),
-        pool_signature,
-    )];
+    let upstreams = pool_addresses
+        .iter()
+        .map(|addr| {
+            Upstream::new(
+                authority_pubkey,
+                addr.to_string(),
+                jds_address.to_string(),
```
plebhash, line 145: each `Upstream` object has 4 elements, where one is `pool_address` and the other is `jds_address`

in theory, when JDC does the fallback, it should re-connect to a different Pool+JDS pair, not just a new Pool

while the original MG test didn't assert for JDC re-connecting to a fallback JDS, I wonder if we shouldn't do that on this Integration Test?

cc @GitGab19

roles/test-utils/mining-device-sv1/src/client.rs
```diff
@@ -244,7 +244,8 @@ impl Client {
             if let Ok(incoming) = recv_incoming.clone().recv().await {
                 Self::parse_message(client.clone(), Ok(incoming)).await;
             } else {
-                warn!("Error reading from socket via `recv_incoming` channel")
+                warn!("Error reading from socket via `recv_incoming` channel");
+                break;
```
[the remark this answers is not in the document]
jbesraa: yup

roles/tests-integration/tests/tproxy_jd_miner.rs
jbesraa: oops that was not intended

roles/tests-integration/lib/mod.rs
jbesraa: that's a valid point, added an assertion for JDS as well.

roles/tests-integration/tests/jdc_fallback.rs
```diff
@@ -0,0 +1,67 @@
...
+        &[
+            (sniffer_addr_1, sniffer_addr_3),
+            (sniffer_addr_2, sniffer_addr_4),
+        ],
+        tp_addr,
+    )
+    .await;
+    // Assert that JDC has connected to the first (Pool,JDS) pair
+    assert_common_message!(&sniffer_1.next_message_from_downstream(), SetupConnection);
+    assert_common_message!(&sniffer_3.next_message_from_downstream(), SetupConnection);
```
plebhash, line 57: I've seen some other tests failing on CI due to `assert_*_message`, and I think we are risking that the same could happen here in an undeterministic way... I think we discussed this on a call recently

there's a small chance of a race-condition here in case we make this assertion before the message actually arrives to sniffer

overall I think we should be very cautious when using these macros, and only use them when we know for sure that the expected message has already arrived

when in doubt, I think it's usually safer to use `wait_for_message_type`

roles/tests-integration/tests/jdc_fallback.rs
```diff
@@ -0,0 +1,67 @@
...
+    let (_translator, sv2_translator_addr) = start_sv2_translator(jdc_addr).await;
+    let _ = start_mining_device_sv1(sv2_translator_addr, true, None).await;
+    // Assert that JDC switched to the second (Pool,JDS) pair
```
plebhash, line 60: maybe we should also assert/`wait_for_message_type` on the `SubmitShareError`?

roles/tests-integration/tests/jdc_fallback.rs
jbesraa: I think for setup_connection this is less risky

roles/tests-integration/tests/jdc_fallback.rs
jbesraa: This would be asserting the sniffer behavior which we are doing in other sniffer tests

roles/tests-integration/tests/jdc_fallback.rs
plebhash: https://github.com/stratum-mining/stratum/actions/runs/13762547660/job/38481652339?pr=1483#step:7:448

```
thread 'jds_should_not_panic_if_jdc_shutsdown' panicked at tests-integration/tests/jd_integration.rs:29:5:
No message received
```

this one failed on [precisely on `SetupConnection`](https://github.com/jbesraa/stratum/blob/11c5c6f6b58d8b1f95fc1716587c9a5c84885cac/roles/tests-integration/tests/jd_integration.rs#L29)

plebhash (approved):
ACK but looks like CI is broken

PR #1491, opened 2025-02-18
unify `coinbase_tx_prefix` and `coinbase_tx_suffix` serialization

### LINKED ISSUE #1490: need to unify `coinbase_tx_prefix` and `coinbase_tx_suffix` serialization strategies on `roles_logic_sv2` (BIP141)
`coinbase_tx_prefix` and `coinbase_tx_suffix` were originally implemented in `roles_logic_sv2::job_creator` in a way that they included BIP141 (segwit marker + flag) fields during serialization

later on it was noted that these fields must be removed during calculation of `merkle_root`, and then #488 introduced a new `struct StrippedCoinbase` to address this

the problem is that `roles_logic_sv2` ended up with fragmented patterns, with a redundant pair of functions responsible for serializing `coinbase_tx_prefix` and `coinbase_tx_suffix`

more context here https://github.com/stratum-mining/stratum/discussions/1256#discussioncomment-11297214

plebhash:
close #1490 

to be merged after #1442 and #1461

---

we introduce a new `struct Coinbase` into `roles_logic_sv2::utils` that replaces the following from `roles_logic_sv2::job_creator`:
- `struct StrippedCoinbase`
- `fn coinbase -> Result<Transaction, Error>`

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/job_creator.rs
+++ b/protocols/v2/roles-logic-sv2/src/job_creator.rs
@@ -9,12 +9,7 @@ use crate::{
 use mining_sv2::NewExtendedMiningJob;
 use nohash_hasher::BuildNoHashHasher;
 use std::{collections::HashMap, convert::TryInto};
-use stratum_common::bitcoin::{
-    blockdata::transaction::{Transaction, TxOut},
-    consensus,
-    consensus::Decodable,
-    Amount,
-};
+use stratum_common::bitcoin::{blockdata::transaction::TxOut, consensus::Decodable, Amount};
 use template_distribution_sv2::{NewTemplate, SetNewPrevHash};
 use tracing::debug;
 
@@ -235,45 +230,6 @@ fn new_extended_job(
     Ok(new_extended_mining_job)
 }
 
-/// Helper type to strip a segwit data from the coinbase_tx_prefix and coinbase_tx_suffix
-/// to ensure miners are hashing with the correct coinbase
-pub fn extended_job_to_non_segwit(
-    job: NewExtendedMiningJob<'static>,
-    full_extranonce_len: usize,
-) -> Result<NewExtendedMiningJob<'static>, Error> {
-    let mut encoded = job.coinbase_tx_prefix.to_vec();
-    // just add empty extranonce space so it can be deserialized. The real extranonce
-    // should be inserted based on the miner's shares
-    let extranonce = vec![0_u8; full_extranonce_len];
-    encoded.extend_from_slice(&extranonce[..]);
-    encoded.extend_from_slice(job.coinbase_tx_suffix.inner_as_ref());
-    let tx: Transaction = consensus::deserialize(&encoded).map_err(|_| Error::InvalidCoinbase)?;
-
-    let coinbase = Coinbase {
-        tx,
-        script_sig_prefix_len: job.coinbase_tx_prefix.len()
-            - (
-                4 + // version
-                1 + // input count
-                32 + // outpoint
-                4 + // input index
-                1
-                // script length byte
-            ),
-    };
-
-    Ok(NewExtendedMiningJob {
-        channel_id: job.channel_id,
-        job_id: job.job_id,
-        min_ntime: job.min_ntime,
-        version: job.version,
-        version_rolling_allowed: job.version_rolling_allowed,
-        merkle_path: job.merkle_path,
-        coinbase_tx_prefix: coinbase.coinbase_tx_prefix()?,
-        coinbase_tx_suffix: coinbase.coinbase_tx_suffix()?,
-    })
-}
-
 // Test
 #[cfg(test)]
 
@@ -284,6 +240,7 @@ pub mod tests {
     use binary_sv2::u256_from_int;
     use quickcheck::{Arbitrary, Gen};
     use std::{cmp, vec};
+    use stratum_common::bitcoin::{blockdata::transaction::Transaction, consensus};
 
     #[cfg(feature = "prop_test")]
     use std::borrow::BorrowMut;
--- a/protocols/v2/roles-logic-sv2/src/job_creator.rs
+++ b/protocols/v2/roles-logic-sv2/src/job_creator.rs
@@ -2,23 +2,18 @@
 //!
 //! This module provides logic to create extended mining jobs given a template from
 //! a template provider as well as logic to clean up old templates when new blocks are mined.
-use crate::{errors, utils::Id, Error};
-use binary_sv2::B064K;
+use crate::{
+    utils::{Coinbase, Id},
+    Error,
+};
 use mining_sv2::NewExtendedMiningJob;
 use nohash_hasher::BuildNoHashHasher;
 use std::{collections::HashMap, convert::TryInto};
-use stratum_common::{
-    bitcoin,
-    bitcoin::{
-        absolute::LockTime,
-        blockdata::{
-            transaction::{OutPoint, Transaction, TxIn, TxOut, Version},
-            witness::Witness,
-        },
-        consensus,
-        consensus::Decodable,
-        Amount,
-    },
+use stratum_common::bitcoin::{
+    blockdata::transaction::{Transaction, TxOut},
+    consensus,
+    consensus::Decodable,
+    Amount,
 };
 use template_distribution_sv2::{NewTemplate, SetNewPrevHash};
 use tracing::debug;
@@ -205,17 +200,16 @@ fn new_extended_job(
         .map_err(|_| Error::TxVersionTooBig)?;
 
     let script_sig_prefix = new_template.coinbase_prefix.to_vec();
-    let script_sig_prefix_len = script_sig_prefix.len() + additional_coinbase_script_data.len();
 
-    let coinbase = coinbase(
+    let coinbase = Coinbase::new(
         script_sig_prefix,
         tx_version,
         new_template.coinbase_tx_locktime,
         new_template.coinbase_tx_input_sequence,
-        coinbase_outputs,
+        coinbase_outputs.to_vec(),
         additional_coinbase_script_data,
         extranonce_len,
-    )?;
+    );
 
     let min_ntime = binary_sv2::Sv2Option::new(if new_template.future_template {
         None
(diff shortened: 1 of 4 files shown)


GitGab19:
Is this ready for review @plebhash ?

plebhash:
> Is this ready for review @plebhash ?

yes

jbesraa:
@plebhash Did you consider using https://rust-unofficial.github.io/patterns/patterns/creational/builder.html for the new `Coinbase` struct? IMO this is a good time to adapt to this pattern.

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -155,10 +158,126 @@ impl<T> Mutex<T> {
...
+    // - input count
+    // - input outPoint
+    // - input index (vout)
+    // - scriptSig length
+    // - scriptSig prefix
+    //
+    // we skip the bip141 marker and flag (which usually come after tx version)
+    // because those are only used for computing the `wtxid`
+    // while the legacy `txid` is what is actually used for computing the merkle root
+    pub fn coinbase_tx_prefix(self) -> Result<B064K<'static>, Error> {
```
jbesraa, line 231:
```suggestion
    pub fn coinbase_tx_prefix(&self) -> Result<B064K<'static>, Error> {
```

Ownership is not needed here. This should allow you to remove all the `.clone()` calls for `Coinbase`.

plebhash:
> @plebhash Did you consider using https://rust-unofficial.github.io/patterns/patterns/creational/builder.html for the new `Coinbase` struct? IMO this is a good time to adapt to this pattern.

I didn't consider this while working on the PR.

After your suggestion, I made some quick sketching but TBH I fail to see the advantages over the current approach.

The commit is on a separate branch https://github.com/plebhash/stratum/commit/e6278fba5ba63ce88efdf7e91d8de7cdcc228a81

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -271,6 +271,67 @@ impl Coinbase {
...
         coinbase_tx_suffix.try_into().map_err(Error::BinarySv2Error)
     }
+
+    /// Reconstructs the coinbase transaction with BIP141 marker and flag bytes.
+    /// This is needed when constructing a block with SegWit transactions.
+    ///
+    /// The function takes the coinbase_tx_prefix, extranonce, and coinbase_tx_suffix
+    /// and constructs a new coinbase transaction with the BIP141 marker and flag bytes
+    /// inserted after the version bytes.
+    pub fn reconstruct_with_bip141(
```
jbesraa, line 281: I think `new_with_bip141` or `new_with_segwit` is more readable

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -271,6 +271,67 @@ impl Coinbase {
         coinbase_tx_suffix.extend_from_slice(&lock_time_u32.to_le_bytes());
         coinbase_tx_suffix.try_into().map_err(Error::BinarySv2Error)
     }
+
+    /// Reconstructs the coinbase transaction with BIP141 marker and flag bytes.
+    /// This is needed when constructing a block with SegWit transactions.
```
jbesraa, line 276: Is this needed because a block contains segwit transactions or because the coinbase is segwit, i.e., even if the block does have segwit txs but coinbase is not, this should not be used right?

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -1331,4 +1408,59 @@ mod tests {
         // m.super_safe_lock(|i| *i = (*i).checked_add(1).unwrap()); // will not compile
         m.super_safe_lock(|i| *i = (*i).checked_add(1).unwrap_or_default()); // compiles
     }
+
+    #[test]
+    fn test_reconstruct_with_bip141() {
```
jbesraa, line 1413: It feels like the new code is testing itself here because you are comparing if the results from `Coinbase::new` and `Coinbase::reconstruct_from..` are similar

protocols/v2/roles-logic-sv2/src/job_creator.rs
```diff
@@ -683,19 +463,32 @@ pub mod tests {
...
-        let stripped = StrippedCoinbaseTx::from_coinbase(coinbase.clone(), 32).unwrap();
-        let prefix = stripped.into_coinbase_tx_prefix().unwrap().to_vec();
-        let suffix = stripped.into_coinbase_tx_suffix().unwrap().to_vec();
+        let tx: Transaction = consensus::deserialize(encoded).unwrap();
+
+        // Calculate script_sig_prefix_len by examining the actual script_sig length
+        // and subtracting the extranonce length (32 in this case)
+        let extranonce_len = 32;
+        let script_sig_len = tx.input[0].script_sig.len();
+        let script_sig_prefix_len = script_sig_len - extranonce_len;
```
jbesraa, line 472: would it make sense to save the right side (`extranonce_len` here) instead of `script_sig_prefix_len`? it feels like we are doing this calculation but are not using it until we call other function. maybe the responsibility of doing this should be on the other function.

jbesraa (commented):
TBH I find the `Coinbase` API a bit confusing to me. There are instances where it is initialized with `Coinbase { a, b }`, others with `Coinbase::new` and also `::reconstruct`.

I would prefer to have a `new` with docs above it indicating is it using segwit or not, and other one ::new_with/out_segwit. It also seems that if it is not segwit the user is expected to build it by themselves which I don't think is optimal.

protocols/v2/roles-logic-sv2/src/utils.rs
plebhash: I expected this new function would spark some important discussions that even touch some spec-related topics.

---

I added this new function yesterday because I did some e2e testing mining on our custom signet and blocks were being rejected with:
```
2025-03-05T01:56:02Z [error] AcceptBlock: bad-witness-nonce-size, CheckWitnessMalleation : invalid witness reserved value size
2025-03-05T01:56:02Z [error] ProcessNewBlock: AcceptBlock FAILED (bad-witness-nonce-size, CheckWitnessMalleation : invalid witness reser
ved value size)
```

These logs didn't happen on the node that received the `SubmitSolution`, but rather on a peer.

(the fact that we didn't catch this on our CI is a bit concerning, but it's probably a separate discussion)

After some research I found this blogpost which described a similar scenario that happened in production last year: https://b10c.me/observations/10-viabtc-blocks-without-witness-data/

---

Coming back to some basic consensus rules:

In order for a block header to be valid, it's `merkle_root` field must **always** be calculated based on the Coinbase's **`txid`**, not **`wtxid`**.

That does not necessarily mean the Coinbase must be stripped of BIP141 data.

Miners could still theoretically produce consensus-valid Coinbases without BIP141 data. This only means that we need to strip BIP141 data from it in order to calculate it's `txid`. 

However, miners are encouraged to always produce Coinbases **with** BIP141, as it would be economically irrational to do otherwise.

On SRI we assume miners will always produce Coinbases with BIP141.

The discussion here is centered around how the Coinbase is serialized for extended jobs to be sent downstream (since mining devices only care about finding a header with a valid `merkle_root`, thus only care about `txid`).

---

Coming back to some historical context for this PR:

`coinbase_tx_prefix` and `coinbase_tx_suffix` were originally implemented in `roles_logic_sv2::job_creator` in a way that they included BIP141 (segwit marker + flag + witness) fields during serialization

however, Sv1 firmware expects `coinbase_tx_prefix` and `coinbase_tx_suffix` to **not** include it (as Sv1 was conceived pre-segwit), because hashing `coinbase_tx_prefix` + `extranonce` + `coinbase_tx_suffix` leads to `wtxid`

so later on it was noted that this was breaking Sv1 compatibility, and then #488 introduced a new `struct StrippedCoinbase` to address this

PR #488 aimed to fix this issue with a very narrow scope, where BIP141 fields were stripped only during translation to Sv1

the Sv2 spec doesn't really say anything about this, and AFAIU it remains an open question, as I noticed that this comment from @Fi3 https://github.com/stratum-mining/stratum/pull/488#discussion_r1133142005 was never properly addressed on PR #488

---

So now we come to a discussion about spec.

The open question is: what assumptions does Sv2 make for `coinbase_tx_prefix` + `coinbase_tx_suffix`?

or more specifically: on `NewExtendedMiningJob` and `DeclareMiningJob`, do we assume they are stripped from BIP141 data?

1. if yes: then every Sv2-compliant software must always re-insert the BIP141 data while reconstructing a block for propagation (which is the direction this PR is going)
2. if no: then functionality for stripping BIP141 data must be restricted to tProxy (which is what the code does before this PR, but with very confusing APIs)

I started this PR going in direction 1, but now I'm having second thoughts.

jbesraa: Maybe as a start lets add the validation functionality from bitcoin core regarding `blocks_without_witness_data` to SRI and then rethink this

plebhash: > Maybe as a start lets add the validation functionality from bitcoin core regarding blocks_without_witness_data to SRI and then rethink this

with the current implementation this error will never happen because SRI always creates a coinbase with full BIP141 data

this error was introduced in my PR while re-constructing the coinbase from serialized bytes, where BIP141 data were stripped

I fixed it with the commit that adds `reconstruct_with_bip141` but that kickstarted the discussion about whether this is a future-proof approach

plebhash: > But it also mean that the TP must strip witness bytes from the prefix and re-add them plus and the reserved value when we submit solution.

I don't think this would imply any changes on TP. The `NewTemplate` message doesn't contain any information about BIP141.

If you check the [`coinbase`](https://github.com/stratum-mining/stratum/blob/da794213ee0d0d17d54aa8d1b1c8c3c0b0e7a91e/protocols/v2/roles-logic-sv2/src/job_creator.rs#L295) function you will see that SRI builds the coinbase from the separate fields that TP sent via `NewTemplate`. This function also uses APIs from `rust-bitcoin` to build a Coinbase transaction with BIP141 support.

In other words, SRI is currently making an opinionated assumption that all Sv2 Coinbases will always contain BIP141 support, and simply removing BIP141 data during Sv1 translation.

And after reflection and discussions, I think this is the safer approach, as always stripping BIP141 data from the coinbase could have unintended consequences in the future.

Issue #1272, opened 2024-12-06
Consider combining some of the github action files


jbesraa:
Currently we add each new action to a new file in [workflows](https://github.com/stratum-mining/stratum/tree/main/.github/workflows). I think we should consider combining some of those flows into a single file.

For example: [test.yaml](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/test.yaml), [rust-msrv.yaml](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/rust-msrv.yaml), [semver-check.yaml](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/semver-check.yaml), [lockfiles.yaml](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/lockfiles.yaml), [fmt.yaml](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/fmt.yaml), [clippy-lint.yaml](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/clippy-lint.yaml) can probably be combined into a single file

jbesraa:
Consider running each job with twice once with rust 1.75(our MSRV) and once with rust stable, beside the MSRV check job which should run just with 1.75

jbesraa:
We did some work on the clippy side, still need to do things on other actions but our CI have already been through a lot of changes in the last couple of milestones. Lets take it a step by step. Closing this for now and will revisit if we think it is needed.

PR #1492, opened 2025-02-18
Use stable rust in `clippy` check and move to a shell script

jbesraa:
Part of #1272 

The most notable fix/change in this pr is adding this https://github.com/stratum-mining/stratum/pull/1492/files#diff-40acd8e96b9e30ae1cfa538fb7ae0cdf80358930dd6f4167375f85980ee019e4 because `tracing` feature is used withing the code but did not exist in Cargo.toml

Probably eventually whats called now `Clippy-lint.yml` would become `rust-checks` and we will have in it clippy/fmt/test/.. all cargo/rust stuff that we want in the same place

### DIFF
--- a/.github/workflows/clippy-lint.yaml
+++ b/.github/workflows/clippy-lint.yaml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.75.0
+          toolchain: stable
           override: true
           components: clippy
       - name: Run Clippy on different workspaces and crates
--- a/.github/workflows/clippy-lint.yaml
+++ b/.github/workflows/clippy-lint.yaml
@@ -28,9 +28,4 @@ jobs:
           override: true
           components: clippy
       - name: Run Clippy on different workspaces and crates
-        run: |
-          cargo clippy --manifest-path=benches/Cargo.toml -- -D warnings -A dead-code
-          cargo clippy --manifest-path=common/Cargo.toml -- -D warnings -A dead-code
-          cargo clippy --manifest-path=protocols/Cargo.toml -- -D warnings -A dead-code
-          cargo clippy --manifest-path=roles/Cargo.toml -- -D warnings -A dead-code
-          cargo clippy --manifest-path=utils/Cargo.toml -- -D warnings -A dead-code
+        run: ./scripts/rust/clippy.sh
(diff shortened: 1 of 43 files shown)


plebhash:
@jbesraa https://github.com/stratum-mining/stratum/issues/1272#issuecomment-2557083525 here you said:

> Consider running each job with twice once with rust 1.75(our MSRV) and once with rust stable

but here we're just abandoning MSRV for stable? what's the rationale?

jbesraa:
> @jbesraa [#1272 (comment)](https://github.com/stratum-mining/stratum/issues/1272#issuecomment-2557083525) here you said:
> 
> > Consider running each job with twice once with rust 1.75(our MSRV) and once with rust stable
> 
> but here we're just abandoning MSRV for stable? what's the rationale?

Added msrv back, so now it is stable + msrv. Got confused a bit. 

I think we need to change the github action settings in the repo because the name of the check stopped being only "clippy-check" but "Clippy Lint / clippy-check ($version)"

plebhash:
can we fix the commit history here?

a few issues are making the review a bit confusing:
- the top commit message is contradicting the latest changes in the PR strategy
- the top commit is titled "Use stable and MSRV versions in clippy CI check", but the other commit is also making changes under the same scope

jbesraa:
> can we fix the commit history here?
> 
> a few issues are making the review a bit confusing:
> 
>     * the top commit message is contradicting the latest changes in the PR strategy
> 
>     * the top commit is titled "Use stable and MSRV versions in clippy CI check", but the other commit is also making changes under the same scope

My bad, updated commits.  
Note that I left only stable version and the rationale is explained in https://github.com/stratum-mining/stratum/pull/1492/commits/851f1691438b5eaf9d9aa9a44ff81a2a5c172e76. I also decided to leave both macos and ubuntu in and not remove macos, mainly due to https://github.com/stratum-mining/stratum/pull/1508/files#diff-adaf0b6cc7ee6f3aca68632c496a4bd7cead4d5ae214049da0881486a50050f5R50 fix that was discovered on macos initially.

jbesraa:
> Finally, the Clippy fixes are in! I've been encountering these for a while. Should we also consider upgrading the MSRV?

sorry, not sure what you mean by "upgrading msrv"?

jbesraa:
This is ready for review

plebhash (approved):

plebhash:
fmt CI is broken

Issue #1503, opened 2025-02-25
CI migration: jds-do-not-stackoverflow-when-no-token MG to Integration Test


Shourya742:
migrate jds-do-not-stackoverflow-when-no-token MG to Integration Test

plebhash:
blocked by #1506, as we need JDS to not respond to `AllocateMiningJobToken`

alternatively, this could also be achieved by mocking a JDS

plebhash:
> > blocked by [#1506](https://github.com/stratum-mining/stratum/issues/1506), as we need JDS to not respond to `AllocateMiningJobToken`
> > alternatively, this could also be achieved by mocking a JDS
> 
> Blocking a message essentially means disconnecting in a way. What I had in mind was whether we could configure the sniffer to disconnect from the upstream connection when needed. For example, after receiving a certain message, the sniffer could automatically disconnect from either the upstream or downstream connection based on a specified value.

as discussed on the dev call today, we'll proceed by implementing this functionality into `Sniffer` by expanding the `InterceptMessage` functionality, but without really dropping the TCP connection

PR #1505, opened 2025-02-25
add-jds-do-not-stackoverflow-when-no-token-test

Shourya742:
closes: #1503


roles/tests-integration/tests/jd_integration.rs
```diff
@@ -26,3 +28,36 @@ async fn jds_should_not_panic_if_jdc_shutsdown() {
...
     assert_common_message!(sniffer.next_message_from_downstream(), SetupConnection);
 }
+
+/// This test ensures that `jd-client` does not panic even if `jd-server` leaves the connection open
+/// after receiving the request for token.
+///
+/// The test verifies whether `jdc` has crashed by attempting to bind to the `jdc` port after 10
+/// seconds of no response from `jd-server`.
+#[tokio::test]
+async fn jds_do_not_stackoverflow_when_no_token() {
```
plebhash, line 38: the original MG had a bit of a nonsensical name that implies that the stack overflow happens on JDS, while it actually happened on JDC

we should avoid propagating this bad name here

plebhash: strictly speaking, it should be `jdc-does-not-...`

roles/tests-integration/tests/jd_integration.rs
```diff
@@ -26,3 +28,43 @@ async fn jds_should_not_panic_if_jdc_shutsdown() {
...
+    start_tracing();
+    let (tp, tp_addr) = start_template_provider(None);
+    let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
+    let (_jds, jds_addr) = start_jds(tp.rpc_info()).await;
+    let block_from_message =
+        sniffer::BlockFromMessage::new(sniffer::MessageDirection::ToDownstream, 81);
```
plebhash, line 44: we shouldn't use literals for message types, because that forces the user to go check the spec to see which message type this is

always use constants

jbesraa:
Do we still need this PR and #1510 after yesterday's discussion?

plebhash:
this doesn't compile

also review points remain unaddressed

plebhash:
> #1543

hmm but then that makes me wonder if we didn't overengineer of #1510

why do we block the entire stream of messages when `InterceptAction::IgnoreFromMessage` is triggered?

why not `InterceptAction::IgnoreMessage` and simply prevent one single message from being relayed forward from `Sniffer`?

plebhash:
> > > #1543
> > 
> > 
> > hmm but then that makes me wonder if we didn't overengineer #1510
> > why do we block the entire stream of messages when `InterceptAction::IgnoreFromMessage` is triggered?
> > why not `InterceptAction::IgnoreMessage` and simply prevent one single message from being relayed forward from `Sniffer`?
> 
> Aren't declare mining jobs async?

I don't understand the question. Can you elaborate?

plebhash:
I tested with the changes from #1556 and I think this is now unblocked

roles/tests-integration/tests/jd_integration.rs
```diff
@@ -55,3 +61,47 @@ async fn jdc_tp_success_setup() {
...
+        )
+        .await;
+
+    jds_jdc_sniffer
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN,
+        )
+        .await;
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
plebhash, line 105: overall I think it makes sense to wait a bit, leaving enough time for the JDC to potentially-but-hopefully-not-panic 

but 10s is way too long for waiting here

I tried removing this line but then the test hangs and a `jd_integration` process is left behind, even if I cancel the test with `ctrl+c`

reducing this sleep to 1s makes the behavior undeterministic: sometimes the test returns and sometimes it hangs

sleeping for 3s makes it return every time (or at least on the 5x I tested this)

so I think we should reduce this to 3s, and add a comment to explain why we're sleeping here

(there's a comment on the top of the test so don't forget to update that too)

plebhash (approved):

PR #1510, opened 2025-02-26
sniffer block from message

### LINKED ISSUE #1506: ITF `Sniffer` needs functionality to block messages
as a pre-requisite for https://github.com/stratum-mining/stratum/issues/1503

we need `Sniffer` to be able to block messages, essentially stopping them from being forwarded

this functionality should be disabled by default, and only be enabled after the user triggers it

Shourya742:
This PR adds Block from message Action to sniffer. Closes #1506

### DIFF
--- a/roles/tests-integration/lib/sniffer.rs
+++ b/roles/tests-integration/lib/sniffer.rs
@@ -48,7 +48,7 @@ enum SnifferError {
 /// queues via [`Sniffer::next_message_from_downstream`] and
 /// [`Sniffer::next_message_from_upstream`], respectively.
 ///
-/// In order to replace the messages sent between the roles, a set of [`InterceptMessage`] can be
+/// In order to replace or ignore the messages sent between the roles, [`InterceptAction`] can be
 /// used in [`Sniffer::new`].
 #[derive(Debug, Clone)]
 pub struct Sniffer {
@@ -58,34 +58,34 @@ pub struct Sniffer {
     messages_from_downstream: MessagesAggregator,
     messages_from_upstream: MessagesAggregator,
     check_on_drop: bool,
-    action: Option<Action>,
+    action: Option<InterceptAction>,
 }
 
 /// Represents an action that [`Sniffer`] can take on intercepted messages.
 #[derive(Debug, Clone)]
-pub enum Action {
+pub enum InterceptAction {
     /// Blocks the message stream after encountering a specific message.
     IgnoreFromMessage(IgnoreFromMessage),
     /// Intercepts and modifies a message before forwarding it.
-    InterceptMessage(Box<InterceptMessage>),
+    ReplaceMessage(Box<ReplaceMessage>),
 }
 
-impl Action {
-    /// Returns the action if it is `IgnoreFromMessage` or `InterceptMessage`  
+impl InterceptAction {
+    /// Returns the action if it is `IgnoreFromMessage` or `ReplaceMessage`  
     /// with the specified message type.
     pub fn find_matching_action(
-        action: &Option<Action>,
+        action: &Option<InterceptAction>,
         msg_type: MsgType,
         direction: MessageDirection,
     ) -> Option<&Self> {
         action.as_ref().and_then(|action| match action {
-            Action::IgnoreFromMessage(bm)
+            InterceptAction::IgnoreFromMessage(bm)
                 if bm.direction == direction && bm.expected_message_type == msg_type =>
             {
                 Some(action)
             }
 
-            Action::InterceptMessage(im)
+            InterceptAction::ReplaceMessage(im)
                 if im.direction == direction && im.expected_message_type == msg_type =>
             {
                 Some(action)
@@ -115,22 +115,22 @@ impl IgnoreFromMessage {
     }
 }
 
-impl From<IgnoreFromMessage> for Action {
+impl From<IgnoreFromMessage> for InterceptAction {
     fn from(value: IgnoreFromMessage) -> Self {
-        Action::IgnoreFromMessage(value)
+        InterceptAction::IgnoreFromMessage(value)
     }
 }
 
 /// Allows [`Sniffer`] to replace some intercepted message before forwarding it.
 #[derive(Debug, Clone)]
-pub struct InterceptMessage {
+pub struct ReplaceMessage {
     direction: MessageDirection,
     expected_message_type: MsgType,
     replacement_message: AnyMessage<'static>,
 }
 
-impl InterceptMessage {
-    /// Constructor of `InterceptMessage`
+impl ReplaceMessage {
+    /// Constructor of `ReplaceMessage`
     /// - `direction`: direction of message to be intercepted and replaced
     /// - `expected_message_type`: type of message to be intercepted and replaced
     /// - `replacement_message`: message to replace the intercepted one
@@ -148,9 +148,9 @@ impl InterceptMessage {
     }
 }
 
-impl From<InterceptMessage> for Action {
-    fn from(value: InterceptMessage) -> Self {
-        Action::InterceptMessage(Box::new(value))
+impl From<ReplaceMessage> for InterceptAction {
+    fn from(value: ReplaceMessage) -> Self {
+        InterceptAction::ReplaceMessage(Box::new(value))
     }
 }
 
@@ -168,7 +168,7 @@ impl Sniffer {
         listening_address: SocketAddr,
         upstream_address: SocketAddr,
         check_on_drop: bool,
-        action: Option<Action>,
+        action: Option<InterceptAction>,
     ) -> Self {
         Self {
             identifier,
@@ -280,7 +280,7 @@ impl Sniffer {
         recv: Receiver<MessageFrame>,
         send: Sender<MessageFrame>,
         downstream_messages: MessagesAggregator,
-        action: Option<Action>,
+        action: Option<InterceptAction>,
     ) -> Result<(), SnifferError> {
         // Blocking flag used in the IgnoreFromMessage action to stop processing messages after the
         // desired interception.
@@ -290,15 +290,18 @@ impl Sniffer {
                 continue;
             }
             let (msg_type, msg) = Self::message_from_frame(&mut frame);
-            let action =
-                Action::find_matching_action(&action, msg_type, MessageDirection::ToUpstream);
+            let action = InterceptAction::find_matching_action(
+                &action,
+                msg_type,
+                MessageDirection::ToUpstream,
+            );
             if let Some(ref action) = action {
                 match action {
-                    Action::IgnoreFromMessage(_) => {
+                    InterceptAction::IgnoreFromMessage(_) => {
                         blocked = true;
                         continue;
                     }
-                    Action::InterceptMessage(intercept_message) => {
+                    InterceptAction::ReplaceMessage(intercept_message) => {
                         let intercept_frame = StandardEitherFrame::<AnyMessage<'_>>::Sv2(
                             Sv2Frame::from_message(
                                 intercept_message.replacement_message.clone(),
@@ -331,7 +334,7 @@ impl Sniffer {
         recv: Receiver<MessageFrame>,
         send: Sender<MessageFrame>,
         upstream_messages: MessagesAggregator,
-        action: Option<Action>,
+        action: Option<InterceptAction>,
     ) -> Result<(), SnifferError> {
         // Blocking flag used in the IgnoreFromMessage action to stop processing messages after the
         // desired interception.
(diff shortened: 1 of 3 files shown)


roles/tests-integration/lib/sniffer.rs
```diff
@@ -59,7 +59,47 @@ pub struct Sniffer {
     messages_from_downstream: MessagesAggregator,
     messages_from_upstream: MessagesAggregator,
     check_on_drop: bool,
-    intercept_messages: Vec<InterceptMessage>,
+    actions: Vec<Action>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[repr(u8)]
+pub enum Action {
```
jbesraa, line 67: Please add docs

roles/tests-integration/lib/sniffer.rs
```diff
@@ -59,7 +59,47 @@ pub struct Sniffer {
     messages_from_downstream: MessagesAggregator,
     messages_from_upstream: MessagesAggregator,
     check_on_drop: bool,
-    intercept_messages: Vec<InterceptMessage>,
+    actions: Vec<Action>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[repr(u8)]
```
jbesraa, line 66: Why do we need this?

roles/tests-integration/lib/sniffer.rs
```diff
@@ -59,7 +59,47 @@ pub struct Sniffer {
     messages_from_downstream: MessagesAggregator,
     messages_from_upstream: MessagesAggregator,
     check_on_drop: bool,
-    intercept_messages: Vec<InterceptMessage>,
+    actions: Vec<Action>,
```
jbesraa, line 62: i would try to avoid the ordering stuff. we can do that by either using `action: Action` here or use a data type that provide a grantee over ordering

jbesraa (changes requested):
Looks good. Can you please add a test or expand another test under `sniffer_integration` to cover `BlockFromMessage` behavior?

roles/tests-integration/lib/sniffer.rs
jbesraa: the `Vec<Action>` is a nice to have but we never needed it really

roles/tests-integration/lib/sniffer.rs
```diff
@@ -59,7 +59,47 @@ pub struct Sniffer {
...
+            (Action::BlockFromMessage(_), Action::InterceptMessage(_)) => std::cmp::Ordering::Less,
+            (Action::InterceptMessage(_), Action::BlockFromMessage(_)) => {
+                std::cmp::Ordering::Greater
+            }
+            _ => std::cmp::Ordering::Equal,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
plebhash, line 90: we should also have docs here, in the same style as `InterceptMessage`

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -142,3 +152,44 @@ async fn test_sniffer_wait_for_message_type_with_remove() {
...
+
+    // `sniffer_b` is placed downstream of `sniffer_a` and should receive nothing.
+    let (sniffer_b, sniffer_b_addr) =
+        start_sniffer("B".to_string(), sniffer_a_addr, false, None).await;
+
+    // Start the Pool, connected to `sniffer_b`.
+    let _ = start_pool(Some(sniffer_b_addr)).await;
+
+    // Assert that `sniffer_b` does not receive any messages, confirming `sniffer_a`'s block works.
+    assert!(sniffer_b.next_message_from_upstream().is_none());
```
plebhash, line 194: after the insights we had from the discussions early this week, I wonder if there's any room for this to ever be a false positive?

in other words: what if this assertion is ok simply because the message hasn't arrived YET?

one way to be 100% is to add a `sniffer_a.wait_for_message_type(MessageDirection::ToDownstream, MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS).await` before this assertion

roles/tests-integration/lib/sniffer.rs
```diff
@@ -59,7 +59,36 @@ pub struct Sniffer {
...
+    action: Option<Action>,
+}
+
+/// Represents an action that [`Sniffer`] can take on intercepted messages.
+#[derive(Debug, Clone)]
+pub enum Action {
+    /// Blocks the message stream after encountering a specific message.
+    BlockFromMessage(BlockFromMessage),
+    /// Intercepts and modifies a message before forwarding it.
+    InterceptMessage(Box<InterceptMessage>),
```
plebhash, line 71: why `Box`?

roles/tests-integration/lib/sniffer.rs
```diff
@@ -215,30 +244,56 @@ impl Sniffer {
...
     ) -> Result<(), SnifferError> {
+        let mut blocked = false;
         while let Ok(mut frame) = recv.recv().await {
+            if blocked {
+                continue;
+            }
             let (msg_type, msg) = Self::message_from_frame(&mut frame);
-            let intercept_message = intercept_messages.iter().find(|im| {
-                im.direction == MessageDirection::ToUpstream && im.expected_message_type == msg_type
+            let action = action.as_ref().and_then(|action| match action {
```
jbesraa, line 255: I would move this `match` into a specific function in `Action::fn_name(msgtype, msg_direction)` so we dont repeat it in the other function

roles/tests-integration/lib/sniffer.rs
```diff
@@ -253,31 +308,58 @@ impl Sniffer {
...
         while let Ok(mut frame) = recv.recv().await {
+            if blocked {
+                continue;
+            }
             let (msg_type, msg) = Self::message_from_frame(&mut frame);
-            let intercept_message = intercept_messages.iter().find(|im| {
-                im.direction == MessageDirection::ToDownstream
-                    && im.expected_message_type == msg_type
+
+            let action = action.as_ref().and_then(|action| match action {
```
jbesraa, line 320: ^^

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -152,3 +152,44 @@ async fn test_sniffer_wait_for_message_type_with_remove() {
...
+        MessageDirection::ToDownstream,
+        MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+    );
+
+    // `sniffer_a` intercepts and blocks `SetupConnectionSuccess` messages.
+    let (_sniffer_a, sniffer_a_addr) = start_sniffer(
+        "A".to_string(),
+        tp_addr,
+        false,
+        Some(sniffer::Action::BlockFromMessage(block_from_message)),
```
jbesraa, line 182: It would be nice to implement `impl From<BlockFromMessage> for Action` and `From<InterceptMessage>`, this way we can just `block_from_message.into()`.

roles/tests-integration/lib/sniffer.rs
jbesraa: it is probably because of  `replacement_message: PoolMessages<'static>`. It feels like something to actually fix in a different PR. i.e., try to optimize `PoolMessages` size.

jbesraa (changes requested):

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -142,3 +142,44 @@ async fn test_sniffer_wait_for_message_type_with_remove() {
...
+/// passes through two sniffers (`sniffer_a` and `sniffer_b`) before reaching the Pool.
+///
+/// - `sniffer_a` is configured to block `SetupConnectionSuccess` messages directed downstream.
+/// - `sniffer_b` should receive no messages after initial setup, ensuring the block works.
+///
+/// **Flow:**  
+/// `TP -> sniffer_a -> sniffer_b -> Pool`  
+#[tokio::test]
+async fn test_sniffer_blocks_message() {
+    start_tracing();
```
jbesraa, line 160: we dont need this, tracing is already initialized in the file

roles/tests-integration/lib/sniffer.rs
jbesraa: I think the naming is a bit confusing around here: Maybe `Action` should be `InterceptAction` and `InterceptMessage` should be `ReplaceMessage`

roles/tests-integration/lib/sniffer.rs
```diff
@@ -59,7 +59,42 @@ pub struct Sniffer {
...
+pub struct BlockFromMessage {
+    direction: MessageDirection,
+    expected_message_type: MsgType,
+}
+
+impl BlockFromMessage {
+    /// Creates a new [`BlockFromMessage`] action.
+    ///
+    /// - `direction`: The direction of the message stream to block.
+    /// - `expected_message_type`: The type of message after which the stream should be blocked.
```
jbesraa, line 110: IMO this is not necessary. while it is good to be verbose in documentation, we don't want to restate the obvious stuff.

roles/tests-integration/lib/sniffer.rs
```diff
@@ -215,30 +256,56 @@ impl Sniffer {
         recv: Receiver<MessageFrame>,
         send: Sender<MessageFrame>,
         downstream_messages: MessagesAggregator,
-        intercept_messages: Vec<InterceptMessage>,
+        action: Option<Action>,
     ) -> Result<(), SnifferError> {
+        let mut blocked = false;
```
jbesraa, line 286: Please add some info above this variables to describe when its value is changed

roles/tests-integration/lib/sniffer.rs
```diff
@@ -58,7 +58,67 @@ pub struct Sniffer {
...
+    /// Blocks the message stream after encountering a specific message.
+    IgnoreFromMessage(IgnoreFromMessage),
+    /// Intercepts and modifies a message before forwarding it.
+    InterceptMessage(Box<InterceptMessage>),
+}
+
+impl Action {
+    /// Returns the action if it is a downstream `IgnoreFromMessage` or `InterceptMessage`  
+    /// with the specified message type.
+    pub fn filter_downstream(
```
plebhash, line 76: what is the purpose of this function?

plebhash: why is it called `filter_downstream` and it's used for both directions?

plebhash: `downstream_or_upstream` feels a bit redundant? every message is either flowing downstream or upstream

roles/tests-integration/lib/sniffer.rs
plebhash: > I think the naming is a bit confusing around here: Maybe `Action` should be `InterceptAction` and `InterceptMessage` should be `ReplaceMessage`

I agree

roles/tests-integration/lib/sniffer.rs
plebhash: from a prompt to `claude-3.7-sonnet`:

> Here are some better name suggestions:
> - `find_matching_action` - Clearly indicates it's searching for an action that matches criteria
> - `get_action_if_matches` - Describes that it returns the action only if it matches
> - `find_action_for_message` - Focuses on finding an action for a specific message
> - `match_action_by_type_and_direction` - Very explicit about the matching criteria
> - `get_applicable_action` - Indicates it returns an action if it's applicable to the given parameters
> - `find_action_for_message_type` - Emphasizes that it's finding an action for a specific message type
>
> My top recommendation would be `find_matching_action` because:
> - It clearly communicates the function's purpose (finding an action that matches criteria)
> - It doesn't imply anything about "downstream" which seems misleading since the function can match actions for any direction
> - It follows common naming conventions for functions that search for something based on criteria

roles/tests-integration/lib/sniffer.rs
```diff
@@ -58,19 +58,79 @@ pub struct Sniffer {
...
+    /// Blocks the message stream after encountering a specific message.
+    IgnoreFromMessage(IgnoreFromMessage),
+    /// Intercepts and modifies a message before forwarding it.
+    ReplaceMessage(Box<ReplaceMessage>),
+}
+
+impl InterceptAction {
+    /// Returns the action if it is `IgnoreFromMessage` or `ReplaceMessage`  
+    /// with the specified message type.
+    pub fn find_matching_action(
```
jbesraa, line 76: Maybe something like the following to reduce duplicated code(and also receive `&self` instead of `Option<action>`, otherwise why would you call this function with a `None` value?):

```suggestion
    pub fn find_matching_action(
    &self,
    msg_type: MsgType,
    direction: MessageDirection,
) -> Option<InterceptAction> {
    match self {
        InterceptAction::IgnoreFromMessage(bm)
        | InterceptAction::ReplaceMessage(bm)
            if bm.direction == direction && bm.expected_message_type == msg_type =>
        {
            Some(action)
        }
        _ => None,
    }
}
```

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -141,3 +141,56 @@ async fn test_sniffer_wait_for_message_type_with_remove() {
...
+async fn test_sniffer_blocks_message() {
+    let (_tp, tp_addr) = start_template_provider(None);
+
+    // Define an action to block SetupConnectionSuccess messages going downstream.
+    let block_from_message = IgnoreFromMessage::new(
+        MessageDirection::ToDownstream,
+        MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+    );
+
+    // `sniffer_a` intercepts and blocks `SetupConnectionSuccess` messages.
```
jbesraa, line 167: `sniffer_a` is just forwarding the messages, I think this is for `sniffer_b`

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -141,3 +141,56 @@ async fn test_sniffer_wait_for_message_type_with_remove() {
...
+    let (sniffer_a, sniffer_a_addr) = start_sniffer("A".to_string(), tp_addr, false, None).await;
+
+    // `sniffer_b` is placed downstream of `sniffer_a` and should receive nothing.
```
jbesraa, line 170: This should receive and block

jbesraa (approved):
Looks good. Added a few nits but still approving this as it blocking a couple of tests. I can handle the nits in a followup

plebhash (approved):

PR #1538, opened 2025-03-07
Improve integration tests

### LINKED ISSUE #1531: flaky tproxy and jd test
https://github.com/stratum-mining/stratum/actions/runs/13693069154/job/38289648289?pr=1518

I think @GitGab19 is investigating this

GitGab19:
This PR adapts the `shares_per_minute` used to compute the downstream diff in the translator on our integration tests. 

By using the same value currently [hardcoded](https://github.com/stratum-mining/stratum/blob/02d14fca6a3caa95b4448dc7bd287e111ed38e92/roles/pool/src/lib/mining_pool/mod.rs#L467) in the Pool, both upstream and downstream diff are now the same.

This means that the diff set to the `sv1_mining_device` in the flaky test referenced on [this issue](https://github.com/stratum-mining/stratum/issues/1531) is now the same of the one set by the `Pool` to the `JDC/Translator`.

So now the first and unique share produced and sent by the sv1_mining_device satisfies both difficulties, and so a `SubmitSharesExtended` message is sent by JDC to the Pool.

Closes #1531


roles/tests-integration/tests/tproxy_jd_miner.rs
plebhash, line 1: why are we merging the two tests back into the same file?

GitGab19: because I'm pretty sure that having them in separate files or not does not change anything wrt tests failures

roles/tests-integration/lib/mod.rs
```diff
@@ -217,9 +217,9 @@ pub async fn start_sv2_translator(upstream: SocketAddr) -> (TranslatorSv2, Socke
     .expect("failed");
     let listening_address = get_available_address();
     let listening_port = listening_address.port();
-    let hashrate = measure_hashrate(1) as f32 / 100.0;
+    let hashrate = measure_hashrate(1) as f32 / 1000.0;
```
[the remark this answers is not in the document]
GitGab19: since I had to put the same `shares_per_minute` as the Pool, which is 1, I thought about decreasing the hashrate as a sort of rebalancing of the diff

plebhash (approved):

Issue #1514, opened 2025-02-27
Move roles integration tests to `test` folder in root


jbesraa:
Currently it is setting inside roles workspace and affects its size unnecessarily

PR #1548, opened 2025-03-10
Exclude integrations tests from tests ci

### LINKED ISSUE #1547: Exclude integration tests from general test ci
A lot of noise is coming from this in CI - integration tests run 3 times on each PR, should keep specific workflow only

jbesraa:
resolves #1547

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -73,7 +73,8 @@ jobs:
           cargo test --manifest-path=benches/Cargo.toml
           cargo test --manifest-path=common/Cargo.toml
           cargo test --manifest-path=protocols/Cargo.toml
-          cargo test --manifest-path=roles/Cargo.toml
+          # Integration tests are run in a separate job
+          cargo test --manifest-path=roles/Cargo.toml --workspace --exclude integration_tests_sv2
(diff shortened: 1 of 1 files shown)


plebhash:
tbh this feels a bit redundant with https://github.com/stratum-mining/stratum/issues/1514

jbesraa:
> tbh this feels a bit redundant with https://github.com/stratum-mining/stratum/issues/1514

I am not very sure on 1514 yet tbh. The nice thing now is that when you are working on the roles, integration test immediately raise a flag if something not working(in local flow) while if we move it out it wont be the case.

jbesraa:
> > > tbh this feels a bit redundant with #1514
> > 
> > 
> > I am not very sure on 1514 yet tbh. The nice thing now is that when you are working on the roles, integration test immediately raise a flag if something not working(in local flow) while if we move it out it wont be the case.
> 
> Can't we just remove tests-integration from workspace member list, and run it manually in ci? We dont need to move it out of roles module, it just wont be part of workspace..

When developing locally `roles` workspace is failing immediately if any changes are not reflected in tests-integration which is great because it pushes devs to amend the tests immediately, what you are suggesting wont address this.

plebhash:
closing in favor of #1554

Issue #1458, opened 2025-02-05
need to standardize dependency (and API re-exporting) patterns


plebhash:
currently most of the `roles` crates import crates from `protocols` in a redundant way

ideally, crates like `network_helpers_sv2` and `roles_logic_sv2` should re-export all the needed items, and we should avoid redundant imports on `roles`

for example: a role shouldn't really need to declare `codec_sv2` as an explicit dependency... that should be an explicit dependency for `network_helpers_sv2`, and if role needs some API from `codec_sv2`, `network_helpers_sv2` should simply re-export it

PR #1549, opened 2025-03-10
Remove protocol deps from `jd-client` crate

jbesraa:
blocked by #1483 
partially addresses https://github.com/stratum-mining/stratum/issues/1458


GitGab19:
This PR has conflicts to be solved

PR #1550, opened 2025-03-10
change `IgnoreFromMessage` to `IgnoreMessage`

### LINKED ISSUE #1506: ITF `Sniffer` needs functionality to block messages

plebhash:
as described here https://github.com/stratum-mining/stratum/issues/1506#issuecomment-2710808939 the previous scope from #1510 was doing more than needed

close #1506

### DIFF
--- a/roles/tests-integration/lib/sniffer.rs
+++ b/roles/tests-integration/lib/sniffer.rs
@@ -64,14 +64,14 @@ pub struct Sniffer {
 /// Represents an action that [`Sniffer`] can take on intercepted messages.
 #[derive(Debug, Clone)]
 pub enum InterceptAction {
-    /// Blocks the message stream after encountering a specific message.
-    IgnoreFromMessage(IgnoreFromMessage),
+    /// Prevents a message from being forwarded and stored into the message aggregator.
+    IgnoreMessage(IgnoreMessage),
     /// Intercepts and modifies a message before forwarding it.
     ReplaceMessage(Box<ReplaceMessage>),
 }
 
 impl InterceptAction {
-    /// Returns the action if it is `IgnoreFromMessage` or `ReplaceMessage`  
+    /// Returns the action if it is `IgnoreMessage` or `ReplaceMessage`
     /// with the specified message type.
     pub fn find_matching_action(
         &self,
@@ -79,7 +79,7 @@ impl InterceptAction {
         direction: MessageDirection,
     ) -> Option<&Self> {
         match self {
-            InterceptAction::IgnoreFromMessage(bm)
+            InterceptAction::IgnoreMessage(bm)
                 if bm.direction == direction && bm.expected_message_type == msg_type =>
             {
                 Some(self)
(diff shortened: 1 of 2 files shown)


roles/tests-integration/lib/sniffer.rs
```diff
@@ -64,22 +64,23 @@ pub struct Sniffer {
 /// Represents an action that [`Sniffer`] can take on intercepted messages.
 #[derive(Debug, Clone)]
 pub enum InterceptAction {
-    /// Blocks the message stream after encountering a specific message.
-    IgnoreFromMessage(IgnoreFromMessage),
+    /// Prevents a message from being forwarded.
+    /// The message will be added to the message aggregator but not sent to the destination.
```
jbesraa, line 68: Why would you save the message? If this is really needed, we should mark the message as `Ignored` or something, otherwise this creates a bit of uncertainty about the sniffer

roles/tests-integration/tests/sniffer_integration.rs
```diff
@@ -154,41 +154,52 @@ async fn test_sniffer_wait_for_message_type_with_remove() {
...
 
+    // Allow time for the message to be intercepted and blocked
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+
     // Assert that `sniffer_c` does not receive any messages, confirming `sniffer_b`'s block works.
     assert!(sniffer_c.next_message_from_upstream().is_none());
+
+    // Assert that `sniffer_b` received the message, while the assertion above proves that it was
+    // successfully ignored
+    assert_common_message!(
```
jbesraa, line 201: If you want to do it for testing, I would add a specified property for `InterceptedMessages` and not the same upstream/downstream vec

roles/tests-integration/lib/sniffer.rs
plebhash: it's not REALLY needed tbh

as you pointed out, my original motivation was mostly the test in `sniffer_integration`

but I just noticed that for `InterceptAction::ReplaceMessage`, we actually store the replacement message into `MessagesAggregator`

so the correct pattern here would be to not store it, or to change the internal representation in `MessagesAggregator` as you pointed out

I tried introducing this:
```diff
+#[derive(Debug, Clone)]
+enum MessageStatus {
+    Forwarded,
+    Ignored,
+    ReplacedBy(AnyMessage<'static>),
+}
+
 #[derive(Debug, Clone)]
 struct MessagesAggregator {
-    messages: Arc<Mutex<VecDeque<(MsgType, AnyMessage<'static>)>>>,
+    messages: Arc<Mutex<VecDeque<(MsgType, AnyMessage<'static>, MessageStatus)>>>,
 }
```

but it started having some consequences that went deep into the macros, exploding the complexity of this change

while I think this is an interesting direction to be explored, it's not worth it for now

so I'm taking the simpler approach and not storing the ignored message

roles/tests-integration/tests/sniffer_integration.rs
plebhash: simplified the approach, as explained here https://github.com/stratum-mining/stratum/pull/1550#discussion_r1987892731

jbesraa (approved):

GitGab19 (approved):

PR #1551, opened 2025-03-10
Pool unblocking + shutdown

jbesraa:
Partially addresses #1320

### DIFF
--- a/roles/pool/src/lib/mining_pool/mod.rs
+++ b/roles/pool/src/lib/mining_pool/mod.rs
@@ -270,47 +270,58 @@ impl Pool {
     async fn accept_incoming_connection(
         self_: Arc<Mutex<Pool>>,
         config: PoolConfig,
+        mut recv_stop_signal: tokio::sync::watch::Receiver<()>,
     ) -> PoolResult<()> {
         let status_tx = self_.safe_lock(|s| s.status_tx.clone())?;
         let listener = TcpListener::bind(&config.listen_address()).await?;
-        info!(
-            "Listening for encrypted connection on: {}",
-            config.listen_address()
-        );
-        while let Ok((stream, _)) = listener.accept().await {
-            let address = stream.peer_addr().unwrap();
-            debug!(
-                "New connection from {:?}",
-                stream.peer_addr().map_err(PoolError::Io)
-            );
+        info!("Pool is running on: {}", config.listen_address());
+        // Run the listener in the background
+        task::spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = recv_stop_signal.changed() => {
+                        info!("Pool is stopping the server after stop shutdown signal received");
+                        break;
+                    },
 
-            let responder = Responder::from_authority_kp(
-                &config.authority_public_key().into_bytes(),
-                &config.authority_secret_key().into_bytes(),
-                std::time::Duration::from_secs(config.cert_validity_sec()),
-            );
-            match responder {
-                Ok(resp) => {
-                    if let Ok((receiver, sender, _, _)) =
-                        Connection::new(stream, HandshakeRole::Responder(resp)).await
-                    {
-                        handle_result!(
-                            status_tx,
-                            Self::accept_incoming_connection_(
-                                self_.clone(),
-                                receiver,
-                                sender,
-                                address
-                            )
-                            .await
-                        );
+                    result = listener.accept() => {
+                        match result {
+                            Ok((stream, _)) => {
+                                let address = stream.peer_addr().unwrap();
+                                info!("New connection from {:?}", stream.peer_addr().map_err(PoolError::Io));
+                                let responder = Responder::from_authority_kp(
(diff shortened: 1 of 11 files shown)


roles/pool/src/lib/mod.rs
```diff
@@ -17,16 +18,30 @@ use tokio::select;
 #[derive(Debug, Clone)]
 pub struct PoolSv2 {
     config: PoolConfig,
+    status_tx: Arc<Mutex<Option<async_channel::Sender<status::Status>>>>,
```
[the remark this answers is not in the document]
jbesraa: It is not about the sender, it is about updating the sender inside `start` function. Otherwise you would need to take ownership in `start(mut self)` function.

roles/pool/src/lib/mod.rs
```diff
@@ -53,54 +68,88 @@ impl PoolSv2 {
...
-                        .is_err()
+            }
+        });
+        Ok(())
+    }
+
+    pub fn shutdown(&self) {
+        info!("Attempting to shutdown pool");
+        if let Ok(status_tx) = &self.status_tx.lock() {
+            if let Some(status_tx) = status_tx.as_ref().cloned() {
```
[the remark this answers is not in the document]
jbesraa: Did you remove the mutex locally and things are working for you?

roles/pool/src/lib/mod.rs
```diff
@@ -148,4 +197,32 @@ mod tests {
         let result = pool.start().await;
         assert!(result.is_err());
     }
+
+    #[tokio::test]
+    async fn shutdown_pool() {
+        let config_path = "config-examples/pool-config-hosted-tp-example.toml";
```
plebhash, line 203: despite the fact that some MG tests do this, overall I feel it's not a good practice to make unit tests depend on hosted infrastructure

can't we use the TP from integration tests for this?

anyways not something I'm strongly opinionated about, but still think it's worth pointing out

roles/pool/src/lib/mod.rs
```diff
@@ -148,4 +197,32 @@ mod tests {
...
+                return;
+            }
+        };
+        let pool_0 = PoolSv2::new(config.clone());
+        let pool_1 = PoolSv2::new(config);
+        assert!(pool_0.start().await.is_ok());
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+        assert!(pool_1.start().await.is_err());
+        pool_0.shutdown();
+        assert!(pool_1.start().await.is_ok());
```
plebhash, line 226: this is failing when I run the tests locally:
```
test tests::shutdown_pool ... FAILED

failures:

---- tests::shutdown_pool stdout ----
thread 'tests::shutdown_pool' panicked at pool/src/lib/mod.rs:226:9:
assertion failed: pool_1.start().await.is_ok()


failures:
    tests::shutdown_pool

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.00s
```

roles/pool/src/lib/mod.rs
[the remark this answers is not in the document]
plebhash: @Shourya742 the commit you shared has a different approach from what you are suggesting here

is your suggestion to simply remove `Mutex` or to replace it with `OnceLock`?

I tried removing `Mutex` and indeed, we need to take ownership on `start` as a trade-off, but also at that point I don't think we would really need `Arc` anymore

I'm not opinionated towards any specific approach, just pointing out that the review is getting a bit confusing here

roles/pool/src/lib/mod.rs
plebhash: adding a 1s sleep after `pool_0.shutdown()` made it work

plebhash: also failed on CI with same error: https://github.com/stratum-mining/stratum/actions/runs/13833913077/job/38704325199?pr=1551#step:7:944

plebhash (approved):
LGTM but looks like fmt CI is broken?

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -270,47 +270,58 @@ impl Pool {
...
         self_: Arc<Mutex<Pool>>,
         config: PoolConfig,
+        mut recv_stop_signal: tokio::sync::watch::Receiver<()>,
     ) -> PoolResult<()> {
         let status_tx = self_.safe_lock(|s| s.status_tx.clone())?;
         let listener = TcpListener::bind(&config.listen_address()).await?;
-        info!(
-            "Listening for encrypted connection on: {}",
-            config.listen_address()
-        );
```
plebhash, line 279: the fact that we're changing this log is prevenging MG `sv1-test` from asserting that `pool_sv2` has properly launched, which is making the test get stuck

so I'm adding to this PR c8a721bd48f918ce684fda7e0320245ae67f8a98 to unblock CI and proceed with merging

PR #1554, opened 2025-03-11
move IT out of roles workspace

Shourya742:
Closes: #1514

This PR removes IT from the roles workspace and updates the CI accordingly.


test/tests-integration/Cargo.toml
plebhash, line 1: should we take this opportunity to rename this `tests-integration` directory?

I remember @GitGab19 saying that he wanted to change it

GitGab19: I would probably call it `integration-tests`, don't remember specific discussions about this

plebhash (approved):

Issue #1543, opened 2025-03-08
JDC cannot accept more than one downstream connection at a time


plebhash:
```rust
#[tokio::test]
async fn jdc_accepts_multiple_downstreams() {
    start_tracing();
    let (tp, tp_addr) = start_template_provider(None);
    let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
    let (_jds, jds_addr) = start_jds(tp.rpc_info()).await;

    let (_jdc, jdc_addr) = start_jdc(pool_addr, tp_addr, jds_addr).await;

    let (sniffer_jdc_first_tproxy, sniffer_jdc_first_tproxy_addr) = start_sniffer("0".to_string(), jdc_addr, false, None).await;
    let (_first_tproxy, _) = start_sv2_translator(sniffer_jdc_first_tproxy_addr).await;

    let (sniffer_jdc_second_tproxy, sniffer_jdc_second_tproxy_addr) = start_sniffer("1".to_string(), jdc_addr, false, None).await;
    let (_second_tproxy, _) = start_sv2_translator(sniffer_jdc_second_tproxy_addr).await;

    sniffer_jdc_first_tproxy.wait_for_message_type(MessageDirection::ToDownstream, MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS).await;

    // never returns
    // sniffer_jdc_second_tproxy.wait_for_message_type(MessageDirection::ToDownstream, MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS).await;
}
```

if the last line on the integration test above is disabled, the test finishes

if it's enabled, the test never returns and eventually the `wait_for_message_type` times out

plebhash:
this can also be reproduced manually, by launching the roles and trying to connect two tProxies to JDC

the second one is never able to connect, indicating that as soon as JDC accepts the first connection, it stops listening on the TCP socket

if we kill the first tProxy, then the second one is able to connect

plebhash:
regardless of whether this was an intentional design decision, one aspect of JDC does not feel right

as soon as (the only supported) downstream disconnects, JDC also disconnects from TP, JDS and Pool and then re-connects to all of them

this is making Integration Tests quite difficult to write, such as #1505

plebhash:
for now, on #1556 let's try to fix this by just prevent JDC from dropping the `TcpListener` when downstream connects

with regards to the disconnection from TP + JDC + Pool, let's take note under #1477 and address this in the future

PR #1556, opened 2025-03-11
Let JDC continue listening to the downstream port but serve only a single client.

Shourya742:
Addresses: #1543

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
+++ b/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
@@ -206,7 +206,7 @@ impl Share {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 /// Basic logic shared between all the channel factories
 struct ChannelFactory {
     ids: Arc<Mutex<GroupId>>,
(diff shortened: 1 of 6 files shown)


jbesraa:
IMO this kind of PRs can be opened on local fork and examined/discussed before making a super draft here. It just creates noise in the PRs list.

plebhash (approved):

jbesraa:
@plebhash whats the motivation for merging this just out of the blue? I dont see any review process and it seems like the pr is changing alot of code?

plebhash:
> @plebhash whats the motivation for merging this just out of the blue? I dont see any review process and it seems like the pr is changing alot of code?

the goal of this PR was to address #1543 with a limited scope, as described in https://github.com/stratum-mining/stratum/issues/1543#issuecomment-2715206785

I apologize for rushing things here, I see how this can seem like it was merged without any kind of review

I merged it aiming to unblock #1505, after testing it both with regards to #1505 and having multiple downstreams connected (where we expect the second one not to be able to connect, which unblocks the `TcpBind` assertion on #1505)

do you have any material objections to the changes on this PR? if yes, please feel free to point and we can address them in future PRs

PR #1561, opened 2025-03-12
remove `translator-proxy-old-share` and revert #910

### LINKED ISSUE #1560: need to remove `translator-proxy-old-share` and revert #910
on #910 I introduced a patch to tProxy based on the false assumption that it should reject Sv1 shares with `"result": false` simply because they belong to an "outdated" job

this PR also introduced `translator-proxy-old-share` MG test, which asserts this behavior

the correct behavior for tProxy should be:
- if Sv2 upstream accepts share, reply with `"result": true`
- if Sv2 upstream rejects share, reply with `"result": false`

however the current architecture of tProxy doesn't allow for this kind of behavior, because the `handle_submit` function (which handles the Sv1 `mining.submit` message) doesn't have any context about the response from the upstream... it simply sends the share into `tx_sv1_bridge` channel in "fire-and-forget" style

therefore IMO the action plan for now is:
- revert the behavior introduced via #910 
- take note of this tProxy limitation (it always replies with `"result": true`) on its `README.md`
- remove `translator-proxy-old-share` MG test

then on some future PR we can make a deeper refactor of tProxy and properly address this

plebhash:
close #1560


GitGab19 (approved):
tACK

PR #1580, opened 2025-03-18
Add `SnifferSV1` to ITF

### LINKED ISSUE #1564: Integration Tests Framework needs a Sv1 Sniffer
pre-requisite for migration of MG `sv1-test`

jbesraa:
resolves #1564


plebhash:
oops, looks like adding c8a721bd48f918ce684fda7e0320245ae67f8a98 to #1551 introduced a conflict here

perhaps it would have been better to have merged this first, sorry about that

roles/roles-utils/network-helpers/src/sv1_connection.rs
```diff
@@ -0,0 +1,171 @@
...
+    io::{AsyncWriteExt, BufReader},
+    net::TcpStream,
+};
+use tokio_util::codec::{FramedRead, LinesCodec};
+
+/// Represents a connection between two roles communicating using SV1 protocol.
+///
+/// This struct can be used read and write messages to the other side of the connection.  The
+/// channel is unidirectional, so you will need two instances of this struct to communicate in both
+/// directions.
```
plebhash, line 14: this comment feels a bit confusing

with `sender` I can send, with `receiver` I can receive... therefore this is bidirectional?

roles/roles-utils/network-helpers/Cargo.toml
```diff
@@ -20,12 +20,16 @@ tokio = { version = "1", features = ["full"] }
...
 
 [features]
 default = ["async-channel", "binary_sv2", "codec_sv2"]
 with_buffer_pool = ["codec_sv2/with_buffer_pool"]
+sv1 = ["sv1_api", "tokio-util", "serde_json"]
 
 [package.metadata.docs.rs]
-features = ["with_buffer_pool"]
\ No newline at end of file
+features = ["with_buffer_pool"]
```
plebhash, line 35: we should add `sv1` feature here so it is added to docs

roles/test-utils/mining-device-sv1/src/client.rs
```diff
@@ -240,14 +269,21 @@ impl Client {
...
-            if let Ok(incoming) = recv_incoming.clone().recv().await {
-                Self::parse_message(client.clone(), Ok(incoming)).await;
-            } else {
-                warn!("Error reading from socket via `recv_incoming` channel");
-                break;
-            }
-        }
+        tokio::select!(
+            _ = tokio::signal::ctrl_c() => {
+                warn!("Stopping sv1 miner");
```
plebhash, line 274: we just log a warning on ctrl+c?

roles/test-utils/mining-device-sv1/src/client.rs
```diff
@@ -196,6 +222,9 @@ impl Client {
               _ = recv_stop_submitting.changed() => {
                 warn!("Stopping miner")
               },
+              _ = tokio::signal::ctrl_c() => {
+                  info!("Stopping miner");
```
plebhash, line 226: we just print a info log on ctrl+c?

roles/test-utils/mining-device-sv1/src/client.rs
```diff
@@ -106,34 +122,44 @@ impl Client {
...
         task::spawn(async move {
-            let mut messages = BufReader::new(reader).lines();
-            while let Ok(message) = messages.next_line().await {
-                match message {
-                    Some(msg) => {
-                        if let Err(e) = sender_incoming.send(msg).await {
-                            error!("Failed to send message to receiver_incoming: {:?}", e);
-                            break; // Exit the loop if sending fails
+            tokio::select!(
+                _ = tokio::signal::ctrl_c() => { },
```
plebhash, line 126: is this intentionally empty?

test/integration-tests/lib/sv1_sniffer.rs
```diff
@@ -0,0 +1,203 @@
...
+                ) => { },
+            };
+        });
+    }
+
+    /// Wait for a specific message to be received from the downstream role.
+    pub async fn wait_for_message_from_downstream(&self, message: &str) {
+        let now = std::time::Instant::now();
+        tokio::select!(
+            _ = tokio::signal::ctrl_c() => { },
```
plebhash, line 86: is this intentionally empty?

test/integration-tests/lib/sv1_sniffer.rs
```diff
@@ -0,0 +1,203 @@
...
+                }
+            } => {}
+        );
+    }
+
+    /// Wait for a specific message to be received from the upstream role.
+    pub async fn wait_for_message_from_upstream(&self, message: &[&str]) {
+        let now = std::time::Instant::now();
+        tokio::select!(
+            _ = tokio::signal::ctrl_c() => { },
```
plebhash, line 110: is this intentionally empty?

roles/test-utils/mining-device-sv1/src/client.rs
plebhash: sorry dumb question.. just refreshed my understanding of `tokio::select` and now I understand that as soon as ctrl+c happens, all the other branches are cancelled

resolving the similar questions I asked in other parts of the PR

test/integration-tests/lib/sv1_sniffer.rs
```diff
@@ -0,0 +1,203 @@
...
+                    sniffer_to_upstream_connection.receiver(),
+                    downstream_to_sniffer_connection.sender(),
+                    messages_from_upstream_sv1
+                ) => { },
+            };
+        });
+    }
+
+    /// Wait for a specific message to be received from the downstream role.
+    pub async fn wait_for_message_from_downstream(&self, message: &str) {
```
plebhash, line 83: I see this is diverging from the Sv2 Sniffer `wait_for_message_type` pattern

here we have two distinct functions, while on Sv2 Sniffer we just have one and take `MessageDirection` as input

is there some fundamental limitation that forces this specific design?

if not, I think maybe it's better to have both following the same patterns, just to make the APIs a bit more intutive

test/integration-tests/lib/sv1_sniffer.rs
```diff
@@ -0,0 +1,203 @@
...
+        recv: Receiver<sv1_api::Message>,
+        send: Sender<sv1_api::Message>,
+        upstream_messages: MessagesAggregatorSV1,
+    ) -> Result<(), SnifferError> {
+        while let Ok(msg) = recv.recv().await {
+            send.send(msg.clone())
+                .await
+                .map_err(|_| SnifferError::DownstreamClosed)?;
+            upstream_messages.add_message(msg.clone()).await;
+            sleep(Duration::from_millis(100)).await;
```
plebhash, line 140: why sleep here?

test/integration-tests/Cargo.toml
```diff
@@ -31,12 +31,17 @@ key-utils = { path = "../../utils/key-utils" }
...
 stratum-common = { path = "../../common" }
 config-helpers = { path = "../../roles/roles-utils/config-helpers" }
 translator_sv2 = { path = "../../roles/translator" }
+sv1_api = { path = "../../protocols/v1", optional = true }
 
 [lib]
 path = "lib/mod.rs"
+
+[features]
+default = []
```
plebhash, line 46:
```suggestion
default = ["sv1"]
```

having to add `--features sv1` to `cargo test` feels a bit counter-intuitive... and forgetting to do it basically disables the tests that use Sv1 Sniffer

test/integration-tests/lib/sv1_sniffer.rs
```diff
@@ -0,0 +1,203 @@
...
+                        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                        continue;
+                    }
+                }
+            } => {}
+        );
+    }
+
+    /// Wait for a specific message to be received from the upstream role.
+    pub async fn wait_for_message_from_upstream(&self, message: &[&str]) {
```
plebhash, line 107: also this takes an array, while `wait_for_message_from_downstream` doesn't

which is yet another reason I believe these should be just one function instead of two (to avoid fragmented patterns)

plebhash:
I added some tracing logs, please cherry-pick in case you agree

ea02800241e217b137670783c18ec4bc20342e4d

58feb8a66ee419795d5bf06a69f91b3f542647ba

roles/roles-utils/network-helpers/src/sv1_connection.rs
frankomosh: from my understanding I think they say a single Channel instance only handles communication in one direction. But yes I agree that the statement could be clearer.

test/integration-tests/Cargo.toml
plebhash: as discussed in a call with @jbesraa , we still only have one single test where we assert any kind of Sv1 behavior

and including the `sv1` feature flag as default would bloat the ITF build process for every other test

so for now, it's ok to leave this as not being a default feature

test/integration-tests/lib/sv1_sniffer.rs
plebhash: as discussed in call with @jbesraa , there are some fundamental limitations coming from `sv1_api::json_rpc` module that are imposing this pattern

test/integration-tests/lib/sv1_sniffer.rs
plebhash: as discussed in call with @jbesraa , there are some fundamental limitations coming from sv1_api::json_rpc module that are imposing this pattern

jbesraa:
> I added some tracing logs, please cherry-pick in case you agree
> 
> [ea02800](https://github.com/stratum-mining/stratum/commit/ea02800241e217b137670783c18ec4bc20342e4d)
> 
> [58feb8a](https://github.com/stratum-mining/stratum/commit/58feb8a66ee419795d5bf06a69f91b3f542647ba)

Squashed the SV1 related one to SV1 commit and  picked SV2 commit in

roles/roles-utils/network-helpers/src/sv1_connection.rs
jbesraa: @frankomosh is right. 
Updated the docs. Let me know if it is still not clear.

test/integration-tests/tests/sv1.rs
```diff
@@ -0,0 +1,27 @@
...
+        .wait_for_message(
+            &[
+                "minimum-difficulty",
+                "version-rolling",
+                "version-rolling.mask",
+                "version-rolling.min-bit-count",
+            ],
+            MessageDirection::ToDownstream,
+        )
+        .await;
```
plebhash, line 26: looks like we're missing `subscribe`, `authorize` and `notify`

`notify` is not on the original MG test, but it's still worth adding since it will happen on this setup

GitGab19 (approved):

Issue #1419, opened 2025-01-29
CI migration:  `jds-receive-solution-while-processing-declared-job` MG to Integration Test


plebhash:
migrate `jds-receive-solution-while-processing-declared-job` MG to Integration Test

plebhash:
looks like this is blocking https://github.com/stratum-mining/stratum/pull/1429

plebhash:
unblocked from #1420 but now blocked by #1506 (`Sniffer` feature for blocking messages)

while the original MG test is mocking a JDC to deliberately skip `ProvideMissingTransactions.Success`, we can achieve this by simply:
- asserting that JDS sent `ProvideMissingTransactions`
- blocking `ProvideMissingTransactions.Success` response from JDC

plebhash:
> we still need to be able to submit a transaction to TP though

actually not really... the intercepted `DeclareMiningJob` could be originally empty, and then modified to contain one

that would also trigger `ProvideMissingTransactions `

PR #1581, opened 2025-03-18
jds receive solution while processing declared job IT

Shourya742:
closes: #1419


test/integration-tests/lib/template_provider.rs
```diff
@@ -149,6 +149,12 @@ impl TemplateProvider {
         client.generate_to_address(101, &address)?;
         Ok(())
     }
+
+    pub fn no_of_blocks(&self) -> Result<i64, corepc_node::Error> {
```
jbesraa, line 153: What is "no_of_blocks"? please use full clear names.
I also dont see this is used anywhere? and it is missing docs.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -74,14 +80,14 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
...
-    let (jds_jdc_sniffer, jds_jdc_sniffer_addr) = start_sniffer(
+    let (jds_jdc_sniffer, jds_jdc_sniffer_bddr) = start_sniffer(
         "JDS-JDC-sniffer".to_string(),
         jds_addr,
         false,
         Some(block_from_message.into()),
     )
     .await;
-    let (_jdc, jdc_addr) = start_jdc(&[(pool_addr, jds_jdc_sniffer_addr)], tp_addr).await;
+    let (_jdc, jdc_addr) = start_jdc(&[(pool_addr, jds_jdc_sniffer_bddr)], tp_addr).await;
```
jbesraa, line 90: this change and everything not in the new test added seems redundant?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -108,3 +114,130 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
...
     assert!(tokio::net::TcpListener::bind(jdc_addr).await.is_err());
 }
+
+///The issue arises from the following message flow. JDC first receives a `NewTemplate`,
+/// prompting it to send a `RequestTransactionData` message to TP. TP responds with
+/// `RequestTransactionDataSuccess`, after which JDC sends a `DeclareMiningJob` to JDS. JDS then
+/// requests missing transactions from JDC via `ProvideMissingTransactions`. While JDC is in the
+/// process of responding with `ProvideMissingTransactionsSuccess`, a block is found, leading JDC
+/// to send a `SubmitSolution` to JDS. However, since JDS has not yet received the requested
+/// transactions from JDC, it results in shutdown. This test tries to mimic the given situation.
```
jbesraa, line 124: I think this is too verbose. the test name is already very clear. if you still wanna add logs for this, please first use "//" rather than "///" and keep it simple and accurate.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -108,3 +114,130 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
...
+
+    // This sniffer sits between `sniffer_a` and `jdc`, replacing `ProvideMissingTransactionSuccess`
+    // with `SubmitSolution`.
+    let (sniffer_b, sniffer_b_addr) = start_sniffer(
+        "B".to_string(),
+        sniffer_a_addr,
+        false,
+        Some(submit_solution_replace.into()),
+    )
+    .await;
```
jbesraa, line 173: I dont understand why you need two sniffers?

jbesraa (changes requested):
Please just fixup your commits into 1 or 2 commits. This is repetitive behavior where you push code in one commit and then remove it in the next and then the pull request is just full of noise and hard to keep track of. Please make it ready for review when it meets the commit history/structure mentioned here https://github.com/stratum-mining/stratum/blob/main/CONTRIBUTING.md

test/integration-tests/tests/jd_integration.rs
jbesraa: `///` is for `pub` modules and `//` is for internal docs

test/integration-tests/tests/jd_integration.rs
jbesraa: no need to test this. we know how the sniffer behaves and we have tests specifically for that in sniffer_integration. you can get away with this test with a single sniffer.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -108,3 +114,127 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
     tokio::time::sleep(Duration::from_secs(3)).await;
     assert!(tokio::net::TcpListener::bind(jdc_addr).await.is_err());
 }
+
+// The issue arises from the following message flow. JDC first receives a `NewTemplate`,
```
jbesraa, line 118: This text is not following any convention or any other test doc style. Please look into other tests and how they are documented and amend accordingly.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -108,3 +114,127 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
...
+        .await;
+
+    sniffer_a
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_PROVIDE_MISSING_TRANSACTIONS,
+        )
+        .await;
+
+    println!("HERE");
```
jbesraa, line 231:
```suggestion
```

jbesraa (commented):
A tip about your commit history: think about your commits in an atomic way. Atomic commits mean that each commit passes all the CI checks. Lets examine your pull request, your first commit is removing a test, if we run the ci on this commit the ci will try to actually run the test because it was only deleted from CI in the second commit. Making your first commit non atomic. So I suggest to fixup the second commit to the first one.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -27,8 +33,8 @@ async fn jds_should_not_panic_if_jdc_shutsdown() {
     // wait for shutdown to complete
     tokio::time::sleep(std::time::Duration::from_secs(1)).await;
     assert!(tokio::net::TcpListener::bind(jdc_addr).await.is_ok());
-    let (sniffer, sniffer_addr) = start_sniffer("0".to_string(), jds_addr, false, None).await;
-    let (_jdc_1, _jdc_addr_1) = start_jdc(&[(pool_addr, sniffer_addr)], tp_addr).await;
+    let (sniffer, sniffer_bddr) = start_sniffer("0".to_string(), jds_addr, false, None).await;
```
jbesraa, line 36: why this change?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -27,8 +33,8 @@ async fn jds_should_not_panic_if_jdc_shutsdown() {
...
+    let (_jdc_1, _jdc_addr_1) = start_jdc(&[(pool_addr, sniffer_bddr)], tp_addr).await;
```
jbesraa, line 37: why this change?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -42,9 +48,9 @@ async fn jdc_tp_success_setup() {
     let (tp, tp_addr) = start_template_provider(None);
     let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
     let (_jds, jds_addr) = start_jds(tp.rpc_info()).await;
-    let (tp_jdc_sniffer, tp_jdc_sniffer_addr) =
+    let (tp_jdc_sniffer, tp_jdc_sniffer_bddr) =
```
jbesraa, line 51: why this change?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -74,14 +80,14 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
         sniffer::MessageDirection::ToDownstream,
         MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN_SUCCESS,
     );
-    let (jds_jdc_sniffer, jds_jdc_sniffer_addr) = start_sniffer(
+    let (jds_jdc_sniffer, jds_jdc_sniffer_bddr) = start_sniffer(
```
jbesraa, line 83: why this change?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -42,9 +48,9 @@ async fn jdc_tp_success_setup() {
     let (tp, tp_addr) = start_template_provider(None);
     let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
     let (_jds, jds_addr) = start_jds(tp.rpc_info()).await;
-    let (tp_jdc_sniffer, tp_jdc_sniffer_addr) =
+    let (tp_jdc_sniffer, tp_jdc_sniffer_bddr) =
         start_sniffer("0".to_string(), tp_addr, false, None).await;
-    let (_jdc, jdc_addr) = start_jdc(&[(pool_addr, jds_addr)], tp_jdc_sniffer_addr).await;
+    let (_jdc, jdc_addr) = start_jdc(&[(pool_addr, jds_addr)], tp_jdc_sniffer_bddr).await;
```
jbesraa, line 53: why this change?

test/integration-tests/tests/jd_integration.rs
jbesraa: this was not resolved

test/integration-tests/tests/jd_integration.rs
```diff
@@ -108,3 +114,124 @@ async fn jdc_does_not_stackoverflow_when_no_token() {
...
+            MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN,
+        )
+        .await;
+
+    sniffer_a
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN_SUCCESS,
+        )
+        .await;
```
plebhash, line 213: despite the fact that JDC does send `AllocateMiningJobToken` twice, I think that waiting for it 2x here is a bit redundant

that's for two different reasons:
- the "extra" `AllocateMiningJobToken` doesn't have any meaningful influence on this scenario
- the internal logic of `wait_for_message_type` only makes sure that the message has arrived once... after the first time, every other call will be non-blocking (even if the message never arrives a second time)

plebhash: well, it works... even if the second message never arrives

just check its implementation

plebhash:
```rust
#[tokio::test]
async fn aaa() {
    start_tracing();
    let (tp, tp_addr) = start_template_provider(None);
    let (sniffer, sniffer_addr) = start_sniffer("0".to_string(), tp_addr, false, None).await;
    let (_pool, _pool_addr) = start_pool(Some(sniffer_addr)).await;

    sniffer
        .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
        .await;

    sniffer
        .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
        .await;

    sniffer
        .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
        .await;
}
```

this test always returns without any issue

plebhash: > it should check for a message, and when one arrives, it becomes ready, continues execution, and simultaneously removes the message from the stack.

`wait_for_message_type_and_clean_queue` does that

plebhash: this times out 

```rust
#[tokio::test]
async fn bbb() {
    start_tracing();
    let (tp, tp_addr) = start_template_provider(None);
    let (sniffer, sniffer_addr) = start_sniffer("0".to_string(), tp_addr, false, None).await;
    let (_pool, _pool_addr) = start_pool(Some(sniffer_addr)).await;

    sniffer
        .wait_for_message_type_and_clean_queue(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
        .await;

    sniffer
        .wait_for_message_type_and_clean_queue(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
        .await;

    sniffer
        .wait_for_message_type_and_clean_queue(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
        .await;
}
```

plebhash: please note that I'm not saying we should use `wait_for_message_type_and_clean_queue` here

what I was originally trying to say is that there's no point on checking for `AllocateMiningJobToken` 2x

plebhash (approved):

Issue #1566, opened 2025-03-12
CI migration: `jds-do-not-fail-on-wrong-tsdatasucc` MG to Integration Test


plebhash:

plebhash:
@GitGab19 did you start working on it?

GitGab19:
No, I did not

PR #1593, opened 2025-03-24
add jds do not fail on wrong tsdatasucc it

Shourya742:
closes: #1566, #1600


GitGab19:
@Shourya742 could you add something about what you did in this PR?

We don't have any detail in the related issue, and I would like to see what's the context to migrate, and how it was done by this PR.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,70 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+// to check whether it remains alive.
+#[tokio::test]
+async fn jds_do_not_fail_on_wrong_tsdatasucc_test() {
+    start_tracing();
+    let (tp, tp_addr) = start_template_provider(None);
+
+    let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
+    let (_jds, jds_addr) = start_jds(tp.rpc_info()).await;
+
+    let submit_solution_replace = ReplaceMessage::new(
```
jbesraa, line 241: this name `submit_solution_replace` is incorrect

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,70 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+    // with `ProvideMissingTransactionSuccess`.
+    let (sniffer_a, sniffer_a_addr) = start_sniffer(
+        "A".to_string(),
+        jds_addr,
+        false,
+        Some(submit_solution_replace.clone().into()),
+    )
+    .await;
+
+    let _ = start_jdc(&[(pool_addr, sniffer_a_addr)], tp_addr).await;
```
jbesraa, line 264: why do you need to provide the pool_addr and sniffer_a_addr here?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,70 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+    // with `ProvideMissingTransactionSuccess`.
+    let (sniffer_b, sniffer_b_addr) = start_sniffer(
+        "B".to_string(),
+        jds_addr,
+        false,
+        Some(submit_solution_replace.into()),
+    )
+    .await;
+
+    let _ = start_jdc(&[(pool_addr, sniffer_b_addr)], tp_addr).await;
```
jbesraa, line 283: why do you need to provide the pool_addr and sniffer_b_addr here?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,70 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+
+    sniffer_b
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_PROVIDE_MISSING_TRANSACTIONS_SUCCESS,
+        )
+        .await;
+
+    assert!(tokio::net::TcpListener::bind(jds_addr).await.is_err());
```
jbesraa, line 292: why is this needed?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,70 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+        "B".to_string(),
+        jds_addr,
+        false,
+        Some(submit_solution_replace.into()),
+    )
+    .await;
+
+    let _ = start_jdc(&[(pool_addr, sniffer_b_addr)], tp_addr).await;
+
+    sniffer_b
```
jbesraa, line 285: why is this needed?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,70 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+        "A".to_string(),
+        jds_addr,
+        false,
+        Some(submit_solution_replace.clone().into()),
+    )
+    .await;
+
+    let _ = start_jdc(&[(pool_addr, sniffer_a_addr)], tp_addr).await;
+
+    sniffer_a
```
jbesraa, line 266: why is this needed?

jbesraa (changes requested):

plebhash:
IIRC on a call we were talking about how the original MG test was a bit nonsensical and it didn't make much sense to try to replicate it identically

@GitGab19 suggested to migrate this with a slightly different approach

I don't remember exactly what it was, but if I had to guess, it was something along the lines of forcing JDS to send a `ProvideMissingTransactions` and replace `ProvideMissingTransactions.Success` with a bad tx?

GitGab19:
> IIRC on a call we were talking about how the original MG test was a bit nonsensical and it didn't make much sense to try to replicate it identically
> 
> @GitGab19 suggested to migrate this with a slightly different approach
> 
> I don't remember exactly what it was, but if I had to guess, it was something along the lines of forcing JDS to send a `ProvideMissingTransactions` and replace `ProvideMissingTransactions.Success` with a bad tx?

Exactly, this was my idea, but if it's too hard to do it we can migrate it as it was on MG.

Let me know @Shourya742 if it's feasible to do it or not

GitGab19:
> Is the behavior in the second variant expected?

Probably that's because the JDS doesn't manage it properly and just skips it (?)

GitGab19:
I would go with 3

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,97 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
 
     assert!(tokio::net::TcpListener::bind(jds_addr).await.is_err());
 }
+
+// This test verifies that JDS does not exit when it receives a `ProvideMissingTransactionsSuccess`
+// different transaction set, than expected.
```
GitGab19, line 229: What does it mean?

Could you change it with something clearer?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,97 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+    let (_pool, pool_addr) = start_pool(Some(tp_addr_1)).await;
+    let (_jds, jds_addr) = start_jds(tp_1.rpc_info()).await;
+
+    let provide_missing_transaction_success_replace = ReplaceMessage::new(
+        MessageDirection::ToUpstream,
+        MESSAGE_TYPE_PROVIDE_MISSING_TRANSACTIONS_SUCCESS,
+        AnyMessage::JobDeclaration(
+            roles_logic_sv2::parsers::JobDeclaration::ProvideMissingTransactionsSuccess(
+                ProvideMissingTransactionsSuccess {
+                    request_id: 2,
```
GitGab19, line 249: Why do you have request_id = 2 here, and not 1?

GitGab19:
I tried to run it locally on my mac, and I'm getting the following error:
```
thread 'jds_receive_solution_while_processing_declared_job_test' panicked at tests/jd_integration.rs:176:5:
assertion failed: tp_2.fund_wallet().is_ok()
test jds_receive_solution_while_processing_declared_job_test ... FAILED

```

Did you notice it as well locally?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
 }
+
+// This test ensures that JDS does not exit when it receives a `ProvideMissingTransactionsSuccess`
+// message containing a different transaction set than expected, where the `tx_short_hash_list` does
+// not match DMJ.
+//
+// It is performing the verification by connecting to JDS after the message exchange
+// to check whether it remains alive.
+#[tokio::test]
+async fn jds_do_not_fail_on_wrong_tsdatasucc_test() {
```
plebhash, line 235: let's give this test a meaningful name that reflects what's actually going on, we don't have to copy the original MG test name verbatum

on a past call we discussed how `tsdatasucc` is kinda nonsensical

plebhash:
fmt CI is broken

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+            roles_logic_sv2::parsers::JobDeclaration::ProvideMissingTransactionsSuccess(
+                ProvideMissingTransactionsSuccess {
+                    request_id: 1,
+                    transaction_list: Seq064K::new(Vec::new()).unwrap(),
+                },
+            ),
+        ),
+    );
+
+    // This sniffer sits between `jds` and `jdc`, replacing `ProvideMissingTransactionSuccess`
```
jbesraa, line 260: This comment says "replacing `ProvideMissingTransactionSuccess` with `ProvideMissingTransactionSuccess`" which does not make sense. I would remove this comment.

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+    // with `ProvideMissingTransactionSuccess`.
+    let (sniffer_a, sniffer_a_addr) = start_sniffer(
```
jbesraa, line 262: There is only a single sniffer in the test, so this can be just (sniffer, sniffer_addr)

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+        .wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_SETUP_CONNECTION)
+        .await;
+
+    sniffer_a
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+        )
+        .await;
+
```
jbesraa, line 284: Why do you have an empty line after every line of code in the test? It makes reading this less. comfortable

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+    sniffer_a
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_PROVIDE_MISSING_TRANSACTIONS_SUCCESS,
+        )
+        .await;
+
+    assert!(tokio::net::TcpListener::bind(jds_addr).await.is_err());
```
jbesraa, line 320: I think this is a "weak" assertion. I would start another JDC and try to connect it to JDS and make sure they exchange SetupConnection/Success

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+        "A".to_string(),
+        jds_addr,
+        false,
+        Some(provide_missing_transaction_success_replace.into()),
+    )
+    .await;
+
+    let (_jdc, jdc_addr) = start_jdc(&[(pool_addr, sniffer_a_addr)], tp_addr_2).await;
+
+    start_sv2_translator(jdc_addr).await;
```
jbesraa, line 272: Do we need to start this translator?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,98 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
 }
+
+// This test ensures that JDS does not exit upon receiving a `ProvideMissingTransactionsSuccess`
+// message containing a transaction set that differs from the `tx_short_hash_list`
+// in the Declare Mining Job.
+//
+// It is performing the verification by connecting to JDS after the message exchange
+// to check whether it remains alive
+#[tokio::test]
+async fn jds_handles_provide_missing_transaction_success_with_unexpected_tx_list() {
```
jbesraa, line 235: I think 
`jds_wont_exit_upon_receiving_unexpected_txids_in_provide_missing_transaction_success` is more accurate

jbesraa (commented):
This commit 29a2b5d3a1f49d2a716caf606abfa80c2c41f0ab says "remove ..." but it is adding a test and not removing 

I think 582d7e508260c274b73216ac7941b05f08935caa and 72a0b2012d6b0c12021166696eafcf74abe6b7ac should be in a single commit

test/integration-tests/tests/jd_integration.rs
jbesraa: I dont understand why adding an empty line makes a difference?

test/integration-tests/tests/jd_integration.rs
```diff
@@ -221,3 +224,112 @@ async fn jds_receive_solution_while_processing_declared_job_test() {
...
+        .await;
+    sniffer_b
+        .wait_for_message_type(
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_SETUP_CONNECTION_SUCCESS,
+        )
+        .await;
+
+    jdc_2.shutdown();
+    tproxy_2.shutdown();
```
GitGab19, line 332: Do we need to explicitly shutdown roles here?

jbesraa: hmm interesting, are you on mac or linux? did you need to shutdown both of them?

jbesraa: Can you try to add `impl Drop from JobDeclaratorClient` and `impl Drop for TranslatorSV2` in the respective crates and call the shutdown from the drop instead? would that work?

jbesraa: Any chance you can rebase on #1592 and remove the shutdown stuff and see if you still encounter the same staling problem?

jbesraa: but why it showed up now and was not a problem yesterday? was it because the second jdc/translator added for the assertion ?

GitGab19: yeah I also tend to prefer the same assertion we used for other tests for coherence

@Shourya742 do you mind using that, removing these shutdowns?

GitGab19 (approved):

jbesraa (approved):

PR #1546, opened 2025-03-10
Move all roles members dependency to workspace 

Shourya742:
closes: #1545

PR #1603, opened 2025-03-27
Use same(1.44.1) tokio version across the project

### LINKED ISSUE #1596: Align tokio versions between the different crates
We use multiple versions(1, 1.38, 1.4..)

Should probably change all of them to use latest version and preferably mark `default-features = false` and import only needed features.

jbesraa:
resolves #1596

### DIFF
--- a/examples/ping-pong-encrypted/Cargo.toml
+++ b/examples/ping-pong-encrypted/Cargo.toml
@@ -13,5 +13,5 @@ noise_sv2 = { path = "../../protocols/v2/noise-sv2" }
 key-utils = { version = "^1.0.0", path = "../../utils/key-utils" }
 network_helpers_sv2 = { version = "3.0.0", path = "../../roles/roles-utils/network-helpers" }
 rand = "0.8"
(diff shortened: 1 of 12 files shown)


jbesraa:
> Instead of making changes like these, we should focus on getting this merged: #1546.

#1546 is a profound change while this pr is just upgrading/aligning tokio versions. I dont see how they are related. 
Also this is mentioned in the description about shared deps:


Specifying a workspace dependency is similar to [package dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) except:

    1. Dependencies from this table cannot be declared as optional
    
    2. [features](https://doc.rust-lang.org/cargo/reference/features.html) declared in this table are additive with the features from [dependencies]
   
   
Which is a deal breaker IMO. 

I would not block this one because of #1546, specially when we also have #1549 and others. 

#1549 #1546 and others needs further discussions and consensus before we merge them.

GitGab19 (approved):

Issue #1523, opened 2025-03-04
need to avoid running Integration Tests in parallel


plebhash:
we recently discovered that running Integration Tests in parallel is the root cause of https://github.com/stratum-mining/stratum/issues/1278

moreover, as soon as we start having more tests with CPU miners, it will be very important to avoid running them in parallel

plebhash:
there's probably multiple ways to achieve this, but I think the easiest option is to simply isolate one test per file

plebhash:
another way seems to be adding this flag to `cargo test`, but I'm not sure how this could interfere with tokio runtimes

```
cargo test -- --test-threads=1
```

plebhash:
another big motivation for this is the fact that if we run Integration Tests in parallel, their tracing logs get mixed up in CI, which kind of defeats the purpose

plebhash:
cross-referencing https://github.com/stratum-mining/stratum/issues/1167

jbesraa:
I think it is safe to close this @plebhash ?

plebhash:
> I think it is safe to close this [@plebhash](https://github.com/plebhash) ?

there's a few things that would be important to fix here before we move on

being able to read isolated tracing logs on CI is very strategic for our next steps, and I'm still afraid that #1278 could come back to haunt us

here's a proposal on how to tackle this with a new approach https://github.com/stratum-mining/stratum/pull/1619

PR #1619, opened 2025-04-03
isolate ITF execution on CI

plebhash:
As described in #1523, parallel execution of Integration Tests can have many unintended consequences.

This discussion originally revolved around how Integration Tests are distributed across files.

This PR takes a different approach, where it doesn't really matter how Integration Tests are organized across files (1 per file, multiple per file, etc).

We take the tradeoff of a slightly more complex Github Action that aims for the following:
- provide 1 parallel CI job for each test function
- all parallel CI jobs benefit from the same cached build (avoids re-building for every test)
- isolated tracing logs per-test

note: the following tests are currently disabled on their respective .rs files, and therefore were left commented out
- `test_jdc_pool_fallback_after_submit_rejection`
- `translation_proxy_and_jd`

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -6,28 +6,494 @@ on:
 name: Integration Tests
 
 jobs:
-  ci:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os:
-          - ubuntu-latest
-        include:
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-gnu
-
+  build:
+    runs-on: ubuntu-latest
     steps:
-      - name: Use stable toolchain
+      - name: Checkout code
         uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: stable
           override: true
 
-      - name: Roles Integration Tests
-        run: |
-         RUST_BACKTRACE=1 RUST_LOG=debug cargo test --manifest-path=test/integration-tests/Cargo.toml --verbose --test '*' -- --nocapture
+      - name: Build project
+        run: cargo build --manifest-path=test/integration-tests/Cargo.toml --verbose
 
-      - name: SV1 Integration Tests
-        run: |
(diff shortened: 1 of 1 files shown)


jbesraa (changes requested):
Looking at the increase in CI checks (up to 32 now), I'm concerned about the noise generated by having a separate job for each test scenario. Since we can typically run specific failing tests locally in isolation for debugging, I'm not sure the individual CI jobs add proportional value.

Anyway, maybe we could reduce the redundancy as most of these jobs seem identical except for the sv1 test (needing a specific feature flag). Would creating a reusable script or using a CI matrix strategy, where we pass the test names as parameters, be a more maintainable approach?

Note that this approach will also require us to change the CI file on each new test we write which is not optimal IMO, specifically for new contributors.

plebhash:
> Anyway, maybe we could reduce the redundancy as most of these jobs seem identical except for the sv1 test (needing a specific feature flag). Would creating a reusable script or using a CI matrix strategy, where we pass the test names as parameters, be a more maintainable approach?

when I wrote this approach, I had the following goals in mind:
- make sure each test is executed in an isolated worker
- make sure tracing logs are isolated for each test
- avoid re-building every time

but I'm sure there's other ways to achieve this.

jbesraa:
I recall we experienced some performance limitations(here and on forks) when we tried running each test in a separate job. It seemed the large number of concurrent jobs occasionally resulted in significant delays before execution, particularly when multiple PRs were queuing up(two PRS would queue 60+ jobs). As this PR focuses on providing detailed logs for tests in CI, it would be good if we could explore some alternative strategies to ensure both performance and log visibility. ATM I am not entirely sure what those alternatives are TBH.

plebhash:
> I recall we experienced some performance limitations(here and on forks) when we tried running each test in a separate job/worker. It seemed the large number of concurrent jobs occasionally resulted in significant delays before execution, particularly when multiple PRs were queuing up(two PRS would queue 60+ jobs). As this PR focuses on providing detailed logs for tests in CI, it would be good if we could explore some alternative strategies to ensure both performance and log visibility. ATM I am not entirely sure what those alternatives are TBH.

I propose we start using SRI community funds to pay for github (team or enterprise) so we don't have to deal with CI bottlenecks

as we move forward to more doing increasingly more test-driven development to enforce spec compliance, I think this is a reasonable investment

cc @pavlenex @GitGab19 @Fi3

Issue #1465, opened 2025-02-09
`roles_logic_sv2`: refactor `channel_logic` module


plebhash:
todo: elaborate on details of what needs to be done here

plebhash:
as noted by @GitGab19 https://github.com/stratum-mining/stratum/pull/1442#discussion_r1958090707

we should investigate why this module is hard-coding `extranonce_len` as `32`

GitGab19:
we need to remove `hex-conservative` dependency: https://github.com/stratum-mining/stratum/pull/1487#discussion_r1961826101

plebhash:
cross referencing: https://github.com/stratum-mining/stratum/issues/1435#issuecomment-2725349612

plebhash:
we should clean up the `info` logs on this module

GitGab19:
While diving into https://github.com/stratum-mining/stratum/issues/1617, I discovered that our Pool is using an hard-coded value of 10 `shares_per_minute` in the `UpdateChannel` handler:
https://github.com/stratum-mining/stratum/blob/a1bca32770ca4b790dca962b01adfb9fb2e788bc/roles/pool/src/lib/mining_pool/message_handler.rs#L95-L96

I tried to use the value taken from its config file, but there's no way to get it from the `PoolChannelFactory`.

Playing around locally, I added a new function like this:
```
pub fn get_shares_per_minute(&self) -> f32 {
        self.inner.share_per_min
}
```

This made me think about the need of some other similar functions on the factory, so I wanted to take note of it here.

GitGab19:
Another example could be a function to get the target of a specific channel, given the `channel_id`.

plebhash:
as noted in #1438, a feature request by DMND is the ability to update channel `extranonce_prefix` "on-the-fly" (after channel opening procedure)

more specifically, they want to be able to update the `additional_coinbase_script_data` (which is part of `extranonce_prefix`) to multiple channels that are already open

this should be taken as a feature request under the scope of this task

PR #1620, opened 2025-04-04
Improve UpdateChannel/SetTarget handling on Pool and tProxy

### LINKED ISSUE #1617: tProxy sends `UpdateChannel` upstream every minute
By looking at the bandwidth measurements on the `benchmarking-tool`, and on the tProxy logs, I noticed that tProxy always sends a `UpdateChannel` message upstream, every minute.

In case the upstream is the Pool, the Pool answers with a `SetTarget` message.

The issue here is that tProxy should send the `UpdateChannel` only when downstream hashrate changes, not every minute.

A side effect of the current behavior is that bandwidth consumption at the miner level is higher than a single downstream connected to a Sv1 pool.

GitGab19:
Close #1617

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
+++ b/protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
@@ -1292,6 +1292,10 @@ impl PoolChannelFactory {
         self.inner.extended_channels.keys().copied().collect()
     }
 
+    pub fn get_shares_per_minute(&self) -> f32 {
+        self.inner.share_per_min
+    }
+
     /// Update coinbase outputs
     pub fn update_pool_outputs(&mut self, outs: Vec<TxOut>) {
         self.pool_coinbase_outputs = outs;
--- a/roles/pool/src/lib/mining_pool/message_handler.rs
+++ b/roles/pool/src/lib/mining_pool/message_handler.rs
@@ -92,8 +92,14 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
 
     fn handle_update_channel(&mut self, m: UpdateChannel) -> Result<SendTo<()>, Error> {
         info!("Received UpdateChannel message");
-        let maximum_target =
-            roles_logic_sv2::utils::hash_rate_to_target(m.nominal_hash_rate.into(), 10.0)?;
+        let shares_per_minute = self
+            .channel_factory
+            .safe_lock(|s| s.get_shares_per_minute())
+            .map_err(|e| Error::PoisonLock(e.to_string()))?;
+        let maximum_target = roles_logic_sv2::utils::hash_rate_to_target(
+            m.nominal_hash_rate.into(),
+            shares_per_minute.into(),
+        )?;
         self.channel_factory
             .safe_lock(|s| s.update_target_for_channel(m.channel_id, maximum_target.clone().into()))
             .unwrap_or_else(|_| {
--- a/roles/translator/src/lib/upstream_sv2/diff_management.rs
+++ b/roles/translator/src/lib/upstream_sv2/diff_management.rs
@@ -13,36 +13,50 @@ use std::{sync::Arc, time::Duration};
 impl Upstream {
     /// this function checks if the elapsed time since the last update has surpassed the config
     pub(super) async fn try_update_hashrate(self_: Arc<Mutex<Self>>) -> ProxyResult<'static, ()> {
-        let (channel_id_option, diff_mgmt, tx_frame) = self_
+        let (channel_id_option, diff_mgmt, tx_frame, last_sent_hashrate) = self_
             .safe_lock(|u| {
                 (
                     u.channel_id,
                     u.difficulty_config.clone(),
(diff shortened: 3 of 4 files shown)


roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -92,18 +92,43 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
...
-            maximum_target,
-        };
-        Ok(SendTo::Respond(Mining::SetTarget(set_target)))
+        let shares_per_minute = self
+            .channel_factory
+            .safe_lock(|s| s.get_shares_per_minute())
+            .map_err(|e| Error::PoisonLock(e.to_string()))?;
+        let current_target = self
+            .channel_factory
+            .safe_lock(|s| s.get_extended_channel_target(&m.channel_id))
```
plebhash, line 101: how are standard channels handled for this scenario?

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
```diff
@@ -1292,6 +1292,17 @@ impl PoolChannelFactory {
...
+    pub fn get_shares_per_minute(&self) -> f32 {
+        self.inner.share_per_min
+    }
+
+    pub fn get_extended_channel_target(&self, channel_id: &u32) -> Option<Target> {
+        self.inner
+            .extended_channels
+            .get(channel_id)
+            .map(|success| success.target.clone().into())
+    }
```
plebhash, line 1304: why no `get_standard_channel_target`?

plebhash:
here's a suggestion for an integration test:

```rust
use const_sv2::{MESSAGE_TYPE_UPDATE_CHANNEL, MESSAGE_TYPE_SET_TARGET};
use integration_tests_sv2::sniffer::MessageDirection;
use integration_tests_sv2::*;

#[tokio::test]
async fn pool_tproxy_update_channel() {
    start_tracing();
    let (_tp, tp_addr) = start_template_provider(None);
    let (_, pool_addr) = start_pool(Some(tp_addr)).await;

    let (sniffer, sniffer_addr) = start_sniffer("".to_string(), pool_addr, false, None);

    let (_translator, tproxy_addr) = start_sv2_translator(sniffer_addr);

    // start first mining device
    let _mining_device_a = start_mining_device_sv1(tproxy_addr, false, None);

    sniffer.wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_UPDATE_CHANNEL).await;
    // maybe assert some message value?

    // start second mining device to increase nominal hash rate
    let _mining_device_b = start_mining_device_sv1(tproxy_addr, false, None);

    sniffer.wait_for_message_type(MessageDirection::ToUpstream, MESSAGE_TYPE_UPDATE_CHANNEL).await;
    // maybe assert some message value?

    sniffer.wait_for_message_type(MessageDirection::ToDownstream, MESSAGE_TYPE_SET_TARGET).await;
    // maybe assert some message value?
}
```

roles/pool/src/lib/mining_pool/message_handler.rs
GitGab19: This made me realize that we don't have a clear way to get the channel type there..

How could I understand which channel type is the one used to receive a message? I think we should discuss about this together, since I feel this is something to consider while working on https://github.com/stratum-mining/stratum/issues/1465

plebhash: hmm cool yeah let's chat about this!

plebhash: I think this is related to the fact that the current Channel Factory design has one parent extended job at factory level, which is used to create child jobs (which could be extended or standard)

GitGab19: I'm gonna drop the first two commits touching the Pool and the ChannelFactory, since the specific issue is already tackled by the last two commits.

protocols/v2/roles-logic-sv2/src/channel_logic/channel_factory.rs
GitGab19: https://github.com/stratum-mining/stratum/pull/1620#discussion_r2033262769

plebhash (approved):

roles/translator/src/lib/upstream_sv2/diff_management.rs
```diff
@@ -13,36 +13,50 @@ use std::{sync::Arc, time::Duration};
...
+            };
+            let message = Message::Mining(Mining::UpdateChannel(update_channel));
+            let either_frame: StdFrame = message.try_into()?;
+            let frame: EitherFrame = either_frame.into();
+
+            tx_frame.send(frame).await.map_err(|e| {
+                super::super::error::Error::ChannelErrorSender(
+                    super::super::error::ChannelSendError::General(e.to_string()),
+                )
+            })?;
```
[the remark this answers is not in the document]
GitGab19: I moved the code which was already there inside the `if`.. why should we change this in this PR?

PR #1622, opened 2025-04-04
rust `1.85` set in cargo-semver-check workflow

### LINKED ISSUE #1621: Semver Check workflow fails on CI
Recently all our PRs are affected by this failure.

It seems that `cargo-semver-checks` version `0.37.0` (which is specified in our current workflow) doesn't support the Rustdoc format `v39` that's being generated by the current Rust `stable` toolchain.

Example here: https://github.com/stratum-mining/stratum/actions/runs/14263791095/job/39980991404?pr=1620

GitGab19:
Closes #1621

### DIFF
--- a/.github/workflows/semver-check.yaml
+++ b/.github/workflows/semver-check.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Install Rust stable
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.85
           override: true
 
       - name: Cache Cargo registry


plebhash (approved):

plebhash:
> Based on the discussion: [#1275 (comment)](https://github.com/stratum-mining/stratum/issues/1275#issuecomment-2532472582), we had fixed the SemVer version. Would it make sense to update it, considering that, in the near future, stable Rustdoc may again stop supporting SemVer v40 also? Instead of updating SemVer, perhaps we could consider fixing the Rust toolchain version to 1.85.

makes sense, could avoid this happening again

GitGab19:
> Based on the discussion: [#1275 (comment)](https://github.com/stratum-mining/stratum/issues/1275#issuecomment-2532472582), we had fixed the SemVer version. Would it make sense to update it, considering that, in the near future, stable Rustdoc may again stop supporting SemVer v40 also? Instead of updating SemVer, perhaps we could consider fixing the Rust toolchain version to 1.85.

Done here: 89fdeddca843f534e58456ffb89f98c59b867f58

Issue #1448, opened 2025-02-04
`roles_logic_sv2`: refactor `utils` module


GitGab19:
This module has the following issues to be tackled:
- it contains several public functions which are only used internally, so probably some of them should be made private. 
- `utils::get_target` is not used anywhere, therefore it should be cleaned up from the codebase
- `utils::Id` and `utils::GroupId` feel like they belong to modules related to channels and job logic

plebhash:
during #1459 we noticed that some unit tests on `utils` module were completely disabled in CI due to `serde` flag gating (which was disabled in CI)

now that we re-enabled it, it became clear that some tests are broken

ref: https://github.com/stratum-mining/stratum/pull/1459#discussion_r1948092820

a few things worth pointing out:
- the `protocols/v2/roles-logic-sv2/reg-test-block.toml` file had been removed from git history
- `gets_merkle_root_from_path` is currently broken, and we're bypassing the original assertion

plebhash:
related to the comment above: https://github.com/stratum-mining/stratum/issues/1268#issuecomment-2613549011

plebhash:
cross referencing https://github.com/stratum-mining/stratum/pull/1624#discussion_r2040569781

GitGab19:
We should also deprecate our `Target` type in favor of the type provided by `rust-bitcoin`

PR #1624, opened 2025-04-05
[WIP] SRI Channel Management Upgrade

### LINKED ISSUE #1465: `roles_logic_sv2`: refactor `channel_logic` module

plebhash:
close #1465

WIP

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
+++ b/protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
@@ -1,0 +1,365 @@
+//! # Extended Channel Factory
+//!
+//! A factory for creating and managing [`ExtendedChannel`]s.
+
+use crate::{
+    channel_management::{
+        chain_tip::ChainTip,
+        extended_channel::{ExtendedChannel, ExtendedChannelError},
+        ShareValidationError, ShareValidationResult,
+    },
+    mining_sv2::{
+        ExtendedExtranonce as ExtendedExtranonceFactory, ExtendedExtranonceError,
+        OpenExtendedMiningChannel, SubmitSharesExtended, Target, UpdateChannel,
+    },
+    template_distribution_sv2::{NewTemplate, SetNewPrevHash as SetNewPrevHashTDP},
+    utils::hash_rate_to_target,
+};
+use binary_sv2::U256;
+use core::ops::Range;
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
+
+/// A factory for creating and managing [`ExtendedChannel`]s.
+///
+/// Allows the user to:
+/// - Create channels
+/// - Remove channels
+/// - Manage channel states upon receipt of:
+///   - [`OpenExtendedMiningChannel`] (create new channel)
+///   - [`UpdateChannel`] (update channel nominal hashrate and target)
+///   - [`NewTemplate`] (update channel template)
+///   - [`SetNewPrevHashTDP`] (update channel chain tip)
+///   - [`SetCustomMiningJob`] (add new custom mining job)
+///   - [`SubmitSharesExtended`] (validate shares)
+///
+/// Only suitable for mining servers, not clients.
+#[derive(Clone, Debug)]
+pub struct ExtendedChannelFactory {
+    extended_channel_factory: Arc<RwLock<HashMap<u32, Arc<RwLock<ExtendedChannel>>>>>,
+    extended_extranonce_factory: Arc<RwLock<ExtendedExtranonceFactory>>,
+    chain_tip: Option<ChainTip>,
+    active_template: Option<NewTemplate<'static>>,
+    future_template: Option<NewTemplate<'static>>,
+    share_batch_size: usize,
+    expected_share_per_minute_per_channel: f64,
+    version_rolling_allowed: bool,
+    // todo
+    // payout_coinbase_outputs: Vec<TxOut>,
+}
+
+impl ExtendedChannelFactory {
+    pub fn new(
+        extended_extranonce_range_0: Range<usize>,
+        extended_extranonce_range_1: Range<usize>,
+        extended_extranonce_range_2: Range<usize>,
+        additional_coinbase_script_data: Option<Vec<u8>>,
+        share_batch_size: usize,
+        expected_share_per_minute_per_channel: f64,
+        version_rolling_allowed: bool,
+    ) -> Result<Self, ExtendedChannelFactoryError> {
+        let extended_extranonce = ExtendedExtranonceFactory::new(
+            extended_extranonce_range_0,
+            extended_extranonce_range_1,
+            extended_extranonce_range_2,
+            additional_coinbase_script_data.clone(),
+        )
+        .map_err(|e| ExtendedChannelFactoryError::ExtendedExtranonceError(e))?;
+
+        Ok(Self {
+            extended_channel_factory: Arc::new(RwLock::new(HashMap::new())),
+            extended_extranonce_factory: Arc::new(RwLock::new(extended_extranonce)),
+            chain_tip: None,
+            active_template: None,
+            future_template: None,
+            share_batch_size,
+            expected_share_per_minute_per_channel,
+            version_rolling_allowed,
+        })
+    }
+
+    pub fn remove_channel(&self, channel_id: u32) -> Result<(), ExtendedChannelFactoryError> {
+        self.extended_channel_factory
+            .write()
+            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?
+            .remove(&channel_id);
+        Ok(())
+    }
+
+    pub fn get_extended_channel(
+        &self,
+        channel_id: u32,
+    ) -> Result<ExtendedChannel, ExtendedChannelFactoryError> {
+        let extended_channel_factory_guard = self
+            .extended_channel_factory
+            .read()
+            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?;
+
+        let extended_channel = extended_channel_factory_guard
+            .get(&channel_id)
+            .ok_or(ExtendedChannelFactoryError::ExtendedChannelNotFound)?;
+        let extended_channel_guard = extended_channel
+            .read()
+            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?;
+        Ok(extended_channel_guard.clone())
+    }
+
+    pub fn has_extended_channel(
+        &self,
+        channel_id: u32,
+    ) -> Result<bool, ExtendedChannelFactoryError> {
+        let res = self
+            .extended_channel_factory
+            .read()
+            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?
+            .contains_key(&channel_id);
+        Ok(res)
+    }
+
+    pub fn set_active_template(&mut self, template: NewTemplate<'static>) {
+        self.active_template = Some(template);
+    }
+
+    pub fn set_future_template(&mut self, template: NewTemplate<'static>) {
+        self.future_template = Some(template);
+    }
+
+    pub fn set_chain_tip(&mut self, chain_tip: ChainTip) {
+        self.chain_tip = Some(chain_tip);
+    }
+
+    pub fn get_chain_tip(&self) -> Option<ChainTip> {
+        self.chain_tip.clone()
+    }
+
+    /// Handle an [`OpenExtendedMiningChannel`] message.
+    ///
+    /// Returns the `channel_id` of the newly created [`ExtendedChannel`].
+    pub fn on_open_extended_mining_channel(
+        &self,
+        m: OpenExtendedMiningChannel<'static>,
+        extended_channel_id: u32,
+    ) -> Result<(), ExtendedChannelFactoryError> {
+        let active_chain_tip = self
+            .chain_tip
+            .clone()
+            .ok_or(ExtendedChannelFactoryError::NoActiveChainTip)?;
+
+        // the spec says the following about user_identity string:
+        // > It is highly recommended that UTF-8 encoding is used.
+        // SRI enforces this here.
+        let user_identity_bytes = m.user_identity.inner_as_ref();
+        let user_identity_string =
+            String::from_utf8(user_identity_bytes.to_vec()).expect("Invalid UTF-8");
+        let user_identity: &'static str = Box::leak(user_identity_string.into_boxed_str());
+
+        let nominal_hashrate = m.nominal_hash_rate;
+        let extranonce_size = m.min_extranonce_size;
+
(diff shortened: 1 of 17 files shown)


protocols/v2/roles-logic-sv2/src/channel_management/group_channel.rs
```diff
@@ -0,0 +1,17 @@
+use crate::channel_management::standard_channel::StandardChannel;
+use crate::mining_sv2::NewExtendedMiningJob;
+use std::collections::HashMap;
+pub struct GroupChannel {
+    channel_id: u32,
+    channels: HashMap<u32, StandardChannel>,
```
plebhash, line 6: There's no need to keep references to the actual `StandardChannel` objects. That is managed at `StandardChannelFactory` level.

A `HashSet<u32>` is sufficient here, as we only need to keep collision-free references to the "child" channel ids within this group.

```suggestion
    channel_ids: HashSet<u32>,
```

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
```diff
@@ -0,0 +1,248 @@
...
+///
+/// Only suitable for mining servers, not clients.
+pub struct ExtendedChannelFactory {
+    channel_id_factory: Arc<RwLock<IdFactory>>,
+    extended_channel_factory: Arc<RwLock<HashMap<u32, Arc<RwLock<ExtendedChannel>>>>>,
+    extended_extranonce_factory: Arc<RwLock<ExtendedExtranonceFactory>>,
+    share_batch_size: usize,
+    expected_share_per_minute_per_channel: f64,
+    version_rolling_allowed: bool,
+}
```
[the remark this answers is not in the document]
plebhash: the alternative approach based on actor-model (as proposed in https://github.com/Shourya742/stratum/commit/004074b81da97de64ecaa0dda482befa9e53572c) removes the need for lock-based synchronization, but it also introduces some trade-offs that have to be taken into consideration:
- we're introducing the notion of "messages" with a new semantic meaning, which can be very confusing to the naive eye, which could misinterpret that them as Sv2 messages
- we're introducing a new burden of channel management to the caller, which can pollute the code on roles even more

while I believe this is a viable design, I'd like to see more compelling reasons to pivot away from the current direction (especially as this will require substantial extra effort to adapt the APIs and could potentially result in delays)

> For example, you might take a read lock on the whole struct to inspect it, and then realize you need to write to one of the nested items (like updating a value in the HashMap inside one of those locks). But Rust won’t let you do that...

hmm I don't think this is correct.

we do not need to hold a write lock on the top-level `HashMap` because we're not mutating the map’s structure, only the individual object behind the nested `RwLock`. 

just check the current implementation of [`ExtendedChannelFactory::on_update_channel`](https://github.com/plebhash/stratum/blob/626029b8756d6ec063a9d623464bf834cb0bb0c5/protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs#L155):
```rust

        // get read access to extended channels map
        let extended_channel_factory_guard = self
            .extended_channel_factory
            .read()
            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?;

        let extended_channel = extended_channel_factory_guard
            .get(&channel_id)
            .ok_or(ExtendedChannelFactoryError::ExtendedChannelNotFound)?;

        // get write access to extended channel
        let mut extended_channel_guard = extended_channel
            .write()
            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?;

        ...

        extended_channel_guard.set_nominal_hashrate(m.nominal_hash_rate);
        extended_channel_guard.set_target(new_target.clone());
```

we only need to hold a write lock on the container `HashMap` if we want to:
- insert a new element
- remove an element

for mutating any element, a read lock on the container is more than sufficient to get a pointer to the nested `RwLock`, and we only need a write lock for the element itself.

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
```diff
@@ -0,0 +1,248 @@
...
+///   - [`UpdateChannel`] (update channel nominal hashrate and target)
+///   - [`NewTemplate`] (update channel template)
+///   - [`SetNewPrevHash`] or [`SetNewPrevHashTDP`] (update channel chain tip)
+///   - [`SetCustomMiningJob`] (add new custom mining job)
+///   - [`SubmitSharesExtended`] (validate shares)
+///
+/// Only suitable for mining servers, not clients.
+pub struct ExtendedChannelFactory {
+    channel_id_factory: Arc<RwLock<IdFactory>>,
+    extended_channel_factory: Arc<RwLock<HashMap<u32, Arc<RwLock<ExtendedChannel>>>>>,
```
jbesraa, line 40: I would avoid the complexity here of having nested `RwLocks` and only have `RwLock` on the outer `HashMap`. In case there is an update, you can remove from `HashMap` and then insert new  (u32, ExtendedChannel) value. Another thing I would consider is using async `Mutex` as our code base and most of the functions under `ExtendedChannelFactory` are async (https://stackoverflow.com/questions/73840520/what-is-the-difference-between-stdsyncmutex-vs-tokiosyncmutex).

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
```diff
@@ -0,0 +1,248 @@
...
+            .map_err(|_| ExtendedChannelFactoryError::PoisonError)?
+            .remove(&channel_id);
+        Ok(())
+    }
+
+    /// Handle an [`OpenExtendedMiningChannel`] message.
+    ///
+    /// Returns the `channel_id` of the newly created [`ExtendedChannel`].
+    pub async fn on_open_extended_mining_channel(
+        &mut self,
```
jbesraa, line 91: One of the main motivations of using mutexes/rwlocks is also to avoid taking `mut self`, so I am very surprised this is needed. can you please elaborate why this is required?

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
```diff
@@ -0,0 +1,248 @@
...
+            channel_id_factory: Arc::new(RwLock::new(IdFactory::new())),
+            extended_channel_factory: Arc::new(RwLock::new(HashMap::new())),
+            extended_extranonce_factory: Arc::new(RwLock::new(extended_extranonce)),
+            share_batch_size,
+            expected_share_per_minute_per_channel,
+            version_rolling_allowed,
+        })
+    }
+
+    pub async fn remove_channel(
```
jbesraa, line 76:
```suggestion
    pub  fn remove_channel(
```

jbesraa: here and elsewhere in this struct the async redundant, we are not using any `await`

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
plebhash: it's not needed indeed, I'll remove it soon

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
plebhash: >  In case there is an update, you can remove from HashMap and then insert new (u32, ExtendedChannel) value.

that would force a lot of extra write locks around `HashMap`, which is exactly what I was trying to avoid

the whole point here is that changing one specific element doesn't block concurrent tasks from changing other elements

> Another thing I would consider is using async Mutex as our code base and most of the functions under ExtendedChannelFactory are async

this would require introducing `tokio` as a dependency of `roles_logic_sv2`, which goes against SRI policy on dependencies for low-level crates

jbesraa: > > In case there is an update, you can remove from HashMap and then insert new (u32, ExtendedChannel) value.
> 
> that would force a lot of extra write locks around `HashMap`, which is exactly what I was trying to avoid
> 
> the whole point here is that changing one specific element doesn't block concurrent tasks from changing other elements
> 

Take into account that the sync `Mutex` would block the whole thread, so I am not fully sure you are achieving what you want here. Maybe it is worth testing this. What do you think about removing the `RwLock` from the whole struct and follow the same pattern as `ExtendedChannelFactory` (what I mentioned in previous comment)? 
> > Another thing I would consider is using async Mutex as our code base and most of the functions under ExtendedChannelFactory are async
> 
> this would require introducing `tokio` as a dependency of `roles_logic_sv2`, which goes against SRI policy on dependencies for low-level crates

While I don't agree `roles-logic` is a low level crate, I also don't think this is a huge issue.

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
```diff
@@ -0,0 +1,248 @@
...
+    share_batch_size: usize,
+    expected_share_per_minute_per_channel: f64,
+    version_rolling_allowed: bool,
+}
+
+impl ExtendedChannelFactory {
+    pub fn new(
+        extended_extranonce_range_0: Range<usize>,
+        extended_extranonce_range_1: Range<usize>,
+        extended_extranonce_range_2: Range<usize>,
```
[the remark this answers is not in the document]
plebhash: indeed would be better encapsulation, but this is something to be improved on `mining_sv2`

I want to avoid leaking the scope of this PR into other crates unless it's absolutely necessary

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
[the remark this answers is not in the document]
plebhash: the main takeaway from my discussion with @Shourya742 was that although this nested `Arc<RwLock>>` approach might be a bit confusing at first (as he refers to anti-pattern), the main concerns around deadlocks described above aren't really true, namely:

> Holding a read lock means you're telling the system "no one is writing to this structure right now"—but then acquiring a write lock on a child item violates that guarantee. This pattern easily leads to deadlocks, especially if other parts of the code try to acquire a write lock on the map while the read lock is still held.It also sorta breaks the mental model of Rust’s safety guarantees: you're effectively mutating state while claiming shared access. It creates subtle concurrency bugs that are hard to trace and debug

we had to clarify a few misconceptions, namely:
A. the assumption that on top of this, the caller would also be doing concurrency management around the Factory (i.e. wrapping it with yet another lock/mutex), which indeed would lead to a lot of redundancy
B. the assumption that in order to mutate individual `HashMap` elements, we would need to take a write lock on the `HashMap`

the point raised by @jbesraa on [this discussion thread](https://github.com/stratum-mining/stratum/pull/1624#discussion_r2040665068) remain open, namely how the usage of `std::sync::RwLock` (as opposed to `tokio::sync::RwLock`) could affect execution on the caller side due to blocking the entire thread, but as we move forward there I feel it's important for all of us to have clarity on point B. listed above

with regards to the Actor-model approach, it remains a viable possibility worth exploring the trade-offs in case we find some fundamental flaws with the lock-based approach here

protocols/v2/roles-logic-sv2/src/job_management.rs
```diff
@@ -0,0 +1,317 @@
...
+/// - Multiple Coinbase Outputs
+/// - Job ID Factory
+/// - Rollable Extranonce Size
+/// - Version Rolling Allowed
+#[derive(Debug)]
+pub struct JobFactory {
+    future_template: Option<NewTemplate<'static>>,
+    active_template: Option<NewTemplate<'static>>,
+    additional_coinbase_outputs: Vec<TxOut>,
+    job_id_factory: Id,
```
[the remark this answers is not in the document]
plebhash: it could be an interesting optimization, but that's scope for https://github.com/stratum-mining/stratum/issues/1448

protocols/v2/roles-logic-sv2/src/channel_management/extended_channel_factory.rs
plebhash: > What do you think about removing the RwLock from the whole struct and follow the same pattern as ExtendedChannelFactory (what I mentioned in previous comment)?

I think that wouldn't really solve much. This approach would simply move the inner `RwLock`s one layer below. If there's indeed some fundamental issue with nesting `RwLock`s, then the issue would remain.

And worse: we would no longer be able to reason about Channel states atomically, as now different state variables could change in a concurrent way.

plebhash: > Take into account that the sync Mutex would block the whole thread, so I am not fully sure you are achieving what you want here. Maybe it is worth testing this.

indeed these locks will block the whole thread, and my mistake of previously marking Factory and Channel methods with `async` keyword indeed made things pretty confusing in that sense... I cleared those up (as well as the `&mut self` keywords) so hopefully we can have more clarity on this

on the other hand, I wouldn't extrapolate that into saying that we're not achieving what we want here.

multi-threaded concurrency remains perfectly doable, as well as usage within `async` runtimes (i.e. calling Factory and Channel methods from `async` contexts)

the only actual restriction with `async` here is single-threaded runtimes

and with regards to this, I see the following options moving forward:
A. simply add comments about this limitation on Rust Docs (single threaded runtimes not supported)
B. allow ourselves to add `tokio` as a dependency to `roles_logic_sv2` so that we can use `tokio::sync::RwLock` and lift this restriction for single-threaded runtimes
C. pivot away from a lock-based approach to an actor-model based approach
D. something else?

while evaluating these different options, I think some more fundamental questions comes to mind:

what are the pre-requisites for `roles_logic_sv2`? is `tokio` a first-class citizen at this level, or do aim to be runtime agnostic?

I feel that concurrency safety is definitely a pre-requisite for this crate, as we need server applications to scale to potentially thousands of client clients at the same time

but there's multiple ways to achieve this, and it's important that we are all aligned on the fundamental assumptions and pre-requisites so we can properly evaluate the trade-offs

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -468,6 +469,37 @@ pub fn hash_rate_from_target(target: U256<'static>, share_per_min: f64) -> Resul
     Ok(result as f64)
 }
 
+/// Converts a `Target` to a `f64` difficulty.
+pub fn target_to_difficulty(target: Target) -> f64 {
```
jbesraa, line 473: https://docs.rs/bitcoin/latest/bitcoin/struct.Target.html#method.difficulty_float

plebhash:
replaced by https://github.com/plebhash/stratum/tree/2025-04-17-channel-management-upgrade-actor-model

Issue #1574, opened 2025-03-17
investigate why Integration Tests are failing


plebhash:
a place where reports from https://github.com/stratum-mining/stratum/issues/1615 become actionable items

plebhash:
I instrumented `test_jdc_pool_fallback_after_submit_rejection` for `tokio-console` and executed the test repeatedly until it would hang:
```
#!/bin/bash

while true; do
    echo "Starting test run at $(date)"
  
    RUSTFLAGS="--cfg tokio_unstable" cargo t test_jdc_pool_fallback_after_submit_rejection -- --nocapture
    
    echo "----------------------------------------"
    sleep 1  # Brief pause before starting next iteration
done 
```

it doesn't hang every time, but after a few iterations, I got a hanging execution

and once again, we have zombie a `noise_connection.rs:72:25`
```
connection: http://127.0.0.1:6669/ (CONNECTED)
views: t = tasks, r = resources
controls: select column (sort) = ←→ or h, l, scroll = ↑↓ or k, j, view details = ↵, invert sort (highest/lowest) = i, scroll to top = gg, scroll to bottom = G, toggle pause = space, quit = q
╭Warnings────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│⚠ 33 tasks have been boxed by the runtime due to their size                                                                                                                                                     │
│⚠ 75 tasks are 1024 bytes or larger                                                                                                                                                                             │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭Tasks (117) ▶ Running (1) ⏸ Idle (114)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│   Warn  ID  State  Name  Total▿ Busy   Sched  Idle   Polls Kind   Location                                                                                         Fields                                      │
│>> ⚠ 1    48 ▶            44m37s 43m59s     1s    37s 14    task   /Users/plebhash/develop/stratum/roles/roles-utils/network-helpers/src/noise_connection.rs:72:25  size.bytes=1240 target=tokio::task          │
```

plebhash:
it's important to note that the investigation above is not reproducing [what happened on CI](https://github.com/stratum-mining/stratum/actions/runs/13902018935/job/38895881297?pr=1540#step:3:7392)

[...]

Issue #1616, opened 2025-04-01
Vardiff needs to be a module of `roles_logic_sv2` used by multiple roles


GitGab19:
We currently have a vardiff functionality baked into our tProxy in two places:
- [`downstream_sv1`](https://github.com/stratum-mining/stratum/blob/main/roles/translator/src/lib/downstream_sv1/diff_management.rs)
- [`upstream_sv2`](https://github.com/stratum-mining/stratum/blob/main/roles/translator/src/lib/upstream_sv2/diff_management.rs)

As we already discussed, we should have a specific module for `vardiff` (probably inside `roles_logic_sv2`), which can be used by multiple roles (e.g. Pool and JDC).

This would also allow us to decouple the translation logic from the aggregation logic inside tProxy, which will make easier the development of a SV2 firmware in the future.

plebhash:
cross referencing https://github.com/stratum-mining/stratum/pull/1625/commits/8a17ebe202a755896e98f3e2f837f756d7909f39#r2031336352

plebhash:
cross referencing https://github.com/stratum-mining/stratum/pull/1572#issuecomment-2730313637

Issue #1477, opened 2025-02-12
Document JDC async primitives


GitGab19:

plebhash:
cross referencing https://github.com/stratum-mining/stratum/pull/1625/commits/8a17ebe202a755896e98f3e2f837f756d7909f39#r2031336352

PR #1625, opened 2025-04-07
Reintroduce TProxy-JD integration test

jbesraa:
Partially addresses #1574 

This was tested on MacOS with M1 and Linux with AMD using the following script without encountering any failures.

```sh
#!/bin/bash

command_to_run="cargo test --test 'jd_tproxy_integration' -- --nocapture"

num_runs=20
success_count=0
fail_count=0

for i in $(seq 1 "$num_runs"); do
  echo "Running command $i of $num_runs..."
  if eval "$command_to_run"; then
    echo "Run $i: Success"
    ((success_count++))
  else
    echo "Run $i: Fail"
    ((fail_count++))
  fi
done

echo "-------------------------"
echo "Total Runs: $num_runs"
echo "Successful Runs: $success_count"
echo "Failed Runs: $fail_count"
echo "-------------------------"
```

### DIFF
--- a/roles/test-utils/mining-device-sv1/src/client.rs
+++ b/roles/test-utils/mining-device-sv1/src/client.rs
@@ -208,6 +208,8 @@ impl Client {
                     warn!("Share channel is not available");
                     break;
                 }
+                // Introduce a delay of 0.2 seconds after sending a share
+                std::thread::sleep(Duration::from_millis(200));
             }
             miner_cloned
                 .safe_lock(|m| m.header.as_mut().map(|h| h.nonce += 1))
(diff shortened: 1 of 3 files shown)


roles/test-utils/mining-device-sv1/src/client.rs
```diff
@@ -208,6 +208,8 @@ impl Client {
                     warn!("Share channel is not available");
                     break;
                 }
+                // Introduce a delay of 0.2 seconds after sending a share
+                std::thread::sleep(Duration::from_millis(200));
```
plebhash, line 212: this is interesting insight!

I think this is perfectly fine as a short-term solution, but it also tells us something deeper about JDC:

the fact that JDC is "vulnerable" to share DoS is proof that JDC lacks proper `vardiff` functionality

this is something we should take note on #1616 and also perhaps #1477

plebhash:
closes #1626

plebhash (approved):
tACK on Mac M2

roles/test-utils/mining-device-sv1/src/client.rs
plebhash: hmm upon reflection not sure how much what I said above makes sense

maybe we're just working around the root cause of the problem

let's see, we can open #1626 again if needed

PR #1639, opened 2025-04-17
Renaming of `translator config` and `mining proxy config`

Sansh2356:
Addresses #1260 .

### DIFF
--- a/roles/translator/src/lib/proxy_config.rs
+++ b/roles/translator/src/lib/proxy_config.rs
@@ -2,7 +2,7 @@ use key_utils::Secp256k1PublicKey;
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize, Clone)]
-pub struct ProxyConfig {
+pub struct TranslatorProxyConfig {
     pub upstream_address: String,
     pub upstream_port: u16,
     pub upstream_authority_pubkey: Secp256k1PublicKey,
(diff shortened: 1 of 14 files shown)


test/integration-tests/lib/mod.rs
```diff
@@ -250,7 +250,7 @@ pub fn start_sv2_translator(upstream: SocketAddr) -> (TranslatorSv2, SocketAddr)
 
     let min_extranonce2_size = 4;
 
-    let config = translator_sv2::proxy_config::ProxyConfig::new(
+    let config = translator_sv2::proxy_config::TranslatorProxyConfig::new(
```
jbesraa, line 253: Can you please change the filename `translator/src/lib/proxy_config.rs` to `translator/src/lib/config.rs`?

roles/translator/src/lib/proxy_config.rs
```diff
@@ -2,7 +2,7 @@ use key_utils::Secp256k1PublicKey;
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize, Clone)]
-pub struct ProxyConfig {
+pub struct TranslatorProxyConfig {
```
jbesraa, line 5: The crate is called `translator_sv2` and the main lib struct is called `TranslatorSV2` so I would go with either `TranslatorConfig` or `TranslatorSV2Config`.

jbesraa (commented):
Thank you for looking into this! 
Just added two small nits.

roles/translator/src/lib/proxy_config.rs
Sansh2356: Kindly check i have done the needful @jbesraa .

plebhash (approved):
commit history looks neat, thanks for addressing the review points

PR #1642, opened 2025-04-21
Move constants to `stratum-common`

jbesraa:
Partially addresses #1135 

IMO trying to fix the sv2.h file and interop test is a waste of time. It is using `cbindgen` v0.21(latest is 0.28) which does not seem to support specific features when building a crate(i.e., cant build `startum-common` only with `constants` feature). I would suggest to remove the CI checks for these and in the future rewrite them properly and add proper CI checks.


plebhash:
> I would suggest to remove the CI checks for these and in the future rewrite them properly and add proper CI checks.

should we add a commit here doing this?

jbesraa:
> > I would suggest to remove the CI checks for these and in the future rewrite them properly and add proper CI checks.
> 
> 
> 
> should we add a commit here doing this?

Happy to do that if we are all on the same page

plebhash (approved):
LGTM but it seems there's still some issues with CI?

jbesraa:
@GitGab19 Can you please remove sv2_header_check from gh settings?

GitGab19:
Just done.

Could you leave a comment under the tracker issue we have for `sv2-ffi` crate regarding this please?

PR #1651, opened 2025-04-28
Remove `SetCustomMiningJob::extranonce_size`

### LINKED ISSUE #1453: upgrade `SetCustomMiningJob` by removing useless `extranonce_size` field
https://github.com/stratum-mining/sv2-spec/pull/121/commits/9a3128e1ae3c835e1af290447f4d70738c2523c5

~~https://github.com/stratum-mining/sv2-spec/pull/121/commits/0fd5cf5d50ed0aa10339c3a676496acf536af94c~~ (this duplicated field is already absent from `SetCustomMiningJob`

jbesraa:
resolves #1453

### DIFF
--- a/protocols/v2/roles-logic-sv2/Cargo.toml
+++ b/protocols/v2/roles-logic-sv2/Cargo.toml
@@ -16,7 +16,7 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 stratum-common = { path = "../../../common", features=["bitcoin", "constants"], version = "^2.0.0"}
 binary_sv2 = { path = "../../../protocols/v2/binary-sv2", version = "^2.0.0" }
 common_messages_sv2 = { path = "../../../protocols/v2/subprotocols/common-messages", version = "^4.0.0" }
-mining_sv2 = { path = "../../../protocols/v2/subprotocols/mining", version = "^3.0.0" }
+mining_sv2 = { path = "../../../protocols/v2/subprotocols/mining", version = "^4.0.0" }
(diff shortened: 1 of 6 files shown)


plebhash (approved):
LGTM but it seems we need to bump MAJOR on `mining_sv2`

jbesraa:
> Changes looks good to me. @jbesraa The reason we're removing the extranonce size from SetCustomMiningJob is that it doesn't make sense for the pool to know about it, since it won't be distributing the job — the JDC will. Am I understanding this correctly?

I have to look closer into the code but it might be tProxy and not JDC, but yes your point is correct.

plebhash:
> The reason we're removing the extranonce size from SetCustomMiningJob is that it doesn't make sense for the pool to know about it, since it won't be distributing the job — the JDC will. Am I understanding this correctly?

`SetCustomMiningJob` is sent over an Extended Channel that's open between JDC and Pool

`extranonce_size` is essentially a property of this channel, and JDC is informed by Pool of this value via `OpenExtendedMiningChannel.Success`

so JDC informing Pool about this again on `SetCustomMiningJob` is redundant

how this custom job is going to be distributed by JDC to machines downstream (and whether there's a tProxy in between or not) is somewhat orthogonal to all this

in theory (even though SRI JDC doesn't do this right now), JDC could be even distributing Standard Jobs downstream, and simply converting the `SubmitSharesStandard` to `SubmitSharesExtended` before submitting them to Pool

jbesraa:
Updated `mining_sv2` version.

PR #1669, opened 2025-05-10
Separate out IsServer and IsClient

pool2win:
This will help us write tests for each easily and also just reduce similar function names in the same module file, even if they belong to different traits.


plebhash:
didn't really dive into a deeper review yet, but at a first glance it seems this PR is mostly about code organization, right?

a few things worth noting:
- the repo has a few application crates that depend on `sv1_api` ([tProxy](https://github.com/stratum-mining/stratum/tree/main/roles/translator), [Sv1 CPU miner](https://github.com/stratum-mining/stratum/tree/main/roles/test-utils/mining-device-sv1), [Integration Tests Sv1 Sniffer](https://github.com/stratum-mining/stratum/blob/main/test/integration-tests/lib/sv1_sniffer.rs)). any breaking changes introduced to `sv1_api` crate implies that the same PR also needs to adapt all those crates that use it as a dependency, otherwise CI won't allow us to merge it
- this is a breaking change that will also force us to bump MAJOR of the crate

pool2win:
Oh man. Let's close this then. I was hoping we'd be exporting things from
lib.rs. Therefore we could reorg internally without breaking clients.

pool2win:
Closing, as this breaks to many dependencies.

Issue #736, opened 2024-01-26
refactor: improve Error handling (clean up `unwrap`s)


plebhash:
as @jakubtrnka pointed out in https://github.com/stratum-mining/stratum/issues/723#issuecomment-1904233521, we should improve error handling across SRI codebase

while I do believe and understand that there are responsible ways to use `unwrap`, as the codebase grows and becomes more complex for newcomers to understand, these `unwrap`s can quickly become footguns with long-term consequences on the maintenability and overall code quality of the project

Fi3:
I think that this is too generic can you be more specific

Fi3:
in particular we have to distinguish between low-level library and application. Now considering that we are going to have mid level library we have 3 "zone"
* low level libraries
* mid level library
* application (mostly used for testing and as reference implementation)

Fi3:
For the libraries every unwrap should be documented if it is not we have to open issues for it. For the application we have `handle_result` that is underused cause is a little bit complex to use. Not sure if we want to improve it (considering that will be used only by application zone), improve it and put it in the mid level library zone, or just leave at it is now.

Fi3:
whatever we decide we should add a code reviews police doc or something like that. For example we can say that every result in the a application zone that is unhandled must be put in a  `handle_result` macro and every exception have to be documented. Or we can decide that we do not care an `unwrap` in application zone are acceptable cause they are no used in production ecc ecc

plebhash:
> For the libraries every unwrap should be documented if it is not we have to open issues for it.

this is a good start! `expect` might also be helpful

> For the application we have `handle_result` that is underused cause is a little bit complex to use.

in my past job, I worked with [`polkadot-sdk`](https://github.com/paritytech/polkadot-sdk) (fka `substrate`), a Rust framework for blockchains which makes heavy usage of macros.

I worked for around ~2y on the go-to-market vertical of @paritytech, and our overall consensus was that our heavy usage of macros on `polkadot-sdk/substrate` was slowing down developer adoption because macros are always harder to understand.

while returning `std::result::Result` does make the code bigger and uglier, it is way easier to understand (and more importantly, to contribute to)

> I think that this is too generic can you be more specific

apologies for that, I will take some time to elaborate more concrete examples across the codebase along with suggestions on what to do about them

I will write more comments on this issue in the future

plebhash:
just found this Discussion started by @xraid , which I think is going in the same direction of this issue so I'm linking here

[...]

PR #1712, opened 2025-05-15
remove redundant map_err's in translators

Shourya742:
Partially Resolves: #736
This PR is part of an effort to improve error handling in the translator. Currently, there are several redundant .map_err calls in the codebase, which make the code harder to read. This PR removes many of them. However, not all instances could be removed due to the use of non-Send types like MutexGuard errors, which cannot be used in Send contexts (e.g., in tokio::spawn). Addressing these cases will require more targeted changes, which we can defer until we work on those parts of the codebase.

### DIFF
--- a/roles/translator/src/lib/upstream_sv2/upstream.rs
+++ b/roles/translator/src/lib/upstream_sv2/upstream.rs
@@ -228,9 +228,7 @@ impl Upstream {
     ) -> ProxyResult<'static, ()> {
         // Get the `SetupConnection` message with Mining Device information (currently hard coded)
         let setup_connection = Self::get_setup_connection_message(min_version, max_version, false)?;
-        let mut connection = self_
-            .safe_lock(|s| s.connection.clone())
-            .map_err(|_e| PoisonLock)?;
+        let mut connection = self_.safe_lock(|s| s.connection.clone())?;
 
         // Put the `SetupConnection` message in a `StdFrame` to be sent over the wire
         let sv2_frame: StdFrame = Message::Common(setup_connection.into()).try_into()?;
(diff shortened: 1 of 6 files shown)


roles/translator/src/lib/upstream_sv2/upstream.rs
```diff
@@ -766,9 +740,7 @@ impl ParseMiningMessagesFromUpstream<Downstream> for Upstream {
                 m.extranonce_size,
             ));
         }
-        self.target
-            .safe_lock(|t| *t = m.target.to_vec())
-            .map_err(|e| RolesLogicError::PoisonLock(e.to_string()))?;
```
GitGab19, line 771: Why were we mapping to the `PoisonLock` from `RolesLogicError`, differently from the other places?

roles/translator/src/lib/upstream_sv2/upstream.rs
```diff
@@ -934,9 +906,7 @@ impl ParseMiningMessagesFromUpstream<Downstream> for Upstream {
         info!("Received SetTarget for channel id: {}", m.channel_id);
         debug!("SetTarget: {:?}", m);
         let m = m.into_static();
-        self.target
-            .safe_lock(|t| *t = m.maximum_target.to_vec())
-            .map_err(|e| RolesLogicError::PoisonLock(e.to_string()))?;
```
GitGab19, line 939: Why were we mapping to the `PoisonLock` from `RolesLogicError`, differently from the other places?

GitGab19 (commented):
Your changes LGTM.

I'm not completely sure about the real benefit of those btw.

Is it only for cleaning up the code or for other particular reasons?

Also, why we mapped all these errors in the first place? Isn't it because we are using the mapped error variant in some other places?

GitGab19 (approved):

PR #1714, opened 2025-05-19
separate vardiff module

Shourya742:
closes: #1616

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
+++ b/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
@@ -12,15 +12,15 @@ use super::{error::VardiffError, Vardiff};
 /// Tracks performance and adjusts the mining target to achieve a desired share rate.
 #[derive(Debug)]
 pub struct VardiffState {
-    /// Current estimated hashrate (H/s) for the connection.
-    pub estimated_channel_hashrate: f32,
+    /// Current estimated hashrate (H/s).
+    pub estimated_hashrate: f32,
     /// Target number of shares to be submitted per minute.
     pub shares_per_minute: f32,
     /// Count of shares received since the last difficulty adjustment.
     pub shares_since_last_update: u32,
     /// Unix timestamp (seconds) of the last difficulty adjustment.
     pub timestamp_of_last_update: u64,
-    /// The current mining target (difficulty) sent to the miner.
+    /// The current mining target.
     pub current_target: Target,
     /// The lowest hashrate (H/s) the system will allow; values below this are clamped.
     pub min_allowed_hashrate: f32,
@@ -31,31 +31,24 @@ impl VardiffState {
     ///
     /// # Arguments
     /// * `shares_per_minute` - The target share submission rate.
-    /// * `estimated_channel_hashrate` - The initial hashrate estimate.
-    pub fn new(
-        shares_per_minute: f32,
-        estimated_channel_hashrate: f32,
-    ) -> Result<Self, VardiffError> {
-        Self::new_with_min(
-            shares_per_minute,
-            estimated_channel_hashrate,
-            DEFAULT_MIN_HASHRATE,
-        )
+    /// * `estimated_hashrate` - The initial hashrate estimate.
+    pub fn new(shares_per_minute: f32, estimated_hashrate: f32) -> Result<Self, VardiffError> {
+        Self::new_with_min(shares_per_minute, estimated_hashrate, DEFAULT_MIN_HASHRATE)
     }
 
     /// Creates a new `VardiffState` with a specific minimum hashrate.
     ///
     /// # Arguments
     /// * `shares_per_minute` - The target share submission rate.
-    /// * `estimated_channel_hashrate` - The initial hashrate estimate.
+    /// * `estimated_hashrate` - The initial hashrate estimate.
     /// * `min_allowed_hashrate` - The minimum hashrate to enforce.
     pub fn new_with_min(
         shares_per_minute: f32,
-        estimated_channel_hashrate: f32,
+        estimated_hashrate: f32,
         min_allowed_hashrate: f32,
     ) -> Result<Self, VardiffError> {
         let current_target =
-            hash_rate_to_target(estimated_channel_hashrate as f64, shares_per_minute as f64)
+            hash_rate_to_target(estimated_hashrate as f64, shares_per_minute as f64)
                 .map_err(|e| VardiffError::HashrateToTargetError(e.to_string()))?
                 .into();
         let timestamp_secs = std::time::SystemTime::now()
@@ -63,7 +56,7 @@ impl VardiffState {
             .as_secs();
 
         Ok(VardiffState {
-            estimated_channel_hashrate,
+            estimated_hashrate,
             shares_per_minute,
             shares_since_last_update: 0,
             timestamp_of_last_update: timestamp_secs,
@@ -95,7 +88,7 @@ impl VardiffState {
 
 impl Vardiff for VardiffState {
     fn hashrate(&self) -> f32 {
-        self.estimated_channel_hashrate
+        self.estimated_hashrate
     }
 
     fn shares_per_minute(&self) -> f32 {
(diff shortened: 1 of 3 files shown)


protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.current_miner_target.clone()
+    }
+
+    fn set_hashrate(&mut self, estimated_channel_hashrate: f32) {
+        self.estimated_channel_hashrate = estimated_channel_hashrate;
+        let current_miner_target = hash_rate_to_target(
+            estimated_channel_hashrate as f64,
+            self.shares_per_minute as f64,
+        )
+        .expect("Cannot convert hash rate to targe")
```
lucasbalieiro, line 79: nit: just a typo here
```suggestion
        .expect("Cannot convert hash rate to target")
```

protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub shares_since_last_update: u32,
+    pub timestamp_of_last_update: u64,
+    pub current_miner_target: Target,
+}
+
+impl VardiffState {
+    pub fn new(shares_per_minute: f32, estimated_channel_hashrate: f32) -> Self {
+        let current_miner_target =
+            hash_rate_to_target(estimated_channel_hashrate as f64, shares_per_minute as f64)
+                .expect("Cannot convert hash rate to target")
```
lucasbalieiro, line 20: quick (maybe dumb) question: should we really` expect()` here on the result of `hash_rate_to_target`? 

Would it make more sense to handle the error and return a fallback target (like the max possible), rather than panicking? Just want to make sure we're okay with this failing at runtime.

plebhash: in the end of the day, it boils down to the edge cases that could really trigger an error on this API call:

https://github.com/stratum-mining/stratum/blob/bbafc8b1ebedfc97f2836294aef0946febc0c8c3/protocols/v2/roles-logic-sv2/src/utils.rs#L324-L337

no sane application would ever try to:
- define some expected `shares_per_minute = 0`
- define some expected `shares_per_minute < 0`
- define or estimate some negative hashrate

all of these potential edge cases are arguably gross API misusage, so I feel that a panic here is somewhat acceptable

roles/translator/src/lib/downstream_sv1/diff_management.rs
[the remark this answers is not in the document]
plebhash: just started looking deeper into this today

still trying to refine my mental model around how this will work wrt aggregated vs non-aggregated modes

protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -0,0 +1,164 @@
...
+
+#[allow(warnings)]
+#[derive(Debug)]
+pub struct VardiffState {
+    pub estimated_channel_hashrate: f32,
+    pub shares_per_minute: f32,
+    pub shares_since_last_update: u32,
+    pub timestamp_of_last_update: u64,
+    pub current_miner_target: Target,
+}
```
GitGab19, line 14: I think we should add a lower limit for difficulty/hashrate here, which can be used later in the `update_hashrate` function to perform some checks.

GitGab19: In this way we can avoid to set a difficulty/target which is too low. This can be configured of course, and we can use a default in case it's not.

Probably the best thing from a UX point of view here is to add the lowest `hashrate` that the implementer wants to support. 

Does it make sense?

GitGab19: This is still to be done, right? Just asking because I saw you pushed something new..

protocols/v2/roles-logic-sv2/src/vardiff/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Increments the shares since the last update.
+    fn increment_shares_since_last_update(&mut self);
+
+    /// Resets the vardiff state for a new calculation cycle.
+    fn reset_counter(&mut self) -> Result<(), VardiffError>;
+
+    /// Updates the hash rate based on recent activity and returns the new hash rate and delta.
+    /// Returns `None` if an update is not yet due.
+    fn update_hashrate(&mut self) -> Result<(), VardiffError>;
```
GitGab19, line 35: Shouldn't this be called:

```suggestion
    fn try_vardiff(&mut self) -> Result<(), VardiffError>;
```

In this case I would also improve the docs on top of the function to better reflect what it does.

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -226,98 +192,21 @@ impl Downstream {
...
+        self_.safe_lock(|d| {
+            let previous_hashrate = d.difficulty_mgmt.hashrate();
+            _ = d.difficulty_mgmt.update_hashrate();
+            let new_hashrate = d.difficulty_mgmt.hashrate();
+            let hashrate_delta = new_hashrate - previous_hashrate;
+            d.upstream_difficulty_config.super_safe_lock(|c| {
+                if c.channel_nominal_hashrate + hashrate_delta > 0.0 {
+                    c.channel_nominal_hashrate += hashrate_delta;
                 } else {
-                    Ok(None)
```
GitGab19, line 318: Why don't we need to return `Ok(None) anymore here?

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -521,7 +411,7 @@ mod test {
         }
         let expected_0s = trailing_0s(expected_target.inner_as_ref().to_vec());
         let actual_0s = trailing_0s(initial_target.inner_as_ref().to_vec());
-        assert!(expected_0s.abs_diff(actual_0s) <= 1);
+        assert!(expected_0s.abs_diff(actual_0s) <= 2);
```
GitGab19, line 414: Why this change?

roles/translator/src/lib/downstream_sv1/downstream.rs
```diff
@@ -105,8 +107,11 @@ impl Downstream {
         extranonce2_len: usize,
         difficulty_mgmt: DownstreamDifficultyConfig,
         upstream_difficulty_config: Arc<Mutex<UpstreamDifficultyConfig>>,
-        last_job_id: String,
```
GitGab19, line 108: Why did you remove this?

protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -0,0 +1,256 @@
...
+
+use super::{error::VardiffError, Vardiff};
+
+/// Represents the dynamic state for a variable difficulty (Vardiff) connection.
+///
+/// Tracks performance and adjusts the mining target to achieve a desired share rate.
+#[derive(Debug)]
+pub struct VardiffState {
+    /// Current estimated hashrate (H/s) for the connection.
+    pub estimated_channel_hashrate: f32,
```
GitGab19, line 16:
```suggestion
    pub estimated_hashrate: f32,
```

protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -0,0 +1,256 @@
...
+const DEFAULT_MIN_HASHRATE: f32 = 1.0;
+
+use super::{error::VardiffError, Vardiff};
+
+/// Represents the dynamic state for a variable difficulty (Vardiff) connection.
+///
+/// Tracks performance and adjusts the mining target to achieve a desired share rate.
+#[derive(Debug)]
+pub struct VardiffState {
+    /// Current estimated hashrate (H/s) for the connection.
```
GitGab19, line 15:
```suggestion
    /// Current estimated hashrate (H/s).
```

protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub estimated_channel_hashrate: f32,
+    /// Target number of shares to be submitted per minute.
+    pub shares_per_minute: f32,
+    /// Count of shares received since the last difficulty adjustment.
+    pub shares_since_last_update: u32,
+    /// Unix timestamp (seconds) of the last difficulty adjustment.
+    pub timestamp_of_last_update: u64,
+    /// The current mining target (difficulty) sent to the miner.
```
GitGab19, line 23:
```suggestion
    /// The current mining target.
```

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -469,11 +360,9 @@ mod test {
             0,
             downstream_conf.clone(),
             Arc::new(Mutex::new(upstream_config)),
-            "0".to_string(),
         );
-        downstream.difficulty_mgmt.min_individual_miner_hashrate = start_hashrate as f32;
```
GitGab19, line 474: Why did you remove this line?

roles/translator/src/lib/downstream_sv1/diff_management.rs
```diff
@@ -469,11 +360,9 @@ mod test {
...
 
-        let total_run_time = std::time::Duration::from_secs(10);
+        let total_run_time = std::time::Duration::from_secs(180);
```
GitGab19, line 365: I don't think this change here is related to any problem you were facing..

GitGab19 (commented):
Left a few nits

PR #1715, opened 2025-05-20
Add `MockUpstream` module

### LINKED ISSUE #1670: Add `MockDownstream`  and `MockUpstream` test module
Main motivation: 
- actively send messages to a specific role and examine its behavior
- can be used in unit tests for testing networking behavior like shutdown

jbesraa:
fully resolves #1670


protocols/v1/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::result_unit_err)]
+#![allow(clippy::result_large_err)]
```
GitGab19, line 2: why do you need this?

test/integration-tests/lib/mock_roles.rs
```diff
@@ -0,0 +1,187 @@
...
+                messages_from_dowsntream.add_message(msg_type, msg);
+                let response = response_messages
+                    .iter()
+                    .find(|(m_type, _)| m_type == &msg_type);
+                if let Some((_, response_msg)) = response {
+                    let message = StandardEitherFrame::<AnyMessage<'_>>::Sv2(
+                        Sv2Frame::from_message(response_msg.clone(), msg_type, 0, false)
+                            .expect("Failed to create the frame"),
+                    );
+                    downstream_sender.send(message).await.unwrap();
```
GitGab19, line 92: Why is this different from the corresponding method of `MockDownstream`?

jbesraa:
Thanks for review
Rebased and addressed comments

test/integration-tests/lib/mock_roles.rs
jbesraa: Well, I am not sure if there is no case. But I took the same approach taken with the `Sniffer` i.e., implement basic features and then gradually add more features if needed.

GitGab19 (approved):

PR #1718, opened 2025-05-21
migrate Pool to new `channels` module of `roles_logic_sv2`

### LINKED ISSUE #1631: need to migrate Pool from `channel_logic` to `channel_management`

plebhash:
close #1631 

to be merged after ~#1672~  #1724

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/channels/server/jobs/standard.rs
+++ b/protocols/v2/roles-logic-sv2/src/channels/server/jobs/standard.rs
@@ -1,7 +1,8 @@
-use crate::channels::server::jobs::JobOrigin;
+use crate::{channels::server::jobs::JobOrigin, utils::deserialize_outputs};
 use binary_sv2::{Sv2Option, U256};
 use mining_sv2::NewMiningJob;
 use stratum_common::bitcoin::transaction::TxOut;
+use template_distribution_sv2::NewTemplate;
 
 /// Abstraction of a standard mining job with:
 /// - the `NewTemplate` message that originated it
@@ -25,6 +26,25 @@ impl<'a> StandardJob<'a> {
     ) -> Self {
         Self {
             origin,
+            extranonce_prefix,
+            coinbase_outputs,
+            job_message,
+        }
+    }
+
+    pub fn from_template(
+        template: NewTemplate<'a>,
+        extranonce_prefix: Vec<u8>,
+        additional_coinbase_outputs: Vec<TxOut>,
+        job_message: NewMiningJob<'a>,
+    ) -> Self {
+        let mut coinbase_outputs = vec![];
+        coinbase_outputs.extend(additional_coinbase_outputs);
+        coinbase_outputs.extend(deserialize_outputs(
+            template.coinbase_tx_outputs.inner_as_ref().to_vec(),
+        ));
+        Self {
+            origin: JobOrigin::NewTemplate(template),
             extranonce_prefix,
             coinbase_outputs,
             job_message,
--- a/protocols/v2/roles-logic-sv2/src/channels/server/jobs/standard.rs
+++ b/protocols/v2/roles-logic-sv2/src/channels/server/jobs/standard.rs
@@ -1,4 +1,4 @@
-use crate::{channels::server::jobs::JobOrigin, utils::deserialize_outputs};
+use crate::utils::deserialize_outputs;
 use binary_sv2::{Sv2Option, U256};
 use mining_sv2::NewMiningJob;
 use stratum_common::bitcoin::transaction::TxOut;
@@ -11,27 +11,13 @@ use template_distribution_sv2::NewTemplate;
 /// - the `NewMiningJob` message to be sent across the wire
 #[derive(Debug, Clone)]
 pub struct StandardJob<'a> {
-    origin: JobOrigin<'a>,
+    template: NewTemplate<'a>,
     extranonce_prefix: Vec<u8>,
     coinbase_outputs: Vec<TxOut>,
     job_message: NewMiningJob<'a>,
 }
 
 impl<'a> StandardJob<'a> {
-    pub fn new(
-        origin: JobOrigin<'a>,
-        extranonce_prefix: Vec<u8>,
-        coinbase_outputs: Vec<TxOut>,
-        job_message: NewMiningJob<'a>,
-    ) -> Self {
-        Self {
-            origin,
-            extranonce_prefix,
-            coinbase_outputs,
-            job_message,
-        }
-    }
-
     pub fn from_template(
         template: NewTemplate<'a>,
         extranonce_prefix: Vec<u8>,
@@ -44,7 +30,7 @@ impl<'a> StandardJob<'a> {
             template.coinbase_tx_outputs.inner_as_ref().to_vec(),
         ));
         Self {
-            origin: JobOrigin::NewTemplate(template),
+            template,
             extranonce_prefix,
             coinbase_outputs,
             job_message,
@@ -67,8 +53,8 @@ impl<'a> StandardJob<'a> {
         &self.job_message
     }
 
-    pub fn get_origin(&self) -> &JobOrigin<'a> {
-        &self.origin
+    pub fn get_template(&self) -> &NewTemplate<'a> {
+        &self.template
     }
 
     pub fn get_merkle_root(&self) -> &U256<'a> {
--- a/roles/pool/src/lib/mining_pool/message_handler.rs
+++ b/roles/pool/src/lib/mining_pool/message_handler.rs
@@ -6,7 +6,7 @@
 //! reacts to various mining-related messages received from a connected downstream miner.
 
 use super::super::mining_pool::Downstream;
-use binary_sv2::{Str0255, Sv2Option};
+use binary_sv2::Str0255;
 use roles_logic_sv2::{
     channels::server::{
         error::{ExtendedChannelError, StandardChannelError},
@@ -129,6 +129,15 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
 
         let mut messages = vec![];
 
+        let group_channel_id = if let Some(group_channel_guard) = &self.group_channel {
+            let group_channel = group_channel_guard
+                .read()
+                .map_err(|e| Error::PoisonLock(e.to_string()))?;
+            group_channel.get_group_channel_id()
+        } else {
+            0
+        };
+
         let open_standard_mining_channel_success = OpenStandardMiningChannelSuccess {
             request_id: incoming.request_id,
             channel_id,
@@ -138,7 +147,7 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
                 .clone()
                 .try_into()
                 .expect("extranonce_prefix must be valid"),
-            group_channel_id: 1, // naive approach, we only support one group channel for now
+            group_channel_id,
         }
         .into_static();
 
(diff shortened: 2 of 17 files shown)


plebhash:
this is also ready for review

just fixed all broken integration tests

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -148,21 +168,86 @@ impl Downstream {
...
+
+        let mut has_cached_template = false;
+        while !has_cached_template {
+            pool.safe_lock(|p| {
+                if p.last_future_template.is_some() {
+                    has_cached_template = true;
+                }
+            })?;
+            tokio::task::yield_now().await;
+        }
```
plebhash, line 209: I had to introduce this because some integration tests were breaking

a client would connect before Pool had received any template, leading to undefined behavior (since we cannot bootstrap a channel without a cached template and prev_hash)

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -148,21 +168,86 @@ impl Downstream {
...
+
+        let mut has_cached_prev_hash = false;
+        while !has_cached_prev_hash {
+            pool.safe_lock(|p| {
+                if p.last_new_prev_hash.is_some() {
+                    has_cached_prev_hash = true;
+                }
+            })?;
+            tokio::task::yield_now().await;
+        }
```
plebhash, line 225: same here

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -125,13 +144,14 @@ pub struct Pool {
...
-    // Shared reference to the factory responsible for creating/managing mining channels
-    // and generating job-related messages.
-    channel_factory: Arc<Mutex<PoolChannelFactory>>,
-    // Stores the template ID associated with the most recently received `SetNewPrevHash` message.
-    last_prev_hash_template_id: u64,
+    downstream_id_factory: IdFactory,
     // Sender channel for reporting status updates and errors to the main monitoring loop.
     status_tx: status::Sender,
+    extranonce_prefix_factory_extended: Arc<Mutex<ExtendedExtranonce>>,
+    extranonce_prefix_factory_standard: Arc<Mutex<ExtendedExtranonce>>,
```
plebhash, line 151: these are enabling unique `extranonce_prefix` allocation across the entire pool

a clone of each is passed down to each `Downstream`

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -148,21 +168,86 @@ impl Downstream {
...
+
+        let share_batch_size = pool.safe_lock(|p| p.share_batch_size)?;
+
+        let mut has_cached_template = false;
+        while !has_cached_template {
+            pool.safe_lock(|p| {
+                if p.last_future_template.is_some() {
+                    has_cached_template = true;
+                }
+            })?;
```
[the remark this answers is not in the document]
plebhash: yeah good catch

added a comment and reduced to one single check against last_prev_hash

plebhash:
I introduced some changes to server-side APIs on #1724, so now this PR needs to be merged after it

protocols/v2/roles-logic-sv2/src/channels/server/jobs/standard.rs
```diff
@@ -62,4 +62,9 @@ impl<'a> StandardJob<'a> {
     pub fn activate(&mut self, min_ntime: u32) {
         self.job_message.min_ntime = Sv2Option::new(Some(min_ntime));
     }
+
+    /// Reverts a future job into a non-future job.
+    pub fn deactivate(&mut self) {
+        self.job_message.min_ntime = Sv2Option::new(None);
+    }
```
GitGab19: The comment says "Reverts a future job into a non-future job.", but the function does the opposite.

roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -54,41 +64,140 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
...
+        .into_static();
+
+        messages.push(Mining::OpenStandardMiningChannelSuccess(
+            open_standard_mining_channel_success,
+        ));
+
+        let last_future_template = self.last_future_template.clone();
+
+        // a future extended job was already created and activated on the group channel
+        // so we take the corresponding standard job to bootstrap the standard channel
```
GitGab19: How do you know the extended job was already activated?

roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -54,41 +64,140 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
...
+        // we deactivate it so we can send it as a future job
+        standard_job.deactivate();
+
+        standard_channel.on_new_template(standard_job.clone(), last_future_template.template_id);
+
+        // send this active job as future job, to be immediately activated with the subsequent
+        // SetNewPrevHash message
+        // (that's the expected flow on firmware side)
+        let mut standard_job_message = standard_job.get_job_message().clone().into_static();
+        standard_job_message.min_ntime = Sv2Option::new(None);
```
GitGab19: Isn't this already done in the `deactivate()` function?

roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -136,25 +389,132 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
...
+                            return Err(Error::FailedToUpdateStandardChannel(e));
+                        }
+                    }
+                }
+            }
+            let new_target = standard_channel.get_target();
+            let set_target = SetTarget {
+                channel_id,
+                maximum_target: new_target.clone().into(),
+            };
```
GitGab19, line 467: I think that blindly trust the client is a bit naive, even if it's more than ok for now. 

But probably the best thing is to check the vardiff state (once we will have it in place) before taking a decision.

Not a concrete suggestion, I just wanted to bring this up to start thinking about the best way to handle this.

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -480,46 +571,84 @@ impl Pool {
...
-                    let downstreams = handle_result!(status_tx, downstreams);
-
-                    for (channel_id, downtream) in downstreams {
-                        let message = Mining::SetNewPrevHash(SetNPH {
-                            channel_id,
-                            job_id,
+
+            let downstreams = handle_result!(status_tx, downstreams);
+
+            for (_downstream_id, downstream) in downstreams {
```
GitGab19, line 589: Just curiosity: have you done any test with a big number of downstreams and channels?

roles/pool/src/lib/mining_pool/message_handler.rs
plebhash: I expected confusion here, sorry for not being able to make it more clear... let's try to clarify it!

the context here is that we are bootstrapping the standard channel, where the message sequence to be sent to the client is always this:
- `OpenStandardMiningChannel.Success`
- `NewMiningJob`
- `SetNewPrevHashMp` (Mining Protocol variant)

if you check a few lines below, you will see that `SetNewPrevHashMp` will be created based on the cached `self.last_new_prev_hash` (which is `SetNewPrevHashTdp`,  the Template Distribution Protocol variant)

we know that a `SetNewPrevHashTdp` always arrives with a `template_id` referencing a future `NewTemplate`, so we know that:
- SOME future template has arrived previously (assuming a sane TP)
- this future template was [processed in the group channel](https://github.com/plebhash/stratum/blob/e4a7a80be930a0bd336a2e9f6b79aa48d5f84e63/roles/pool/src/lib/mining_pool/mod.rs#L702)
- this future template [was activated on the group channel upon arrival of `SetNewPrevHashTdp`](https://github.com/plebhash/stratum/blob/e4a7a80be930a0bd336a2e9f6b79aa48d5f84e63/roles/pool/src/lib/mining_pool/mod.rs#L595)

I was trying to closely follow the fact that spec says that each standard channel always belongs to some group channel, so I designed `StandardChannel::on_new_template` in a convoluted way to force the user to always convert the extended job from the group channel

however, the fact that this is causing confusion is making me re-evaluate this decision... I'll do some experiments and report back here

roles/pool/src/lib/mining_pool/message_handler.rs
plebhash: yes, this is redundant indeed

but as stated in the other comment, I will try to remove this convoluted design of converting extended jobs into standard

protocols/v2/roles-logic-sv2/src/channels/server/jobs/standard.rs
plebhash: yes, this comment is backwards indeed

but as stated in the other comment, I will try to remove this convoluted design of converting extended jobs into standard

roles/pool/src/lib/mining_pool/mod.rs
plebhash: not really

as far as I remember, the most diverse scenarios I tested were:
- BMM + Sv2 CPU miner (which means two separate downstreams, each one with a standard channel)
- tProxy with bitaxe + BMM100 (which means two separate downstreams, one with extended and one with standard channel)
- 2 tProxies, each one with a bitaxe (which means two separate downstreams, each one with an extended channel)
- 1 tProxy with 2 bitaxes (which means one downstream and one extended channel, not really different from having only one bitaxe)

overall, this validates that multiple clients work fine, although the scale is quite small (only 2)... I can try to expand this a bit, with something like 5-10 tProxies with one CPU miner each and see how it behaves

testing multiple channels under the same client will only be possible once we implement non-aggregated mode into tProxy

roles/pool/src/lib/mining_pool/message_handler.rs
GitGab19: I don't think you need to change that.

Following your explanation it makes sense, it has only to be detailed a bit more through comments maybe.

roles/pool/src/lib/mining_pool/mod.rs
GitGab19: I was asking this because the nested loop made me think about the risk of bottlenecks here.

What would you say about it?

roles/pool/src/lib/mining_pool/message_handler.rs
plebhash: ok I just pushed some new commits here, namely:
- 0d3f4dfe3e2b1ca9fd3a9deddc2248b40bb81b83
- 056802ba0600378f1ddadebcaff1c469b6fe39c4
- 3ff64b00e4ea164d0c2278563182e7c53b0d0b93

now we don't need this convoluted logic for converting extended job from group channel into standard, with the extra deactivation step

now, `channels::server::StandardChannel` also has a `JobFactory` (previously called `ExtendedJobFactory`, but now expanded so it also creates standard jobs)

now, as new templates arrive, we check if the `Downstream` was created with `SetupConnection.REQUIRES_STANDARD_JOBS` flag, and based on that we send one extended job or multiple standard jobs (created directly from the template)

the same goes for `SetNewPrevHash`

roles/pool/src/lib/mining_pool/mod.rs
plebhash: I see your point, and indeed now we have computational complexity `O(x * y)` for both `Pool::on_new_prev_hash` and `Pool::on_new_template`, where `x` is number of clients and `y` is number of channels per client

but this is not really a consequence of the new channel APIs, but rather a consequence of the fact that this migration is getting rid of the assumption that each client will have only one channel, which we can see in parts of the code such as this:

https://github.com/stratum-mining/stratum/blob/291e5aacb56a56789e2d3828cc37bbe3cfdc7d03/roles/pool/src/lib/mining_pool/mod.rs#L550-L570

when this Pool was designed, a non-aggregated proxy use-case wasn't considered, and now we're kind of ossified around this assumption... not so much with regards to the ability of handling multiple channels per client, but definitely with regards to the ability of handling multiple channels per client **efficiently** and without bottlenecks

MAYBE we could mitigate this bottleneck by spawning one task for each client, aiming to enable asynchronous processing of `NewTemplate` and `SetNewPrevHash`? 

(that is, spawning inside the `for (_downstream_id, downstream) in downstreams` loops of `Pool::on_new_prev_hash` and `Pool::on_new_template`)

I tried to do some quick experimentation in this direction but things broke all over the place, so it's not really trivial... and if we decided to explore this, we should also consider the danger of introducing new vectors for zombie tasks

another alternative would be to stick with the original assumption that every client will always only support one single channel, and reject requests to open new channels after the first one has already been open

finally, we could also decide to live with this bottleneck in SRI Pool?

plebhash: > finally, we could also decide to live with this bottleneck in SRI Pool?

I mean, for now this is a theoretical bottleneck... so maybe we can try to expand the analysis

these nested loops are happening on the context where we need to process `NewTemplate` and `SetNewPrevHash` for all clients and all channels

inside the channel management layer, processing `SetNewPrevHash` is pretty straightforward, as it consists of mostly `O(1)` lookups under the hood

and inside the channel management layer, processing `NewTemplate` could have a `O(n)` in the case of Standard Channels, because of merkle root calculation

but aside from that, everything else is `O(1)`

---

so if we consider the complexity from the application layer, worst case is `O(n * x * y)`, where:
- `n` is the number of merkle branches in the template
- `x` is the number of clients
- `y` is the number of channels per client

roles/pool/src/lib/mining_pool/message_handler.rs
plebhash: > I don't think you need to change that.
> 
> Following your explanation it makes sense, it has only to be detailed a bit more through comments maybe.

oh I only saw this now

overall I think the APIs ended up a bit cleaner after the changes described above, but also happy to drop them if requested

roles/pool/src/lib/mining_pool/mod.rs
GitGab19: I think we can live with this now, and better evaluate in the future if this becomes a bottleneck or not 👍

Issue #1652, opened 2025-04-28
coinbase output pool configuration


jakubtrnka:
I'm not sure if this really is an issue or just my misunderstanding.

I don't like very much how the coinbase transaction is configured at the pool app

```toml
# List of coinbase outputs used to build the coinbase tx
# ! Right now only one output is supported, so comment all the ones you don't need !
# For P2PK, P2PKH, P2WPKH, P2TR a public key is needed. For P2SH and P2WSH, a redeem script is needed.  
coinbase_outputs = [
    #{ output_script_type = "P2PK", output_script_value = "0372c47307e5b75ce365daf835f226d246c5a7a92fe24395018d5552123354f086" },
    #{ output_script_type = "P2PKH", output_script_value = "0372c47307e5b75ce365daf835f226d246c5a7a92fe24395018d5552123354f086" },
    #{ output_script_type = "P2SH", output_script_value = "00142ef89234bc95136eb9e6fee9d32722ebd8c1f0ab" },
    #{ output_script_type = "P2WSH", output_script_value = "00142ef89234bc95136eb9e6fee9d32722ebd8c1f0ab" },
    { output_script_type = "P2WPKH", output_script_value = "036adc3bdf21e6f9a0f0fb0066bf517e5b7909ed1563d6958a10993849a7554075" },
]
```

It's an input that is not user friendly. It actually requires some knowledge to get to it.
1. You want to start mining. You have got an bitcoin address. 
    1. In better case you have the pubkey available. But you need to understand how to get to it. Does your wallet allow access to that information?
    2. In worse case, you don't even have the public key. just the pubkey hash (there is no technical need for the public key)
2. There are multiple pubkey types. compressed, uncompressed, x-only coordinate. Each type has a different bitcoin address associated to it. You need to understand the specifics in order to do it right.
3. mainnet, testnet... they also have a different addresses
4. Even if you have the pubkey, there is no checksum. You may introduce a typo and mine into a black-hole without realizing it.

I'd prefer having the option to go both low-level and high-level:
1. low-level: Allow the user to specify the full ScriptPubKey to be used in the coinbase tx. Without any checks. Responsibility fully lies on the user.
2. high-level: Configuration input would be the "the thing that actually represents the output". i.e. legacy bitcoin address for P2PKH, bech32 address of a given length for segwit, taproot etc. Those addresses would be validated at the start and would be translated to a given scriptpubkey.

Don't force users to translate their addresses into some difficult intermediary item

some ideas:

[...]

Issue #1637, opened 2025-04-15
version bumps for v1.4.0


plebhash:
similar to #1345

plebhash:
#1629 is bumping `const_sv2` from v4.0.0 to v5.0.0

plebhash:
#1651 is forcing us to bump `mining_sv2` from v3.0.0 to v4.0.0

plebhash:
`sv1_api` is about to be bumped MAJOR via https://github.com/stratum-mining/stratum/pull/1668#discussion_r2082501242

from v1.0.1 to v2.0.0

plebhash:
> `sv1_api` is about to be bumped MAJOR via [#1668 (comment)](https://github.com/stratum-mining/stratum/pull/1668#discussion_r2082501242)
> 
> from v1.0.1 to v2.0.0

#1668 is no longer introducing the version bump

hiding the comments for now

plebhash:
#1728 forced us to bump:
- `codec_sv2` from v2.0.0 to v2.1.0
- `framing_sv2` from v5.0.0 to v5.1.0
- `noise_sv2` from v1.3.0 to v1.4.0
- `common_messages_sv2` from v5.0.0 to v5.1.0
- `template_distribution_sv2` from v3.0.0 to v3.1.0
- `sv2_ffi` from v2.0.0 to v2.1.0
- `network_helpers_sv2` from v.3.0.0 to v3.1.0

---

`const_sv2` was DEPRECATED

---

`job_declaration_sv2` and `mining_sv2` were both bumped from v4.0.0 to v4.1.0 BY MISTAKE, which is being reverted via #1749

PR #1719, opened 2025-05-22
Refactor and unification of `CoinbaseOutput` type

apoelstra:
Currently there are three `CoinbaseOutput` types: one in protocols/v2/roles-logic-sv2 which I believe is intended to be the "real" type, one in `roles/roles-utils/config-helpers` which I believe is intended to be a mirror of the coinbase outputs specified in the TOML config, and one in `roles/pool` which I believe should've been combined with the second one in #1507 but was missed.

However, both the "real" and "config" types are identical and neither has any invariants enforced on them; they are just a pair of strings. This PR combines these all into one type. In my next PR I will add invariants and make it impossible to construct this type, even from serde, except with a valid coinbase output. I will then extend the type to allow parsing not only the current type/value format, but also a new descriptor format which will cover more usecases and be more compatible with wallets.

This is a first step toward #1652.


protocols/v2/roles-logic-sv2/src/errors.rs
```diff
@@ -74,12 +74,6 @@ pub enum Error {
     InvalidCoinbase,
     /// Value remaining in coinbase output was not correctly updated (it's equal to 0)
     ValueRemainingNotUpdated,
-    /// Unknown script type in config
-    UnknownOutputScriptType,
-    /// Invalid `output_script_value` for script type. It must be a valid public key/script
-    InvalidOutputScript,
-    /// Empty coinbase outputs in config
-    EmptyCoinbaseOutputs,
```
GitGab19, line 82: This removal should be taken into account under https://github.com/stratum-mining/stratum/issues/1637.

I think it should force us to bump MAJOR for `roles_logic_sv2` before releasing.

Why `semver-check` was not triggered in this case? cc @plebhash 
Am I missing something?

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -221,92 +219,6 @@ fn reduce_path<T: AsRef<[u8]>>(coinbase_id: [u8; 32], path: &[T]) -> [u8; 32] {
...
-                    &Secp256k1::<All>::new(),
-                    pub_key,
-                    None,
-                ))
-            }
-            _ => Err(Error::UnknownOutputScriptType),
-        }
-    }
-}
-
```
GitGab19, line 309: Same doubt on this

GitGab19 (commented):
Left a couple of comments about bumping MAJOR for `roles_logic_sv2` before next release, but I would like a double check by @plebhash on this.

LGTM.

protocols/v2/roles-logic-sv2/src/errors.rs
apoelstra: Good catch. Yes, this is definitely an API break in roles-logic-sv2.

plebhash: in other words, this API breaking change will also go into the v2.0.0 -> v3.0.0 version bumping when we publish it to crates.io

roles/roles-utils/config-helpers/Cargo.toml
```diff
@@ -13,4 +13,5 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 [dependencies]
 serde = { version = "1.0.89", features = ["derive","alloc"], default-features = false }
+miniscript = { version = "12.3.2", default-features = false, features = [ "no-std" ] }
```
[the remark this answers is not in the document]
plebhash: > it's used in the line use miniscript::bitcoin in roles/roles-utils/config-helpers/src/coinbase_output/mod.rs.

that's on the commit afterwards, I overlooked that... sorry!

PR #1720, opened 2025-05-25
Allow specifying descriptors (including `raw` and `addr`) as coinbase outputs

### LINKED ISSUE #1652: coinbase output pool configuration

apoelstra:
Fixes #1652.

Draft because I need to update the example configuration files and the documentation. But I am using this with my Bitaxes. Tried both the `raw` and the `addr` descriptor and was able to use a vanitygenned address which previously was impossible to specify.

### DIFF
--- a/roles/jd-client/config-examples/jdc-config-hosted-example.toml
+++ b/roles/jd-client/config-examples/jdc-config-hosted-example.toml
@@ -26,18 +26,13 @@ tp_authority_public_key = "9bwHCYnjhbHm4AS3pWg9MtAH83mzWohoJJJDELYBqZhDNqszDLc"
 jdc_signature = "JDC"
 
 # Solo Mining config
-# List of coinbase outputs used to build the coinbase tx in case of Solo Mining (as last-resort solution of the pools fallback system)
-# ! Put your Extended Public Key or Script as output_script_value !
-# ! Right now only one output is supported, so comment all the ones you don't need !
-# For P2PK, P2PKH, P2WPKH, P2TR a public key is needed. For P2SH and P2WSH, a redeem script is needed.  
-coinbase_outputs = [
-    #{ output_script_type = "P2PK", output_script_value = "0372c47307e5b75ce365daf835f226d246c5a7a92fe24395018d5552123354f086" },
-    #{ output_script_type = "P2PKH", output_script_value = "0372c47307e5b75ce365daf835f226d246c5a7a92fe24395018d5552123354f086" },
-    #{ output_script_type = "P2SH", output_script_value = "00142ef89234bc95136eb9e6fee9d32722ebd8c1f0ab" },
-    #{ output_script_type = "P2WSH", output_script_value = "00142ef89234bc95136eb9e6fee9d32722ebd8c1f0ab" },
-    { output_script_type = "P2WPKH", output_script_value = "036adc3bdf21e6f9a0f0fb0066bf517e5b7909ed1563d6958a10993849a7554075" },
-    #{ output_script_type = "P2TR", output_script_value = "036adc3bdf21e6f9a0f0fb0066bf517e5b7909ed1563d6958a10993849a7554075" },
-]
+# Coinbase output used to build the coinbase tx in case of Solo Mining (as last-resort solution of the pools fallback system)
+#
+# Coinbase outputs are specified as descriptors. A full list of descriptors is available at
+#     https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki#appendix-b-index-of-script-expressions
+# Although the `musig` descriptor is not yet supported and the legacy `combo` descriptor never
+# will be. If you have an address, embed it in a descriptor like `addr(<address here>)`.
+coinbase_output = "addr(tb1qa0sm0hxzj0x25rh8gw5xlzwlsfvvyz8u96w3p8)"
 
 # Enable this option to set a predefined log file path.
 # When enabled, logs will always be written to this file.
--- a/roles/pool/src/lib/config.rs
+++ b/roles/pool/src/lib/config.rs
@@ -22,6 +22,7 @@ pub struct PoolConfig {
     authority_public_key: Secp256k1PublicKey,
     authority_secret_key: Secp256k1SecretKey,
     cert_validity_sec: u64,
+    #[serde(alias = "coinbase_output")] // only one is allowed, so don't make the user type the plural
     #[serde(deserialize_with = "config_helpers::deserialize_vec_exactly_1")]
     coinbase_outputs: Vec<CoinbaseOutput>,
     pool_signature: String,
--- a/roles/pool/src/lib/config.rs
+++ b/roles/pool/src/lib/config.rs
@@ -22,6 +22,7 @@ pub struct PoolConfig {
     authority_public_key: Secp256k1PublicKey,
     authority_secret_key: Secp256k1SecretKey,
     cert_validity_sec: u64,
+    #[serde(deserialize_with = "config_helpers::deserialize_vec_exactly_1")]
     coinbase_outputs: Vec<CoinbaseOutput>,
     pool_signature: String,
     shares_per_minute: f32,
@@ -31,6 +32,10 @@ pub struct PoolConfig {
 
 impl PoolConfig {
     /// Creates a new instance of the [`PoolConfig`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if `coinbase_outputs` is empty.
     pub fn new(
         pool_connection: ConnectionConfig,
         template_provider: TemplateProviderConfig,
@@ -39,6 +44,10 @@ impl PoolConfig {
         shares_per_minute: f32,
         share_batch_size: usize,
     ) -> Self {
+        assert!(
+            !coinbase_outputs.is_empty(),
+            "set of coinbase outputs must be nonempty"
+        );
         Self {
             listen_address: pool_connection.listen_address,
             tp_address: template_provider.address,
--- a/roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
+++ b/roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
@@ -1,8 +1,6 @@
 mod errors;
 mod serde_types;
 
-use core::convert::TryFrom;
-
 use miniscript::{
     bitcoin::{Script, ScriptBuf},
     DefiniteDescriptorKey, Descriptor,
@@ -20,14 +18,6 @@ pub struct CoinbaseOutput {
 }
 
 impl CoinbaseOutput {
-    /// Creates a new [`CoinbaseOutput`] from a script type and value.
-    pub fn new(output_script_type: String, output_script_value: String) -> Result<Self, Error> {
-        Self::try_from(serde_types::LegacyCoinbaseOutput {
-            output_script_type,
-            output_script_value,
-        })
-    }
-
     /// Creates a new [`CoinbaseOutput`] from a descriptor string.
     pub fn from_descriptor(s: &str) -> Result<Self, Error> {
         let desc = s.parse::<Descriptor<DefiniteDescriptorKey>>()?;
--- a/roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
+++ b/roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
@@ -2,7 +2,7 @@ mod errors;
 mod serde_types;
 
 use miniscript::{
-    bitcoin::{Script, ScriptBuf},
+    bitcoin::{address::NetworkUnchecked, Address, Network, Script, ScriptBuf},
     DefiniteDescriptorKey, Descriptor,
 };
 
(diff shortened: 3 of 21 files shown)


apoelstra:
> I assume you're pointing the Bitaxes to tProxy?

Yes.
> are you hooking tProxy straight to Pool, or are you also deploying JDS+JDC?

Yep, straight to Pool. Sometimes the job IDs change and the bitaxe keeps doing its original work, leading to a fairly high number of "wrong job ID" errors and lost work. With Bitaxes I don't really care because their main purpose is learning, but I would like to investigate this next.

> are you connected to some testnet, signet (without signature), regtest or mainnet?

Mainnet. I am using Sjors' fork of Core.

apoelstra:
Thanks, that sounds great. Added you on Signal. As I said there, it'll probably be a few weeks before I have time to play with my mining set up (other than getting this PR through) but I would like to join eventually.

GitGab19:
Hey @apoelstra, any updates on this one?

It would be cool to have this one before next release.

apoelstra:
What is the timeline for the next release? I think I can work on this today or tomorrow. Sorry for the delay. Work has been pretty hectic for the last 3 weeks.

GitGab19:
I would say we are targeting the 2nd week of July (between 7th and 10th).

No worries btw, if you can do it before that it would be awesome, otherwise it's fine also after that.

apoelstra:
Rebased only; haven't yet added the new tests and documentation that I need to in order to take this out of draft.

apoelstra:
Was able to connect my bitaxes to this with the `addr` descriptor in the example pool (which points to an address I control by the way; I'm happy to change this in case we worry about people mining to the default address).

test/integration-tests/lib/mo
```diff
@@ -0,0 +1 @@
+
```
lucasbalieiro: I think this file got uploaded by accident

roles/pool/src/lib/config.rs
```diff
@@ -20,6 +20,7 @@ pub struct PoolConfig {
     authority_public_key: Secp256k1PublicKey,
     authority_secret_key: Secp256k1SecretKey,
     cert_validity_sec: u64,
+    #[serde(deserialize_with = "config_helpers::deserialize_vec_at_least_1")]
```
lucasbalieiro: shouldn't we be using

```suggestion
    #[serde(deserialize_with = "config_helpers::deserialize_vec_exactly_1")]
```

to strictly enforce what the TOML comment is stating?

```not_rust
# ! Right now only one output is supported, so comment all the ones you don't need !
```
As it stands, the pool still starts even if I configure multiple coinbase outputs, which goes against that guidance.

apoelstra: Good question! I wasn't sure whether to change the code to enforce the comment, or leave it be to match the existing logic. I'm happy to go either way (it's a trivial change, as you can see).

plebhash: yeah, we should go with only 1

tbh this shouldn't even be a `Vec`, but maybe that's out of scope

apoelstra: I can update `deserialize_vec_exactly_1` to allow a singleton vector **or** a single object. Then we will continue to support the vec syntax for backward compatibility (and forward compat in case we ever decide a reasonable use of multiple coinbase output scripts).

plebhash: maybe I should have elaborated a bit more on my message above

the reason I believe this shouldn't be a `Vec` is because we're only really parsing the script from the config file, while the output value remains undetermined (because template revenue can only be discovered at runtime)

on SRI we have basically fours scenarios where outputs are parsed from config files:
A. Pool for mining with JD
B. JDS for mining with JD
C. JDC for solo-mining as JD-fallback
D. Pool for mining without JD

for scenarios A and B, the spec imposes that only one single output can be used for pool payout (because otherwise there would be no simple way to coordinate this payout over the JD message flow)

(we could potentially merge Pool and JDs someday, so scenarios A and B could become the same, although that's not too relevant here)

for scenarios C and D, I guess one could argue that some kind of arbitrary rule could be established to allow distributing the template revenue across more than one single output

however, that remains purely speculative

ANYWAYS, all this rambling is just to clarify why I believe the `Vec<CoinbaseOutput>` should have never been a `Vec` in the first place

but to replace `Vec<CoinbaseOutput>` with `CoinbaseOutput` (or even `CoinbaseScriptPubkey` with a more accurate naming) would imply also changing some other parts of the code (namely `get_txout` and `get_coinbase_output` functions and where they are called), which would probably break tests and make things way more complicated... so that's why I also feel this is a bit out of scope and probably not worth the effort now

> I can update `deserialize_vec_exactly_1` to allow a singleton vector **or** a single object. Then we will continue to support the vec syntax for backward compatibility (and forward compat in case we ever decide a reasonable use of multiple coinbase output scripts).

and yeah that sounds like a good way to move forward 👍

apoelstra: Ok, great -- and I should also change the example config files to use the non-`Vec` syntax?

apoelstra: Done. I also serde-aliased `coinbase_outputs` to `coinbase_output`.

A followup PR can rename the internal field names from `coinbase_outputs` to `coinbase_output` and change their type from `Vec<T>` to `T` and so on. This seems like a pretty mechanical thing and is maybe worth filing a "good for new contributors" issue about.

plebhash: awesome, thanks a lot!

I see we broke some tests though :(

apoelstra: This error failure is pretty frustrating. The issue is that I'm using an internal untagged enum to implement "deserialize as either a vec or singleton", but in serde when you use an untagged enum then it throws away all meaningful error information and instead shows the user an ugly meaningless error.

There is a massive crate [serde_with](https://docs.rs/serde_with/latest/serde_with/struct.OneOrMany.html) which uses proc macros and [basically mirrors all of serde](https://docs.rs/serde_with/latest/src/serde_with/content/de.rs.html#25-56) to avoid this, but this is a giant dependency for something that really should be trivial. (There is also this classic serde issue https://github.com/serde-rs/serde/issues/1907 requesting this common functionality, closed as WONTFIX because it's "adequately supported by the `deserialize_with` attribute", with no explanation of what that means).

apoelstra: We have a couple choices here:

* Live with the terrible error messages
* Take an alternate approach where we have two fields, `coinbase_outputs` which is a `Vec` and `coinbase_output` which is a non-`Vec`, and extra logic to enforce that only one is populated (which I think would wind up requiring us to do the refactoring to remove `Vec` entirely, or something close to it)
* Leave the array syntax as-is

plebhash: the fact that we're making progress towards a cleaner codebase that doesn't rely on `Vec<CoinbaseOutput>` is very good, especially in terms of UX improvement

so that makes me lean towards the first approach, which feels like a reasonable trade-off that allows us to keep going in this direction

I created #1793 so we can continue making progress in this direction while unblocking your contribution here

for now, we could add a commit adapting those tests to assert against the meaningless error messages, while adding a comment providing context and a reference to #1793... I can do that in case you're too busy with other stuff

apoelstra: Perfect. I'll do it because I'd prefer to edit my own commits to ensure they all compile and pass tests (though I am only checking this with default feature combos).

roles/jd-client/config-examples/jdc-config-hosted-example.toml
```diff
@@ -26,18 +26,13 @@ tp_authority_public_key = "9bwHCYnjhbHm4AS3pWg9MtAH83mzWohoJJJDELYBqZhDNqszDLc"
...
-    { output_script_type = "P2WPKH", output_script_value = "036adc3bdf21e6f9a0f0fb0066bf517e5b7909ed1563d6958a10993849a7554075" },
-    #{ output_script_type = "P2TR", output_script_value = "036adc3bdf21e6f9a0f0fb0066bf517e5b7909ed1563d6958a10993849a7554075" },
-]
+# Coinbase output used to build the coinbase tx in case of Solo Mining (as last-resort solution of the pools fallback system)
+#
+# Coinbase outputs are specified as descriptors. A full list of descriptors is available at
+#     https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki#appendix-b-index-of-script-expressions
+# Although the `musig` descriptor is not yet supported and the legacy `combo` descriptor never
+# will be. If you have an address, embed it in a descriptor like `addr(<address here>)`.
+coinbase_output = "addr(bc1pandyt0shm5d0l99ppsp3knm7n7wynfyy3khazzvc7m033j9ulupscyve35)"
```
GitGab19: Should we put an address coming from the same wallet used for the current configs here?

Should it be a `testnet` one since we are referring to `testnet4` in our [getting-started guide](https://stratumprotocol.org/blog/getting-started)?

apoelstra: Yeah, using a testnet address seems like a good idea. This PR introduces hooks to allow preventing testnet addresses from being used an mainnet (but does not actually use these hooks!) so this will eventually have the nice side-effect of preventing people from using the example configs without changing the coinbase address.

apoelstra: Done. I used exactly the same address as was implied by the old config. This also has the nice effect that I didn't have to change any unit tests, so you can see that my new code generates the new scriptpubkeys as the old code.

GitGab19: > so this will eventually have the nice side-effect of preventing people from using the example configs without changing the coinbase address

Can you elaborate more on this? 

I haven't reviewed this PR yet (I'm gonna do it tomorrow morning), but does your comment mean that it's not gonna work on testnet4? Or did you mean that the user would have to change the address with a custom mainnet one when testing SRI on mainnet?

apoelstra: @GitGab19 I have added an `ok_for_mainnet` boolean to coinbase outputs. This is set to "false" for non-mainnet addresses but "true" for everything else (it should also be set to "false" when the user provides non-mainnet xpubs within a descriptor, I suppose, but I didn't implement that).

Eventually I would like this boolean to trigger an error if the boolean is set to "false" when the user attempts to start up on mainnet.

But testnets are a free-for-all. If you connect to testnet, or testnet4, or signet or mutinynet or regtest or whatever, there should be no error conditions. (a) it's basically impossible to distinguish testnets reliably and to figure out whether an address is appropriate for a given testnet, and (b) it really doesn't matter if iwe let the user mess up and lose money on a testnet.

roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
```diff
@@ -13,108 +12,362 @@ pub use errors::Error;
...
+                    return Err(Error::AddrDescriptorNChildren(tree.args.len()));
+                }
+                if !tree.args[0].args.is_empty() {
+                    return Err(Error::AddrDescriptorGrandchild);
+                }
+
+                let addr = tree.args[0].name.parse::<Address<NetworkUnchecked>>()?;
+                Ok(Self {
+                    script_pubkey: addr.assume_checked_ref().script_pubkey(),
+                    ok_for_mainnet: addr.is_valid_for_network(Network::Bitcoin),
```
GitGab19, line 67: This is the key point where we set this new `ok_for_mainnet` boolean to false in case the address is not a mainnet one, right?

roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
```diff
@@ -13,108 +12,362 @@ pub use errors::Error;
...
+                    return Err(Error::RawDescriptorNChildren(tree.args.len()));
+                }
+                if !tree.args[0].args.is_empty() {
+                    return Err(Error::RawDescriptorGrandchild);
+                }
+
+                let bytes = Vec::<u8>::from_hex(tree.args[0].name)?;
+                Ok(Self {
+                    script_pubkey: ScriptBuf::from(bytes),
+                    // Users of hex scriptpubkeys are on their own.
```
GitGab19, line 84: Does it mean we cannot check the mainnet-validity of the scriptpubkeys?

apoelstra: Correct. A hex scriptpubkey has no network flag of any form -- a scriptpubkey looks exactly the same no matter what network it's intended for.

roles/roles-utils/config-helpers/src/coinbase_output/mod.rs
apoelstra: Yep. And as I mentioned above, there's actually another check I should do in the general descriptor case, where I should iterate through all the `DefiniteDescriptorKey`s and look at the network flag on the xpubs. But I think this is out-of-scope for this PR since it's a bit inconvenient to do and I think I want to make some upstream rust-miniscript changes to support it better.

apoelstra: https://github.com/rust-bitcoin/rust-miniscript/issues/844

GitGab19:
@apoelstra could you please fix the conflicts here? 

Otherwise I can do it for you if you're busy.

apoelstra:
Rebased.

PR #1728, opened 2025-05-30
realocate constants to their specific crates

plebhash:
IMO the approach of #1642 was suboptimal

moving the contents from `const_sv2` into `stratum_common` was a bit redundant, since we ended up still having all constants concentrated in one single place

constants should live in the specific crates where they have contextual meaning


protocols/v2/sv2-ffi/Cargo.toml
GitGab19, line 1: Shouldn't we bump its version to `2.1.0`?

roles/roles-utils/network-helpers/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "network_helpers_sv2"
-version = "3.0.0"
+version = "3.0.1"
```
GitGab19: Why did you bump PATCH and not MINOR here?

GitGab19:
LGTM, I just left a couple of comments to be reviewed before merging this.

roles/roles-utils/network-helpers/Cargo.toml
plebhash: because no constants were introduced to the crate, we only adjusted the import path

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> - MAJOR version when you make incompatible API changes
> - MINOR version when you add functionality in a backward compatible manner
> - PATCH version when you make backward compatible bug fixes

but yeah, strictly speaking there was no bugfix here... so perhaps MINOR would be more adequate

protocols/v2/sv2-ffi/Cargo.toml
plebhash: hmm indeed I overlooked the version control here

it would have originally done a PATCH bump, but perhaps MINOR is better

plebhash: updated to bump MINOR

roles/roles-utils/network-helpers/Cargo.toml
plebhash: updated to bump MINOR

Issue #1725, opened 2025-05-29
Add Dedicated Test Suite for Vardiff to Ensure Stable and Accurate Difficulty Adjustment


Shourya742:
After the Vardiff module is integrated into the roles logic via (https://github.com/stratum-mining/stratum/pull/1714), we should begin developing a dedicated test suite to verify that difficulty adjustments are calculated accurately and remain stable over time. The goal is to ensure that Vardiff responds correctly to changes in miner hashrate without causing erratic fluctuations, maintaining consistent target share submission rates and overall system reliability across a variety of scenarios.

PR #1729, opened 2025-06-03
Add vardiff test framework

Shourya742:
Partially solves: #1725

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
+++ b/protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
@@ -268,3 +268,49 @@ fn test_try_vardiff_no_shares_more_than_60s_decrease<V: Vardiff>(vardiff: &mut V
     );
     assert_eq!(vardiff.shares_since_last_update(), 0);
 }
+
+fn test_try_vardiff_with_less_spm_than_expected<V: Vardiff>(vardiff: &mut V) {
+    let initial_hashrate = vardiff.hashrate();
+
+    assert_eq!(initial_hashrate, 1000.0);
+
+    let simulation_duration = 60;
+    // testing case when realized_shares_per_minute / shares_per_minute = 0.4
+    simulate_shares_and_wait(vardiff, 4, simulation_duration);
+
+    let hashrate_after_60s = vardiff.try_vardiff().expect("try_vardiff failed").unwrap();
+
+    assert_eq!(hashrate_after_60s, 400.0);
+
+    let simulation_duration = 120;
+    // testing case when realized_shares_per_minute / shares_per_minute = 0.5
+    simulate_shares_and_wait(vardiff, 10, simulation_duration);
+
+    let hashrate_after_120s = vardiff.try_vardiff().expect("try_vardiff failed").unwrap();
+
+    assert_eq!(hashrate_after_120s, 200.0);
+
+    let simulation_duration = 180;
+    // testing case when realized_shares_per_minute / shares_per_minute = 0.55
+    simulate_shares_and_wait(vardiff, 16, simulation_duration);
+
+    let hashrate_after_180s = vardiff.try_vardiff().expect("try_vardiff failed").unwrap();
+
+    assert_eq!(hashrate_after_180s, 106.0);
+
+    let simulation_duration = 240;
+    // testing case when realized_shares_per_minute / shares_per_minute = 0.7
+    simulate_shares_and_wait(vardiff, 28, simulation_duration);
+
+    let hashrate_after_240s = vardiff.try_vardiff().expect("try_vardiff failed").unwrap();
+
+    assert_eq!(hashrate_after_240s, 74.2);
+
+    let simulation_duration = 300;
+    // testing case when realized_shares_per_minute / shares_per_minute = 0.85
+    simulate_shares_and_wait(vardiff, 42, simulation_duration);
+
+    let hashrate_after_300s = vardiff.try_vardiff().expect("try_vardiff failed").unwrap();
+
+    assert_eq!(hashrate_after_300s, 62.327995);
+}
--- a/protocols/v2/roles-logic-sv2/src/vardiff/test/classic.rs
+++ b/protocols/v2/roles-logic-sv2/src/vardiff/test/classic.rs
@@ -11,8 +11,8 @@ use super::{
     test_try_vardiff_no_shares_less_than_30s_decrease,
     test_try_vardiff_no_shares_more_than_60s_decrease,
     test_try_vardiff_stable_hashrate_minimal_change_or_no_change,
-    test_try_vardiff_with_shares_30_to_60s, test_try_vardiff_with_shares_less_than_30,
-    test_try_vardiff_with_shares_more_than_60s, Vardiff,
+    test_try_vardiff_with_less_spm_than_expected, test_try_vardiff_with_shares_30_to_60s,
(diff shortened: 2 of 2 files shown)


protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        );
+        assert!(
+            diff_percentage < 20.0,
+            "Change should be minimal for stable rate if any"
+        );
+        assert_eq!(vardiff.shares_since_last_update(), 0)
+    } else {
+        assert!(
+            true,
+            "No update for stable hashrate is acceptable if deviation is low"
```
plebhash: I don't understand this... why `assert!` on `true`?

the sentence in the string is also not clear, are we saying that if deviation is low, updates are unacceptable? or that it's acceptable that no updates happened?

and overall, do we expect an update to happen or not?

plebhash: and since hashrate didn't change, shouldn't we simply assert that `vardiff.try_vardiff()` returns `Ok(None)`?

I mean, I understand the API was designed to be robust against statistically insignificant changes, but the execution of this test is deterministic so I don't get why we would leave different `if / else` statements here?

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        "Should not update within 15s; elapsed 10s"
+    );
+}
+
+pub fn test_try_vardiff_stable_hashrate_minimal_change_or_no_change<V: Vardiff>(vardiff: &mut V) {
+    let initial_hashrate = vardiff.hashrate();
+
+    let simulation_duration_secs = 70;
+    let expected_shares_for_duration =
+        (TEST_SHARES_PER_MINUTE * simulation_duration_secs as f32 / 60.0).round() as u32;
```
plebhash: given the linearity of this operation, what's the difference between setting a 70s duration and something much smaller, like 5s?

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+use std::{thread, time::Duration};
+
+use crate::VardiffState;
+mod classic;
+
+use super::Vardiff;
+
+pub const TEST_INITIAL_HASHRATE: f32 = 1000.0;
+pub const TEST_SHARES_PER_MINUTE: f32 = 10.0;
+pub const TEST_MIN_ALLOWED_HASHRATE: f32 = 10.0;
```
plebhash, line 11: I did `cargo t` on this branch and I feel the test duration is a bit prohibitive.

waiting 70s or even 250s for a test to finish its execution feels a bit too much.

but also, I don't see any fundamental reason that forces us to do this

I mean, sure, expected share rate is set on a per-minute basis for UX reasons, but that doesn't mean that we must test scenarios for multiple minutes to assert the sanity of the APIs

if we remember that all math done here is linear, we can simply tweak these constants in order to allow us to perform our simulations under smaller execution timespans

plebhash: for example, if `TEST_SHARES_PER_MINUTE` is 60_000, we expect 1_000 shares per second

if in 1 second we only get 10 shares, that should be a trigger to lower the difficulty

if in 1 second we get 5_000 shares, that should be a trigger to raise the difficulty

if we want to emulate scenarios of "long" timespans, we can do ~10s

overall, we have plenty of room to play around with magnitudes, while trying to optimize for test execution time

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        new_hashrate, TEST_MIN_ALLOWED_HASHRATE,
+        "Hashrate should be clamped to minimum"
+    );
+    assert_eq!(
+        vardiff.hashrate(),
+        TEST_MIN_ALLOWED_HASHRATE,
+        "Stored hashrate should be clamped"
+    );
+    assert_eq!(vardiff.shares_since_last_update(), 0);
+}
```
plebhash: as discussed in the call, it seems this test belongs in a different file?

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+        new_hashrate > initial_hashrate,
+        "Estimated hashratem should increase"
+    );
+    assert!(
+        vardiff.target() < initial_target,
+        "Target should become harder (smaller value)"
+    );
+    assert_eq!(vardiff.shares_since_last_update(), 0);
+}
+
```
plebhash, line 183: I see tests to assert target increase and decrease

but I don't see anything asserting that the vardiff converges to a correct hashrate estimation?

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
plebhash: or even 1s?

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
plebhash: ok I see the minimum window we have is 15s

https://github.com/stratum-mining/stratum/blob/e5fd8b7a35ba8f7c13e25608a8ec7f56116cb1d3/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs#L157-L159

@GitGab19 I assume this is important for the correct behavior of the algorithm, right?

if so, then we can still make test durations much smaller, sticking to 15s as the default execution time and making artithmetic adaptations around `TEST_SHARES_PER_MINUTE`

bit-aloo: I’ve added a test configuration for this, which will also allow delta_time <= 0 in testing scenarios.

bit-aloo: Although the linearity seems tempting to exploit for testing, there’s a fundamental reason we need to wait for this duration our logic is designed to trigger at specific time intervals. Anyway, I’ve updated the tests to specifically target those wait points, and they’re now more intuitive to understand. The downside is that we still need to wait at least 60 seconds for those points to be reached. I initially considered adding some test-specific logic to bypass this, but that felt like heavy lifting given the current setup.

protocols/v2/roles-logic-sv2/src/vardiff/test/mod.rs
bit-aloo: I understand your point about the convergence test. I’m planning to add that in a separate PR since it will require more work. This PR focuses on adding the basic skeleton for the testing framework for the vardiff module and strictly handling edge cases in the vardiff calculation. We already have a convergence mechanism in the difficulty management of tproxy, but it's somewhat unreliable. Also, simulating the entire hashing process is influenced by variable processing speeds, which makes it tricky. Designing such tests to be deterministic across the system would need a more careful and thorough review.

plebhash: when I wrote this comment originally, I wasn't necessarily thinking about dynamically testing with real hashrate from a CPU miner

my mindset was more around statically running the tests with specific values, such that we can assert that the vardiff algo converges into the right direction on its hashrate estimation

plebhash (approved):

Issue #1438, opened 2025-02-03
Add a way for the pool to change `coinbase_script_additional_data` of an already opened channel


GitGab19:
Third task from the original PR: https://github.com/stratum-mining/stratum/pull/1248#issue-2656301508

plebhash:
@Fi3 can you elaborate on the motivation for this?

plebhash:
> I need this to put a commitment of the diff assigned by the pool in the share is only needed if you want verify work provided by miners. Not sure if you want to add it on SRI. If you don't I will need to maintain a separate roles logic and merge upstream when you modify roles logic.

I am writing a proposal for a new Channel Management module for `roles_logic_sv2`. The APIs will be very different, but I will accommodate this as a feature request.

plebhash:
> I need this to put a commitment of the diff assigned by the pool in the share is only needed if you want verify work provided by miners. Not sure if you want to add it on SRI. If you don't I will need to maintain a separate roles logic and merge upstream when you modify roles logic.

@Fi3 how is this difficulty encoded?

Sv2 Extended Extranonces are limited to 32 bytes, due to the fact that `OpenStandardMiningChannel.Success` is `B0_32` and that needs to fit the entire Extended Extranonce

is 32 bytes enough to commit difficulty? if it's compact encoded (like `nbits`) then I guess we're fine, but if it's a full target (256 bits) then we already used all the available space (and there's nothing left for search space allocation)

plebhash:
I reached out to @Fi3 over Discord and he confirmed that it will be compact encoded like `nbits`, so only 32 bits (or 4 bytes) are necessary for this

plebhash:
given that `additional_coinbase_script_data` is now embedded into `extranonce_prefix`, the APIs introduced in #1672 should be sufficient for this purpose

more specifically, the application layer can call:
- `roles_logic_sv2::channels::client::extended::ExtendedChannel::set_extranonce_prefix`
- `roles_logic_sv2::channels::client::standard::StandardChannel::set_extranonce_prefix`

these methods change the channel state so that every job created afterwards are associated with the new `extranonce_prefix`

it's up to the API caller to ensure unique search space allocation, since the new `extranonce_prefix` are not necessarily the ones originally assigned via `mining_sv2::ExtendedExtranonce`

it's also up to the application layer to ensure that the client gets a `SetExtranoncePrefix` message with a notification of this change BEFORE receiving any new jobs

---

[...]

PR #1730, opened 2025-06-03
add verification on `roles_logic_sv2::channels::*::standard::StandardChannel::set_extranonce_prefix`

plebhash:
otherwise it would be possible to set a new `extranonce_prefix` that's bigger than 32 bytes, which would violate the spec and would not fit a `SetExtranoncePrefix` message

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/channels/server/standard.rs
+++ b/protocols/v2/roles-logic-sv2/src/channels/server/standard.rs
@@ -10,7 +10,7 @@ use crate::{
     },
     utils::{bytes_to_hex, hash_rate_to_target, target_to_difficulty, u256_to_block_hash},
 };
-use mining_sv2::{SubmitSharesStandard, Target};
+use mining_sv2::{SubmitSharesStandard, Target, MAX_EXTRANONCE_LEN};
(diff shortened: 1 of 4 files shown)


plebhash:
after merging this, we should be able to close #1438

protocols/v2/roles-logic-sv2/src/channels/server/standard.rs
```diff
@@ -126,8 +126,17 @@ impl<'a> StandardChannel<'a> {
...
+        &mut self,
+        extranonce_prefix: Vec<u8>,
+    ) -> Result<(), StandardChannelError> {
+        if extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
+            return Err(StandardChannelError::NewExtranoncePrefixTooLarge);
+        }
+
         self.extranonce_prefix = extranonce_prefix;
+
+        Ok(())
```
GitGab19, line 139: Add the same check on client version of this function.

plebhash: done

GitGab19 (approved):

PR #1738, opened 2025-06-07
update imports in `roles` to reflect library reorganization

### LINKED ISSUE #1458: need to standardize dependency (and API re-exporting) patterns

lucasbalieiro:
closes #1458 ;

* **Reorganized exports:**

  * `roles_logic_sv2` now re-exports `binary_sv2` and `framing_sv2`.
  * `network_helpers_sv2` now re-exports `codec_sv2`.

* **Updated dependencies in `roles/` crates:**

  * Removed direct imports of `binary_sv2`, `framing_sv2`, and `codec_sv2`.
  * Updated all imports to go through `roles_logic_sv2` and `network_helpers_sv2` instead.


plebhash:
hey @Shourya742 can you have a look at this when you have some time?

I gave some pointers to @lucasbalieiro , which is how these dependencies currently make more sense in my mind.

and from a quick review, it seems he implemented them correctly here.

however I recall from past discussions you had a different view on `framing_sv2` and `codec_sv2`? IIRC you pointed out that `codec_sv2` should re-export `framing_sv2`

at the time, that also made sense to me... but that was before `network_helpers_sv2` entered on my radar for this discussion 

later on I noticed that `codec_sv2` (and consequently `noise_sv2`) are only imported on `network_helpers_sv2`, but not on `roles_logic_sv2`... which kind of makes sense, since `roles_logic_sv2` doesn't really try to actually send encoded messages over the wire

and then I also noticed that `framing_sv2` is used in isolation (that is: without `codec_sv2`) on `roles_logic_sv2`, more specifically for the `parsers` module... which also kind of makes sense

so if we funneled `framing_sv2` through `codec_sv2`, we would risk "bloating" `roles_logic_sv2` with the extra code of `codec_sv2`, even though none of it's APIs are really used there

anyways, I'm not hardly opinionated on this, as in the end of the day, I feel that the line between `codec_sv2`, `framing_sv2`, and `binary_sv2` is extremely blurry and sometimes I even wonder if those shouldn't be one unified crate

so I don't really care if we choose a different approach and "bloat" `roles_logic_sv2` as a trade-off for a more coherent crate organization... just felt the need to clarify the rationale of this PR here

plebhash:
this needs to be rebased

Issue #1704, opened 2025-05-14
Standardize `args` Module Structure


Shourya742:
The way CLI arguments are declared and parsed varies slightly across role implementations. While similar patterns are used, the file structures and organization differ.

Having inconsistent patterns across roles makes the codebase harder to read and maintain. If we adopt a specific approach in one role, we should apply it consistently across all roles.

To address this, we should standardize CLI argument handling by introducing a standalone `args` module at the root level of each role's binary crate. This module should include:

* A `struct Args` annotated with `clap` macros
* A `fn process_cli_args()` function that returns a `*Config` object

This standardized pattern is already in use in some roles:

* [`roles/translator`](https://github.com/stratum-mining/stratum/tree/main/roles/translator)
* [`roles/jd-client`](https://github.com/stratum-mining/stratum/tree/main/roles/jd-client)

PR #1739, opened 2025-06-07
Refactor CLI argument parsing with `clap`

### LINKED ISSUE #1703: Use `clap` for Managing CLI Arguments
Currently, most of our roles parse CLI arguments directly using `std::env::args()`, which adds unnecessary complexity and boilerplate.

[`clap`](https://crates.io/crates/clap) is a widely used crate that provides powerful abstractions for defining and managing CLI arguments. It allows argument parsing to be handled declaratively, often reducing the logic to a few macros.

This approach has already been introduced by @Fi3 in the implementation of `roles/test-utils/mining-device` via [PR #819](https://github.com/stratum-mining/stratum/pull/819).

We should standardize the use of `clap` across all roles for consistency and maintainability.

lucasbalieiro:
closes #1703 

This pull request refactors the CLI argument parsing across multiple modules to use the [`clap`](https://crates.io/crates/clap) crate. The goal is to improve **maintainability**, **consistency**, and **readability** by replacing custom argument parsing logic with `clap`’s declarative approach.

* Migrated from manual CLI parsing to `clap`-based parsing.
* Reduced redundancy and simplified argument handling across modules.
* Preserved existing command names and default values to ensure backward compatibility.

**Testing:**

* Verified functionality by following the instructions in the [Getting Started guide](https://stratumprotocol.org/getting-started/#ii-getting-started---running-all-roles).
* Confirmed that no CLI APIs were broken as a result of this refactor.


lucasbalieiro:
@Shourya742, do you think it makes sense to address #1704 in this PR, since it involves many of the same areas? Or would it be better to keep this focused and tackle that separately?

Also, do you prefer having each role refactor in a separate commit?

plebhash (approved):

PR #1740, opened 2025-06-09
update imports in `roles` to reflect library reorganization

### LINKED ISSUE #1458: need to standardize dependency (and API re-exporting) patterns

lucasbalieiro:
As discussed in #1738, this PR proposes an alternative approach to close #1458.

All roles were updated to follow the new import scheme proposed in #1738:
import scheme

---

**Note to Reviewers:**
I had to update the feature flags in the `Cargo.toml` files because some crates are no longer imported directly. As a result, their features need to be forwarded through intermediate crates until they reach their final destination.

Because of this changes in the feature flags, CI required me to major bump the version of `netwrok_helpers_sv2` to `4.0.0`


lucasbalieiro:
@Shourya742 , test module updated: [abfb149](https://github.com/stratum-mining/stratum/pull/1740/commits/abfb149efb8d8c8e0973320ad23e15304894dbad)

Also updated the examples `ping-pong` and `ping-pong-ecrypted`:  [6a54ea1](https://github.com/stratum-mining/stratum/pull/1740/commits/6a54ea1529052cbee7d2a1fb38cb00e7067ef7c9)

plebhash:
as discussed in the call today, here's an alternative path:

image

PR #1744, opened 2025-06-11
update imports in `roles` and `tests` to reflect library reorganization

### LINKED ISSUE #1458: need to standardize dependency (and API re-exporting) patterns

lucasbalieiro:
As discussed in #1740, this PR proposes an alternative solution to close #1458.

All roles have been updated to follow the new import scheme proposed in #1740:

image

---

### Notes for Reviewers

During the migration to the new import structure, I observed that `stratum-common` was previously used solely to re-export the `bitcoin` and `secp256k1` crates.

With the new structure, however, `roles-logic-sv2` is now a dependency of `stratum-common`. This introduces a circular dependency if `roles-logic-sv2` continues to access the `bitcoin` crate via `stratum-common`.

To break this cycle, I removed the `bitcoin` and `secp256k1` re-exports from `stratum-common`. Any crates that previously relied on those re-exports now import the required crates directly.

---

### Version Bumps

* `stratum-common`: **2.0.0 → 3.0.0**
  Reason: Removed the `bitcoin` feature and introduced new features

* `network-helpers-sv2`: **3.1.0 → 4.0.0**
  Reason: Removed the `binary_Sv2` feature

### DIFF
--- a/protocols/v2/codec-sv2/examples/encrypted.rs
+++ b/protocols/v2/codec-sv2/examples/encrypted.rs
@@ -24,14 +24,16 @@ use codec_sv2::{
 };
 #[cfg(feature = "noise_sv2")]
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
+
+#[cfg(feature = "noise_sv2")]
+use noise_sv2::{ELLSWIFT_ENCODING_SIZE, INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE};
+
 use std::convert::TryInto;
 #[cfg(feature = "noise_sv2")]
 use std::{
     io::{Read, Write},
     net::{TcpListener, TcpStream},
 };
-#[cfg(feature = "noise_sv2")]
-use stratum_common::{ELLSWIFT_ENCODING_SIZE, INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE};
 
 // Arbitrary message type.
 // Supported Sv2 message types are listed in the [Sv2 Spec Message
--- a/roles/Cargo.toml
+++ b/roles/Cargo.toml
@@ -8,7 +8,8 @@ members = [
     "test-utils/mining-device-sv1",
     "translator",
     "jd-client",
-    "jd-server"
+    "jd-server",
+    "roles-utils/network-helpers"
 ]
 
 [profile.dev]
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -4,7 +4,7 @@ version = "1.0.1"
 edition = "2021"
 
 [dependencies]
-stratum-common = { path = "../common", features=["bitcoin"]}
+bitcoin = { version = "0.32.5", optional = true }
 async-std={version = "1.10.0", features = ["attributes"]}
 criterion = "0.5.1"
 async-channel = "1.4.0"
--- a/common/Cargo.lock
+++ b/common/Cargo.lock
@@ -3,10 +3,92 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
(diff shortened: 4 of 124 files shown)


lucasbalieiro:
> and maybe I would have re-exported bitcoin from roles_logic_sv2

Yeah, this looks like a better approach.

Since I’m rebasing this branch anyway, I’ll move `bitcoin` into `roles_logic_sv2` and feature-flag it via `stratum-common -> roles_logic_sv2`.

Would you be okay taking another look after that?

plebhash:
> @plebhash, after seeing all the heavy lifting we're doing to make this work, accessing Bitcoin through `roles-logic` and adding new features, I’m now more inclined to keep `network-helpers` and `roles-logic-sv2` as separate crates, added to the `roles` crate’s `Cargo.toml`. Even with this setup, we aren’t really avoiding the need to initialize separate codecs for both `roles-logic` and `network-helpers`. So, let’s just keep both as part of `roles`, import all protocol crates related components directly from `roles-logic-sv2`, and leave `network-helpers` as it is.

AFAIU that's what we are already doing here.

just to be clear, are you suggesting some kind of adjustments to this PR?

plebhash:
> we aren’t really avoiding the need to initialize separate codecs for both roles-logic and network-helpers.

I'm not sure I follow what you mean here.

in practice, `codec_sv2` is only really used on `network_helpers_sv2`

for `roles_logic_sv2`, `codec_sv2` is simply a re-exporting entry-point to the other crates

roles/Cargo.toml
```diff
@@ -8,7 +8,8 @@ members = [
     "test-utils/mining-device-sv1",
     "translator",
     "jd-client",
-    "jd-server"
+    "jd-server",
+    "roles-utils/network-helpers"
```
plebhash, line 12: why was this needed?

protocols/v2/codec-sv2/examples/encrypted.rs
plebhash, line 1: it seems this was leftover from #1728 

let's remove this change from this commit, and fix it under the scope of #1753

roles/Cargo.toml
lucasbalieiro: the rust compiler explicitly instructed me to take this action. To reproduce the issue, try running the following CI command on a nightly toolchain:

```bash
cargo fmt --all --manifest-path=benches/Cargo.toml -- --check
```

Make sure that `roles-utils/network-helpers` is **NOT** declared as a workspace member in the `Cargo.toml`. You'll encounter this error:

```bash
$ cargo fmt --all --manifest-path=benches/Cargo.toml -- --check
`cargo metadata` exited with an error: error: current package believes it's in a workspace when it's not:
current:   /home/lucasbalieiro/Projects/stratum/roles/roles-utils/network-helpers/Cargo.toml
workspace: /home/lucasbalieiro/Projects/stratum/roles/Cargo.toml

this may be fixable by adding `roles-utils/network-helpers` to the `workspace.members` array of the manifest located at: /home/lucasbalieiro/Projects/stratum/roles/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

protocols/v2/codec-sv2/examples/encrypted.rs
lucasbalieiro: Ok, force-pushed a rebase to castch-up with main and done this revert in the process.

roles/Cargo.toml
plebhash: ok it seems that's because before this PR, all crates under `roles` would directly list `network_helpers_sv2` as a dependency

that would include `network_helpers_sv2` as a member of the workspace

but now, it's imported into the crates of `roles` via `stratum-common`, which makes it no longer belong to any workspace, unless it's listed there

plebhash (approved):

protocols/v2/codec-sv2/examples/encrypted.rs
lucasbalieiro: I had to roll back due to CI issues when running the MSRV 1.75 Check.

However, I’ve separated this change into its own commit:
[c300c5](https://github.com/stratum-mining/stratum/pull/1744/commits/c300c5ddf7a612c6d3ee964381d56fd944616cdb)

Issue #1746, opened 2025-06-13
Update Vardiff


Shourya742:
Based on recent discussions, we have decided that Vardiff should no longer manage the target and hashrate. Both of these should be handled as part of the channel implementation. We need to update Vardiff accordingly and make the necessary accompanying changes

PR #1747, opened 2025-06-13
Update vardiff

Shourya742:
closes: #1746

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
+++ b/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
@@ -24,7 +24,6 @@ impl VardiffState {
     /// Creates a new `VardiffState` with the default minimum hashrate.
     ///
     /// # Arguments
-    /// * `shares_per_minute` - The target share submission rate.
     /// * `estimated_hashrate` - The initial hashrate estimate.
     pub fn new() -> Result<Self, VardiffError> {
         Self::new_with_min(DEFAULT_MIN_HASHRATE)
@@ -33,8 +32,6 @@ impl VardiffState {
     /// Creates a new `VardiffState` with a specific minimum hashrate.
     ///
     /// # Arguments
-    /// * `shares_per_minute` - The target share submission rate.
-    /// * `estimated_hashrate` - The initial hashrate estimate.
     /// * `min_allowed_hashrate` - The minimum hashrate to enforce.
     pub fn new_with_min(min_allowed_hashrate: f32) -> Result<Self, VardiffError> {
         let timestamp_secs = std::time::SystemTime::now()
@@ -93,7 +90,7 @@ impl Vardiff for VardiffState {
     /// deviation from the target rate is significant enough (based on internal,
     /// time-sensitive thresholds), it estimates a new hashrate and applies it.
     ///
-    /// It returns `Ok(Some(new_hashrate))` when an update occurs,
+    /// It returns `Ok(Some(new_hashrate, new_target))` when an update occurs,
     /// `Ok(None)` when conditions don't warrant an update, and
     /// `Err` for actual processing errors.
     fn try_vardiff(
--- a/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
+++ b/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
@@ -12,16 +12,12 @@ use super::{error::VardiffError, Vardiff};
 /// Tracks performance and adjusts the mining target to achieve a desired share rate.
 #[derive(Debug)]
 pub struct VardiffState {
-    /// Current estimated hashrate (H/s).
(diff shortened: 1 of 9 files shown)


protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -12,16 +12,12 @@ use super::{error::VardiffError, Vardiff};
 /// Tracks performance and adjusts the mining target to achieve a desired share rate.
 #[derive(Debug)]
 pub struct VardiffState {
-    /// Current estimated hashrate (H/s).
-    pub estimated_hashrate: f32,
     /// Target number of shares to be submitted per minute.
     pub shares_per_minute: f32,
```
plebhash: IIRC we also discussed about removing this?

bit-aloo: I though we discussed to persist this at both the places.

plebhash: isn't that a bit confusing?

what's the advantage of having this parameter living redundantly in two different places?

bit-aloo: Agree, Updated: ffd95e7

protocols/v2/roles-logic-sv2/src/vardiff/mod.rs
```diff
@@ -38,7 +26,12 @@ pub trait Vardiff: Debug + Send {
 
     /// Checks performance and potentially adjusts difficulty, returning the new
     /// hashrate if an update occurred.
-    fn try_vardiff(&mut self) -> Result<Option<f32>, VardiffError>;
+    fn try_vardiff(
```
plebhash, line 29: we should update the comments here to reflect the new return type

protocols/v2/roles-logic-sv2/src/vardiff/classic.rs
```diff
@@ -146,7 +96,12 @@ impl Vardiff for VardiffState {
     /// It returns `Ok(Some(new_hashrate))` when an update occurs,
     /// `Ok(None)` when conditions don't warrant an update, and
     /// `Err` for actual processing errors.
-    fn try_vardiff(&mut self) -> Result<Option<f32>, VardiffError> {
+    fn try_vardiff(
```
plebhash, line 96: we should update the comments here to reflect the new return type

bit-aloo: Ah!! missed it, thanks for pointing out.

bit-aloo: Updated: f0dbf00

plebhash: actually, I started to think a few steps down the line, when we use the return type from `try_vardiff` to actually change the channel state

we already have the `update_channel` method, which is called when the server receives an `UpdateChannel` message from the client

the `update_channel` method takes the following parameters:
- `nominal_hashrate`
- `requested_max_target`

this method internally calls `hash_rate_to_target`, so I feel it's redundant to return `Target` in `try_vardiff`

we can re-purpose the `update_channel` method so it's also used for updates coming from vardiff... we would only need to cache the  `requested_max_target` that came via `Open*MiningChannel` or  the last `UpdateChannel` into the channel state

please check d6a840d3802584828bc8f48cfdfece3507d4f303

there, I made `update_channel` take an `requested_max_target: Option<Target>`... so that when the server receives an `UpdateChannel` message, this method is called with `Some`

but when we just need to update the channel state because of vardiff, we call `update_channel` with `None`, and the internal logic takes the cached value... the application layer would send a `CloseChannel` in case of error, because that would violate the `requested_max_target`

but as stated above, I feel it's redundant for `try_vardiff` to return `Target`, because we would be calling `hash_rate_to_target` 2x

please check 635f723fbbb7bbc5e58f74a344f8fec363a0f6d0 where I reverted the return type of `try_vardiff`

let me know if you agree with this approach, and if yes feel free to cherry-pick these commits into the PR

bit-aloo: Agree, update_channel becomes an abstraction to update channel nominal hashrate and target based on vardiff value or downstream relevant messages.

bit-aloo: Updated

plebhash: should we also cherry-pick d6a840d3802584828bc8f48cfdfece3507d4f303 here?

that way channel APIs are already adapted

bit-aloo: done

PR #1750, opened 2025-06-18
Simplify network helpers

Shourya742:
This PR restructures the network_helpers noise_connection flow to make it simpler and more reliable. Previously, we were starting background tasks before completing the handshake, which made the flow confusing and likely caused ghost tasks. Now, the handshake is fully completed first, and only then are the read/write tasks spawned. We also removed the shared state wrapped in sync primitives instead, each stream gets its own cloned state, avoiding unnecessary locking since encryption and decryption are independent in Noise. Overall, this simplifies the architecture, hopefully eliminated zombie task, and makes the flow much easier to reason about.

### DIFF
--- a/roles/roles-utils/network-helpers/src/lib.rs
+++ b/roles/roles-utils/network-helpers/src/lib.rs
@@ -1,19 +1,10 @@
-use binary_sv2::{Deserialize, GetSize, Serialize};
 pub mod noise_connection;
 pub mod plain_connection;
 #[cfg(feature = "sv1")]
 pub mod sv1_connection;
 
-use async_channel::{Receiver, RecvError, SendError, Sender};
-use codec_sv2::{
-    noise_sv2::{ELLSWIFT_ENCODING_SIZE, INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE},
-    Error as CodecError, HandShakeFrame, HandshakeRole, StandardEitherFrame,
-};
-use futures::lock::Mutex;
-use std::{
-    convert::TryInto,
-    sync::{atomic::AtomicBool, Arc},
-};
+use async_channel::{RecvError, SendError};
+use codec_sv2::Error as CodecError;
 
 #[derive(Debug)]
 pub enum Error {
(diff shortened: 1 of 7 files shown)


roles/roles-utils/network-helpers/src/noise_connection.rs
```diff
@@ -44,138 +61,143 @@ impl Connection {
...
-                )
-                .await?
+        })
+    }
+
+    fn spawn_writer<'a, Message: Serialize + Deserialize<'a> + GetSize + Send + 'static>(
+        mut writer: OwnedWriteHalf,
+        mut write_state: State,
+        address: std::net::SocketAddr,
+        receiver_outgoing: Receiver<StandardEitherFrame<Message>>,
```
lucasbalieiro, line 187: Is there a reason to not do a `receiver_outgoing.close()` when you capture an error?

Similar to what you did with `sender_incoming` inside the `fn spawn_reader()`?

bit-aloo: Not really necessary, since once we exit the loop, the receiver is dropped, and the other places where it's used are ephemeral and get dropped immediately as well.

plebhash:
still unable to kill SRI Pool after connecting and disconnecting Sv2 CPU Miner

plebhash:
correction: I was testing with `regtest` and CPU miner connected to SRI Pool before it had received a `NewTemplate` and `SetNewPrevHash`

disconnecting the CPU miner in this state is what cause the inability to kill Pool

therefore not necessarily related to `network_helpers_sv2`

roles/roles-utils/network-helpers/src/lib.rs
```diff
@@ -41,81 +32,3 @@ impl<T> From<SendError<T>> for Error {
         Error::SendError
     }
 }
-
-trait SetState {
-    async fn set_state(self_: Arc<Mutex<Self>>, state: codec_sv2::State);
-}
```
plebhash, line 47: IIRC this was the reason the project was locked into 1.75 as MSRV, right?

PR #1752, opened 2025-06-19
Handle overlflow in u256

### LINKED ISSUE #1751: `hashrate_from_target` can overlfow and cause panics
While testing #1737 locally, I got the following panic:

```
2025-06-19T08:39:54.443532Z DEBUG pool_sv2::mining_pool: Starting vardiff updates for downstream: 1
2025-06-19T08:39:54.443559Z DEBUG vardiff: Hashrate update check triggered:
            - Elapsed time: 60s
            - Shares since last update: 4
            - Realized shares per minute: 4.0000
            - Current miner target: Target { head: 340282366920938463463374607431768211455, tail: 340282366920938463463374607431768211455 }
thread 'tokio-runtime-worker' panicked at /home/gitgab19/.cargo/registry/src/index.crates.io-6f17d22bba15001f/primitive-types-0.13.1/src/lib.rs:44:1:
arithmetic operation overflow
2025-06-19T08:40:04.444013Z DEBUG pool_sv2::mining_pool: Received downstream message type: 22, payload: [2, 0, 0, 0, 82, 202, 173, 75, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255]
2025-06-19T08:40:04.444049Z ERROR pool_sv2::mining_pool: Error: PoisonLock("poisoned lock: another task failed inside")
thread 'tokio-runtime-worker' panicked at pool/src/lib/mining_pool/mod.rs:407:17:
not yet implemented
```

The problem sits in the `hashrate_from_target` functions (defined [here](https://github.com/stratum-mining/stratum/blob/bff96e42651e8fe2bbc38f771af4d5bb3a1d4854/protocols/v2/roles-logic-sv2/src/utils.rs#L388)) which can produce an overflow which is not correctly managed.

Specifically, the overflow can be generated in the following line:
```
    let target_plus_one =
        U256Primitive::from_big_endian(target_arr.as_ref()) + U256Primitive::one();
```

If target is set to the maximum possible value (`U256::MAX`), adding one to it will overflow the 256-bit integer, causing a panic (as seen in the logs above). This panic is not caught by the function’s error handling, which only uses checked arithmetic for multiplication and division, but not for addition.

GitGab19:
Closes #1751

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/utils.rs
+++ b/protocols/v2/roles-logic-sv2/src/utils.rs
@@ -413,8 +413,9 @@ pub fn hash_rate_from_target(target: U256<'static>, share_per_min: f64) -> Resul
         return Err(Error::HashrateError(InputError::DivisionByZero));
     }
     let shares_occurrency_frequence = from_u128_to_u256(shares_occurrency_frequence);
-    let target_plus_one =
-        U256Primitive::from_big_endian(target_arr.as_ref()) + U256Primitive::one();
+    let target_plus_one = U256Primitive::from_big_endian(target_arr.as_ref())
+        .checked_add(U256Primitive::one())
+        .ok_or(Error::HashrateError(InputError::ArithmeticOverflow))?;
     let denominator = target_plus_one
         .checked_mul(shares_occurrency_frequence)
         .and_then(|e| e.checked_div(U256Primitive::from(100)))
--- a/protocols/v2/roles-logic-sv2/src/utils.rs
+++ b/protocols/v2/roles-logic-sv2/src/utils.rs
@@ -1194,4 +1194,26 @@ mod tests {
             max_difficulty
         );
     }
+
+    #[test]
+    fn test_hash_rate_from_target_with_max_target() {
+        use binary_sv2::U256;
+        // This is the maximum value for a 256-bit unsigned integer
+        let max_u128 = 340282366920938463463374607431768211455u128;
+        // Compose the bytes for U256::MAX
+        let mut max_bytes = [0u8; 32];
+        max_bytes[..16].copy_from_slice(&max_u128.to_be_bytes());
+        max_bytes[16..].copy_from_slice(&max_u128.to_be_bytes());
+        let target = U256::from(max_bytes);
+        let share_per_min = 4.0;
+        let result = hash_rate_from_target(target, share_per_min);
+        assert!(
+            matches!(
+                result,
+                Err(Error::HashrateError(InputError::ArithmeticOverflow))
+            ),
+            "Expected ArithmeticOverflow error, got: {:?}",
+            result
+        );
+    }
 }


protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -413,8 +413,9 @@ pub fn hash_rate_from_target(target: U256<'static>, share_per_min: f64) -> Resul
         return Err(Error::HashrateError(InputError::DivisionByZero));
     }
     let shares_occurrency_frequence = from_u128_to_u256(shares_occurrency_frequence);
-    let target_plus_one =
-        U256Primitive::from_big_endian(target_arr.as_ref()) + U256Primitive::one();
+    let target_plus_one = U256Primitive::from_big_endian(target_arr.as_ref())
+        .checked_add(U256Primitive::one())
+        .ok_or(Error::HashrateError(InputError::ArithmeticOverflow))?;
```
plebhash, line 414: I guess this approach requires some adaptations to vardiff, because in case this edge case happens again, the only difference now is that vardiff will have to deal with a `InputError::ArithmeticOverflow` instead of a panic

since this happens when target is `U256::MAX`, wouldn't it be better to simply not add one?

```
    let target_plus_one = if target == U256Primitive::MAX {
        // Use target directly when at maximum value to avoid overflow
        target
    } else {
        U256Primitive::from_big_endian(target_arr.as_ref()) + U256Primitive::one()
    };
```

since the formula is:
```
h = (2^256 - t) / (s * (t + 1))
```

we would still get a reasonably good estimation of hashrate, with the benefit of not having to treat `InputError::ArithmeticOverflow` on the vardiff algorithm

plebhash: actually, whenever target is max, `h` will always be ~0

```
h = (2^256 - (2^256 - 1)) / (s * ((2^256 - 1) + 1))
h = (2^256 - 2^256 + 1) / (s * 2^256)
h = 1 / (s * 2^256)
```

so an alternative approach here could also be:
```
    // Handle edge case where target is at maximum value
    if target == U256Primitive::MAX {
        // When target is maximum, any hash is valid, so required hashrate approaches 0
        return Ok(f64::MIN_POSITIVE); 
    }
    
    let target_plus_one = U256Primitive::from_big_endian(target_arr.as_ref()) + U256Primitive::one();
```

plebhash: when I wrote the comments above, I had not yet checked this fallback mechanism on vardiff:

https://github.com/stratum-mining/stratum/blob/bff96e42651e8fe2bbc38f771af4d5bb3a1d4854/protocols/v2/roles-logic-sv2/src/vardiff/classic.rs#L129-L141

so take the suggestions above as optional

plebhash (approved):

PR #1758, opened 2025-06-20
run Integration Tests via `cargo nextest`

plebhash:
benefits:

## automatically retry undeterministic tests

SRI has flaky integration tests, which we are ok to live with for now
but if a test fails more than 3 times, it's safe to assume it's failing deterministically
and that's a reliable indication that we shouldn't merge the PR

achieved via `retries = { backoff = "fixed", count = 3, delay = "2s" }`  on `.config/nextest.toml`

## automatically kill frozen tests

if a test runs for 60s, it's marked as `slow`
if it runs for 120s, it's killed

achieved via `slow-timeout = { period = "60s", terminate-after = 2 }` on `.config/nextest.toml`

## isolated test execution

no more logs of different tests mixed together, and most importantly, no mixing of CPU hashpower

achieved via `test-threads = 1` on `.config/nextest.toml`

## full CPU resource allocation for each test

each test now has all CPU cores available for it's execution

achieved via:
- `threads-required = "num-cpus"` on `.config/nextest.toml`
- changing all integration tests from `#[tokio::test]` to `#[tokio::test(flavor = "multi_thread")]`


plebhash:
hmm looks like 3 retries is not enough?

https://github.com/stratum-mining/stratum/actions/runs/15787922050/job/44508236501?pr=1758

```
TRY 4 TMT [ 120.010s] integration_tests_sv2::jd_provide_missing_transaction jds_ask_for_missing_transactions
```

plebhash:
ok I reverted the changes related to full CPU resource allocation for each test

namely:
- `threads-required = "num-cpus"` on `.config/nextest.toml`
- changing all integration tests from `#[tokio::test]` to `#[tokio::test(flavor = "multi_thread")]`

now, the flaky test was re-executed and the second time allowed it to pass

it seems there's something about how the tokio runtime distributes the workload across different threads

if we allow multiple threads for the runtime, the failures become deterministic

if we allow only one single thread for the runtime, the failures become non-deterministic and the retry strategy starts to work

plebhash:
> This looks awesome! I really liked the idea of having a test summary at the end. It also correctly identified the flaky test jd_tproxy_integration and provided timestamps showing how long each test took.

indeed!

thanks for suggesting `cargo nextest`, I wish we had introduced this earlier.

GitGab19 (approved):

PR #1760, opened 2025-06-23
introduce custom difficulty TP on Integration Tests

### LINKED ISSUE #1515: share == block on ITF
during this discussion https://github.com/stratum-mining/stratum/pull/1354#discussion_r1973802583 I realized something about our integration tests:

while most of the time, we want the CPU miner to simply submit a share, it turns out that regtest difficulty is so low (compared to the [difficulty internally established](https://github.com/stratum-mining/stratum/blob/883bfa59108cce119700606527507a859f08d000/roles/tests-integration/lib/mod.rs#L219) by `start_sv2_translator`) that we always end up finding an actual block (instead of just a share)

I was able to validate this by running `translate_sv1_to_sv2_successfully` and later checking the TP logs on `debug.log`

while this has not been a problem so far, I think it's still important to report it so we can start thinking about it, because finding an actual block has many (potentially undesired) implications on message flow

for example, finding a block implies:
- JDC sending `SubmitSolution` to JDS
- JDC sending `SubmitSolution` to TP
- trigger TP to send new `SetNewPrevHash`

sometimes we might not want these messages to happen

plebhash:
currently, Integration Tests always launch a Template Provider on `regtest` mode, which means every share is a block

as described in #1515, this is not always desired

sometimes, we want a share to be a share, and a block to be found only after some actual mining (or perhaps the test doesn't require any blocks to be mined, in which case it's better that it never happens)

this PR introduces `enum DifficultyLevel` to the APIs that launch the Template Provider on Integration tests

it has three levels available:
- `DifficultyLevel::Low`: TP in `regtest` mode (every share is a block)
- `DifficultyLevel::Mid`: TP in `signet` mode with genesis difficulty (most of the time, a CPU should find a block in a minute or less)
- `DifficultyLevel::High`: TP in `signet` mode with pre-mined blocks raising network difficulty to bitaxe level of 77761.11 (most of the time, a CPU should take a REALLY long time to find a block)

Note: `signet` mode has `-signetchallenge=51`, which means no signature is needed on the coinbase.

We also introduce the possibility of setting a custom timeout on `Sniffer` (which defaults to 60s), in case we need to wait for more than 60s.

close #1515


lucasbalieiro (approved):
This is a nice feature!

Seeing that `tar.gz` file gave me [`XZ Utils` backdoor](https://nvd.nist.gov/vuln/detail/cve-2024-3094) vibes 😂!

For future reference, this is the `shasum` of the file at this commit https://github.com/stratum-mining/stratum/pull/1760/commits/9795833807c733afa6c2bd756b50f73a9324ae48:
```bash
~/vm$ shasum -a 256 high_diff_chain.tar.gz
19c20e9c691d12f51e392218f6516b39f3370d323e14e325e2b22a9e2bb8011c  high_diff_chain.tar.gz
```

The contents appear consistent with expected file types::
```bash
~/vm/high_diff_chain/blocks$ file *
blk00000.dat: data
index:        directory
rev00000.dat: data
xor.dat:      data


~/vm/high_diff_chain/blocks/index$ file *
000005.ldb:      LevelDB table data
000255.ldb:      LevelDB table data
000257.ldb:      LevelDB table data
000260.ldb:      LevelDB table data
000261.log:      data
CURRENT:         ASCII text
LOCK:            empty
MANIFEST-000259: data


~/vm/high_diff_chain/chainstate$ file *
000289.log:      data
000290.ldb:      LevelDB table data
CURRENT:         ASCII text
LOCK:            empty
MANIFEST-000287: data
```
Also, done some testing locally and everything work as expected!

lucasbalieiro:
> if you can come up with a solution I can cherry-pick it into the PR

i think i have a proposal:

I’ve set up a demo repo to showcase the functionality: [test_binary_trigger](https://github.com/lucasbalieiro/test_binary_trigger).

Check the closed PRs to see the bot in action:

* The bot compares the commit to the main branch and checks for changes in binary files using [`git diff --numstat`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-code--numstatcode).
* If a binary file is added, edited, or deleted, the bot comments on the PR like this:
  image

If multiple commits or force pushes affect binary files, the bot updates the comment, avoiding duplicate comments.

Examples:

* **Updating a bin file:** [#2](https://github.com/lucasbalieiro/test_binary_trigger/pull/2)
* **Adding a non-bin file (no comment triggered):** [#3](https://github.com/lucasbalieiro/test_binary_trigger/pull/3)
* **Mixed changes (bin file added, non-bin file edited):** [#4](https://github.com/lucasbalieiro/test_binary_trigger/pull/4)
  The comment was edited instead of creating a new one: [#1 - comment edit](https://github.com/lucasbalieiro/test_binary_trigger/pull/1#issuecomment-2997816494).

### Implementation:

To implement this, just copy the code from this gist: [binary_trigger.yml](https://gist.github.com/lucasbalieiro/012a9d17ce7490234c03a880904b0ea3), and add a new file to `.github/workflows`.

### Alternatives I considered:

1. **Committing the bin hash and checking it with CI (against a `bin_hashes.txt` file):**
   This would work but requires manually updating the hash for each binary file, which can become cumbersome. An attacker could also modify the hash in the text file, bypassing detection.

2. **Using a third-party service (e.g., CDN or cloud storage) to store the bin file's hash:**
   This adds unnecessary complexity—bin files need to be uploaded, and hashes updated in the repo. Plus, an attacker could modify the hash in the third-party service.

### Why this approach:

* It’s self-contained, simple, and doesn’t rely on third-party services.
* No extra steps needed to upload binaries or update hashes.
* The bot JUST ALERTS the reviewer to binary file changes, leaving them to decide if the changes are valid.

This is the simplest solution I came up with, without adding unnecessary overhead.

plebhash:
@lucasbalieiro feels a bit overengineering, no?

overall, I'm still trying to understand the security implications here... if some PR changed the `.tar.gz`, people running integration tests would be running malicious code?

why not simply make the CI verify some checksum on this specific file?

plebhash:
> Why do we need a tar file, its of considerable size or 12.7 MB? to get the chainstate?

this chain has ~46k blocks

currently the decompressed sizes are:
- `blocks`: 26Mb
- `chainstate`: 3.9Mb

I guess we could simplify things and just avoid compressing them overall.

lucasbalieiro (approved):
I verified that the newly introduced parameters, such as the sniffer timeout and the `DifficultyLevel` added to the template provider, do not change the default behavior of existing tests.

The purpose of the PR is clearly defined and the code is well structured. Also it add tests to verify is the difficulty is accurately asserted

From my perspective, this PR is ready to be merged.

---

### Notes on Chainstate Additions

For future reference, here are the SHA256 checksums of the binary files added to the repository:

```
e7d6ce2c74091683c4240d89a26ea7bf79f18371f94564a22257819a880bcf82  ./blocks/rev00000.dat
973768f7261d1beb241e42ae822484bd28b8158fcaa46e6d4a13ce8b2dd209f5  ./blocks/xor.dat
712cded78455684a858a1ce2957dd4d73ad551a4be0de44437d021a0d61ef34e  ./blocks/blk00000.dat
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  ./blocks/.lock
ba5a07e9e5b30eb57a97db43e99cb2d1fffc544393bf93c609c6192f9c235f40  ./blocks/index/MANIFEST-000259
8e0aa525dc73725df59ef8be9e6a7f73b66e6728fd04a9ce1d65bc55308f3c78  ./blocks/index/000255.ldb
529d0e962d3c22d907fc52433f3a9e6d6146837536c9ad1214067750471164e2  ./blocks/index/000260.ldb
923a6842b44e988bcca3704a07c4402d12a6b4176600c22e2ac69e57495eae6f  ./blocks/index/CURRENT
26e0f5789f2daf6871a3ebe4ef0f34fc36d75a2af625a0110430d1ccba91599f  ./blocks/index/000261.log
9ff3e08c37178a75b558b97ee3d30d22629402a94bc55b0da228ef88b568470b  ./blocks/index/000257.ldb
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  ./blocks/index/LOCK
5355a61461e66ee56704c9e8079500b1426ba06cba9bce31e787c78d31ca920e  ./blocks/index/000005.ldb
5dc4ac7094fb55567a14768ba3ba123e605221ed519562414dfc9049d8655b8d  ./chainstate/000290.ldb
faa457e2645ad20fc7f9656205537d0aa3b880600c9459fefde567b09009940c  ./chainstate/000289.log
f10d3bfae67806d64c76681b90019371441121bf61ed169a83c7e17ed638a555  ./chainstate/CURRENT
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  ./chainstate/LOCK
b226852cbb38e943d95fab32ee55df0e1daf199bc8d744c400618a11f3618391  ./chainstate/MANIFEST-000287
```

---

### Chainstate Size Optimization Attempts

I made Several attempts  to reduce the size of the chainstate data added to the repository (in addition to compacting the folder):

* I attempted to isolate the earliest block introducing difficulty `77761.11239`, which allowed for removal of preceding blocks. However, this only saved a few kilobytes.
* I also tried trimming the data back to the earliest block with a difficulty of `1.00`, assuming it would still pose a reasonable challenge for CPU miners. This reduced the size by approximately 3 MB which is not significantly impactful.

---

I wrote a local test to verify the intended behavior. It confirmed that the CPU miner required more work to submit a solution at higher difficulties compared to the lowest setting, validating the difficulty logic in practice.

test/integration-tests/tests/jdc_block_propogation.rs
GitGab19, line 1: Can you fix the typo in the filename?

test/integration-tests/tests/jdc_block_propogation.rs
```diff
@@ -8,15 +9,20 @@ use stratum_common::roles_logic_sv2::{job_declaration_sv2::*, template_distribut
 #[tokio::test]
 async fn propogated_from_jdc_to_tp() {
```
GitGab19:
```suggestion
async fn propagated_from_jdc_to_tp() {
```

test/integration-tests/tests/jds_block_propogation.rs
```diff
@@ -8,15 +9,20 @@ use stratum_common::roles_logic_sv2::{job_declaration_sv2::*, template_distribut
 #[tokio::test]
 async fn propogated_from_jds_to_tp() {
```
GitGab19:
```suggestion
async fn propagated_from_jds_to_tp() {
```

We should also fix the typo in the filename.

PR #1761, opened 2025-06-24
Correct Coinbase Output Length Calculation in `Pool` Role Implementation

### LINKED ISSUE #1762: Wrong Coinbase Output Length Calculation in Pool Role Implementation
While working on a issue of another project: [#24](https://github.com/vinteumorg/pleblottery/issues/24), I encountered a discrepancy between the current implementation and the specification regarding how the `coinbase_output_max_additional_size` field should be calculated in the context of Coinbase transactions.

I was investigating how to implement the `CoinbaseOutputConstraint` message in the project. To get a better understanding of how this works, I referred to the SRI implementation and drew some inspiration from it.

https://github.com/stratum-mining/stratum/blob/6cecfd1ba22b7072f7520a87598d391652365765/roles/pool/src/lib/mod.rs#L74-L75

However, after reviewing the code, I realized that there seems to be a mistake in how the code is currently calculating the size of the coinbase outputs. The current implementation on the main branch calculates the size based on the length of the vector returned by `get_coinbase_output(&config)?;`, rather than the actual size of the `TxOut` elements themselves. This approach doesn't seem to match the specification and might lead to incorrect size calculations.

I confirmed this issue with @plebhash , who agreed that this is an incorrect approach at the time. According to the specification:

> *The maximum additional serialized bytes which the pool will add in coinbase transaction outputs.*

This field (`coinbase_output_max_additional_size`) should reflect the actual serialized size of the transaction outputs, not the length of the vector containing them.

lucasbalieiro:
closes #1762 

This pull request addresses an issue with the coinbase output length calculation in the `Pool` role.

### Problem:

The code was previously using the length of the `Vec<TxOut>` (number of outputs) to estimate the coinbase output size:

```rust
let coinbase_output_len = coinbase_output_result.len() as u32;
```

However, this approach only reflects the count of outputs, not their actual serialized size, which is critical for accurate calculation.

### Solution:

We now calculate the total serialized size by summing the sizes of individual `TxOut` elements. This aligns with the field’s intended purpose:

> *The maximum additional serialized bytes which the pool will add in coinbase transaction outputs.*

### Logs:

Here are the results with and without the fix:

**With the fix:**
image

**Without the fix:**
image

### DIFF
--- a/roles/pool/src/lib/mod.rs
+++ b/roles/pool/src/lib/mod.rs
@@ -72,7 +72,10 @@ impl PoolSv2 {
 
         // Prepare coinbase output information required by TemplateRx.
         let coinbase_output_result = get_coinbase_output(&config)?;
-        let coinbase_output_len = coinbase_output_result.len() as u32;
+        let coinbase_output_len = coinbase_output_result
+            .iter()
+            .map(|output| output.size() as u32)
+            .sum();
         let tp_authority_public_key = config.tp_authority_public_key().cloned();
         let coinbase_output_sigops = coinbase_output_result
             .iter()


plebhash (approved):
we found this while working on https://github.com/vinteumorg/pleblottery/issues/24

also correlates to https://github.com/stratum-mining/stratum/pull/1759, more specifically https://github.com/stratum-mining/stratum/pull/1759/commits/891e04e7fd434ef8eea8629d753c0a8c0ece5249

plebhash:
@lucasbalieiro can you create an issue describing the problem and link to this PR?

overall we try to create issues describing the problems as much as possible, just to make it easier to keep track of things

lucasbalieiro:
> @lucasbalieiro can you create an issue describing the problem and link to this PR?
> 
> overall we try to create issues describing the problems as much as possible, just to make it easier to keep track of things

Thanks for the heads up;

I opened the issue: #1762

GitGab19 (approved):

PR #1764, opened 2025-06-25
Refactor network helpers pass 2

Shourya742:
We're still encountering a non-deterministic error on SRI (thanks to @plebhash and @lucasbalieiro for pointing that out), which continues to stem from the network_helpers. My initial suspicion that the issue was due to senders and receivers from network_helpers not being properly dropped during shutdown was only partially correct.

In reality, some structures still hold ownership of the channel endpoints from network_helpers, which prevents tasks from exiting cleanly. This leads to stale loops that persist even after a disconnection or a SIGTERM is received.

This PR explicitly closes all relevant channels on any downstream disconnection or upon receiving a SIGTERM, ensuring a full shutdown of the connection. While this should resolve the immediate issue, we also need to reassess the widespread use of 'static lifetimes throughout the codebase. These overly broad lifetimes are unintentionally prolonging resource ownership beyond necessity, contributing to memory retention and leaks like this.


plebhash:
ok looking deeper here now

I guess it's the same solution indeed... just making sure all channels are closed on every possible scenario to ensure no channels are left behind

@lucasbalieiro can you test this with your deterministic setup?

PR #1774, opened 2025-06-30
Refactor network helpers (Take 3)

Shourya742:
This PR combines the previously separate reader and writer tasks into a single tokio::select! block. It also adds a SIGTERM handling arm for graceful termination and introduces a writable_len method in the decoder to determine the required number of bytes to read from the socket buffer without altering the decoder's internal state. Makes reader cancellation safe on application layer, but stream is cancellation unsafe.


roles/roles-utils/network-helpers/src/noise_connection.rs
```diff
@@ -148,81 +155,88 @@ impl Connection {
             }
         };
 
-        debug!("Handshake completed with state: {:?}", state);
+        error!("Handshake completed with state: {:?}", state);
```
lucasbalieiro: should this really be an `error!`?

bit-aloo: Shoot, PR should still be in draft... This I added just to span these logs explicitly

PR #1775, opened 2025-06-30
add check on coinbase output script pubkey for custom job on Pool

### LINKED ISSUE #1776: pool needs to check for presence of expected script pubkey of custom job coinbase outputs
now that discussions around https://github.com/stratum-mining/sv2-spec/issues/133 have been solved, Pool needs to have a check for the presence of the expected script pubkey on the custom job coinbase outputs

plebhash:
close #1776

### DIFF
--- a/roles/pool/src/lib/mining_pool/message_handler.rs
+++ b/roles/pool/src/lib/mining_pool/message_handler.rs
@@ -11,7 +11,7 @@ use std::{
     sync::{Arc, RwLock},
 };
 use stratum_common::roles_logic_sv2::{
-    bitcoin::Amount,
+    bitcoin::{consensus::Decodable, transaction::TxOut, Amount},
     channels::server::{
         error::{ExtendedChannelError, StandardChannelError},
         extended::ExtendedChannel,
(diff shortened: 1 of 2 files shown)


GitGab19 (approved):

roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -881,10 +881,33 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
...
+                    request_id: m.request_id,
+                    channel_id: m.channel_id,
+                    error_code: "pool-payout-script-missing"
+                        .to_string()
+                        .try_into()
+                        .expect("error code must be valid string"),
+                };
+                return Ok(SendTo::Respond(Mining::SetCustomMiningJobError(error)));
+            }
+        }
```
bit-aloo:
```suggestion
let missing_script = self.empty_pool_coinbase_outputs.iter().find(|pool_output| {
    !custom_job_outputs
        .iter()
        .any(|custom_output| custom_output.script_pubkey == pool_output.script_pubkey)
});

if let Some(_) = missing_script {
    error!("SetCustomMiningJobError: pool-payout-script-missing");

    let error = SetCustomMiningJobError {
        request_id: m.request_id,
        channel_id: m.channel_id,
        error_code: "pool-payout-script-missing"
            .to_string()
            .try_into()
            .expect("error code must be valid string"),
    };

    return Ok(SendTo::Respond(Mining::SetCustomMiningJobError(error)));
}
```
We can use find + any same complexity, but more ergonomic.

PR #1772, opened 2025-06-27
fix coinbase deserialization

plebhash:
close #1771

Issue #1771, opened 2025-06-27
bad coinbase deserialization is breaking JD


plebhash:
on #1745 we changed the coinbase serialization/deserialization logic of JDC and JDS, as an adaptation to spec changes of https://github.com/stratum-mining/sv2-spec/pull/138

we started making heavy usage of `roles_logic_sv2::utils::deserialize_outputs`, which turns out to be buggy

this is leading to Pool rejecting shares from JDC, as each side was ending up with different serialized outputs for the custom job (which we're not catching on CI)

we should:
- stop using `roles_logic_sv2::utils::deserialize_outputs` in favor of `Vec::<TxOut>::consensus_decode` (or `TxOut::consensus_decode` for template outputs, see https://github.com/Sjors/bitcoin/issues/92)
- introduce an integration test that validates that we're able to do the full-cycle of JD mining

PR #1777, opened 2025-06-30
re-introduce dedicated function to deserialize outputs from `NewTemplate` message

plebhash:
as the outcome of https://github.com/Sjors/bitcoin/issues/92

---

to be merged after #1775

---

in https://github.com/stratum-mining/stratum/pull/1772 we realized that `deserialize_outputs` was introducing issues into the code

that's because this function can only deserialize outputs that were serialized into an opaque blob, not following Bitcoin consensus rules (setting a `VarInt` prefix as the size of the array, followed by the serialized bytes... or in other words, `rust-bitcoin` serializing a `Vec<TxOut>`)

this function was being used to deserialize outputs from `SetCustomMiningJob`, which are serialized under Bitcoin consensus rules (`Vec<TxOut>`)

`NewTemplate.coinbase_tx_outputs` however follows a different rule, where `NewTemplate.coinbase_tx_outputs_count` carries the size of the array, and `NewTemplate.coinbase_tx_outputs` simply has the raw bytes (no `VarInt` prefix to indicate the size of the array)

in other words, `NewTemplate.coinbase_tx_outputs` MUST NOT be deserialized via `Vec::<TxOut>::consensus_decode`, and it needs a dedicated function to do so

but `SetCustomMiningJob.coinbase_tx_outpus` MUST be deserialized via `Vec::<TxOut>::consensus_decode` (and this is what was causing problems on #1771)


protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -55,6 +57,48 @@ impl Default for Id {
...
+        // Move the cursor forward by the size of this TxOut
+        cursor += len;
+        deserialized_outputs.push(out);
+    }
+
+    if deserialized_outputs.len() != coinbase_tx_outputs_count as usize {
+        return Err(Error::FailedToDeserializeCoinbaseOutputs);
+    }
+
+    Ok(deserialized_outputs)
```
bit-aloo, line 99: Just curious, can we use a Cursor directly here, since consensus_decode already accepts any type that implements Read? This way, we wouldn't need to manually update the buffer position, as consensus_decode would advance the cursor on its own.

Something like this:

```rust
pub fn deserialize_template_outputs_1(
    serialized_outputs: Vec<u8>,
    coinbase_tx_outputs_count: u32,
) -> Result<Vec<TxOut>, Error> {
    let mut cursor = Cursor::new(serialized_outputs);
    let mut deserialized_outputs = Vec::new();

    while let Ok(out) = TxOut::consensus_decode(&mut cursor) {
        deserialized_outputs.push(out);
    }

    Ok(deserialized_outputs)
}
```

plebhash: sure, feel free to send a PR with this improvement

but please make sure this is not left out, because that's the reason `coinbase_tx_outputs_count` is used as an input parameter for this function:
```rust
    if deserialized_outputs.len() != coinbase_tx_outputs_count as usize {
        return Err(Error::FailedToDeserializeCoinbaseOutputs);
    }
```

PR #1785, opened 2025-07-03
Implements roles functionality to write logs to file and set log level

### LINKED ISSUE #677: Feature flag to write logs to files instead of just stdout
# Background
Currently, when a role is ran all logs are printed to `stdout`.

# Problem
There is no way to write the logs to a local file. It would be nice to have the option to write the logs to a local file so it would be easier to understand how each role is behaving, avoiding having to scroll back in history that is not persistent.

# Solution
Add a feature flag to give the user the option to write logs to 1) `stdout` and a local file and 2) only a local file (silent mode).

Consider using `-v` flag for verbose (print logs to `stdout`) and `-f <file-path>` flag to write to local file @ `file-path`.

lucasbalieiro:
closes #677

introduces improved logging capabilities across multiple components, providing more flexibility and control over log output. Key updates include:

* **Configurable Log Levels**
  Users can now set the verbosity of logs using levels such as `debug`, `info`, `warn`, and `error`.

* **Optional Log File Output**
  Logs can optionally be written to a file via the `-f` flag, in addition to standard output (`stdout`), supporting better log persistence and analysis.

* **Verbose Mode**
  When `-v` is used in combination with `-f`, logs are written to both `stdout` and the specified log file.

These enhancements do **not** affect the existing "Getting Started" tutorial or existing default behavior.

### Example Usage

```bash
cargo run -- --config config-examples/pool-config-local-tp-example.toml -v -f pool.log --log-level debug
```

This will:

* Write logs to both `stdout` and `pool.log`
* Set the log level to `DEBUG`

### Default Behavior

* **Log Level:**
  Defaults to `INFO` if no level is specified or if an invalid level is provided.

* **Log Output:**
  If `-f` is not set, logs are printed only to `stdout`.

* **Verbose Mode (`-v`):**
  Requires `-f` to be set. When used, logs are written to both `stdout` and the specified file.

* **No Flags Provided:**
  If none of the new flags are used, behavior remains unchanged: logs are printed to `stdout` at `INFO` level, maintaining compatibility with existing usage and tutorials.


lucasbalieiro:
force-pushed the `-v` flag specified in the issue. Now I think the issue can be closed if this PR is merged

roles/jd-client/src/logging.rs
```diff
@@ -0,0 +1,52 @@
+use std::{fs::OpenOptions, io, path::PathBuf, str::FromStr};
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{fmt, prelude::*, EnvFilter, Registry};
+
+/// Initialize logging to stdout and optionally to a file.
```
bit-aloo: Because the setup is repetitive, we should move init_logging into a shared common/utils crate that every role can import. I still need to review the init_logging function in more detail.

plebhash: [`config-helpers`](https://github.com/stratum-mining/stratum/tree/main/roles/roles-utils/config-helpers) could be a destination for this

lucasbalieiro: force-pushed
moved `logging.rs` to [`config_helpers`](https://github.com/stratum-mining/stratum/tree/main/roles/roles-utils/config-helpers). Update all roles accordingly

lucasbalieiro: > Maybe we can have log dir defined in config.toml files as well.

@Shourya742 I'm a bit unsure about this approach.

What would be the default behavior if the log directory is specified in config.toml?
Would the app still start with only stdout logging unless the -f flag is explicitly passed?
Or would it automatically start logging to the file specified in config.toml, even without -f?

bit-aloo: Logging to stdout should be the default behavior. The log directory should be an optional field in the configuration. If specified, logs should be written to that directory; otherwise, stdout should be used. Currently, if we run the application with the `-f` flag, it blocks the terminal and no logs are visible which results in a poor user experience. Ideally, we should also support a `--detach` mode in the future to run the process in the background. The `--verbose` flag also feels ambiguous it appears to route logs to all data sinks, but the behavior isn’t very clear.

Here’s how I think the logging architecture should behave:

1. Stdout logging should be the default unless the user explicitly detaches from the process.
2. A log file path option should be available in the config file(optional), with command-line arguments taking precedence.
3. I'm unsure how we want to define the verbose flag whether it should just enable multiple output sinks or something else.
4. In Rust, we have `RUST_LOG` for specifying log levels, and most logging implementations respect that by default. Still, having a `--log-level` CLI flag (with a corresponding config field, defaulting to `info`) is also useful.
5. To improve the logging infrastructure, we can later add spans to key modules. This would allow integration with observability tools like OpenTelemetry for deeper visibility into system behavior.

Let me know your thoughts on this.

lucasbalieiro: > Let me know your thoughts on this.

I'm generally indifferent in these cases, but considering the solution proposed in the issue and your approach, yours seems better.

I've force-pushed a few modifications:

* Removed the `--verbose` flag that was proposed by the issue.
* Added the `log_file` parameter to the `roles` config files. It’s commented out (optional) by default and includes explanatory notes:

  ```toml
  # Enable this option to set a predefined log file path.
  # When enabled, logs will always be written to this file.
  # The CLI option --log-file (or -f) will override this setting if provided.
  # log_file = "./mining-proxy.log"
  ```
* Kept the `-f` flag: if the config file already specifies a log file, the `-f` flag will override it. 
* Log levels are now controlled through the `RUST_LOG` environment variable. If it’s not set, it defaults to `INFO`.

roles/pool/src/args.rs
```diff
@@ -18,16 +18,27 @@ pub struct Args {
...
-    let config: PoolConfig = Config::builder()
+    let mut config: PoolConfig = Config::builder()
         .add_source(File::new(config_path, FileFormat::Toml))
         .build()
         .and_then(|settings| settings.try_deserialize::<PoolConfig>())
         .expect("Failed to load or deserialize config");
+
+    if let Some(log_file) = &args.log_file {
+        config.set_log_dir(Some(log_file.clone()));
+    }
```
bit-aloo:
```suggestion
config.set_log_dir(args.log_file);
```

roles/translator/src/args.rs
```diff
@@ -41,6 +47,11 @@ pub fn process_cli_args<'a>() -> ProxyResult<'a, TranslatorConfig> {
         .build()?;
 
     // Deserialize settings into TranslatorConfig
-    let config = settings.try_deserialize::<TranslatorConfig>()?;
+    let mut config = settings.try_deserialize::<TranslatorConfig>()?;
+
+    if let Some(log_file) = &args.log_file {
+        config.log_file = Some(log_file.clone());
+    }
```
bit-aloo:
```suggestion
config.set_log_dir(args.log_file);
```

roles/roles-utils/config-helpers/src/logging.rs
```diff
@@ -0,0 +1,41 @@
+use std::{fs::OpenOptions, io, path::PathBuf, str::FromStr};
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{fmt, prelude::*, EnvFilter, Registry};
+
+/// Initialize logging to stdout and optionally to a file.
+///
+/// If `log_file` is Some, logs will be written to both stdout and the file.
+/// If `log_level` is not provided or is invalid, it defaults to "info".
+pub fn init_logging(log_file: Option<&PathBuf>) {
```
bit-aloo:
```suggestion
pub fn init_logging(log_file: Option<&Path>) {
```

roles/roles-utils/config-helpers/src/logging.rs
```diff
@@ -0,0 +1,41 @@
...
+    let rust_log = std::env::var("RUST_LOG").unwrap_or_else(|_| "info".to_string());
+    let log_level_filter = LevelFilter::from_str(&rust_log).unwrap_or(LevelFilter::INFO);
+    let env_filter = EnvFilter::new(log_level_filter.to_string());
+
+    let subscriber: Box<dyn tracing::Subscriber + Send + Sync> = match log_file {
+        Some(path) => {
+            // Log to both file and stdout
+            let path = path.clone();
```
bit-aloo:
```suggestion
            let path = path.to_owned();
```

roles/pool/src/lib/config.rs
```diff
@@ -109,6 +113,15 @@ impl PoolConfig {
     pub fn set_tp_address(&mut self, tp_address: String) {
         self.tp_address = tp_address;
     }
+
+    /// Sets the log directory.
+    pub fn set_log_dir(&mut self, log_dir: Option<PathBuf>) {
+        self.log_file = log_dir;
+    }
+    /// Returns the log directory.
+    pub fn log_dir(&self) -> Option<&PathBuf> {
```
bit-aloo:
```suggestion
    pub fn log_dir(&self) -> Option<&Path> {
```

roles/pool/src/lib/config.rs
```diff
@@ -109,6 +113,15 @@ impl PoolConfig {
...
+        self.log_file.as_ref()
```
bit-aloo:
```suggestion
        self.log_file.as_deref()
```

roles/mining-proxy/src/lib/mod.rs
```diff
@@ -108,6 +108,7 @@ pub struct MiningProxyConfig {
     pub downstream_share_per_minute: f32,
     pub expected_total_downstream_hr: f32,
     pub reconnect: bool,
+    pub log_file: Option<PathBuf>,
```
bit-aloo:
```suggestion
    log_file: Option<PathBuf>,
```
as we have a getter

roles/translator/src/main.rs
```diff
@@ -1,24 +1,24 @@
...
 /// defined in `translator_sv2::TranslatorSv2`. Errors during startup are logged.
 #[tokio::main]
 async fn main() {
-    tracing_subscriber::fmt::init();
-
     let proxy_config = match process_cli_args() {
         Ok(p) => p,
         Err(e) => panic!("failed to load config: {e}"),
     };
+    init_logging(proxy_config.log_file.as_ref());
```
bit-aloo: lets use getter

roles/jd-client/src/args.rs
```diff
@@ -41,6 +47,11 @@ pub fn process_cli_args<'a>() -> ProxyResult<'a, JobDeclaratorClientConfig> {
         .build()?;
 
     // Deserialize settings into JobDeclaratorClientConfig
-    let config = settings.try_deserialize::<JobDeclaratorClientConfig>()?;
+    let mut config = settings.try_deserialize::<JobDeclaratorClientConfig>()?;
+
+    if let Some(log_file) = args.log_file {
+        config.set_log_file(log_file);
+    }
```
bit-aloo:
```suggestion
   config.set_log_file(args.log_file);
```

roles/jd-client/src/lib/config.rs
```diff
@@ -163,6 +166,13 @@ impl JobDeclaratorClientConfig {
             _ => Ok(result),
         }
     }
+
+    pub fn log_file(&self) -> Option<&PathBuf> {
```
bit-aloo:
```suggestion
    pub fn log_file(&self) -> Option<&Path> {
```

roles/jd-client/src/lib/config.rs
```diff
@@ -163,6 +166,13 @@ impl JobDeclaratorClientConfig {
...
+        self.log_file.as_ref()
```
bit-aloo:
```suggestion
        self.log_file.as_deref()
```

roles/jd-server/src/args.rs
```diff
@@ -46,11 +54,16 @@ pub fn process_cli_args() -> Result<JobDeclaratorServerConfig, JdsError> {
...
+    let mut config = settings
         .try_deserialize::<JobDeclaratorServerConfig>()
         .map_err(|e| {
             error!("Failed to deserialize config: {}", e);
             JdsError::BadCliArgs
         })?;
+
+    if let Some(log_file) = args.log_file {
+        config.set_log_file(log_file);
+    }
```
bit-aloo:
```suggestion
     config.set_log_file(args.log_file);
```

roles/jd-server/src/lib/config.rs
```diff
@@ -143,6 +145,12 @@ impl JobDeclaratorServerConfig {
             _ => Ok(result),
         }
     }
+    pub fn log_file(&self) -> Option<&PathBuf> {
```
bit-aloo:
```suggestion
    pub fn log_file(&self) -> Option<&Path> {
```

roles/jd-server/src/lib/config.rs
```diff
@@ -143,6 +145,12 @@ impl JobDeclaratorServerConfig {
...
+        self.log_file.as_ref()
```
bit-aloo:
```suggestion
        self.log_file.as_deref()
```

roles/mining-proxy/src/args.rs
```diff
@@ -42,11 +48,16 @@ pub fn process_cli_args() -> Result<MiningProxyConfig, Error> {
...
+    let mut config = settings
         .try_deserialize::<MiningProxyConfig>()
         .map_err(|e| {
             error!("Failed to deserialize config: {}", e);
             Error::BadCliArgs
         })?;
+
+    if let Some(log_file) = args.log_file {
+        config.log_file = Some(log_file);
+    }
```
bit-aloo:
```suggestion
        config.log_file(log_file);
```

Issue #1780, opened 2025-07-02
Refactor network-utils `Connection` type to return Reader/Writer directly instead of returning channels


coleFD:
The network-utils crate is very useful for developers, but currently it forces the implementation to spin up 2 async tasks and rely on async channels to communicate with client. This requires 2 sequential async awaits for every Tcp read and write which is not very performant. Instead, the crate should return the a type similar to a typical tcp stream type (one that implements the `AsyncRead` and `AsyncWrite`). In addition, the type returned should implement `into_split()` (see https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.into_split), so that the user has the option to separate the reader and writer into different tasks if warranted.

PR #1792, opened 2025-07-08
add noise stream

Shourya742:
closes: #1780

This PR introduces `NoiseTcpStream`, wrapper around a `TcpStream` that performs a noise handshake and enables encrypted, frame-based communication over TCP .

The design includes:

* A one-time Noise handshake as either `Initiator` or `Responder`
* Framed reading and writing of `StandardEitherFrame<Message>` types
* Explicit read and write halves (`NoiseTcpReadHalf`, `NoiseTcpWriteHalf`)
* Support for both blocking and non-blocking (`try_*`) frame I/O

The `NoiseTcpStream` is not a byte stream in the traditional sense. It encapsulates:

* A stateful encryption/decryption pipeline via the Noise protocol
* Frame reassembly logic required by the SV2 codec
* Buffers and internal state tightly coupled to message boundaries

Therefore, we expose only high-level, frame-based I/O methods:

* `read_frame()` and `write_frame()` (async)
* `try_read_frame()` and `try_write_frame()` (non-blocking)

### DIFF
--- a/protocols/v2/codec-sv2/src/decoder.rs
+++ b/protocols/v2/codec-sv2/src/decoder.rs
@@ -184,7 +184,15 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
         }
     }
 
-    /// Provides with the amount of bytes to read.
+    /// Returns the number of bytes expected in the next read operation.
+    ///
+    /// This value indicates how many more bytes are required to complete the
+    /// current Noise-encrypted frame. It is used to determine the exact size
+    /// of the writable buffer that should be passed to the underlying stream
+    /// during reading.
+    ///
+    /// The returned length dynamically updates as data is received and processed,
+    /// and ensures that we only read as much as needed to complete the frame.
     pub fn writable_len(&self) -> usize {
         self.missing_noise_b
     }
--- a/protocols/v2/codec-sv2/src/decoder.rs
+++ b/protocols/v2/codec-sv2/src/decoder.rs
@@ -184,6 +184,11 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
         }
     }
 
+    /// Provides with the amount of bytes to read.
+    pub fn writable_len(&self) -> usize {
+        self.missing_noise_b
+    }
+
     /// Provides a writable buffer for receiving incoming Noise-encrypted Sv2 data.
     ///
     /// This buffer is used to store incoming data, and its size is adjusted based on the number
--- a/roles/roles-utils/network-helpers/src/lib.rs
+++ b/roles/roles-utils/network-helpers/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod noise_connection;
+pub mod noise_stream;
 pub mod plain_connection;
 #[cfg(feature = "sv1")]
 pub mod sv1_connection;
--- a/roles/roles-utils/network-helpers/src/noise_connection.rs
+++ b/roles/roles-utils/network-helpers/src/noise_connection.rs
@@ -1,20 +1,15 @@
 #![allow(clippy::new_ret_no_self)]
-use crate::Error;
+use crate::{
+    noise_stream::{NoiseTcpReadHalf, NoiseTcpStream, NoiseTcpWriteHalf},
+    Error,
+};
 use async_channel::{unbounded, Receiver, Sender};
 use codec_sv2::{
     binary_sv2::{Deserialize, GetSize, Serialize},
-    noise_sv2::{ELLSWIFT_ENCODING_SIZE, INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE},
-    HandShakeFrame, HandshakeRole, StandardEitherFrame, StandardNoiseDecoder, State,
+    HandshakeRole, StandardEitherFrame,
 };
-use std::{convert::TryInto, sync::Arc};
-use tokio::{
-    io::{AsyncReadExt, AsyncWriteExt},
-    net::{
-        tcp::{OwnedReadHalf, OwnedWriteHalf},
-        TcpStream,
-    },
-    task,
-};
+use std::sync::Arc;
+use tokio::{net::TcpStream, task};
 use tracing::{debug, error};
 
 pub struct Connection;
@@ -35,35 +30,8 @@ impl<Message> ConnectionState<Message> {
     }
 }
 
-async fn send_message<'a, Message: Serialize + Deserialize<'a> + GetSize + Send + 'static>(
-    writer: &mut OwnedWriteHalf,
-    msg: StandardEitherFrame<Message>,
-    state: &mut State,
-    encoder: &mut codec_sv2::NoiseEncoder<Message>,
-) -> Result<(), Error> {
-    let buffer = encoder.encode(msg, state)?;
-    writer
-        .write_all(buffer.as_ref())
-        .await
-        .map_err(|_| Error::SocketClosed)?;
-    Ok(())
-}
-
-async fn receive_message<'a, Message: Serialize + Deserialize<'a> + GetSize + Send + 'static>(
-    reader: &mut OwnedReadHalf,
-    state: &mut State,
-    decoder: &mut StandardNoiseDecoder<Message>,
-) -> Result<StandardEitherFrame<Message>, Error> {
-    let writable = decoder.writable();
-    reader
-        .read_exact(writable)
-        .await
-        .map_err(|_| Error::SocketClosed)?;
-    decoder.next_frame(state).map_err(Error::CodecError)
-}
-
 impl Connection {
-    pub async fn new<'a, Message: Serialize + Deserialize<'a> + GetSize + Send + 'static>(
+    pub async fn new<Message>(
         stream: TcpStream,
         role: HandshakeRole,
     ) -> Result<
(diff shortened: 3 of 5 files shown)


roles/roles-utils/network-helpers/src/noise_stream.rs
```diff
@@ -0,0 +1,186 @@
...
+            },
+        })
+    }
+}
+
+impl<Message> NoiseTcpWriteHalf<Message>
+where
+    Message: Serialize + for<'a> Deserialize<'a> + GetSize + Send + 'static,
+{
+    pub async fn write_frame(&mut self, frame: StandardEitherFrame<Message>) -> Result<(), Error> {
```
coleFD, line 175: We should add a doc comment here that this is not cancellation safe as tokio does for `.write_all`

roles/roles-utils/network-helpers/src/noise_stream.rs
```diff
@@ -0,0 +1,186 @@
...
+            .map_err(|_| Error::SocketClosed)?;
+        Ok(())
+    }
+}
+
+impl<Message> NoiseTcpReadHalf<Message>
+where
+    Message: Serialize + for<'a> Deserialize<'a> + GetSize + Send + 'static,
+{
+    pub async fn read_frame(&mut self) -> Result<StandardEitherFrame<Message>, Error> {
```
coleFD, line 222: this is also not cancellation safe, but because `.writable()` returns a mutable buffer maintained by `decoder` as long as `.next_frame()` is called, those partially written bytes are accounted for in subsequent attempts. Basically if this function gets cancelled with partially written bytes, the next call to `.writable()` will return the same buffer size as the previous failed attempt, which is incorrect because some bytes were actually written. The `missing_bytes` estimate only updates on calls to `.next_frame()`, so I think it could be cancellation safe as long as `.next_frame` is before the initial call to `.writable()` since it would reset `missing_bytes` based on the partially written buffer that was cancelled in the previous read. I will test this out to confirm this doesnt break anything.

But if this function is not cancellation safe, it should never be used in a `tokio::select!`

roles/roles-utils/network-helpers/src/noise_stream.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl<Message> NoiseTcpWriteHalf<Message>
+where
+    Message: Serialize + for<'a> Deserialize<'a> + GetSize + Send + 'static,
+{
+    pub async fn write_frame(&mut self, frame: StandardEitherFrame<Message>) -> Result<(), Error> {
+        let buf = self.encoder.encode(frame, &mut self.state)?;
+        self.writer
+            .write_all(buf.as_ref())
+            .await
```
coleFD, line 179: can you add a sync function for nonblocking writes using `.try_write()`.

roles/roles-utils/network-helpers/src/noise_stream.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl<Message> NoiseTcpReadHalf<Message>
+where
+    Message: Serialize + for<'a> Deserialize<'a> + GetSize + Send + 'static,
+{
+    pub async fn read_frame(&mut self) -> Result<StandardEitherFrame<Message>, Error> {
+        let writable = self.decoder.writable();
+        self.reader
+            .read_exact(writable)
+            .await
```
coleFD: can you add a nonblocking equivalent of this function as well using `.try_read()`

roles/roles-utils/network-helpers/src/noise_stream.rs
```diff
@@ -0,0 +1,186 @@
...
+        Ok(())
+    }
+}
+
+impl<Message> NoiseTcpReadHalf<Message>
+where
+    Message: Serialize + for<'a> Deserialize<'a> + GetSize + Send + 'static,
+{
+    pub async fn read_frame(&mut self) -> Result<StandardEitherFrame<Message>, Error> {
+        let writable = self.decoder.writable();
```
coleFD: correct me if i am wrong but `.writable()` allocates a buffer of an estimated size. This means a single `.read_exact()` may not load enough bytes because the buffer may be too small. Subsequent calls to `.writable()` will improve the guess, so to be able to return a `Message` you will need to wrap the calls to `.writable()` and `.read_exact()` in a bounded loop that loops until it either successfully decodes a Message or until the bounds our met. Bounds are warranted to prevent any possible DDoS

My personal implementation that SEEMS to work looks something like this, but please let me know if i am misguided here
```
        // not sure if 3 is enough but we want to prevent stuck loops no matter what.
        // as long as `attempts` is at least the number of attempts required for the longest message, we are good
        let mut attempts = 3;
        while attempts > 0 {
            let buf = self.decoder.writable();
            self.io.read_exact(buf).await?;
            match self.decoder.next_frame(&mut self.state) {
                Ok(frame) => {
                    debug!("msg len: {}", frame.encoded_length());
                    return Ok(frame);
                }
                Err(codec_sv2::Error::MissingBytes(n)) => {
                    attempts -= 1;
                    debug!("bytes left: {}", n);
                    continue;
                }
                Err(e) => {
                    return Err(anyhow::anyhow!(
                        "Failed to decode first handshake message: {}",
                        e
                    ));
                }
            }
        }
        Err(anyhow::anyhow!("Failed to decode message"))
```

coleFD: edit: this does not include the cancellation safe method discussed in the previous comment but the final implementation should address it

roles/roles-utils/network-helpers/src/noise_stream.rs
bit-aloo: Yes, exactly. But you can have a reference to future inside the select block, to atleast evade the future drop.

roles/roles-utils/network-helpers/src/noise_stream.rs
bit-aloo: Agreed. In our previous network helpers implementation, we were looping over the buffer in a separate task, which caused issues with infinite loops that prevented the runtime from exiting. However, I’m not sure if hardcoding a value is the ideal solution. I’d prefer an approach where the reader half maintains an internal buffer, uses read to fill it, and once the buffer is full, flushes it, updates the decoder's state, and sends the message or an error.

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -184,6 +184,11 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
         }
     }
 
+    /// Provides with the amount of bytes to read.
+    pub fn writable_len(&self) -> usize {
+        self.missing_noise_b
+    }
+
```
GitGab19: We need to bump MINOR version number of `codec_sv2` for this.

I'll take note of this in the tracker issue #1801 once this is merged.

protocols/v2/codec-sv2/src/decoder.rs
```diff
@@ -184,6 +184,11 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
         }
     }
 
+    /// Provides with the amount of bytes to read.
```
GitGab19: Can you expand the description here with a bit more of context?

roles/roles-utils/network-helpers/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 pub mod noise_connection;
+pub mod noise_stream;
```
GitGab19, line 2: We need to bump MINOR version number of `network_helpers_sv2` for this.

I'll take note of this in the tracker issue #1801 once this is merged.

roles/roles-utils/network-helpers/src/noise_stream.rs
```diff
@@ -0,0 +1,330 @@
...
+                if n == 0 {
+                    return Err(Error::SocketClosed);
+                }
+
+                self.bytes_read += n;
+            }
+
+            self.decoder
+                .writable()
+                .copy_from_slice(&self.current_frame_buf[..]);
```
coleFD, line 247: is there a reason to resize the buffer instead of maintaining a constant buffer length and doing a `.copy_from_slice(&self.current_frame_buf[..expected]`? Resizing is more expensive

bit-aloo: There’s not much difference in practice resize usually doesn’t reallocate unless the expected size exceeds the current capacity. It simply updates the length and zero-fills the extra bytes, which adds negligible overhead compared to I/O. It's also simpler and avoids manual slice handling. We could have preallocated current_frame_buf with the maximum expected frame size, but in practice, the allocator stabilizes after a few calls if resizing is needed. I will initialize the buffer with a set capacity.

roles/roles-utils/network-helpers/src/noise_connection.rs
```diff
@@ -35,35 +30,8 @@ impl<Message> ConnectionState<Message> {
...
-    reader
-        .read_exact(writable)
-        .await
-        .map_err(|_| Error::SocketClosed)?;
-    decoder.next_frame(state).map_err(Error::CodecError)
-}
-
 impl Connection {
-    pub async fn new<'a, Message: Serialize + Deserialize<'a> + GetSize + Send + 'static>(
+    pub async fn new<Message>(
```
GitGab19, line 34: @Shourya742 wasn't this a breaking change? Since you changed the signature..

bit-aloo: Yup, in a way like to you need deserialized message to be static, and we have them as static.

PR #1795, opened 2025-07-08
isolate `channels_sv2` and `parsers_sv2` into standalone crates

plebhash:
this will provide improved modularity into UniFFI bindings


.github/workflows/release-libs.yaml
GitGab19, line 1: There's a conflict to be solved here, but I just want to recall you to add the new crates in the appropriate order:

- `channel-sv2` depends on `binary_sv2`, `common_messages_sv2`, `mining_sv2`, `template_distribution_sv2`, `job_declaration_sv2`
- `parsers-sv2` depends on `binary_sv2`, `framing_sv2`, `common_messages_sv2`, `mining_sv2`, `template_distribution_sv2`, `job_declaration_sv2`

protocols/v2/channels-sv2/src/merkle_root.rs
GitGab19, line 1: Why not using the ones defined in `utils.rs`?

protocols/v2/channels-sv2/src/target.rs
GitGab19, line 1: Why not using the ones defined it `utils.rs`?

protocols/v2/channels-sv2/src/template.rs
GitGab19, line 1: Why not using the one defined in `utils.rs`?

roles/translator/src/lib/status.rs
```diff
@@ -221,5 +221,6 @@ pub async fn handle_error(
         Error::Sv1MessageTooLong => {
             send_status(sender, e, error_handling::ErrorBranch::Break).await
         }
+        Error::Parser(_) => send_status(sender, e, error_handling::ErrorBranch::Continue).await,
```
GitGab19: Why here you used `error_handling::ErrorBranch::Continue` while in the Pool you used the `Break`?

GitGab19 (commented):
I left a couple of minor comments, but it LGTM.

protocols/v2/channels-sv2/src/merkle_root.rs
plebhash: I assume you're referring to `roles_logic_sv2::utils`

remember that this PR is isolating `roles_logic_sv2::channels` into a standalone crate

now, `roles_logic_sv2` depends on `channels_sv2` so it can re-export it

if we set `channels_sv2` to depend on `roles_logic_sv2`, that would create a circular dependency that the compiler wouldn't allow

---

moreover, after this PR, `roles_logic_sv2` will have the following modules (ignoring the general-purpose `utils`):
- `channel_logic` + `job_creator` (which are legacy and only still exist because JDC depends on it)
- `handlers`
- `vardiff`

so at this point, this makes me wonder if it doesn't make sense to eventually give `handlers` and `vardiff` their own dedicated crates and deprecate the "umbrella" `roles_logic_sv2`?

we would have to re-think the re-exporting strategy again, probably moving the re-exportings to `stratum-common/core`

anyways, that's a future discussion, but hopefully it clarifies why `channels_sv2` has the following modules that were taken from `roles_logic_sv2`:
- `merkle_root`
- `target` 
- `template`

protocols/v2/channels-sv2/src/target.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1795#discussion_r2198790608

protocols/v2/channels-sv2/src/template.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1795#discussion_r2198790608

roles/translator/src/lib/status.rs
plebhash: tbh I didn't think too much about this, I guess this was just an auto-complete that I overlooked

given that this error variant previously fell under `Error::RolesSv2Logic`, which leads to a `Break`, then indeed this should have been a `Break`

I'm rebasing accordingly

plebhash:
adding the new crates to `semver` CI is causing the following issue:

```
    Building channels_sv2 v0.1.0 (current)
       Built [   8.266s] (current)
     Parsing channels_sv2 v0.1.0 (current)
      Parsed [   0.019s] (current)
error: channels_sv2 not found in registry (crates.io). For workarounds check https://github.com/obi1kenobi/cargo-semver-checks#does-the-crate-im-checking-have-to-be-published-on-cratesio
```

I guess the only options are:
- postpone adding them to `semver` CI until they're published via `release-libs` CI
- manually publish them now, and keep track of potential version bumps on #1801

protocols/v2/channels-sv2/src/merkle_root.rs
GitGab19: Crystal clear.

Wrt having crates for `handlers` and `vardiff` I'm not against that for sure.

Let's keep discussing about it.

GitGab19 (approved):

Issue #1701, opened 2025-05-14
Reconsider `PoolChannelFactory` Initialization Design in JDC


Shourya742:
In JDC, the `PoolChannelFactory` is initialized within the `upstream` subsystem and then passed to the `downstream` subsystem after a channel is opened. A more logical design would involve initializing the `PoolChannelFactory` directly in the `downstream` subsystem, and then sending relevant channel update messages upstream as needed. This design should be reconsidered for better modularity and data flow.

plebhash:
I think this issue is somewhat outdated, as I assume eventually JDC is no longer going to rely on `PoolChannelFactory`

however, this is the closest subissue of #1674 for this comment 

as noted on https://github.com/stratum-mining/stratum/pull/1800#issuecomment-3076084703, `get_txout` should be deprecated, as it's a bit misleading to create a `TxOut` with 0 value to later replace it with template revenue

Issue #1674, opened 2025-05-14
Refactor: JDC


Shourya742:
This tracker captures the technical debt and design issues identified in the JDC role. The goal is to simplify, decouple, and improve configurability and ergonomics across the system.

plebhash:
#1773 showed a bug on TP (that's already been solved)

however, as pointed out [here](https://github.com/stratum-mining/stratum/issues/1773#issuecomment-3048585994), it doesn't make sense to crash the entire application just because it received a `RequestTransactionDataError `

lucasbalieiro:
> it seems that the introduction of logic for rejection of stale shares on Pool is not generating undesired triggers for fallback on JDC
> 
> [@lucasbalieiro](https://github.com/lucasbalieiro) can provide more context

Yes, I can confirm that, while testing things on the issue #1773 , this error is being triggered intermittently during prolonged test sessions.I haven’t had time to dive deeper, but I captured the following logs from the most recent run where the issue reproduced after several hours of operation:

#### JDC
The logs show that:

* A `stale-share` error is received,
* The JDC attempts to change Pool,
* Hours after, the Reader shuts down unexpectedly.


```txt
[2m2025-07-11T03:03:44.106600Z[0m [32m INFO[0m [2mjd_client::job_declarator::message_handler[0m[2m:[0m Received `DeclareMiningJobSuccess` with id 215
[2m2025-07-11T03:03:44.106832Z[0m [32m INFO[0m [2mjd_client::job_declarator::message_handler[0m[2m:[0m Received `AllocateMiningJobTokenSuccess` with id: 0
[2m2025-07-11T03:03:44.106851Z[0m [32m INFO[0m [2mjd_client::upstream_sv2::upstream[0m[2m:[0m Sending set custom mining job
[2m2025-07-11T03:03:44.107179Z[0m [32m INFO[0m [2mjd_client::upstream_sv2::upstream[0m[2m:[0m Received SetCustomMiningJobSuccess for channel id: 1 for job id: 429
[2m2025-07-11T03:03:44.108247Z[0m [33m WARN[0m [2mjd_client::upstream_sv2::upstream[0m[2m:[0m Extended job received from upstream, proxy ignore it, and use the one declared by JOB DECLARATOR
[2m2025-07-11T03:03:44.109411Z[0m [33m WARN[0m [2mjd_client::upstream_sv2::upstream[0m[2m:[0m SNPH received from upstream, proxy ignored it, and used the one declared by JDC
[...]

PR #1800, opened 2025-07-10
standardize rewards to pool/solo mining into a single `scriptPubKey`

### LINKED ISSUE #1793: standardize rewards to pool/solo mining into a single `scriptPubKey`
some apps under `roles` are parsing the script from the config file into a `Vec<CoinbaseOutput>`, while the output value remains undetermined (because template revenue can only be discovered at runtime)

on SRI we have basically fours scenarios where outputs are parsed from config files:
A. Pool for mining with JD
B. JDS for mining with JD
C. JDC for solo-mining as JD-fallback
D. Pool for mining without JD

for scenarios A and B, the Sv2 spec imposes that only one single output can be used for pool payout (because otherwise there would be no simple way to coordinate this payout over the JD message flow)

(we could potentially merge Pool and JDs someday, so scenarios A and B could become the same, although that's not too relevant here)

for scenarios C and D, maybe one could argue that some kind of arbitrary rule could be established to allow distributing the template revenue across more than one single output... however, that remains purely speculative, and also diverges a bit from a KISS philosophy

---

recently on @apoelstra 's contribution #1720, we started taking some steps towards simplifying things

now, the config files of Pool, JDS and JDC no longer use the `coinbase_outputs = [ ... ]` `toml` syntax that implies an array, which (on top of the descriptors, the main contribution of this PR), is a great improvement in terms of UX

however, in order to keep complexity manageable on this PR, we made some trade-offs, namely:
- as a transition strategy, we use a temporary `config_helpers::deserialize_vec_exactly_1` as the deserializer function, which will still output a `Vec<CoinbaseOutput>` (with a single element)
- we adapted some assertions against error messages to bypass the fact that `serde` is throwing meaningless error messages due to the `untagged` enum inside `config_helpers::deserialize_vec_exactly_1`

---

so the task here is:
- rename `config_helpers::CoinbaseOutput` into `config_helpers::CoinbaseRewardScript`  to make it clear that:
  - this doesn't really carry any amount yet (and therefore it's not really an output)
  - this is the script where the Pooled/Solo reward will end up going to
- adapt every place in Pool, JDC and JDS where they expect a `Vec<CoinbaseOutput>` so it expects a `CoinbaseRewardScript`

lucasbalieiro:
closes #1793

This refactor modifies the handling of coinbase outputs across multiple roles by replacing the `CoinbaseOutput` type with `CoinbaseRewardScript`. The change simplifies the configuration and enforces a single coinbase output script, eliminating the need for a vector of outputs.

Additionally, the `deserialize_vec_exactly_1` function is removed since it is no longer necessary.

Adjusted the usage of `Vec<TxOut>` to `TxOut` only to keep consistency with the `CoinbaseRewardScript` where possible.

In the config files for `Pool`, `JDS`, `JDC` the `coinbase_output` was renamed to `coinbase_reward_script`

deprecate `Pool::get_coinbase_output` and `JDS::get_txout`

### DIFF
--- a/roles/jd-server/src/lib/config.rs
+++ b/roles/jd-server/src/lib/config.rs
@@ -238,7 +238,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_txout_non_empty() {
+    fn test_get_non_empty_coinbase_reward_script() {
         let pk = TEST_PK_HEX
             .parse::<bitcoin::PublicKey>()
             .expect("Failed to parse public key");
@@ -263,7 +263,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_txout_empty() {
+    fn test_get_coinbase_reward_script_empty() {
         let error =
             load_coinbase_config_str("\"\"").expect_err("cannot parse config with empty txout");
         assert_eq!(
@@ -273,7 +273,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_txout_invalid_miniscript() {
+    fn test_get_invalid_miniscript_in_coinbase_reward_script() {
         let error = load_coinbase_config_str(&format!("\"INVALID\""))
             .expect_err("Cannot parse config with bad miniscript");
         assert_eq!(
@@ -283,7 +283,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_txout_invalid_value() {
+    fn test_get_invalid_value_in_coinbase_reward_script() {
         let error = load_coinbase_config_str(&format!("\"wpkh({TEST_INVALID_PK_HEX})\""))
             .expect_err("Cannot parse config with bad pubkeys");
         assert_eq!(


roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -79,8 +79,8 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
 
         // note: the fact that we're parsing a Vec<TxOut> from the config file is a bit of a hack
         // so while we don't clean that up, we only set the value of the first output
```
plebhash: this comment can be removed now

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -790,8 +790,8 @@ impl Pool {
                     // note: the fact that we're parsing a Vec<TxOut> from the config file is a bit
                     // of a hack so while we don't clean that up, we only set
                     // the value of the first output
```
plebhash: this comment can be removed now

lucasbalieiro:
@plebhash , thanks for the insights!

I think this is ready for review now!

plebhash:
> @plebhash , thanks for the insights!
> 
> I think this is ready for review now!

I guess the PR description is now outdated, which could be a bit misleading to the review process.

lucasbalieiro:
> I guess the PR description is now outdated, which could be a bit misleading to the review process.

Adjusted

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -1305,7 +1291,7 @@ mod test {
         let _coinbase_tx_value_remaining: u64 = 625000000;
         let _coinbase_tx_outputs_count = 0;
         let coinbase_tx_locktime = 0;
-        let coinbase_tx_outputs: Vec<bitcoin::TxOut> = super::get_coinbase_output(&config);
+        let coinbase_tx_output: bitcoin::TxOut = super::get_coinbase_output(&config);
```
bit-aloo: type can be removed here

lucasbalieiro:
rebased and addressed @Shourya742's comment

roles/pool/src/lib/config.rs
```diff
@@ -22,9 +22,7 @@ pub struct PoolConfig {
     authority_public_key: Secp256k1PublicKey,
     authority_secret_key: Secp256k1SecretKey,
     cert_validity_sec: u64,
-    #[serde(alias = "coinbase_output")] // only one is allowed, so don't make the user type the plural
-    #[serde(deserialize_with = "config_helpers::deserialize_vec_exactly_1")]
-    coinbase_outputs: Vec<CoinbaseOutput>,
+    coinbase_output: CoinbaseRewardScript,
```
plebhash: let's also rename `coinbase_output` into `coinbase_reward_script`

this should also be reflected on the `toml` files

lucasbalieiro: addressed in 3 new commits
https://github.com/stratum-mining/stratum/pull/1800/commits/940268f575c024e245782febfb5c3c7299a32533
https://github.com/stratum-mining/stratum/pull/1800/commits/33df8483a07042ab8db9b570910bed57a85bb53b
https://github.com/stratum-mining/stratum/pull/1800/commits/03dbd1110a31e39b2e49d667323e0607ff3975e0

test/integration-tests/lib/mod.rs
plebhash, line 1: let's rename the instances of `coinbase_output` in this file into `coinbase_reward_script`

plebhash:
while doing my final review here, I started feeling that `get_coinbase_output` (on Pool) and `get_txout` (on JDS and JDC) still feel quite redundant

so I went ahead and tried to get rid of them.

---

I was successful for Pool, so I'd encourage the following commits to be cherry-picked here:
- 8340b4d663772cb4f50ef016d6ab4f82ba33edcf
- 311196fb9c8d97f98df1e9cbbb9869953114aebb

---

for JDC that was not possible, mostly due to the fact that `PoolChannelFactory` expects `TxOut` with 0 amount, to be replaced internally... I'll take note on #1701, which feels slightly outdated but it's the best approximation sub-issue of #1674

---

for JDS it was possible, however JDS contains some unit tests that depend on `get_txout`

tbh, those unit tests are pretty generic and I feel they should belong to `config-helpers` rather than JDS

@lucasbalieiro can you try to migrate these unit tests from JDS to `config-helpers`?
- `test_get_txout_non_empty`
- `test_get_txout_empty`
- `test_get_txout_invalid_miniscript`
- `test_get_txout_invalid_value`

obviously, they should also be renamed, since strictly speaking, we don't really a need `get_txout` to cover for the functionality on those tests 

and then, once those tests are migrated, let's also get rid of `get_txout` in JDS in favor of parsing the `CoinbaseRewardScript` into a `TxOut` in the place where it's needed

test/integration-tests/lib/mod.rs
lucasbalieiro: addressed. instances renamed in https://github.com/stratum-mining/stratum/pull/1800/commits/34f6739f63e100cd9ff8b9ec86252c756a1547e4

plebhash:
@lucasbalieiro thanks for the deeper analysis 

I think the approach you took makes sense

plebhash (approved):

GitGab19 (approved):

PR #1806, opened 2025-07-15
Refactor handler

Shourya742:
This PR aims to relax the strict constraints around handler implementations in the `protocol` crate, particularly concerning synchronicity and return types.

Handlers are no longer required to take `Arc<Mutex<Self>>`; instead, they now operate with a mutable reference (`&mut self`) for state mutation. This change simplifies the design and improves ergonomics.

We are also moving away from the `SendTo_` abstraction in favor of returning `Option<Vec<Message>>`. The goal is to delegate message routing decisions to the upstream implementor rather than enforcing it within the library.

Despite these refactors, users still need to implement the relevant traits and invoke the appropriate `handle_*_message` methods.

We explored exposing a single `handle_message` method to unify the interface, but this approach would have forced implementors to implement all handler traits, which is not practical. You can find the associated code changes here: (https://github.com/Shourya742/stratum/pull/20)

### DIFF
--- a/protocols/v2/handlers-sv2/src/mining.rs
+++ b/protocols/v2/handlers-sv2/src/mining.rs
@@ -83,10 +83,13 @@ pub trait HandleMiningMessagesFromClientSync {
                 }
                 _ => Err(Error::UnexpectedMessage(MESSAGE_TYPE_SET_CUSTOM_MINING_JOB)),
             },
+            CloseChannel(m) => self.handle_close_channel(m),
 
             _ => Err(Error::UnexpectedMessage(0)),
         }
     }
+
+    fn handle_close_channel(&mut self, msg: CloseChannel) -> Result<(), Error>;
 
     fn handle_open_standard_mining_channel(
         &mut self,
@@ -178,11 +181,14 @@ pub trait HandleMiningMessagesFromClientAsync {
                     }
                     _ => Err(Error::UnexpectedMessage(MESSAGE_TYPE_SET_CUSTOM_MINING_JOB)),
                 },
+                CloseChannel(m) => self.handle_close_channel(m).await,
 
                 _ => Err(Error::UnexpectedMessage(0)),
             }
         }
     }
+
+    async fn handle_close_channel(&mut self, msg: CloseChannel) -> Result<(), Error>;
 
     async fn handle_open_standard_mining_channel(
         &mut self,
@@ -445,7 +451,6 @@ pub trait HandleMiningMessagesFromServerAsync {
                     }
                     _ => Err(Error::UnexpectedMessage(MESSAGE_TYPE_SET_GROUP_CHANNEL)),
                 },
-
                 _ => Err(Error::UnexpectedMessage(0)),
             }
         }


GitGab19:
@Shourya742 I think we should try to explore the possibility of sending messages inside the handlers, since it was the original request.

Maybe we could also consider to provide a double option, with one async trait which can be used to send messages directly there, and one trait which returns `Option<SameMessageReceived>` as we were discussing during yesterday's call.

plebhash:
> one trait which returns `Option<SameMessageReceived>` as we were discussing during yesterday's call.

maybe I'm missing something, but wouldn't the caller context already have `SameMessageReceived`?

if that's the case, what would be the value of returning the same message again, instead of a simple `Ok` that signals successful execution?

GitGab19:
@Shourya742 I think we should have:
- async --> sends messages internally and returns `Result<(), Error>`
- sync --> returns `Result<Vec<MessageToBeSent>, Error>`

As pointed out by @plebhash , the caller already has the message in its context (in our current status the caller has the frame), so it makes no sense to return the same message in the sync case.

I don't know if returning the Vec<MessageToBeSent> would work well for every case we have, especially in `NewTemplate` or `SetNewPrevHash` ones. We need to think a bit more about these cases I guess.

protocols/v2/handlers-sv2/src/job_declaration.rs
```diff
@@ -0,0 +1,117 @@
...
+    MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN, MESSAGE_TYPE_ALLOCATE_MINING_JOB_TOKEN_SUCCESS,
+    MESSAGE_TYPE_DECLARE_MINING_JOB, MESSAGE_TYPE_DECLARE_MINING_JOB_ERROR,
+    MESSAGE_TYPE_DECLARE_MINING_JOB_SUCCESS, MESSAGE_TYPE_PROVIDE_MISSING_TRANSACTIONS,
+    MESSAGE_TYPE_PROVIDE_MISSING_TRANSACTIONS_SUCCESS, MESSAGE_TYPE_PUSH_SOLUTION, *,
+};
+use parsers_sv2::JobDeclaration;
+
+#[trait_variant::make(ParseJobDeclarationMessagesFromUpstreamAsync: Send)]
+pub trait ParseJobDeclarationMessagesFromUpstreamSync {
+    fn handle_job_declaration_message(
```
plebhash: hmm I've never done this before, so I'm not speaking from experience

but this looks different from the example in the [blogpost](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/)?

```rust
#[trait_variant::make(HttpService: Send)]
pub trait LocalHttpService {
    async fn fetch(&self, url: Url) -> HtmlBody;
}

...

pub trait HttpService: Send {
    fn fetch(
        &self,
        url: Url,
    ) -> impl Future<Output = HtmlBody> + Send;
}
```

bit-aloo: Oopsie! I noticed this while implementing the new handlers for the new tproxy. It's been fixed now, I forgot to push earlier, sorry for inconvenience.

protocols/v2/handlers-sv2/src/template_distribution.rs
```diff
@@ -0,0 +1,213 @@
...
+            TemplateDistribution::RequestTransactionDataSuccess(_) => Err(
+                Error::UnexpectedMessage(MESSAGE_TYPE_REQUEST_TRANSACTION_DATA_SUCCESS),
+            ),
+            TemplateDistribution::RequestTransactionDataError(_) => Err(Error::UnexpectedMessage(
+                MESSAGE_TYPE_REQUEST_TRANSACTION_DATA_ERROR,
+            )),
+        }
+    }
+
+    fn handle_coinbase_out_data_size(
```
plebhash:
```suggestion
    fn handle_coinbase_output_constraints(
```

protocols/v2/handlers-sv2/src/template_distribution.rs
```diff
@@ -0,0 +1,213 @@
...
+                    Error::UnexpectedMessage(MESSAGE_TYPE_REQUEST_TRANSACTION_DATA_SUCCESS),
+                ),
+                TemplateDistribution::RequestTransactionDataError(_) => Err(
+                    Error::UnexpectedMessage(MESSAGE_TYPE_REQUEST_TRANSACTION_DATA_ERROR),
+                ),
+            }
+        }
+    }
+
+    async fn handle_coinbase_out_data_size(
```
plebhash:
```suggestion
    async fn handle_coinbase_output_constraints(
```

protocols/v2/handlers-sv2/src/error.rs
```diff
@@ -0,0 +1,16 @@
+use parsers_sv2::ParserError;
+
+#[derive(Debug)]
+pub enum HandlerError {
+    UnexpectedMessage(u8),
+    ParserError(ParserError),
+    OpenStandardMiningChannelError,
+    OpenExtendedMiningChannelError,
```
GitGab19: Why do we have these two errors here?

bit-aloo: Because in the Mining handler, we typically handle the channel opening state within the handler’s own routing method.

coleFD:
looks great!

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+
+#[derive(PartialEq, Eq)]
+pub enum SupportedChannelTypes {
+    Standard,
+    Extended,
+    Group,
+    GroupAndExtended,
+}
+
+pub trait ParseMiningMessagesFromDownstreamSync<Up: D>
```
GitGab19: Why did you put `<Up: D> here?

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+    fn handle_mining_message(&mut self, message_type: u8, payload: &mut [u8]) -> Result<(), Error> {
+        let parsed: Mining = (message_type, payload).try_into()?;
+        self.dispatch_mining_message(parsed)
+    }
+
+    fn dispatch_mining_message(&mut self, message: Mining) -> Result<(), Error> {
+        let (channel_type, work_selection) =
+            (self.get_channel_type(), self.is_work_selection_enabled());
+
+        use Mining::*;
```
GitGab19, line 39: What is this?

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+        match message {
+            OpenStandardMiningChannel(m) => {
+                if !self.is_downstream_authorized(&m.user_identity)? {
+                    // Add correct error type
+                    return Err(Error::OpenStandardMiningChannelError);
+                }
```
GitGab19: I feel this check imposes a specific direction about authorization mechanism to any implementer of this trait. I would probably remove it.

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+                    SupportedChannelTypes::Extended => Err(Error::UnexpectedMessage(
+                        MESSAGE_TYPE_OPEN_STANDARD_MINING_CHANNEL,
+                    )),
+                }
+            }
+            OpenExtendedMiningChannel(m) => {
+                if !self.is_downstream_authorized(&m.user_identity)? {
+                    // Add correct Error type
+                    return Err(Error::OpenExtendedMiningChannelError);
+                }
```
GitGab19: Same here

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+            (self.get_channel_type(), self.is_work_selection_enabled());
+
+        async move {
+            use Mining::*;
+            match message {
+                OpenStandardMiningChannel(m) => {
+                    if !self.is_downstream_authorized(&m.user_identity)? {
+                        // Add correct error type
+                        return Err(Error::OpenStandardMiningChannelError);
+                    }
```
GitGab19: Same here

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+                        SupportedChannelTypes::Extended => Err(Error::UnexpectedMessage(
+                            MESSAGE_TYPE_OPEN_STANDARD_MINING_CHANNEL,
+                        )),
+                    }
+                }
+                OpenExtendedMiningChannel(m) => {
+                    if !self.is_downstream_authorized(&m.user_identity)? {
+                        // Add correct Error type
+                        return Err(Error::OpenExtendedMiningChannelError);
+                    }
```
GitGab19: Same here

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    fn handle_submit_shares_standard(&mut self, msg: SubmitSharesStandard) -> Result<(), Error>;
+
+    fn handle_submit_shares_extended(&mut self, msg: SubmitSharesExtended) -> Result<(), Error>;
+
+    fn handle_set_custom_mining_job(&mut self, msg: SetCustomMiningJob) -> Result<(), Error>;
+}
+
+#[trait_variant::make(Send)]
+pub trait ParseMiningMessagesFromDownstreamAsync<Up: D>
```
GitGab19: Why did you put `<Up: D> here?

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    async fn handle_submit_shares_extended(
+        &mut self,
+        msg: SubmitSharesExtended,
+    ) -> Result<(), Error>;
+
+    async fn handle_set_custom_mining_job(&mut self, msg: SetCustomMiningJob) -> Result<(), Error>;
+}
+
+pub trait ParseMiningMessagesFromUpstreamSync<Down: D>
```
GitGab19: Why did you put `<Down: D> here?

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,547 @@
...
+        msg: SetCustomMiningJobError,
+    ) -> Result<(), Error>;
+
+    fn handle_set_target(&mut self, msg: SetTarget) -> Result<(), Error>;
+
+    fn handle_set_group_channel(&mut self, msg: SetGroupChannel) -> Result<(), Error>;
+}
+
+#[trait_variant::make(Send)]
+pub trait ParseMiningMessagesFromUpstreamAsync<Down: D>
```
GitGab19: Why did you put `<Down: D> here?

protocols/v2/handlers-sv2/src/lib.rs
GitGab19, line 1: I would probably change `Parse` with `Handle` (e.g. `HandleCommonMessagesFrom..`).

Also, I think we should always use `Client` and `Server` as terminology. A mix between client, downstream, server, and upstream could create some confusion to implementers of this crate imo.

protocols/v2/handlers-sv2/src/mining.rs
bit-aloo: We have this in the role_logic_sv2 handlers as well, but we're not really using any generics per se. So it might be okay to remove it.

protocols/v2/handlers-sv2/src/mining.rs
bit-aloo: Just importing Mining message variants.

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,503 @@
...
+
+                SetCustomMiningJob(m) => match (channel_type, work_selection) {
+                    (SupportedChannelTypes::Extended, true)
+                    | (SupportedChannelTypes::GroupAndExtended, true) => {
+                        self.handle_set_custom_mining_job(m).await
+                    }
+                    _ => Err(Error::UnexpectedMessage(MESSAGE_TYPE_SET_CUSTOM_MINING_JOB)),
+                },
+
+                _ => Err(Error::UnexpectedMessage(0)),
```
GitGab19, line 186: You are missing here `CloseChannel` which can be also sent by the client to the server: https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md#539-closechannel-client---server-server---client

protocols/v2/handlers-sv2/src/mining.rs
```diff
@@ -0,0 +1,503 @@
...
+
+            SetCustomMiningJob(m) => match (channel_type, work_selection) {
+                (SupportedChannelTypes::Extended, true)
+                | (SupportedChannelTypes::GroupAndExtended, true) => {
+                    self.handle_set_custom_mining_job(m)
+                }
+                _ => Err(Error::UnexpectedMessage(MESSAGE_TYPE_SET_CUSTOM_MINING_JOB)),
+            },
+
+            _ => Err(Error::UnexpectedMessage(0)),
```
GitGab19, line 88: You are missing here `CloseChannel` which can be also sent by the client to the server: https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md#539-closechannel-client---server-server---client

bit-aloo: Added here:  e0a5ff4

protocols/v2/handlers-sv2/src/mining.rs
bit-aloo: Added here: e0a5ff4

GitGab19 (approved):

PR #1808, opened 2025-07-21
ressurect `test/scale`

### LINKED ISSUE #1105: `test/scale` crate out of date
# Background
`stratum/test` contains three crates:
1. `config/`: Contains each role's toml config for the `message-generator` test stack.
2. `message-generator`: Generates messages to test SRI.
3. `scale`: Outputs the time spent sending 1,000,000 `SubmitSharesStandard` throughout the system. It contains a `main.rs` binary, which means it will generate a `Cargo.lock` when built. Thi

When working on enforcing no changes to `Cargo.lock` in #1039 (also related to #1044 and #1102), each crate was investigated to see which contains a `main.rs` to enforce the no change in lock files during the `pre-push` call to `scripts/clippy-fmt-and-test.sh`.

# Problem
It is unclear exactly how/when `scale` should be used. When running `cargo build` in `test/scale`, a versioning error on the `network_helpers_sv2` is encountered. This indicates to me that this crate is very out of date and is not really used.

```sh
cargo build
    Updating crates.io index
error: failed to select a version for the requirement `network_helpers_sv2 = "^0.1"`
candidate versions found which didn't match: 2.0.0
location searched: /Users/rachelrybarczyk/Development/StratumV2/Dev/stratum/roles/roles-utils/network-helpers
required by package `scale v1.0.0 (/Users/rachelrybarczyk/Development/StratumV2/Dev/stratum/test/scale)`
```

# Solution
Understand what `test/scale` is used for. If it is needed, update it so it runs properly. If it is not needed, remove it and all references to it.

lucasbalieiro:
Closes #1105;

- Updates the crate code to catch-up with the SRI changes
- Updates Clap to use a more recent version
- Added the crate to fmt and clippy CI process;

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ cobertura.xml
 **/template-provider
 stratum-message-generator
 *.log
+test/scale/Cargo.lock
--- a/.github/workflows/fmt.yaml
+++ b/.github/workflows/fmt.yaml
@@ -35,3 +35,4 @@ jobs:
           cargo fmt --all --manifest-path=roles/Cargo.toml -- --check
           cargo fmt --all --manifest-path=utils/Cargo.toml -- --check
           cargo fmt --all --manifest-path=test/integration-tests/Cargo.toml -- --check
+          cargo fmt --all --manifest-path=test/scale/Cargo.toml -- --check
--- a/scripts/rust/clippy.sh
+++ b/scripts/rust/clippy.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-workspaces=("benches" "common" "roles" "protocols" "utils" "test/integration-tests")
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -20,4 +20,5 @@ cobertura.xml
 /test/integration-tests/template-provider
 **/template-provider
 stratum-message-generator
-*.log
\ No newline at end of file
+*.log
+test/scale/Cargo.lock
```
GitGab19, line 24: shouldn't we commit the `Cargo.lock`? 

cc @plebhash @Shourya742

bit-aloo: Yes, for library crates we don't, but for binary we do.

lucasbalieiro:
As we discussed during the dev call: closing this in favor #1809

PR #1812, opened 2025-07-24
Rename `config-helpers` to `config_helpers_sv2` and remove of unused `roles-logic-sv2`

### LINKED ISSUE #1811: `config-helpers` is not published to crates.io
While working on #1799 , I noticed we never published `config-helpers` crate.

In order to have roles into a new repo, we need to get crates from crates.io.

Because of this every crate consumed by roles have to be published there.

While publishing it, I noticed that there's an old `roles-logic-sv2` import in its `Cargo.toml` which is not used at all. Also, to maintain coherence, crate name should be `config_helpers_sv2`.

GitGab19:
Closes #1811


GitGab19:
I'm testing it locally and it happens only if we remove `roles_logic_sv2 = { path = "../../../protocols/v2/roles-logic-sv2" }` from the `Cargo.toml`.

But at the same time there's no real usage of `roles-logic-sv2` there, that's why I was removing it.

plebhash:
make sure it's covered on [`release-libs.yaml`](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/release-libs.yaml)

GitGab19:
@plebhash thanks for your suggestions!

I updated those tests in https://github.com/stratum-mining/stratum/pull/1812/commits/6799aef54274b9e3461ba4f7eb897d8ce3158fcf.

plebhash (approved):

PR #1818, opened 2025-07-29
Update roles rust edition to 2021

### LINKED ISSUE #1666: Upgrade `rust` edition in Roles from `2018`
Should upgrade rust version across the different `Cargo.toml's`. Consider using `2024` or `2021`.
```
edition = "2018"
```

lucasbalieiro:
closes #1666 ;

Updates all roles crates to edition 2021 where needed;

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "stratum-common"
 version = "3.0.0"
-edition = "2018"
+edition = "2021"
 description = "SV2 pool role"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/stratum-mining/stratum"
--- a/protocols/v1/Cargo.toml
+++ b/protocols/v1/Cargo.toml
@@ -2,7 +2,7 @@
 name = "sv1_api"
 version = "2.0.0"
 authors = ["The Stratum V2 Developers"]
-edition = "2018"
+edition = "2021"
 readme = "README.md"
 description = "API for bridging SV1 miners to SV2 pools"
 documentation = "https://docs.rs/sv1_api"
--- a/protocols/v2/binary-sv2/Cargo.toml
+++ b/protocols/v2/binary-sv2/Cargo.toml
@@ -2,7 +2,7 @@
 name = "binary_sv2"
 version = "3.0.0"
 authors = ["The Stratum V2 Developers"]
-edition = "2018"
+edition = "2021"
 readme = "README.md"
 description = "Sv2 data format"
 documentation = "https://docs.rs/binary_sv2"
--- a/protocols/v2/binary-sv2/codec/Cargo.toml
+++ b/protocols/v2/binary-sv2/codec/Cargo.toml
@@ -2,7 +2,7 @@
 name = "binary_codec_sv2"
 version = "2.0.0"
 authors = ["The Stratum V2 Developers"]
-edition = "2018"
+edition = "2021"
 readme = "README.md"
 description = "Sv2 data format"
 documentation = "https://docs.rs/binary_codec_sv2"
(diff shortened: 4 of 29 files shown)


lucasbalieiro:
Side note:

My initial attempt was to update to edition 2024, which I successfully did. However, when running the CI locally, I was blocked by the MSRV, which doesn’t support `edition2024` yet. it was only stabilized in Rust 1.85.

https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024

plebhash:
> For the role crates, MSRV doesn’t really matter since they’re just applications built on top of our protocol layer.

if a dependency is attached to some MSRV, doesn't that mean the crate that depends on it is also bound to that MSRV?

in other words: what happens if I try to build `pool_sv2` with `rustc` 1.74.0?

lucasbalieiro:
> Thanks for this! Could you run `cargo fix --edition` once and check if the lock file changes? Also, it'd be great if you can take a quick look to see if any new features in the edition might actually help us. For the role crates, MSRV doesn’t really matter since they’re just applications built on top of our protocol layer. But for the protocol crates, we do need to stick to an MSRV mainly to guarantee that downstream users can compile them with any stable compiler starting from 1.75. That also means we shouldn't use any syntax or features that aren’t supported in 1.75.

@Shourya742, thanks for the review!

I’ve completed the migration using `cargo fix --edition`, and it didn’t require any changes to the lock file.

I also took a quick look at the changes in the 2021 edition and didn’t notice anything particularly useful for our case.
[Edition Guide - Rust 2021](https://doc.rust-lang.org/stable/edition-guide/rust-2021/index.html)

The primary reason for opening this PR is that I noticed some mismatched editions between roles, with some on 2018 and others on 2021.

This is more of a housekeeping thing.

GitGab19:
LGTM, but I'm curious to know: have you tried to bump also protocol crates' editions during your testing?

GitGab19:
Why not? 
For coherence among our entire codebase I think it would make sense to have 2021 edition also there.

lucasbalieiro:
> Why not? For coherence among our entire codebase I think it would make sense to have 2021 edition also there.

protocols edition bumped in: [upgrade protocols to rust edition 2021](https://github.com/stratum-mining/stratum/pull/1818/commits/177bce9b3b3460e9acd9ba4fd1e2072bb5f41ad5)

GitGab19:
We still have `stratum-common`, `integration_tests_sv2`, `buffer_sv2`, and `buffer-fuzz` with `2018` edition.

lucasbalieiro:
> We still have `stratum-common`, `integration_tests_sv2`, `buffer_sv2`, and `buffer-fuzz` with `2018` edition.

@GitGab19, the `buffer` ones are a bit tricky.

`buffer-fuzz` relies on a specific toolchain that doesn't support the 2021 edition, which in turn prevents `buffer_sv2` from being upgraded as well.


I added the commits for `stratum-common` and `integration-tests`: [a7db02b](https://github.com/stratum-mining/stratum/pull/1818/commits/a7db02b09e896a2d28586ff71ab39521a21d67f6) [feaf4c6](https://github.com/stratum-mining/stratum/pull/1818/commits/feaf4c69f6725c0d731cd2cafbdde1f8263c6e9e)

GitGab19:
Can you elaborate more on the toolchain? Which toolchain is used there?

lucasbalieiro:
> Can you elaborate more on the toolchain? Which toolchain is used there?

Apologies 😅, while writing up the explanation, I realized that the `buffer-fuzz` toolchain is actually using nightly, whereas I had been running `cargo fix --edition` with the stable toolchain.

```bash
$ cat utils/buffer/fuzz/rust-toolchain.toml 
[toolchain]
channel = "nightly-2020-04-06" #here is the toolchain for the buffer-fuzz
```

I've just updated the buffer crate accordingly.

GitGab19 (approved):

PR #1824, opened 2025-08-01
remove pool and miner tag from `extranonce_prefix`

### LINKED ISSUE #1823: do pool and miner tag on `coinbase_prefix` (not `extranonce_prefix`)
on #1537 we changed `mining_sv2::ExtendedExtranonce::new` such that we can embed `additional_coinbase_script_data` as part of `extranonce_prefix`

in practice, this was inteded to be used by pools and miners to write their own "tags" into the coinbase as part of the `extranonce_prefix`

prior to this change, we used to do this by placing such data inside `coinbase_prefix`, right after BIP34 and before `extranonce_prefix`

it's worth pointing out that `extranonce_prefix` is encoded as `B0_32` on:
- `OpenExtendedMiningChannel.Success`
- `OpenStandardMiningChannel.Success`
- `SetExtranoncePrefix`

which means now we only have 32 bytes to be used for:
- writing pool tag
- writing miner/JDC tag
- allocating unique search space across different channels
- rolling bytes (on extended channels)

some time after merging #1537, I started to realize the many undesired consequences of this, given that now the pool and miner tags are competing for space that was originally meant for search space allocation and rolling

but most importantly, we are also at risk of breaking compatibility with Sv1 devices, for which we have not done extensive testing after merging #1537

some issues that already serve as indications of undesired consequences on Sv1:
- https://github.com/stratum-mining/stratum/issues/1657
- https://github.com/jstefanop/Apollo-Miner-Binaries/issues/19

with that in mind, I feel #1537 was a bad idea, and we should revert those changes by:
- still allowing `mining_sv2::ExtendedExtranonce::new` to take a static prefix, but limited to 2 bytes max, and with the sole purpose of distinguishing between different mining servers (and not writing abritrary ASCII text to identify pool or miner)
- adapting channel constructors of `channels_sv2` so they take arbitrary strings to be placed inside `coinbase_prefix` of jobs, after BIP34 bytes

this has implications on JD, and the following scenarios will apply:
-   A: No JD (Pool-generated template), **only Pool tag**: this is pretty much identical to what already happens in Sv1. On block explorers, Pool can be identified by that, or payout address (which should probably take precedence).
-   B: JD (Miner-generated template), **only Miner tag**: On block explorers, Pool can only be inferred from payout address.
-   C: JD (Miner-generated template), **Pool + Miner tag**: this is similar to what Ocean is doing, but it will require a protocol extension (since canonical Sv2 JD message flow doesn't have a dedicated fields for this).

Image

plebhash:
close #1823

## `channels_sv2`

we introduce new constructors for:
- `channels_sv2::server::extended::ExtendedChannel`
- `channels_sv2::server::standard::StandardChannel`
- `channels_sv2::server::group::GroupChannel`

the `new` constructor variant is made private, in favor of the following public constructor variants:
- `new_for_pool`: takes a `pool_tag_string: String` to be written into the coinbase `scriptSig` for non-JD jobs
- `new_for_job_declaration_client`: takes a `pool_tag_string: Option<String>` and a `miner_tag_string: String` to be written into the coinbase `scriptSig` for JD jobs

providing `Some` as `pool_tag_string` into `new_for_job_declaration_client` will depend on a protocol extension that allows JDS to send this string via `AllocateMiningJobToken.Success` (there's no dedicated field for this on the canonical JD spec)

`JobFactory` takes these strings and adds delimiters for easy distinction on block explorers. The delimiters are:

`/pool_tag/miner_tag/`

in case `None` is provided (or an empty string), the delimiters are still written, therefore the following combinations are possible:
- `///`
- `/pool_tag//` e.g.: non-JD jobs (assuming non-empty `pool_tag`)
- `//miner_tag/` e.g.: JD jobs without protocol extension (assuming non-empty `miner_tag`)
- `/pool_tag/miner_tag/` e.g.: JD jobs with protocol extension (assuming non-empty `pool_tag` and `miner_tag`)

moreover, now we write tags and extranonce into scriptSig via `OP_PUSHBYTES_*` opcodes, as opposed to simply writing the raw data

## `mining_sv2`

we still allow arbitrary data into the `ExtendedExtranonce::new` constructor, but now it's called `static_prefix` and it's limited to 2 bytes max

this is inteded to allow unique `extranonce_prefix` allocation for different mining servers, and no longer meant for writing `pool_tag` or `miner_tag`

### DIFF
--- a/roles/pool/src/lib/mining_pool/mod.rs
+++ b/roles/pool/src/lib/mining_pool/mod.rs
@@ -937,6 +937,8 @@ impl Pool {
         shares_per_minute: f32,
         recv_stop_signal: tokio::sync::watch::Receiver<()>,
     ) -> Result<Arc<Mutex<Self>>, PoolError> {
+        // range_1 is used for dynamically allocating extranonce_prefix across different channels
+        // from these 8 bytes, the first 2 bytes are statically defined by static_prefix
         let range_1_start = 0;
         let range_1_end = 8;
 


protocols/v2/subprotocols/mining/src/lib.rs
```diff
@@ -513,9 +513,9 @@ impl ExtendedExtranonce {
             return Err(ExtendedExtranonceError::InvalidRanges);
         }
 
-        if let Some(additional_coinbase_script_data) = additional_coinbase_script_data.clone() {
-            if additional_coinbase_script_data.len() > range_1.end - range_1.start {
-                return Err(ExtendedExtranonceError::InvalidAdditionalCoinbaseScriptDataLength);
+        if let Some(static_prefix) = static_prefix.clone() {
+            if static_prefix.len() > 2 {
+                return Err(ExtendedExtranonceError::InvalidStaticPrefixLength);
             }
```
plebhash: hmm maybe we're leaving behind some edge cases here

plebhash: solved

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -46,16 +46,65 @@ impl JobIdFactory {
...
+        let mut pool_miner_tag = vec![];
+        pool_miner_tag.extend_from_slice(b"/");
+        if let Some(pool_tag_string) = &self.pool_tag_string {
+            pool_miner_tag.extend_from_slice(pool_tag_string.as_bytes());
+        }
+        pool_miner_tag.extend_from_slice(b"/");
+        if let Some(miner_tag_string) = &self.miner_tag_string {
+            pool_miner_tag.extend_from_slice(miner_tag_string.as_bytes());
+        }
+        pool_miner_tag.extend_from_slice(b"/");
```
GitGab19, line 89: Why not adding the delimiters only if there are tag values?

GitGab19: Oh I guess it's to make it easier to compute the right number of bytes used for them in `coinbase_tx_prefix` function, right?

protocols/v2/channels-sv2/src/server/jobs/factory.rs
GitGab19, line 1: Is it feasible to add a unit test here to make sure we build a valid coinbase tx with a couple of pool or miner tags examples?

protocols/v2/handlers-sv2/Cargo.toml
GitGab19, line 1: Let's ensure we don't need to bump MAJOR also for `handlers_sv2` here.

protocols/v2/parsers-sv2/Cargo.toml
GitGab19, line 1: Let's ensure we don't need to bump MAJOR also for `parsers_sv2` here.

protocols/v2/roles-logic-sv2/Cargo.toml
GitGab19, line 1: Let's ensure we don't need to bump MAJOR also for `roles_logic_sv2` here.

protocols/v2/subprotocols/mining/src/lib.rs
```diff
@@ -311,9 +311,9 @@ impl From<&mut ExtendedExtranonce> for Extranonce {
 ///   the upstream.
 /// - `range_1` → `0..16`: The pool P increments these bytes to ensure each downstream gets a unique
 ///   extended extranonce search space. The pool could optionally choose to set some fixed bytes as
-///   `additional_coinbase_script_data` (smaller than 16 bytes), which are set on the beginning of
-///   this range and will not be incremented. Usually, these bytes are used to add an identifier for
-///   the pool.
+///   `static_prefix` (smaller than 16 bytes), which are set on the beginning of this range and will
```
GitGab19: Why 16 bytes?

In the PR description you mentioned 2 bytes for this.

protocols/v2/channels-sv2/src/server/jobs/factory.rs
plebhash: it makes it easier to distinguish edge cases

for example, if you see `/some_tag/`, did that come from pool or miner?

on the other hand:

if you see `/some_tag//`, you know it came from pool and miner didn't write anything

if you see `//some_tag/`, you know it came from miner and pool didn't write anything

---

anyways, this is just something that felt right in my mind when I was writing this after OS suggestion... and it's already the third iteration (before it was `<pool_tag>miner_tag|` and then it was `<pool_tag>|miner_tag|`)

so I'm not too opinionated

protocols/v2/handlers-sv2/Cargo.toml
plebhash: the main reasons for bumping MAJOR on `mining_sv2` were:

- renaming `ExtendedExtranonceError::InvalidAdditionalCoinbaseScriptDataLength` into `ExtendedExtranonceError::InvalidStaticPrefixLength`
- renaming `MAX_EXTRANONCE_LEN` into `FULL_EXTRANONCE_LEN` 

none of these are re-exported by `handlers_sv2`, therefore no public APIs of `handlers_sv2` were broken

protocols/v2/parsers-sv2/Cargo.toml
plebhash: the main reasons for bumping MAJOR on `mining_sv2` were:

- renaming `ExtendedExtranonceError::InvalidAdditionalCoinbaseScriptDataLength` into `ExtendedExtranonceError::InvalidStaticPrefixLength`
- renaming `MAX_EXTRANONCE_LEN` into `FULL_EXTRANONCE_LEN` 

none of these are re-exported by `parsers_sv2`, therefore no public APIs of `parsers_sv2` were broken

protocols/v2/roles-logic-sv2/Cargo.toml
plebhash: indeed `roles_logic_sv2` re-exports the entire `mining_sv2` and `channels_sv2` crates, so I would agree a MAJOR bump would be needed

but I'm curious why didn't semver CI catch it?

protocols/v2/subprotocols/mining/src/lib.rs
plebhash: yeah I overlooked this

fixed

protocols/v2/channels-sv2/src/server/jobs/factory.rs
plebhash: https://github.com/stratum-mining/stratum/pull/1824/commits/972735591c50852f6285bc9dc3387b30506bd714

protocols/v2/subprotocols/mining/src/lib.rs
```diff
@@ -513,9 +513,9 @@ impl ExtendedExtranonce {
             return Err(ExtendedExtranonceError::InvalidRanges);
         }
 
-        if let Some(additional_coinbase_script_data) = additional_coinbase_script_data.clone() {
-            if additional_coinbase_script_data.len() > range_1.end - range_1.start {
-                return Err(ExtendedExtranonceError::InvalidAdditionalCoinbaseScriptDataLength);
+        if let Some(static_prefix) = static_prefix.clone() {
+            if static_prefix.len() > 2 || static_prefix.len() > range_1.end - range_1.start {
```
bit-aloo: we can club these conditions

protocols/v2/subprotocols/mining/src/lib.rs
```diff
@@ -662,15 +662,14 @@ impl ExtendedExtranonce {
...
-            if let Some(additional_data) = &self.additional_coinbase_script_data {
-                self.range_1.start + additional_data.len()
-            } else {
-                self.range_1.start
-            };
+        let extended_part_start = if let Some(static_prefix) = &self.static_prefix {
+            self.range_1.start + static_prefix.len()
+        } else {
+            self.range_1.start
+        };
```
bit-aloo: let extended_part_start =
    self.range_1.start + self.static_prefix.as_ref().map_or(0, |p| p.len());

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -930,34 +937,39 @@ impl Pool {
...
         let range_2 = std::ops::Range {
             start: range_1_end,
-            end: MAX_EXTRANONCE_LEN,
+            end: FULL_EXTRANONCE_LEN,
         };
 
+        // simulating a scenario where there are multiple mining servers
+        // this static prefix allows unique extranonce_prefix allocation
+        // for this mining server
+        let static_prefix = vec![0x01];
```
bit-aloo: shouldn't we put this in config?

protocols/v2/subprotocols/mining/src/lib.rs
plebhash: done

protocols/v2/subprotocols/mining/src/lib.rs
plebhash: done

roles/pool/src/lib/mining_pool/mod.rs
plebhash: 62c62f4094b3bdcdaf7e247e99fd80eca95201c0

protocols/v2/channels-sv2/src/server/jobs/factory.rs
GitGab19: I think it makes perfectly sense 👍

roles/pool/src/lib/mining_pool/mod.rs
```diff
@@ -930,34 +937,39 @@ impl Pool {
         shares_per_minute: f32,
         recv_stop_signal: tokio::sync::watch::Receiver<()>,
     ) -> Result<Arc<Mutex<Self>>, PoolError> {
-        let range_0 = std::ops::Range { start: 0, end: 0 };
+        let range_1_start = 0;
+        let range_1_end = 8;
```
GitGab19, line 943: Could you just add a comment here explaining what are these 8 bytes? 

I know we have docs in `mining_sv2` crate for this, but an immediate description could be helpful for everyone in the future.

protocols/v2/roles-logic-sv2/Cargo.toml
GitGab19: It could be the version of `semver-check` we use on our CI which could be not able to recognize this kind of cases..

GitGab19: I would take note of this in the tracker issue we have btw

roles/pool/src/lib/mining_pool/mod.rs
plebhash: bae1cfcfe29fbfe0016cb6e8542514a71f5e0fe3

PR #1826, opened 2025-08-05
fix overlooked consequences of `SetupConnection.REQUIRES_WORK_SELECTION`

### LINKED ISSUE #1825: overlooked consequences of `SetupConnection.REQUIRES_WORK_SELECTION`
if a client connects to Pool with `SetupConnection.REQUIRES_WORK_SELECTION`, that means it will work on custom jobs

we overlooked many consequences of this:
- Standard Channels should not be allowed
- there's no need to send future `NewExtendedMiningJob` and `SetNewPrevHash` after `OpenExtendedMiningChannel.Success`
- there's no need to send `NewExtendedMiningJob` when a `NewTemplate` arrives
- there's no need to call `channels_sv2::server::extended::ExtendedChannel::on_new_template` when a `NewTemplate` arrives
- `channels_sv2::server::extended::ExtendedChannel::on_set_new_prev_hash` doesn't have to return an error if the corresponding `template_id` does not exist on the `JobStore`

plebhash:
close #1825

### DIFF
--- a/roles/pool/src/lib/mining_pool/message_handler.rs
+++ b/roles/pool/src/lib/mining_pool/message_handler.rs
@@ -74,6 +74,20 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
 
         info!("Received OpenStandardMiningChannel: {}", incoming);
 
+        if self.requires_custom_work {
+            error!("OpenStandardMiningChannel: Standard Channels are not supported for this connection");
+            let open_standard_mining_channel_error = OpenMiningChannelError {
+                request_id,
+                error_code: "standard-channels-not-supported-for-custom-work"
+                    .to_string()
+                    .try_into()
+                    .expect("error code must be valid string"),
+            };
+            return Ok(SendTo::Respond(Mining::OpenMiningChannelError(
+                open_standard_mining_channel_error,
+            )));
+        }
+
         let last_future_template = self.last_future_template.clone();
         let last_set_new_prev_hash_tdp = self.last_new_prev_hash.clone();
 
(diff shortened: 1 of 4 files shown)


protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -185,15 +185,15 @@ pub fn has_requires_std_job(flags: u32) -> bool {
 /// Helper function to check if `REQUIRES_VERSION_ROLLING` bit flag present.
 pub fn has_version_rolling(flags: u32) -> bool {
     let flags = flags.reverse_bits();
-    let flags = flags << 1;
+    let flags = flags << 2;
```
GitGab19, line 188: Why this change?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -185,15 +185,15 @@ pub fn has_requires_std_job(flags: u32) -> bool {
...
     let flag = flags >> 31;
     flag != 0
 }
 
 /// Helper function to check if `REQUIRES_WORK_SELECTION` bit flag present.
 pub fn has_work_selection(flags: u32) -> bool {
     let flags = flags.reverse_bits();
-    let flags = flags << 2;
+    let flags = flags << 1;
```
GitGab19, line 196: Why this change?

protocols/v2/subprotocols/common-messages/src/setup_connection.rs
bit-aloo: Swapping the flags, earlier it was wrong

GitGab19: How was it wrong? I'm asking because we never got any issue with this in the past afaik

plebhash: `has_version_rolling` would return true if `REQUIRES_WORK_SELECTION` was set, and vice-versa

we never had problems with this because up until now, these functions never used anywhere, aside from unit tests that asserted this flipped behavior

run a `grep` for `has_work_selection` and `has_version_rolling` and you'll see it's not used anywhere

PR #1832, opened 2025-08-13
strip BIP141 from Extended Jobs

### LINKED ISSUE #1399: Issues with Translator and Braiins SV2 pool
1. After the translator is started, it tries to open an extended mining channel with the user "ABC". Such user is not known to the server and rejects the chanel. The translator thereafter terminates.
2. I didn't find a way to configure the user. Neither do I think it's necessary for the translator to open any channel when there is no downstream mining session active.
 
3. If I create the user ABC. the translator connects and then it crashes with the following error:
```
thread 'tokio-runtime-worker' panicked at translator/src/lib/proxy/next_mining_notify.rs:22:10:
failed to convert extended job to non segwit: InvalidCoinbase
```

plebhash:
on #1399 we realized Braiins Pool was stripping BIP141 bytes from Extended Jobs

this PR introduces some changes on how SRI deals with this:

now `channels_sv2::server::jobs::extended::ExtendedJob` keeps a representation of `coinbase_tx_prefix` and `coinbase_tx_suffix` WITH BIP141 bytes, however its internal `NewExtendedMiningJob` message carries `coinbase_tx_prefix` and `coinbase_tx_suffix` WITHOUT BIP141 bytes

that means a Mining Server can easily reconstruct the coinbase for block propagation with BIP141 bytes

but the `NewExtendedMiningJob` message will be sent with those bytes stripped away, making it easier for the Mining Client to find the coinbase `txid` (and not `wtxid`) for merkle root calculation

this follows the same pattern that Braiins is currently doing.

moreover, we also introduce `channels_sv2::client::bip141::try_strip_bip141`

this new helper function will take a `NewExtendedMiningJob` and return `Ok(None)` in case the original message was already stripped of BIP141 bytes, and `Ok(Some(NewExtendedMiningJob))` in case those bytes were stripped away

it's meant to be used by Mining Clients to make sure they are able to remain flexible by being able to deal with stripped and non-stripped jobs

since there's a high likelihood that `roles_logic_sv2::job_creator` could be eventually deprecated, this new helper function acts as an alternative to `roles_logic_sv2::job_creator::{StrippedCoinbaseTx, extended_job_to_non_segwit}`

this PR also applies this new stripping function to old tProxy, so it should be reproduced in the new tProxy cc @GitGab19 

---

close #1399


plebhash:
cc @coleFD FYI this is somewhat correlated to what you did on #488

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -341,7 +388,7 @@ impl JobFactory {
     fn custom_coinbase_tx_prefix(
         &self,
         m: SetCustomMiningJob<'_>,
-    ) -> Result<B064K<'static>, JobFactoryError> {
+    ) -> Result<Vec<u8>, JobFactoryError> {
```
bit-aloo, line 378: What was the reason for changing the return types here?

protocols/v2/channels-sv2/src/client/bip141.rs
```diff
@@ -0,0 +1,86 @@
...
+                    .map_err(|_| StripBip141Error::FailedToDeserializeCoinbaseVersion)?,
+                input: Decodable::consensus_decode(&mut decoder)
+                    .map_err(|_| StripBip141Error::FailedToDeserializeCoinbaseInputs)?,
+                output: Decodable::consensus_decode(&mut decoder)
+                    .map_err(|_| StripBip141Error::FailedToDeserializeCoinbaseOutputs)?,
+                lock_time: Decodable::consensus_decode(&mut decoder)
+                    .map_err(|_| StripBip141Error::FailedToDeserializeCoinbaseLockTime)?,
+            }
+        }
+    };
```
bit-aloo: What was the reason behind using both, rather than just one (the Error block implementation)?

protocols/v2/channels-sv2/src/client/bip141.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let txid = coinbase.compute_txid();
+    let wtxid = coinbase.compute_wtxid();
+
+    match txid.to_raw_hash() == wtxid.to_raw_hash() {
```
bit-aloo: lets go for an early return kinda style here, and then rest of the method:
```rust
    if coinbase.compute_txid().to_raw_hash() == coinbase.compute_wtxid().to_raw_hash() {
        return Ok(None);
    }
```

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -218,6 +218,20 @@ impl JobFactory {
...
+
+        // strip bip141 witness bytes from coinbase_tx_suffix
+        let locktime_position = coinbase_tx_suffix.len() - 4;
+
+        // strip witness count (1), witness length (1) and witness data (32)
+        let mut coinbase_tx_suffix_stripped_bip141 =
+            coinbase_tx_suffix[..locktime_position - 1 - 1 - 32].to_vec();
+        coinbase_tx_suffix_stripped_bip141
+            .extend_from_slice(&coinbase_tx_suffix[locktime_position..]);
+
```
bit-aloo: This part is duplicated both in the strip method and here. let’s move it to a common place. Also, can we define proper constants for these numbers and then use them here?
Something like this:
```rust
    const MARKER_FLAG_OFFSET: usize = 4;
    const MARKER_FLAG_LEN: usize = 2;
    const WITNESS_COUNT_LEN: usize = 1;
    const WITNESS_LEN_LEN: usize = 1;
    const WITNESS_DATA_LEN: usize = 32;
    const LOCKTIME_LEN: usize = 4;
```

protocols/v2/channels-sv2/src/client/bip141.rs
GitGab19, line 1: Given its importance, It would be good to have a unit test covering this function

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -142,8 +142,8 @@ impl JobFactory {
             self.coinbase_tx_suffix(template.clone(), additional_coinbase_outputs.clone())?;
         let merkle_path = template.merkle_path.clone();
         let merkle_root = merkle_root_from_path(
-            coinbase_tx_prefix.inner_as_ref(),
-            coinbase_tx_suffix.inner_as_ref(),
+            &coinbase_tx_prefix,
+            &coinbase_tx_suffix,
```
GitGab19, line 147: Aren't these two containing witness?

They are coming from `coinbase_tx_prefix()` and `coinbase_tx_suffix()` and it seems we have witness data there

protocols/v2/channels-sv2/src/server/jobs/factory.rs
GitGab19: Yeah probably we could have a function in the `bip141` module specific for this

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -283,6 +307,19 @@ impl JobFactory {
...
+
+        // strip bip141 witness bytes from coinbase_tx_suffix
+        let locktime_position = coinbase_tx_suffix.len() - 4;
+
+        // strip witness count (1), witness length (1) and witness data (32)
+        let mut coinbase_tx_suffix_stripped_bip141 =
+            coinbase_tx_suffix[..locktime_position - 1 - 1 - 32].to_vec();
+        coinbase_tx_suffix_stripped_bip141
+            .extend_from_slice(&coinbase_tx_suffix[locktime_position..]);
+
```
GitGab19: See https://github.com/stratum-mining/stratum/pull/1832/files#r2276344244

protocols/v2/channels-sv2/src/server/jobs/factory.rs
plebhash: to keep it consistent with the non-custom variant of this method

plebhash: overall it's way easier to manipulate `Vec<u8>` for general purposes

if the end goal is to serialize into a `B064K` to use this in a message, that can be done locally on the context where such message is being built

protocols/v2/channels-sv2/src/server/jobs/factory.rs
plebhash: `merkle_root_from_path` internally calls `compute_txid`, which always returns the correct `txid` (regardless of whether witness was stripped or not)

https://github.com/stratum-mining/stratum/blob/c3b15a0046bb6ab75d14620a3a272dc08bf26fa0/protocols/v2/channels-sv2/src/merkle_root.rs#L41

protocols/v2/channels-sv2/src/client/bip141.rs
plebhash: switched to the second approach only

protocols/v2/channels-sv2/src/client/bip141.rs
plebhash: done

protocols/v2/channels-sv2/src/server/jobs/factory.rs
plebhash: done

protocols/v2/channels-sv2/src/client/bip141.rs
plebhash: done

but also moved the module to the root of the crate, instead of inside of `client`

protocols/v2/channels-sv2/src/server/jobs/factory.rs
plebhash: the `bip141` module was moved to the root of the crate, instead of inside of `client` module

protocols/v2/channels-sv2/src/bip141.rs
```diff
@@ -0,0 +1,181 @@
...
+///
+/// If the coinbase transaction is already stripped of bip141, returns `Ok(None)`.
+/// If the coinbase transaction is not stripped, returns `Ok(Some(NewExtendedMiningJob))` with the
+/// stripped coinbase transaction.
+#[allow(clippy::type_complexity)]
+pub fn try_strip_bip141(
+    coinbase_tx_prefix: Vec<u8>,
+    coinbase_tx_suffix: Vec<u8>,
+) -> Result<Option<(Vec<u8>, Vec<u8>)>, StripBip141Error> {
+    let mut encoded = vec![];
```
bit-aloo: Let’s add this condition. I know it’s probably not necessary, but just in case, no panics:
```rust
    if coinbase_tx_prefix.len() < MARKER_FLAG_OFFSET + MARKER_FLAG_LEN
        || coinbase_tx_suffix.len() < WITNESS_COUNT_LEN + WITNESS_LEN_LEN + WITNESS_DATA_LEN + LOCKTIME_LEN
    {
        return Err(StripBip141Error::FailedToDeserializeCoinbaseVersion);
    }

```

bit-aloo:
```suggestion
    let mut encoded = Vec::with_capacity(
        coinbase_tx_prefix.len() + FULL_EXTRANONCE_LEN + coinbase_tx_suffix.len(),
    );
```

protocols/v2/channels-sv2/src/bip141.rs
```diff
@@ -0,0 +1,181 @@
...
+/// This helps calculate the coinbase `txid` (instead of `wtxid`) for merkle root calculation, and
+/// it is particularly helpful for translation to Sv1.
+///
+/// If the coinbase transaction is already stripped of bip141, returns `Ok(None)`.
+/// If the coinbase transaction is not stripped, returns `Ok(Some(NewExtendedMiningJob))` with the
+/// stripped coinbase transaction.
+#[allow(clippy::type_complexity)]
+pub fn try_strip_bip141(
+    coinbase_tx_prefix: Vec<u8>,
+    coinbase_tx_suffix: Vec<u8>,
```
bit-aloo: Let’s make them slices, since there are two possible outcomes: either we return an error, or we produce a stripped version. In the error case, we don’t need to allocate new space, and in the stripped version case, we’ll be creating new vectors from the slices anyway.
```suggestion
    coinbase_tx_prefix: &[u8],
    coinbase_tx_suffix: &[u8],
```

protocols/v2/channels-sv2/src/bip141.rs
plebhash: I don't see what kind of panics we'd be avoiding with this

protocols/v2/channels-sv2/src/server/jobs/factory.rs
GitGab19: Oh I see, I missed that!

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -218,6 +219,16 @@ impl JobFactory {
             self.coinbase_tx_prefix(template.clone(), additional_coinbase_outputs.clone())?;
         let coinbase_tx_suffix =
             self.coinbase_tx_suffix(template.clone(), additional_coinbase_outputs.clone())?;
+
+        println!("aaaa {:?}", coinbase_tx_prefix);
+        println!("bbbb {:?}", coinbase_tx_suffix);
```
GitGab19: Remove these logs

plebhash: oops done

protocols/v2/channels-sv2/src/bip141.rs
bit-aloo: Overflow panics can occur because the length of any of these lists may be smaller than the constant values used here, which would cause a panic. Since this is a public method intended for implementors, they can pass in any type of slice, including empty ones. In such cases, it will simply panic.

plebhash: if any of these edge cases happened, we would either fail to deserialize the coinbase or return `Ok(None)` way before we try to use those indexes

bit-aloo: hmm... thats true as well... consider it non blocking

PR #1833, opened 2025-08-13
feat: add stratum-translation library for SV1↔SV2 protocol translation

GitGab19:
This PR introduces a new crate called `stratum_translation` which contains main functions to go from Sv2 to Sv1 and the other way around.

It's a first step towards providing a library which simplifies the translation between Sv1 and Sv2 for any kind of implementers.

This could be used from our JDC as well in the future, to enable Sv1 support directly into it without the need for the tProxy.
But it could also make life easier for firmware developers to add Sv2 support more quickly in the future.

I will start by using this new crate in the new tProxy during my work on https://github.com/stratum-mining/stratum/pull/1791


roles/roles-utils/stratum-translation/src/utils.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    if pdiff > target {
+        let diff = pdiff / target;
+        Ok(diff.low_u64() as f64)
+    } else {
+        let diff = target / pdiff;
+        let diff = diff.low_u64() as f64;
+        Ok(1.0 / diff)
+    }
+}
```
plebhash: not sure what the usage you're intending for this function

but it might be worth considering that [`bitcoin::Target`](https://docs.rs/bitcoin/latest/bitcoin/struct.Target.html) already has all this functionality out of the box

I left some comments about this on `channels_sv2`: https://github.com/stratum-mining/stratum/blob/c3b15a0046bb6ab75d14620a3a272dc08bf26fa0/protocols/v2/channels-sv2/src/server/extended.rs#L53

roles/roles-utils/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -0,0 +1,224 @@
...
+        let coinbase_suffix = vec![
+            0xff, 0xff, 0xff, 0xff, // sequence
+            0x01, // output count
+            0x00, 0xf2, 0x05, 0x2a, 0x01, 0x00, 0x00, 0x00, // value (50 BTC)
+            0x19, // script length
+            0x76, 0xa9, 0x14, // OP_DUP OP_HASH160 <20 bytes>
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x88, 0xac, // OP_EQUALVERIFY OP_CHECKSIG
+            0x00, 0x00, 0x00, 0x00, // locktime
+        ];
```
plebhash, line 196: the spec has `coinbase_prefix` with a different meaning (e.g.: `SetCustomMiningJob.coinbase_prefix` and `NewTemplate.coinbase_prefix`)

usually I try to always stick with `coinbase_tx_prefix` when that's the intended contextual meaning

roles/roles-utils/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -0,0 +1,224 @@
...
+        let new_prev = SetNewPrevHash {
+            channel_id: 1,
+            job_id: 456,
+            prev_hash: [0x02u8; 32].into(),
+            min_ntime: 1234567890,
+            nbits: 0x1d00ffff,
+        };
+
+        // Create a realistic coinbase with proper structure
+        let mut coinbase_prefix = vec![0x01, 0x00, 0x00, 0x00]; // version
```
plebhash: the spec has `coinbase_prefix` with a different meaning (e.g.: `SetCustomMiningJob.coinbase_prefix` and `NewTemplate.coinbase_prefix`)

usually I try to always stick with `coinbase_tx_prefix` and `coinbase_tx_suffix` when that's the intended contextual meaning

roles/roles-utils/stratum-translation/src/sv1_to_sv2.rs
plebhash, line 1: should we try to support Standard Channels here?

roles/roles-utils/stratum-translation/src/sv2_to_sv1.rs
plebhash, line 1: should we try to support Standard Channels here?

roles/roles-utils/stratum-translation/src/utils.rs
plebhash: turned into an issue for visibility https://github.com/stratum-mining/stratum/issues/1834

GitGab19: This function is used to set the initial difficulty to the Sv1 downstream in the tProxy (the current one is using it [here](https://github.com/GitGab19/stratum/blob/c3b15a0046bb6ab75d14620a3a272dc08bf26fa0/roles/translator/src/lib/downstream_sv1/diff_management.rs#L150)).

But I agree, we should replace this with the primitives coming from rust-bitcoin.

I'll add a comment on top of the function and take a note of this under the issue you created.

GitGab19: I just noticed we have a function to do this in `channels_sv2::target`. I can make that module public and use that function. 

So that when we are gonna tackle #1834 we are gonna change things only there.

roles/roles-utils/stratum-translation/src/sv1_to_sv2.rs
GitGab19: How can we achieve that if the client rolls the extranonce?

Were you thinking about a scenario where we set the `extranonce2_size` to 0?

plebhash: yes, the only way I can imagine this working is if `extranonce2_size` is 0

obviously, that's not going to work on legacy unmodified Sv1 firmware, I've already seen it in practice

however, if this lib is ever going to be used by developers who are writing the Sv1 firmware themselves (and they can implement it in a way that accepts `extranonce2_size = 0`), that leaves Standard Channels as viable option

GitGab19: Even if we suppose that the miner can work with an `extranonce2_size` = 0, there's a problem related to the `mining.notify` message.

We don't have a field for the `merkle_root` there, we only have the `merkle_branches`, `coinb1`, and `coinb2`.

I'm arriving to the conclusion that standard channels with Sv1 would require a different message and a completely different way to compute the hash (basically by implementing the HOM into the Sv1 firmware).

plebhash: damn, that's true!

well, I guess Standard and Group Channels are out of the picture for this kind of approach then

plebhash (approved):

roles/roles-utils/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -0,0 +1,239 @@
...
+/// # Panics
+/// This function will panic if the conversion from extended job to non-segwit fails.
+pub fn build_sv1_notify_from_sv2(
+    new_prev_hash: SetNewPrevHash<'static>,
+    new_job: NewExtendedMiningJob<'static>,
+    clean_jobs: bool,
+) -> Result<server_to_client::Notify<'static>> {
+    // TODO change this to use the new channels_sv2::client::bip141::try_strip_bip141
+    // introduced in https://github.com/stratum-mining/stratum-mining/pull/1832
+    //let new_job = try_strip_bip141(new_job);
```
plebhash, line 30:
```suggestion
    let new_job = try_strip_bip141(new_job);
```

plebhash:
```suggestion
    let new_job = match try_strip_bip141(new_job).map_err(...) {
        Some(new_job_stripped) = new_job_stripped,
        None = new_job,
    };
```

pseudocode suggestion as we'll probably need to:
- add new error variant to `crate::error`
- maybe clone `new_job`

GitGab19: Done

PR #1838, opened 2025-08-15
fix sigops calculation for `CoinbaseOutputContraints`

### LINKED ISSUE #1837: incorrect sigops calculation for `CoinbaseOutputConstraints `
BIP 141 introduced a weighted sigop cost system where sigops in scriptSig/scriptPubKey count at 4x weight

currently in SRI, despite the fact that in we create the coinbase as a SegWit transaction, we're only counting sigops of specific output scripts, without applying the proper weight:

https://github.com/stratum-mining/stratum/blob/863f070d44f3952b697d5624e87d5ea33de8a8f7/roles/pool/src/lib/mod.rs#L83

https://github.com/stratum-mining/stratum/blob/863f070d44f3952b697d5624e87d5ea33de8a8f7/roles/jd-client/src/lib/template_receiver/mod.rs#L253-L259

[`bitcoin::Transaction::total_sigop_cost`](https://docs.rs/bitcoin/latest/bitcoin/struct.Transaction.html#method.total_sigop_cost) is a better way because it takes the entire transaction structure into consideration and also makes sure to use BIP141 counting rules.

refs:
- https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#user-content-Sigops
- https://bitcoin.stackexchange.com/questions/117356/what-is-sigop-signature-operation#117359

plebhash:
close #1837

### DIFF
--- a/roles/jd-client/src/lib/template_receiver/mod.rs
+++ b/roles/jd-client/src/lib/template_receiver/mod.rs
@@ -16,8 +16,12 @@ use stratum_common::{
     roles_logic_sv2::{
         self,
         bitcoin::{
+            absolute::LockTime,
+            blockdata::witness::Witness,
             consensus::{deserialize, serialize, Encodable},
-            Amount, TxOut,
+            script::ScriptBuf,
+            transaction::{OutPoint, Transaction, Version},
+            Amount, Sequence, TxIn, TxOut,
         },
         codec_sv2::{HandshakeRole, Initiator, StandardEitherFrame, StandardSv2Frame},
         handlers::{template_distribution::ParseTemplateDistributionMessagesFromServer, SendTo_},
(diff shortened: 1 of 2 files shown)


lucasbalieiro:
Concept ACK

I agree with the change from `count_sigops` to `total_sigops_count`.

Confirmed that Bitcoin Core has the same behavior on current `main`, where it counts the SigOps and applies a scale factor:
https://github.com/bitcoin/bitcoin/blob/7b4a1350dfd6b3892a9c314eeff28b22ef0c1a73/src/consensus/tx_verify.cpp#L143

BIP141 also distinguishes between the concepts of SigOp *count* and SigOp *cost*, so this change aligns with the spec.

I’ll do some additional testing later with practical config changes, but for now I have a small **nit**:
We could use rust-bitcoin’s provided defaults instead of manually creating empty vectors. For example:

```rust
version: Version::non_standard(2),
lock_time: LockTime::from_consensus(0),
input: vec![TxIn {
    previous_output: OutPoint::null(),
    script_sig: vec![].into(),
    sequence: Sequence(0),
    witness: Witness::from(vec![vec![0; 32]]),
}],
output: vec![empty_coinbase_output],
```

Could be written as:

```rust
version: Version::TWO,
lock_time: LockTime::ZERO,
input: vec![TxIn {
    previous_output: OutPoint::null(),
    script_sig: ScriptBuf::new(),
    sequence: Sequence::MAX,
    witness: Witness::default(),
}],
output: vec![empty_coinbase_output],
```

In practice, this wouldn’t change behavior, but it may be slightly more idiomatic.

roles/jd-client/src/lib/template_receiver/mod.rs
```diff
@@ -250,14 +254,28 @@ impl TemplateRx {
                             deserialize(&jds_coinbase_outputs).expect("Invalid coinbase output");
 
                         let mut coinbase_output_max_additional_size = 0;
-                        let mut coinbase_output_max_additional_sigops = 0;
 
-                        for output in deserialized_jds_coinbase_outputs {
+                        for output in deserialized_jds_coinbase_outputs.clone() {
```
bit-aloo:
```suggestion
                      let coinbase_output_max_additional_size: usize = deserialized_jds_coinbase_outputs.iter().map(|o| o.size()).sum();
```

bit-aloo: Clone can be avoided.

Issue #1810, opened 2025-07-22
finish Rust Docs for `channels_sv2` crate


plebhash:
as I'm working on UniFFI bindings for `channels_sv2`, I visited https://docs.rs/channels_sv2 and noticed the Rust Docs for this crate are very incomplete

we should complete them to make sure this crate follows the same standards we established for the other crates

PR #1840, opened 2025-08-18
Turns `channel_sv2::client::*` `no_std` compatible

### LINKED ISSUE #1822: make `channels_sv2` compatible with `no_std`
in order to build firmware that's compatible with [`rust-embedded`](https://github.com/rust-embedded), crates have to be compatible with `no_std`

we already had some contributions adapting some crates with this purpose:
- https://github.com/stratum-mining/stratum/pull/1394
- https://github.com/stratum-mining/stratum/pull/1346
- https://github.com/stratum-mining/stratum/pull/1398
- https://github.com/stratum-mining/stratum/pull/1315
- https://github.com/stratum-mining/stratum/pull/1238
- https://github.com/stratum-mining/stratum/pull/1230
- https://github.com/stratum-mining/stratum/pull/932

`channels_sv2` could be useful for firmware, but it's not yet compatible with `no_std`

this task should only cover `client` and `chain_tip` modules... all other modules should be simply disabled from compilation under `no_std`

lucasbalieiro:
closes #1822 ;

- Adds `hashbrown ` to replace `HashMap` and `HashSet`
- Replacing `std::io::Cursor` with `bitcoin::io::Cursor`
- Replacing `std` types and macros with `alloc` equivalents.

Updated auxiliary modules ( like merkle_root, bip141) to be `no_std` compatible types because `channels_sv2` depends on them.

### DIFF
--- a/protocols/v2/channels-sv2/Cargo.toml
+++ b/protocols/v2/channels-sv2/Cargo.toml
@@ -22,3 +22,9 @@ job_declaration_sv2 = { path = "../subprotocols/job-declaration", version = "^4.
 tracing = { version = "0.1"}
 bitcoin = { version = "0.32.5" }
 primitive-types = "0.13.1"
+hashbrown = { version = "0.15.5", optional = true }
+
+
+[features]
+default = []
+no_std = ["hashbrown"]
--- a/protocols/v2/channels-sv2/src/client/extended.rs
+++ b/protocols/v2/channels-sv2/src/client/extended.rs
@@ -3,6 +3,8 @@
 //! This module provides an abstraction over the state of an [Sv2](https://stratumprotocol.org/specification)
 //! **Extended Channel** within a mining client.
 
+extern crate alloc;
+use super::HashMap;
 use crate::{
     bip141::try_strip_bip141,
     chain_tip::ChainTip,
@@ -13,6 +15,7 @@ use crate::{
     merkle_root::merkle_root_from_path,
     target::{bytes_to_hex, target_to_difficulty, u256_to_block_hash},
 };
+use alloc::{format, string::String, vec, vec::Vec};
 use binary_sv2::{self, Sv2Option};
 use bitcoin::{
     blockdata::block::{Header, Version},
@@ -23,7 +26,6 @@ use mining_sv2::{
     NewExtendedMiningJob, SetNewPrevHash as SetNewPrevHashMp, SubmitSharesExtended, Target,
     FULL_EXTRANONCE_LEN,
 };
-use std::{collections::HashMap, convert::TryInto};
 use tracing::debug;
 
 /// A type alias representing an extended mining job tied to a specific `extranonce_prefix`.
--- a/protocols/v2/channels-sv2/src/lib.rs
+++ b/protocols/v2/channels-sv2/src/lib.rs
@@ -12,11 +12,16 @@
 //! - Standard, extended, and group channel support
 //! - Share accounting
 //! - Job store abstractions
+#![cfg_attr(feature = "no_std", no_std)]
+
+#[cfg(not(feature = "no_std"))]
+pub mod server;
+
+#[cfg(not(feature = "no_std"))]
+pub mod template;
 
 pub mod bip141;
 pub mod chain_tip;
 pub mod client;
 mod merkle_root;
-pub mod server;
 pub mod target;
-pub mod template;
--- a/protocols/v2/channels-sv2/src/lib.rs
+++ b/protocols/v2/channels-sv2/src/lib.rs
@@ -12,6 +12,7 @@
 //! - Standard, extended, and group channel support
 //! - Share accounting
 //! - Job store abstractions
+//! - [`client`] module is `no_std` compatible. To enable it build the crate with `no_std` feature.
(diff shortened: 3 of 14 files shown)


protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -10,17 +12,18 @@ use crate::{
...
     blockdata::block::{Header, Version},
     hashes::sha256d::Hash,
     CompactTarget, Target as BitcoinTarget,
 };
+use hashbrown::HashMap;
 use mining_sv2::{
     NewExtendedMiningJob, SetNewPrevHash as SetNewPrevHashMp, SubmitSharesExtended, Target,
     FULL_EXTRANONCE_LEN,
 };
-use std::{collections::HashMap, convert::TryInto};
```
plebhash, line 26: what's the replacement for `std::convert::TryInto`?

plebhash: I would avoid simply replacing `std::collections::HashMap` for `hashbrown::HashMap`

instead, I would do something like this:
```rust
#[cfg(not(feature = "no_std"))]
type HashMap<K, V> = std::collections::HashMap<K, V>;

#[cfg(feature = "no_std")]
type HashMap<K, V> = hashbrown::HashMap<K, V>;
```

protocols/v2/channels-sv2/Cargo.toml
```diff
@@ -22,3 +22,9 @@ job_declaration_sv2 = { path = "../subprotocols/job-declaration", version = "^4.
 tracing = { version = "0.1"}
 bitcoin = { version = "0.32.5" }
 primitive-types = "0.13.1"
+hashbrown = "0.15.5"
+
+[features]
+default = ["std"]
+std = []
```
plebhash: we can make `hashbrown` optional by flipping the flag logic and having a `no_std` flag instead of `std`

```suggestion
hashbrown = { version = "0.15.5", optional = true }

[features]
default = []
no_std = ["hashbrown"]
```

protocols/v2/channels-sv2/src/lib.rs
```diff
@@ -1,7 +1,13 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(feature = "std")]
+pub mod server;
+
+#[cfg(feature = "std")]
+pub mod template;
```
plebhash:
```suggestion
#![cfg_attr(feature = "no_std", no_std)]

#[cfg(not(feature = "no_std"))]
pub mod server;

#[cfg(not(feature = "no_std"))]
pub mod template;
```

protocols/v2/channels-sv2/src/client/extended.rs
plebhash: note: this is based on the fact that we have a `no_std` flag (instead of `std` proposed here)... that allows us to only use `hashbrown::HashMap` when explicitly defined

this is better than forcing a fragmented pattern across the crate, since we still rely on `std::collections::HashMap` for the `server` module, regardless of flags

plebhash: same goes for all proposed usage of `hashbrown::{HashMap, HashSet}` on `group.rs`, `share_accounting.rs` and `standard.rs`

lucasbalieiro: About the `std::convert::TryInto`, it does not need a replacement because it already comes on the prelude of `core::prelude::v1` since version 1.55

https://github.com/rust-lang/rust/blob/9eb4a2652031ed5ba97c29ef21c79db1645f7883/library/core/src/prelude/mod.rs#L49

lucasbalieiro:
@plebhash , thanks for the review and the insights.

Addressed the points you mentioned.

And the type alias for `HashMap` and `HashSet`, I added into a newly created file`prelude.rs` file and updated where needed.

protocols/v2/channels-sv2/src/aliases.rs
plebhash, line 1: hmm I'm afraid this could be a bit misleading?

here we're setting aliases that are universal for the crate, which leads the reader to believe they will take place everywhere

but all around the `server` module we're still sticking to `std::collections::HashMap`, independently of `no_std` flag (which is how it should be, since we don't really have realistic `no_std` use cases for Mining Servers)

I understand the urge to reduce code repetition, but in this case perhaps it's best to be repetitive-but-explicit rather than implicit-but-misleading

plebhash:
in the spirit of #1810, let's make sure we add some comments about `no_std` compatibility

not only on the code, but also on the crate's `README.md`

protocols/v2/channels-sv2/src/aliases.rs
lucasbalieiro: Thanks, addressed it by removing the `aliases.rs` file.

And sent the types to `src/client/mod.rs` so its now only visible inside `clients` module. Avoiding the confusion.

plebhash:
I wanted to write this while I was approving the PR and I forgot, so here it goes:

we try to be as conservative as possible when introducing new dependencies to low level crates, mostly for security reasons

however `hashbrown` can be seen as an exception to this rule, given that it's maintained by the Core Rust team and it's a dependency of `std` itself

Issue #1821, opened 2025-07-30
make `parsers_sv2` compatible with `no_std`


plebhash:
in order to build firmware that's compatible with [`rust-embedded`](https://github.com/rust-embedded), crates have to be compatible with `no_std`

we already had some contributions adapting some crates with this purpose:
- https://github.com/stratum-mining/stratum/pull/1394
- https://github.com/stratum-mining/stratum/pull/1346
- https://github.com/stratum-mining/stratum/pull/1398
- https://github.com/stratum-mining/stratum/pull/1315
- https://github.com/stratum-mining/stratum/pull/1238
- https://github.com/stratum-mining/stratum/pull/1230
- https://github.com/stratum-mining/stratum/pull/932

when these contributions happened, `parsers` was still a module of `roles_logic_sv2`.

recently, we made `parsers_sv2` into a standalone crate.

this crate would be useful for firmware, since it would allow conversion of raw bytes into Rust-typed messages from the `subprotocol` crates (`mining_sv2`, `job_declaration_sv2` and `template_distribution_sv2`, which are already `no_std` compatible)

plebhash:
while we're working on this, it would be good to also do the following:
- evaluate the `unimplemented!` lines (if they can be cleaned up, or receive some meaningful implementation)
- crate an `examples` directory to showcase basic usage of the APIs

Cyber-Lord:
I'll like to take on this or be fully part of the implementation.

PR #1842, opened 2025-08-22
make `parsers_sv2` `no_std` compatible

lucasbalieiro:
partially addresses #1821 .


protocols/v2/parsers-sv2/src/error.rs
```diff
@@ -24,5 +24,3 @@ impl std::fmt::Display for ParserError {
         }
     }
 }
-
-impl std::error::Error for ParserError {}
```
plebhash, line 28: we don't need to replace this one?

lucasbalieiro: I wasn't able to find a suitable replacement for this one.
And based on the usage of the `ParserError` on our codebase its seems that we don't need this `impl`.

I could replace it with the version from `core` but `error` in `core` is unstable on our MSRV.

So I thought of just deleting it. 

We could add some feature flag like we did recently with `channels_sv2 (#1840)`.

What you think?

plebhash: well, if the removal is not causing problems anywhere, I guess that's fine

plebhash (approved):

PR #1845, opened 2025-08-26
fix `SetCustomMiningJob` edge cases

### LINKED ISSUE #1844: need to fix `SetCustomMiningJob` edge cases
when a Pool's Client:
- establishes a Connection with `REQUIRES_WORK_SELECTION` bit flag set
- opens an Extended Channel
- does not send a `SetCustomMiningJob`

any call to `channels_sv2::server::ExtendedChannel::get_active_job()` will return an error, because the Extended Channel is in a state where there's no active jobs

plebhash:
close #1844


GitGab19 (approved):
ACK

PR #1843, opened 2025-08-24
New Job Declarator Client

Shourya742:
This PR introduces the new Job Declarator Client (JDC).
It adopts a new architecture that makes the flow leaner and easier to comprehend, while leveraging the new channel APIs and other improvements in the SRI stack.

This is still a work in progress, with ongoing polishing of rough edges.

Please review the overall diff rather than going commit by commit, the holistic view will make the changes clearer.

Currently, we have channel id unique across downstream, which gonna fix and make it unique per downstream.

This PR closes: https://github.com/stratum-mining/stratum/issues/1674 and all its sub-issues.

WIP

* [x] Make fallback deterministic (currently it explodes into a cascade of errors, requiring the setup to re-run multiple times).
* [x] Add improved documentation explaining the flow.
* [x] Update README.
* [x] Replace the old JDC with the new one and update integration tests.
* [x] Update CI Workflow to include new-jdc.

Closes: #1897

PR #1846, opened 2025-08-28
2025 08 28 channels sv2 adaptations for new jdc

plebhash:
this is a pre-requisite for #1843

JDC needs a way to produce a `SetCustomMiningJob` from a `NewTemplate`

previously, this was achievable via `channels_sv2::server::extended::ExtendedChannel`, which is flawed for a couple different reasons:

first, this assumes each Extended Channel that JDC established with a client will generate a different `SetCustomMiningJob`

second, the constructor for this channel abstraction is not meant to use difficulty target set by a server

---

notes for usage on #1843 cc @Shourya742 

JDC should get a dedidated `JobFactory` that will be responsible for doing `new_custom_job` to generate the `SetCustomMiningJob` to be sent to Pool

once JDC receives `SetCustomMiningJob.Success`, its `channels_sv2::client::extended::ExtendedChannel` must be updated with `on_set_custom_mining_job_success`

also, when shares are being forwarded to Pool, it's important to make sure they have the correct `channel_id` and `job_id` that correlate to this Channel, and the `job_id` assigned via `SetCustomMiningJob.Success`


protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -371,7 +467,7 @@ impl<'a> ExtendedChannel<'a> {
 
         // create the header for validation
         let header = Header {
-            version: Version::from_consensus(share.version as i32),
+            version: BlockVersion::from_consensus(share.version as i32),
```
GitGab19, line 501: Why this change?

GitGab19 (approved):
LGTM

protocols/v2/channels-sv2/src/client/extended.rs
plebhash: because now we have two imports with the same name:
- `bitcoin::transaction::Version` (which was added on this PR)
- `bitcoin::block::Version as BlockVersion` (which was renamed on this PR)

protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -18,16 +18,23 @@ use crate::{
...
 use mining_sv2::{
-    NewExtendedMiningJob, SetNewPrevHash as SetNewPrevHashMp, SubmitSharesExtended, Target,
-    FULL_EXTRANONCE_LEN,
+    NewExtendedMiningJob, SetCustomMiningJob, SetCustomMiningJobSuccess,
+    SetNewPrevHash as SetNewPrevHashMp, SubmitSharesExtended, Target, FULL_EXTRANONCE_LEN,
 };
 use tracing::debug;
 
+// import for docs only
+#[allow(unused_imports)]
```
lucasbalieiro: **nit:** you can skip the unused import by linking it straight in the doc like this:

```rust
/// A [`SetNewPrevHash`](template_distribution_sv2::SetNewPrevHash) (Template Distribution Protocol) message should …
```

protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -252,6 +255,136 @@ impl<'a> ExtendedChannel<'a> {
...
+
+    /// Handles a [`ChainTip`] update.
+    ///
+    /// To be used by a Sv2 Job Declarator Client, which should never receive a
+    /// [`SetNewPrevHash`](SetNewPrevHashMp) (Mining Protocol) message, or will most likely
+    /// ignore it if it does.
+    ///
+    /// So a [`SetNewPrevHash`](template_distribution_sv2::SetNewPrevHash) (Template Distribution
+    /// Protocol) message should be converted into a [`ChainTip`] and passed to this function.
+    pub fn on_chain_tip_update(&mut self, chain_tip: ChainTip) -> Result<(), ExtendedChannelError> {
```
GitGab19, line 370: How is this gonna be used?

It seems pretty similar to the `on_set_new_prev_hash` function we already have (?)

plebhash: the `on_set_new_prev_hash` method exists for scenarios where this is being used on a client that is not doing custom work

a JDC shouldn't expect to receive a `SetNewPrevHashMp`

@Shourya742 pointed out that he was getting `NoChainTip` errors when trying to do share validation for custom jobs on the upstream channel

Initially, I considered to set `self.chain_tip` inside `on_set_custom_mining_job_success`, given that `SetCustomMiningJob` has the 3 needed fields to build a `ChainTip`. However, that still left the problem that we had no way of moving past custom jobs to stale, flushing seen shares, etc.

So now `on_set_custom_mining_job_success` simply asserts presence and sanity of `self.chain_tip`.

And `on_chain_tip_update` is meant to be used by JDC when a `SetNewPrevHashTdp` arrives, updating `self.chain_tip` and doing the usual janitor work on past jobs.

GitGab19: Crystal clear, thanks!

GitGab19 (approved):

PR #1866, opened 2025-09-12
Storage trait

average-gary:
This doesn't modify anything in existing channel implementation but proposed the trait with some slight modification to the API.

### DIFF
--- a/.github/workflows/clippy-lint.yaml
+++ b/.github/workflows/clippy-lint.yaml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.89
           override: true
           components: clippy
       - name: Run Clippy on different workspaces and crates
--- a/protocols/v2/handlers-sv2/src/common.rs
+++ b/protocols/v2/handlers-sv2/src/common.rs
@@ -1,24 +1,28 @@
-use crate::error::HandlerError as Error;
 use common_messages_sv2::{
     ChannelEndpointChanged, Reconnect, SetupConnectionError, SetupConnectionSuccess, *,
 };
 use core::convert::TryInto;
 use parsers_sv2::CommonMessages;
 
+use crate::error::HandlerErrorType;
+
 pub trait HandleCommonMessagesFromServerSync {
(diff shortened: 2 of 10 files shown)


average-gary:
> but I worry that we could end up in an unproductive cycle of PR reviews that are blindly fed back into AI, only to dream and drift away into different directions

ACK. Apologies and I will do better moving forward.

> it seems the trait is not generic, but opinionated about the available storage backends listed in the ShareAccountingConfig enum

I see what you mean, anyone implementing the trait as is would need to modify this enum.

plebhash:
I guess the two flavors of `ShareValidationError` would need to be unified into the new root `share_accounting` module

protocols/v2/channels-sv2/src/share_accounting.rs
```diff
@@ -0,0 +1,47 @@
+use bitcoin::hashes::sha256d::Hash;
+
+pub trait ShareAccountingTrait {
```
plebhash:
```suggestion
pub trait ShareAccounting {
```

no need to have `Trait` in the name

protocols/v2/channels-sv2/src/share_accounting.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn get_share_work_sum(&self) -> u64;
+
+    /// Returns the highest difficulty found across all users and shares.
+    fn get_best_diff(&self) -> f64;
+
+    /// Updates the best difficulty if the new value is higher (backward compatibility).
+    fn update_best_diff(&mut self, diff: f64);
+}
+
+pub trait ShareAccountingClientTrait: ShareAccountingTrait {
```
plebhash:
```suggestion
pub trait ShareAccountingClient: ShareAccounting {
```

protocols/v2/channels-sv2/src/share_accounting.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn new() -> Self;
+}
+
+pub trait ShareAccountingServerTrait: ShareAccountingTrait {
```
plebhash:
```suggestion
pub trait ShareAccountingServer: ShareAccounting {
```

PR #1884, opened 2025-09-17
`JobStore` as generic

### LINKED ISSUE #1883: need to replace `JobStore` trait object for generic on `channels_sv2`
currently, server channel abstractions `channels_sv2` carry a `JobStore` as a trait object

https://github.com/stratum-mining/stratum/blob/ebccc10d021263efa584fa21de76e43161397268/protocols/v2/channels-sv2/src/server/extended.rs#L86-L94

there's no strong reason to do it this way instead of leveraging generics

moreover, this approach has the downside of forcing dynamic dispatch, which adds runtime overhead

plebhash:
close #1883

### DIFF
--- a/roles/pool/src/lib/mining_pool/message_handler.rs
+++ b/roles/pool/src/lib/mining_pool/message_handler.rs
@@ -100,7 +100,7 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
             // we only create one group channel for all standard channels
 
             let group_channel_id = self.channel_id_factory.next();
-            let job_store = Box::new(DefaultJobStore::new());
+            let job_store = DefaultJobStore::new();
 
             let mut group_channel = GroupChannel::new_for_pool(
                 group_channel_id,
@@ -132,7 +132,7 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
             .to_vec();
 
         let channel_id = self.channel_id_factory.next();
-        let job_store = Box::new(DefaultJobStore::new());
+        let job_store = DefaultJobStore::new();
         let mut standard_channel = match StandardChannel::new_for_pool(
             channel_id,
             user_identity,
@@ -314,7 +314,7 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
         };
 
         let channel_id = self.channel_id_factory.next();
-        let job_store = Box::new(DefaultJobStore::new());
+        let job_store = DefaultJobStore::new();
         let mut extended_channel = match ExtendedChannel::new_for_pool(
             channel_id,
             user_identity,
--- a/roles/pool/src/lib/mining_pool/mod.rs
+++ b/roles/pool/src/lib/mining_pool/mod.rs
@@ -43,7 +43,10 @@ use stratum_common::{
         self,
         bitcoin::{Amount, TxOut},
         channels_sv2::server::{
-            extended::ExtendedChannel, group::GroupChannel, standard::StandardChannel,
+            extended::ExtendedChannel,
+            group::GroupChannel,
+            jobs::{extended::ExtendedJob, job_store::DefaultJobStore, standard::StandardJob},
+            standard::StandardChannel,
         },
         codec_sv2::{
             self, binary_sv2::U256, HandshakeRole, Responder, StandardEitherFrame, StandardSv2Frame,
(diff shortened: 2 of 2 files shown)


colerob1776:
looks good to me

protocols/v2/channels-sv2/src/server/standard.rs
```diff
@@ -89,12 +92,16 @@ pub struct StandardChannel<'a> {
     nominal_hashrate: f32,
     share_accounting: ShareAccounting,
     expected_share_per_minute: f32,
-    job_store: Box<dyn JobStore<StandardJob<'a>>>,
+    job_store: J,
     job_factory: JobFactory,
     chain_tip: Option<ChainTip>,
+    phantom: PhantomData<&'a ()>,
```
bit-aloo, line 98: why covariant ? shouldn't we want Job and channel to be invariant?

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -83,22 +83,29 @@ use tracing::debug;
...
     requested_max_target: Target,
     target: Target, // todo: try to use Target from rust-bitcoin
     nominal_hashrate: f32,
-    job_store: Box<dyn JobStore<ExtendedJob<'a>>>,
+    job_store: J,
     job_factory: JobFactory,
     share_accounting: ShareAccounting,
     expected_share_per_minute: f32,
     chain_tip: Option<ChainTip>,
+    phantom: PhantomData<&'a ()>,
```
bit-aloo, line 102: why covariant? shouldn't we want Job and channel to be invariant?

protocols/v2/channels-sv2/src/server/group.rs
```diff
@@ -59,15 +61,22 @@ use std::collections::{HashMap, HashSet};
...
+where
+    J: JobStore<ExtendedJob<'a>>,
+{
     group_channel_id: u32,
     standard_channel_ids: HashSet<u32>,
     job_factory: JobFactory,
-    job_store: Box<dyn JobStore<ExtendedJob<'a>>>,
+    job_store: J,
     chain_tip: Option<ChainTip>,
+    phantom: PhantomData<&'a ()>,
```
bit-aloo, line 73: why covariant? shouldn't we want Job and channel to be invariant?

protocols/v2/channels-sv2/src/server/standard.rs
plebhash: not having this line results in the following compilation error:

```
error[E0392]: parameter `'a` is never used
  --> v2/channels-sv2/src/server/standard.rs:83:28
   |
83 | pub struct StandardChannel<'a, J>
   |                            ^^ unused parameter
   |
   = help: consider removing `'a`, referring to it in a field, or using a marker such as `PhantomData`

error[E0560]: struct `server::standard::StandardChannel<'a, J>` has no field named `phantom`
   --> v2/channels-sv2/src/server/standard.rs:218:13
    |
218 |             phantom: PhantomData,
    |             ^^^^^^^ `server::standard::StandardChannel<'a, J>` does not have this field
    |
    = note: all struct fields are already assigned

Some errors have detailed explanations: E0392, E0560.
For more information about an error, try `rustc --explain E0392`.
error: could not compile `channels_sv2` (lib) due to 2 previous errors
```

what alternative do you suggest?

bit-aloo: Line is important for associating jobs lifetime to channels lifetime. To make lifetime associativity invariant we can have `PhantomData<&'a mut ()>,`, I hope this would work. I was curious for initial choice of covariance..

protocols/v2/channels-sv2/src/server/group.rs
colerob1776: I just realized the `JobStore` trait returns a reference to a Job in `get_active_job()` which is limiting when using a shared job store. I imagine that trait method should return `Option<T>` where `T` handles access to the underlying job. I can do that in another issue, but just wanted to point it out

colerob1776: by invariant are you referrring to `PhantomData<&'a mut ()>`? From my understanding that would enforce JobStore and Job to have the same lifetime right?

Edit: I meant to say Channel and Job to have the same lifetime

protocols/v2/channels-sv2/src/server/standard.rs
plebhash: > I was curious for initial choice of covariance

it wasn't deliberate, tbh wasn't even on my radar.

added `mut` on all instances of `PhantomData`

protocols/v2/channels-sv2/src/server/group.rs
plebhash: ~~I think it would enforce Channel, JobStore and Job all to have the same lifetime~~

nevermind, I think this is not correct

to be honest, overall this topic is way outside my rust comfort zone

bit-aloo: JobStore, and Channel?

plebhash: lifetime `'a` is set for all 3: Channel, JobStore and Job

while for Channel it's set via `PhantomData`, which now carries `&'a mut ()`

protocols/v2/channels-sv2/src/server/standard.rs
plebhash: according to https://doc.rust-lang.org/nomicon/subtyping.html

`&'a mut ()` is also covariant over `'a`?

image

@Shourya742 are you sure we're achieving the desired outcome here?

bit-aloo: Invariant to type T

plebhash: but most importantly, what's the tangible problem we're aiming to solve here?

like I mentioned above, the original rationale for using `PhantomData` was simply to allow `StandardChannel` to have some struct member associated with `'a` lifetime, which was previously achieved with `Box<dyn JobStore<ExtendedJob<'a>>>` which now turned into a generic `J`, and not using `PhantomData` would prevent the crate from compiling

from my research, it seems that using either `PhantomData<&'a ()>` OR `PhantomData<&'a mut ()>` will have `StandardChannel` being covariant over `'a`, which is exactly the same behavior we had prior to this PR, when we had `job_store: Box<dyn JobStore<ExtendedJob<'a>>>`

so whatever lifetime variance problems we had before, we'll continue having... but we're also not introducing anything new or different here with regards to that

moreover, the goal of this PR is simply to make `job_store` generic instead of trait object, so trying to solve lifetime issues (which as we already know, go way deeper than `channels_sv2` crate) feels a bit out of scope?

bit-aloo: I agree. Since the channel is responsible for producing jobs, it’s appropriate for those jobs to have lifetimes covariant with the channel’s lifetime meaning they can be shorter. This reflects the fact that consumers of the channel only receive jobs. If, instead, the channel were storing jobs originating from elsewhere, invariance would be required to ensure safety. But in the server case, covariance should be ok.

PR #1890, opened 2025-09-19
Add close channel method to pool

Shourya742:
This PR extends the legacy HandlerSv2 (old) trait `ParseMiningMessagesFromDownstream` to support `CloseChannel` message. While we could update the Pool to use the new HandlerSv2 APIs, doing so would require removing a significant amount of existing code. Instead, this PR adds CloseChannel handling to all required implementors (such as Pool and the legacy JDC).

### DIFF
--- a/roles/jd-client/src/lib/downstream.rs
+++ b/roles/jd-client/src/lib/downstream.rs
@@ -909,7 +909,7 @@ impl ParseMiningMessagesFromDownstream<UpstreamMiningNode> for DownstreamMiningN
         &mut self,
         m: CloseChannel,
     ) -> Result<SendTo<UpstreamMiningNode>, Error> {
-        info!("Received Close Channle: {m}");
+        info!("Received Close Channel: {m}");
         Ok(SendTo::None(None))
     }
 }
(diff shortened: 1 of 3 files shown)


roles/jd-client/src/lib/downstream.rs
```diff
@@ -903,6 +903,15 @@ impl ParseMiningMessagesFromDownstream<UpstreamMiningNode> for DownstreamMiningN
         warn!("Ignoring SetCustomMiningJob");
         Ok(SendTo::None(None))
     }
+
+    /// Handle close channel
+    fn handle_close_channel(
+        &mut self,
+        m: CloseChannel,
+    ) -> Result<SendTo<UpstreamMiningNode>, Error> {
+        info!("Received Close Channle: {m}");
```
GitGab19:
```suggestion
        info!("Received Close Channel: {m}");
```

roles/pool/src/lib/mining_pool/message_handler.rs
```diff
@@ -998,4 +998,12 @@ impl ParseMiningMessagesFromDownstream<()> for Downstream {
         };
         Ok(SendTo::Respond(Mining::SetCustomMiningJobSuccess(success)))
     }
+
+    fn handle_close_channel(&mut self, m: CloseChannel) -> Result<SendTo<()>, Error> {
+        info!("Received Close Channle: {m}");
```
GitGab19:
```suggestion
        info!("Received Close Channel: {m}");
```

PR #1902, opened 2025-09-23
Persistence trait

average-gary:
After digging into the ShareAccounting more, I noticed there were calls within the hot path for mining messages. Notably, `validate_share`.

Since this validation and mining messages in general are on the critical path, I don't think #1866 with Results for ShareAccounting functions is the correct approach. 

In this approach, the Channel constructor would pass in a Sender for the Channel to use to pass data through the a Persistence implementation that is then responsible for handling it's own threads, state, errors, etc.

### DIFF
--- a/protocols/v2/channels-sv2/src/persistence.rs
+++ b/protocols/v2/channels-sv2/src/persistence.rs
@@ -34,6 +34,8 @@ pub enum ShareAccountingEvent {
         user_identity: String,
         /// Work value of the accepted share (difficulty as floating point)
         share_work: f64,
+        /// Sequence number of the share
+        share_sequence_number: u32,
         /// Hash of the accepted share (for duplicate detection)
         share_hash: Hash,
         /// Total shares accepted after this update
--- a/protocols/v2/channels-sv2/src/server/share_accounting.rs
+++ b/protocols/v2/channels-sv2/src/server/share_accounting.rs
@@ -136,6 +136,7 @@ where
             channel_id,
             user_identity: user_identity.to_string(),
             share_work,
+            share_sequence_number,
             share_hash,
             total_shares_accepted: self.shares_accepted,
             total_share_work_sum: self.share_work_sum,
--- a/protocols/v2/channels-sv2/src/client/share_accounting.rs
+++ b/protocols/v2/channels-sv2/src/client/share_accounting.rs
@@ -102,6 +102,7 @@ where
             channel_id,
             user_identity: user_identity.to_string(),
             share_work,
+            share_sequence_number,
             share_hash,
             total_shares_accepted: self.shares_accepted,
             total_share_work_sum: self.share_work_sum,
--- a/roles/stratum-apps/src/share_persistence/mod.rs
+++ b/roles/stratum-apps/src/share_persistence/mod.rs
@@ -97,6 +97,7 @@ impl ShareFileHandler {
                 channel_id,
                 user_identity,
                 share_work,
+                share_sequence_number,
                 share_hash,
                 total_shares_accepted,
                 total_share_work_sum,
@@ -105,10 +106,11 @@ impl ShareFileHandler {
             } => {
                 let result = self.file.write_all(
                     format!(
-                        "ShareAccepted: channel_id: {}, user_identity: {}, share_work: {}, share_hash: {}, total_shares_accepted: {}, total_share_work_sum: {}, timestamp: {:?}, block_found: {}\n",
+                        "ShareAccepted: channel_id: {}, user_identity: {}, share_work: {}, share_sequence_number: {}, share_hash: {}, total_shares_accepted: {}, total_share_work_sum: {}, timestamp: {:?}, block_found: {}\n",
                         channel_id,
                         user_identity,
                         share_work,
+                        share_sequence_number,
                         share_hash,
                         total_shares_accepted,
                         total_share_work_sum,


average-gary:
The latest push implements a simple file writer Persistence which outputs a file. 
```
ShareAccepted: channel_id: 1, share_work: 1, share_sequence_number: 1, share_hash: 00000000026fd2491f671027c9f92f95edcfd487b9ab313e3c8bfee782a95233, total_shares_accepted: 1, total_share_work_sum: 1, timestamp: SystemTime { tv_sec: 1758806242, tv_nsec: 877614000 }
BestDifficultyUpdated: channel_id: 1, new_best_diff: 105.05410207283688, previous_best_diff: 0, timestamp: SystemTime { tv_sec: 1758806242, tv_nsec: 877618000 }
ShareAccepted: channel_id: 1, share_work: 1, share_sequence_number: 2, share_hash: 000000005e15aca4e127fe29c208d802b4cc7b2d8c30d226b56051e287cef015, total_shares_accepted: 2, total_share_work_sum: 2, timestamp: SystemTime { tv_sec: 1758806250, tv_nsec: 256314000 }
```

Of note, I think the Persistence implementation here could use the `status_tx` channel for errors. I'm not sure if SRI would want this basic implementation as part of the repo but I can improve on the error handling if so.

protocols/v2/channels-sv2/src/persistence.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Optional method called when the persistence handler is being dropped.
+    ///
+    /// Implementations can use this for cleanup operations, but should not block.
+    fn shutdown(&self) {}
+
+    #[cfg(feature = "persistence")]
+    fn new(sender: Self::Sender) -> Self;
+
+    #[cfg(not(feature = "persistence"))]
+    fn new() -> Self;
```
plebhash: AFAIU, this seems to be the root-cause motivation for introducing the `persistence ` feature flag.

I feel this is making the overall design a bit convoluted and it would be better if we could avoid it.

Just make one single constructor that takes an `Option`

```rust
fn new(sender: Option<Self::Sender>) -> Self;
```

protocols/v2/channels-sv2/src/persistence.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// No-op method for share accounting events.
+    pub fn persist_event(&self, _event: ShareAccountingEvent) {}
+
+    /// No-op flush method.
+    pub fn flush(&self) {}
+
+    /// No-op shutdown method.
+    pub fn shutdown(&self) {}
+}
```
plebhash: why is there a standalone `impl` block detached from the trait?

```suggestion
```

protocols/v2/channels-sv2/src/persistence.rs
plebhash: or even better, no need for `Option`

just keep this one:
```rust
fn new(sender: Self::Sender) -> Self;
``` 

then we always do `NoPersistance::new(())`, which looks a bit weird but should be hidden away from users that choose `NoPersistance::default()`

plebhash: overall, the main point is that having several different constructor flavors and feature flag is way too convoluted and should be avoided

protocols/v2/channels-sv2/src/persistence.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    #[cfg(feature = "persistence")]
+    fn new(_sender: Self::Sender) -> Self {
+        Self
+    }
+
+    #[cfg(not(feature = "persistence"))]
+    fn new() -> Self {
+        Self
+    }
```
plebhash:
```suggestion
    fn new(_sender: Self::Sender) -> Self {
        Self
    }
```

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -186,7 +188,159 @@ where
             Some(miner_tag_string),
         )
     }
+}
+
+#[cfg(feature = "persistence")]
```
plebhash: I don't think it's a good idea to have these additional constructor flavors

nor the feature flag

maybe this was somewhat inspired by the fact that we have `new_for_pool` and `new_for_job_declarator_client`, but that was a deliberate way to avoid footguns on the user side with regards to coinbase tags

but this is different here... if the user doesn't want persistance, they can simply pass `persistence: NoPersistance::default()` on the constructors that already exist, just like they already do with `DefaultJobStore`

plebhash: think of this as a multidimensional matrix of code maintainability

by introducing new feature flags and different constructor flavors, you're adding new dimensions to this matrix

so now we're making code complexity explode in multiple dimensions

all while the problems we're trying to solve with these new things actually have much simpler solutions

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -83,7 +84,7 @@ use tracing::debug;
 /// - the channel's job factory
 /// - the channel's chain tip
 #[derive(Debug)]
-pub struct ExtendedChannel<'a, J>
+pub struct ExtendedChannel<'a, J, P = NoPersistence>
```
plebhash: this syntax is only useful for `struct`s that support raw initialization without a constructor

for example, if this pseudocode was a valid usage pattern:
```rust
let extended_channel = ExtendedChannel {
   ... // lots of parameters but no `persistence` 
}
```

however, this `struct` is meant to always be initialized with a constructor, so `P = NoPersistence` serves no purpose

```suggestion
pub struct ExtendedChannel<'a, J, P>
```

protocols/v2/channels-sv2/Cargo.toml
```diff
@@ -28,3 +28,4 @@ hashbrown = { version = "0.15.5", optional = true }
 [features]
 default = []
 no_std = ["hashbrown"]
+persistence = []
```
plebhash:
```suggestion
```

I don't think this new feature flag is a good idea, see other comments

protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -87,6 +91,38 @@ impl<'a> ExtendedChannel<'a> {
...
+            extranonce_prefix,
+            rollable_extranonce_size,
+            target,
+            nominal_hashrate,
+            version_rolling,
+            future_jobs: HashMap::new(),
+            active_job: None,
+            past_jobs: HashMap::new(),
+            stale_jobs: HashMap::new(),
+            share_accounting: ShareAccounting::new(channel_id, P::default()),
```
plebhash: why `P::default()`?

plebhash: ah nevermind... this was supposed to be the "no persistence" constructor

I guess it falls into the other comments on the topic

protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -88,10 +93,11 @@ impl<'a> ExtendedChannel<'a> {
         nominal_hashrate: f32,
         version_rolling: bool,
         rollable_extranonce_size: u16,
+        persistence: P,
     ) -> Self {
         Self {
             channel_id,
-            user_identity,
+            user_identity: user_identity.clone(),
```
plebhash: why are we introducing a `clone` here?

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -237,21 +244,27 @@ where
 
         Ok(Self {
             channel_id,
-            user_identity,
+            user_identity: user_identity.clone(),
```
plebhash: why the extra `clone`?

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -237,21 +244,27 @@ where
...
             phantom: PhantomData,
         })
     }
+}
 
+impl<'a, J, P> ExtendedChannel<'a, J, P>
+where
+    J: JobStore<ExtendedJob<'a>>,
+    P: Persistence,
+{
```
plebhash: why are we introducing this new codeblock? 

it's identical to what started on line 107

roles/jd-server/src/lib/config.rs
plebhash, line 1: why is this file being touched?

it seems to have nothing to do with the changes introduced by the PR?

protocols/v2/channels-sv2/src/client/standard.rs
```diff
@@ -54,30 +55,34 @@ pub struct StandardChannel<'a> {
...
         user_identity: String,
         extranonce_prefix: Vec<u8>,
         target: Target,
         nominal_hashrate: f32,
+        persistence: P,
     ) -> Self {
         Self {
             channel_id,
-            user_identity,
+            user_identity: user_identity.clone(),
```
plebhash: why the extra `clone`?

protocols/v2/channels-sv2/src/persistence.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///
+    /// This is typically something like `tokio::sync::mpsc::UnboundedSender<ShareAccountingEvent>`
+    /// or `async_channel::Sender<ShareAccountingEvent>`.
+    type Sender: Clone + Send + 'static;
+
+    /// Sends a share accounting event for persistence.
+    ///
+    /// This method MUST be non-blocking and infallible from the caller's perspective.
+    /// Any persistence failures should be handled internally (e.g., logged, retried, etc.).
+    fn persist_event(&self, event: ShareAccountingEvent);
```
plebhash: @average-gary have you done an actual implementation of this?

if `persist_event` is non-async and we use `tokio::sync::mpsc::UnboundedSender<ShareAccountingEvent>` or `async_channel::Sender<ShareAccountingEvent>` as `Sender`, then how are we able to call `send(...).await`?

plebhash: nevermind, just saw the usage on Pool

roles/pool/src/lib/share_persistence/mod.rs
plebhash, line 1: I guess this could be a module of `stratum-apps`, so that it can be re-used across apps without having to copy-paste the code?

cc @GitGab19

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -44,6 +46,43 @@ use crate::{
...
+
+    fn shutdown(&self) {
+        match self {
+            PoolPersistence::File(p) => p.shutdown(),
+            PoolPersistence::None(p) => p.shutdown(),
+        }
+    }
+
+    fn new(sender: Self::Sender) -> Self {
+        PoolPersistence::File(ShareFilePersistence::new(sender))
```
plebhash: if we're hardcoding the `File` variant into the constructor, what's the point of having both options?

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -143,13 +185,40 @@ impl ChannelManager {
             downstream_receiver,
         };
 
+        // Initialize persistence based on config
```
plebhash, line 153: I don't see any changes to the `.toml` config files?

plebhash: I was able to enable it by manually adding it

if the goal is not to enable it by default, we can leave it commented out

but it should be still on the `.toml` files under `config-examples`, ideally with a comment explaining

protocols/v2/channels-sv2/src/server/extended.rs
average-gary: Was mistake during refactor. Fixed now.

roles/jd-server/src/lib/config.rs
average-gary: Cargo clippy/fmt issue with wrong rust version. Should be fixed.

roles/pool/src/lib/channel_manager/mod.rs
average-gary: This made me reconsider the trait abstraction. 

Introduced in 9e94c36d3f8b9c5f463ecf5b9ad9469ded8a3e32 we have a `Persistence<T>` enum which creation logic for Enabled/Disabled (removing the need for the previous `NoPersistence` struct) and `T` is `PersistenceHandler` that defines the actual persistence functions.

I think this makes things more legible and idiomatic.

roles/pool/src/lib/channel_manager/mod.rs
average-gary: Good catch. Had a locally ignored config I was using for testing.

roles/pool/src/lib/share_persistence/mod.rs
GitGab19: I haven't looked into this yet.

If this module is abstract enough and can be used as it is from other apps, it should definitely be a module of `stratum-apps` 👍

average-gary: Ack. Will refactor as a module in `stratum-apps` and integrate into other roles.

average-gary: Looks like the tightest dependency was the StatusSender. Removing that makes it quite generic. Working on more commits to make it a module.

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -143,13 +150,40 @@ impl ChannelManager {
...
+        let persistence = if let Some(path) = config.share_persistence_file_path() {
+            info!("Initializing file-based share persistence: {}", path);
+            let mut share_file_handler = ShareFileHandler::new(path).await;
+            let sender = share_file_handler.get_sender();
+            let receiver = share_file_handler.get_receiver();
+
+            // Spawn the share file handler task
+            task_manager.spawn(async move {
```
average-gary, line 161: During review, I identified a potential data loss issue with share persistence during pool shutdown.

 ### Current Behavior

 The share persistence task runs in the background with a 1024-event buffer. During Pool shutdown:

(roles/pool/src/lib/mod.rs:165-167):
```rust
  task_manager.abort_all().await;   // Immediately aborts all tasks
  task_manager.join_all().await;    // Waits for aborted tasks
``` 
 Any pending share events in the channel buffer are lost when the task is aborted.

### Impact

 - When: Only during pool shutdown (Ctrl+C, crashes)
 - What's lost: Up to 1024 pending share events
 - Severity: Depends on use case - critical if used for payment calculation, minor if just for auditing

### Possible Solution 
I think this can be accomplished in a follow up PR as follows:

Store the persistence task handle separately and add graceful shutdown:
 1. Stop accepting new mining work
 2. Drop persistence field to signal channel closure
 3. Wait for persistence task to drain its queue naturally
 4. Then abort remaining tasks

This ensures all shares are persisted before shutdown completes.

Alternatively, we can just document for implementers as this is an application side consideration/trade off and not protocol specific.

average-gary:
The fmt check is failing but I don't see that issue in the current commit

Tested with BitAxe locally.

GitGab19:
> The fmt check is failing but I don't see that issue in the current commit
> 
> Tested with BitAxe locally.

On our CI we use `nightly` for `fmt`.

Try by running `cargo +nightly fmt` locally, I just did it locally and it fixed the error you're getting on CI.

protocols/v2/channels-sv2/src/server/share_accounting.rs
```diff
@@ -112,14 +119,30 @@ impl ShareAccounting {
     /// - Records the share hash to detect duplicates.
     pub fn update_share_accounting(
```
EthnTuttle, line 120: Will need sequence number for unique share correlation in an eHash implementation. Channel id + sequence num = unique share identifier between an Upstream/Downstream.

average-gary:
This PR becomes an application concern when #1966 lands. Closing.

PR #1903, opened 2025-09-23
rfc: add metal GPU miner (native sv2)

Sjors:
Make laptops hot again! 🔥

Only for macOS. Off by default, can be enabled with `--features gpu-metal`.

On my M4 MacBook Pro I get about **627 MH/s from the GPU** and **221 MH/s from the CPU**. 

Impressive, given that the CPU has dedicated sha256 instructions. That said, after #1901 I can live with just the CPU performance.

Nonce space is divided between GPU and CPU.

This add `nTime` rolling (one second per wall clock second) to not run out of nonce space when template updates are sparse.

Mostly vibe coded with OpenAI GPT-5.

### DIFF
--- a/roles/Cargo.lock
+++ b/roles/Cargo.lock
@@ -581,6 +581,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "block"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


Shourya742:
@Sjors Can you open this against the [sv2-apps repo](https://github.com/stratum-mining/sv2-apps)

Sjors:
I don't think the marginal benefit is worth the complexity, so just closing this.

PR #1905, opened 2025-09-23
ci(msrv): enforce locked builds in MSRV workflow

Sjors:
Add `--locked` to `cargo build` and `cargo test --no-run` steps in `rust-msrv.yaml` (that have a committed lock file).

Prevents CI from rewriting lockfiles.

Hopefully this produces more better error messages if you accidentally bump a crate.

### DIFF
--- a/.github/workflows/rust-msrv.yaml
+++ b/.github/workflows/rust-msrv.yaml
@@ -27,20 +27,20 @@ jobs:
       - name: Build Protocols
         run: cargo build --manifest-path=protocols/Cargo.toml
       - name: Build Roles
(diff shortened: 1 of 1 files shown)


Sjors:
@lucasbalieiro thanks for the clarification. I removed `--locked` from the crates for which we don't commit a lock file.

https://github.com/stratum-mining/stratum/compare/0463c0c2bc1447c86ed12c88922ca215e0978a0d..02efa2535da422860a6379d9fa92fd68faa07fc4

lucasbalieiro (approved):

Issue #1880, opened 2025-09-16
JDC: Enforce that upstream channel opening is only permitted when at least one downstream is connected without panic


Shourya742:
Follow more here: https://github.com/stratum-mining/stratum/pull/1843#discussion_r2350156129.

PR #1917, opened 2025-09-26
Enforce upstream channel opening

Shourya742:
Closes: #1880

### DIFF
--- a/test/integration-tests/lib/mod.rs
+++ b/test/integration-tests/lib/mod.rs
@@ -168,7 +168,6 @@ pub fn start_jdc(
     );
     let shares_per_minute = 10.0;
     let shares_batch_size = 1;
-    let min_extranonce_size = 4;
     let user_identity = "IT-test".to_string();
     let jdc_signature = "JDC".to_string();
     let jd_client_proxy = JobDeclaratorClientConfig::new(
@@ -181,7 +180,6 @@ pub fn start_jdc(
         tp_config,
         upstreams,
         jdc_signature,
-        min_extranonce_size,
         None,
     );
     let ret = jd_client_sv2::JobDeclaratorClient::new(jd_client_proxy);


roles/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -692,6 +700,8 @@ impl ChannelManager {
                                 upstream_message.user_identity =
                                     self.user_identity.clone().try_into()?;
                                 upstream_message.request_id = 1;
+                                upstream_message.min_extranonce_size =
+                                    self.min_extranonce_size + upstream_message.min_extranonce_size;
```
GitGab19: I would remove the `self.min_extranonce_size` from the ChannelManager in favor of a hardcoded reasonable value.

That would simplify the life to the user who might not know which value to put there.

In the tProxy I defined a constant of 4 bytes to be used for internal search space allocation: https://github.com/stratum-mining/stratum/blob/562c29e573a29e8e6a754b8a8429f28fcf8a1983/roles/translator/src/lib/sv2/channel_manager/channel_manager.rs#L30

I would use again 4 bytes for the JDC, which would cover ~4billions of downstream channels.

bit-aloo: Users can either add that to the config file or stick with the default, which should be fine. Hardcoding makes it rigid.

roles/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -675,12 +682,13 @@ impl ChannelManager {
                         format!("{}#{}", x.user_identity.as_utf8_or_hex(), downstream_id);
                     x.user_identity = user_identity.try_into()?;
 
-                    let downstream_msg = Mining::OpenExtendedMiningChannel(x.clone()).into_static();
+                    let downstream_msg = x.clone().into_static();
```
GitGab19:
```suggestion
                    let downstream_msg = downstream_channel_request.clone().into_static();
```

roles/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -735,12 +745,13 @@ impl ChannelManager {
                     let user_identity = format!("{:?}#{}", x.user_identity, downstream_id);
                     x.user_identity = user_identity.try_into()?;
 
-                    let downstream_msg = Mining::OpenStandardMiningChannel(x.clone()).into_static();
+                    let downstream_msg = x.clone().into_static();
```
GitGab19:
```suggestion
                    let downstream_msg = downstream_channel_request.clone().into_static();
```

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -113,7 +118,9 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
                     Err(e) => {
                         warn!("Failed to build extranonce factory: {e:?}");
                         self.upstream_state.set(UpstreamState::NoChannel);
-                        return (self.upstream_state.get(), None, None);
+                        let close_channel =
+                            close_channel_msg(msg.channel_id, "downstream not available");
```
GitGab19:
```suggestion
                            create_close_channel_msg(msg.channel_id, "downstream not available");
```

roles/jd-client/src/lib/utils.rs
```diff
@@ -437,6 +438,65 @@ impl AtomicUpstreamState {
...
+            PendingChannelRequest::StandardChannel(_) => MESSAGE_TYPE_OPEN_STANDARD_MINING_CHANNEL,
+        }
+    }
+}
+
+/// Creates a [`CloseChannel`] message for the given channel ID and reason.
+///
+/// The `msg` is converted into a [`Str0255`] reason code.  
+/// If conversion fails, this function will panic.
+pub(crate) fn close_channel_msg<'a>(channel_id: u32, msg: &'a str) -> CloseChannel<'a> {
```
GitGab19:
```suggestion
pub(crate) fn create_close_channel_msg<'a>(channel_id: u32, msg: &'a str) -> CloseChannel<'a> {
```

roles/jd-client/src/lib/channel_manager/mod.rs
GitGab19: I still think we should follow what we did in tProxy for the same reasons described above.

We are already giving to users too many config parameters to think about imo, and this should be an opinionated decision we should take (as we did for tProxy).

bit-aloo: Removed the min_extranonce from config, and added a constant. If somebody might need it in future we can add it.

Issue #1877, opened 2025-09-15
JDC `handle_set_extranonce_prefix` should not assume `full_extranonce_len == 32`


plebhash:
as described in #1861, it is not safe to assume that every Pool will establish Extended Channels such that 
`full_extranonce_size = size(extranonce_prefix) + extranonce_size == 32`

the implementation of `handle_set_extranonce_prefix` introduced in #1843 is making such assumption

plebhash:
closed via https://github.com/stratum-mining/stratum/pull/1944

Issue #1841, opened 2025-08-22
Enable Serializing and Encrypting Separately in Codec encoder API


colerob1776:
When broadcasting jobs, it is most performant to serialize once, and share the serialized bytes to each client tcp writer, so it would be good to allow this in the api as `encode(&mut self, data: Arc<Vec<u8>>)` and `encode(&mut self, data: &[u8])`

plebhash:
thanks for sharing @colerob1776 

so if I understand correctly, you want to add new APIs into `codec_sv2` that would take a frame that's already been serialized and then encrypt it?

if you feel this will be valuable for your strategy, and it can still maintain the current APIs, I guess there's no problems with having this contribution

---

I'm curious how you're planning to deal with this on a Channel level, though?

`channel_id` wouldn't be the only field that differs... there's also `NewMiningJob.merkle_root` and `job_id`

currently, `channels_sv2` is designed in a way that the different Channel states are not necessarily synchronized with regards to `job_id`

when a new Channel is created, the `job_id`s are created from scratch on this particular Channel... but it sounds like you want a 1:1 correspondence between `job_id` and `template_id`? or at least you'd need some strategy to enforce this field is synchronized across all Channels

moreover, the Channel states must always be updated in order to allow them to validate shares... and currently, the APIs are shaped such that updating Channel states pretty much correlates to the task of creating the not-yet-serialized `NewMiningJob/NewExtendedMiningJob` and `SetNewPrevHash` messages (with the appropriate `channel_id`, `job_id` and `merkle_root` values)

how would you approach this?

colerob1776:
Yes, we will not be able to use `channel_sv2` for our particular implementation. We are planning to maintain a centralized job store that will retain a 1:1 job_id to template_id as you mentioned so job state is detached from channel state. In this layout, channel state only retains the extranonce1, but the centralized job store also knows about individual extranonces. When a job comes in, if it is an extended channel, all clients receive the same job (assuming they all expect the same template) with the channel_id copies for each channel. For standard channels, the merkle root will be calculated and copied into the buffer in it's respective spot as well

Issue #1861, opened 2025-09-08
need to re-evaluate strategies around extranonce sizes on `channels_sv2`


plebhash:
while reviewing #1791 I tried connecting new tProxy to Braiins Pool, which led me to find yet another incompatibility issue

aside from BIP141 (which is not the topic of discussion here), they also have another peculiar way of sending Extended Jobs

basically, they're not doing search space allocation over Extranonce Prefix... instead, they send `OpenExtendedMiningChannel.Success` with an empty `extranonce_prefix`, and the jobs for this channel will have a unique field in `scriptSig` (which is serialized inside `coinbase_tx_prefix`).

I did some experiments and noticed that the jobs they send have a pretty long `scriptSig` of 70 bytes.

They also send a `OpenExtendedMiningChannel.Success.extranonce_size = 6`.

With the old tProxy, this wasn't an issue, because the old Channel APIs didn't enforce a Full Extranonce Size of 32 bytes.

But now, we're forcing `full_extranonce = extranonce_prefix + extranonce` (or `extranonce1 + extranonce2` in Sv1-land) to ALWAYS be 32 bytes long (despite the fact that this channel was opened with an empty `extranonce_prefix` and only expects a 6 bytes `extranonce`). 

With the 70 bytes already allocated for the `scriptSig`, when we add 32 of  `full_extranonce`, we end up violating the max 100 bytes consensus rule. As a result, we end up with an error that's quite similar to what was reported in #1399 :

Inside `validate_sv1_share`, new tProxy calls `merkle_root_from_path`, which errors out when we try to use Rust Bitcoin's `consensus::deserialize` to get a `Transaction` object, with error message:
```
"data not consumed entirely when explicitly deserializing"
```

But differently from #1399, this has nothing to do with BIP141, which was already fixed on old tProxy via #1832.

This is happening now because new tProxy ~~is~~ was forcing Sv1 workers to work with `extranonce1` + `extranonce2` that always sum up to 32 bytes, and when we try to reconstruct the Coinbase for share validation, we violate consensus (because `scriptSig` is 102 bytes) and get this error.

I guess this is a sign that we should re-evaluate how we're treating the Extranonce Prefix inside `channel_sv2` and `mining_sv2`. Perhaps we don't need to force a full allocation of 32 bytes every time.

plebhash:
I've been investigating this for the past few days, here's an analysis of how we can move away from the assumption that `extranonce_prefix` + `extranonce` is always 32 bytes.

[...]

PR #1921, opened 2025-09-28
Fix set extranonce jdc

Shourya742:
closes: #1877

blocked on: #1841


plebhash:
> @plebhash, for this I had to tweak the channel APIs a bit. Since the `set_new_extranonce` API inherently expects the extranonce length to be `MAX_EXTRANONCE_LEN` (or atleast it has some checks for that), I introduced a `full_extranonce_length` field inside the channels. This replaces the hardcoded `MAX_EXTRANONCE_LEN`, specifically for extended channels.
> 
> Do you think this is the right direction, or do you have a better idea for supporting extranonce lengths other than `MAX_EXTRANONCE_LEN`?

no, I think we're making some mistakes here

let me work on a PR that addresses the notes around `channels_sv2` crate from this comment: https://github.com/stratum-mining/stratum/issues/1861#issuecomment-3294257582

hopefully that will give you a more clear picture

Issue #1881, opened 2025-09-16
pool needs to use `handlers_sv2` crate


plebhash:
as a replacement to `roles_logic_sv2::handlers` module

Issue #1924, opened 2025-09-30
Improved Extranonce Prefix Generation


coleFD:
The current method for extranonce_prefix generation simply increments a uint, which has size/resiliency trade-offs. Theoritically the extranonce_prefix for a single server only needs 2 bytes since the servers will not run more than 65k connections. Incrementing a u16 for extranonce prefix is not wise because connections drop and reconnect all the time meaning extranonce_prefixes will wrap on saturation and get reused which duplicates work. To fix this, the current method allocates a larger extranonce prefix so that it will never wrap, but this comes at the cost of coinbase space.

A more efficient method is to use a constant size bitvec with length equal to the number of connections the stratum server would like to allow. When a client connects, the extranonce prefix will be the index of an open slot in the bitvec. When a client disconnects, the slot can be freed so future connection can reuse the extranonce prefix.

Issue #1885, opened 2025-09-17
new lib crate: Sv2 ↔ Bitcoin Core IPC


plebhash:
up until now, we had two options for deploying a Sv2 Template Provider:
- build a modified `bitcoind` from [`sv2` branch of @Sjors fork](https://github.com/Sjors/bitcoin/tree/sv2) that listens on a dedicated Sv2 port
- build a `bitcoin-mine` sidecar from [`sv2-ipc` branch of @Sjors fork](https://github.com/Sjors/bitcoin/tree/sv2-ipc), which connects via [`capnproto`](https://capnproto.org/) protocol over IPC to `bitcoin-node` that's built from early development iterations of Bitcoin Core

with https://github.com/bitcoin/bitcoin/pull/31802 merged, Bitcoin Core v30.0 is about to add `bitcoin-node` to the official releases, meaning that essentially any software can leverage `capnproto` over IPC and benefit from Bitcoin Core mining interface.

some early work has already been done for this, namely:
- [`github.com/Sjors/sv2-tp`](https://github.com/Sjors/sv2-tp) is a dedicated repo for the C++ implementation of `bitcoin-mine` sidecar mentioned above (now renamed as `sv2-tp`)
- @rustaceanrob has done some [experimental rust integration](https://github.com/rustaceanrob/bitcoin-ipc/tree/mining), described in detail in https://github.com/bitcoin-core/libmultiprocess/issues/174
- @Shourya742 has done some [experimental integration into SRI](https://github.com/Shourya742/stratum/tree/2025-05-05-template-provider-role), where a new crate provides an executable that binds to Bitcoin Core over IPC and listens on a dedicated Sv2 port to perform the Sv2 Template Distribution Protocol

this issue is meant to track the work where, inspired by the references above, we build a crate that can be used as a library to be integrated into Rust apps.

this library shall act as a "translation" layer between Bitcoin Core's `capnproto` Mining Interface over IPC and Sv2 Template Distribution Protocol.

SRI apps using this library (e.g.: JDC, Pool) shall be able to connect to `bitcoin-node` (from official v30+ Bitcoin Core releases) over a UNIX socket and leverage APIs providing Sv2 Template Distribution Protocol messages over an `async` runtime.

rustaceanrob:
[...]

PR #1926, opened 2025-09-30
Refactor Pool

Shourya742:
closes: stratum-mining/stratum#1881.

### DIFF
--- a/roles/pool/src/lib/channel_manager/mod.rs
+++ b/roles/pool/src/lib/channel_manager/mod.rs
@@ -61,8 +61,6 @@ pub struct ChannelManagerData {
     // Mapping of `(downstream_id, channel_id)` → vardiff controller.
     // Each entry manages variable difficulty for a specific downstream channel.
     vardiff: HashMap<(u32, u32), VardiffState>,
-    // Channel_id to downstream_id map
-    channel_id_to_downstream_id: HashMap<u32, u32>,
     // Coinbase outputs
     coinbase_outputs: Vec<u8>,
     // Last new prevhash
@@ -93,7 +91,7 @@ pub struct ChannelManager {
 }
 
 impl ChannelManager {
-    /// Constructor method used to instantiate the Channel Manager
+    /// Constructor method used to instantiate the ChannelManager
     #[allow(clippy::too_many_arguments)]
     pub async fn new(
         config: PoolConfig,
@@ -136,7 +134,6 @@ impl ChannelManager {
             extranonce_prefix_factory_standard,
             downstream_id_factory: AtomicUsize::new(1),
             vardiff: HashMap::new(),
-            channel_id_to_downstream_id: HashMap::new(),
             coinbase_outputs,
             last_future_template: None,
             last_new_prev_hash: None,
@@ -334,26 +331,14 @@ impl ChannelManager {
         Ok(())
     }
 
-    // Removes a downstream entry from the Channel Manager’s state.
+    // Removes a Downstream entry from the ChannelManager’s state.
     //
     // Given a `downstream_id`, this method:
-    // 1. Removes the corresponding downstream from the `downstream` map.
-    // 2. Cleans up all associated channel mappings (both standard and extended) by removing their
-    //    entries from `channel_id_to_downstream_id`.
+    // 1. Removes the corresponding Downstream from the `downstream` map.
     #[allow(clippy::result_large_err)]
     fn remove_downstream(&self, downstream_id: u32) -> PoolResult<()> {
         self.channel_manager_data.super_safe_lock(|cm_data| {
-            if let Some(downstream) = cm_data.downstream.remove(&downstream_id) {
-                downstream.downstream_data.super_safe_lock(|ds_data| {
-                    for k in ds_data
-                        .standard_channels
-                        .keys()
-                        .chain(ds_data.extended_channels.keys())
-                    {
-                        cm_data.channel_id_to_downstream_id.remove(k);
-                    }
-                });
-            }
+            cm_data.downstream.remove(&downstream_id);
         });
         Ok(())
     }
@@ -361,7 +346,7 @@ impl ChannelManager {
     // Handles messages received from the TP subsystem.
     //
     // This method listens for incoming frames on the `tp_receiver` channel.
-    // - If the frame contains a TemplateDistribution message, it forwards it to the   template
+    // - If the frame contains a TemplateDistribution message, it forwards it to the template
     //   distribution message handler.
     // - If the frame contains any unsupported message type, an error is returned.
     async fn handle_template_provider_message(&mut self) -> PoolResult<()> {
--- a/roles/pool/src/lib/template_receiver/mod.rs
+++ b/roles/pool/src/lib/template_receiver/mod.rs
@@ -140,14 +140,14 @@ impl TemplateReceiver {
         Err(PoolError::Shutdown)
     }
 
-    /// Start unified message loop for template receiver.
+    /// Start unified message loop for TemplateReceiver.
     ///
     /// Responsibilities:
     /// - Run handshake (`setup_connection`)
     /// - Send [`CoinbaseOutputConstraints`]
     /// - Handle:
-    ///   - Messages from template provider
-    ///   - Messages from channel manager
+    ///   - Messages from Template Provider
+    ///   - Messages from ChannelManager
     ///   - Shutdown signals (upstream/job-declarator fallback)
     pub async fn start(
         mut self,
@@ -211,7 +211,7 @@ impl TemplateReceiver {
     ///
     /// Routes:
     /// - `Common` messages → handled locally
-    /// - `TemplateDistribution` messages → forwarded to channel manager
+    /// - `TemplateDistribution` messages → forwarded to ChannelManager
     /// - Unsupported messages → logged and ignored
     pub async fn handle_template_provider_message(&mut self) -> PoolResult<()> {
         let mut sv2_frame = self.template_receiver_channel.tp_receiver.recv().await?;
@@ -329,7 +329,7 @@ impl TemplateReceiver {
         Ok(())
     }
 
-    // Performs the initial handshake with template provider.
+    // Performs the initial handshake with Template Provider.
     pub async fn setup_connection(&mut self, addr: String) -> PoolResult<()> {
         let socket: SocketAddr = addr.parse().map_err(|_| {
             error!(%addr, "Invalid socket address");


roles/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -0,0 +1,1043 @@
...
+        let messages = self.channel_manager_data.super_safe_lock(|channel_manager_data| {
+            let Some(downstream) = channel_manager_data.downstream.get_mut(&downstream_id) else {
+                return Err(PoolError::DownstreamIdNotFound);
+            };
+
+            if downstream.requires_custom_work.load(Ordering::SeqCst) {
+                error!("OpenStandardMiningChannel: Standard Channels are not supported for this connection");
+                let open_standard_mining_channel_error = OpenMiningChannelError {
+                    request_id,
+                    error_code: "standard-channels-not-supported-for-custom-work"
```
lucasbalieiro, line 94: The spec doesn’t currently list this as a possible error code:
https://github.com/stratum-mining/sv2-spec/blob/main/05-Mining-Protocol.md#536-openminingchannelerror-server---client

Maybe we could consider either adding a new error code for this scenario there, or rephrasing the spec so it’s clearer that implementations aren’t limited strictly to the listed values. 

As it stands, the phrasing made me think only those values would be spec-compliant for the `OpenMiningChannelError` message, but none of them seems like a good fit here.

lucasbalieiro: Oh no, I just found out  that the Section https://github.com/stratum-mining/sv2-spec/blob/main/03-Protocol-Overview.md#35-error-codes gives us permission to create our own error codes. Sorry

roles/pool/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pool_sv2"
-version = "0.1.3"
+version = "0.1.0"
```
plebhash: why are we walking backwards in time?
```suggestion
version = "1.0.0"
```

roles/pool/src/lib/utils.rs
```diff
@@ -0,0 +1,344 @@
...
+            "writer_task",
+            spawned_at = %format!("{}:{}", caller.file(), caller.line())
+        )));
+    }
+}
+
+/// Deserializes a raw coinbase output into a list of transaction outputs.
+pub fn deserialize_coinbase_outputs(coinbase_output: &[u8]) -> Vec<TxOut> {
+    bitcoin::consensus::deserialize(coinbase_output).expect("Invalid coinbase output")
+}
```
plebhash: plase see #1934

roles/pool/src/lib/mod.rs
```diff
@@ -1,265 +1,177 @@
...
-    /// - Sets up a server for downstream miners to connect.
-    /// - Creates multiple channels for communication between components.
-    /// - Monitors system health and handles shutdown conditions.
-    pub async fn start(&self) -> Result<(), PoolError> {
-        let config = self.config.clone();
-        // Channels for internal communication between Template Receiver and downstream miners.
-        let (status_tx, status_rx) = unbounded(); // Monitors status of both upstream and downstream.
+    /// Starts the Pool main loop.
+    pub async fn start(&self) -> PoolResult<()> {
+        let miner_coinbase_outputs = vec![self.config.get_txout()];
```
plebhash: `miner_coinbase_outputs`?

roles/pool/src/lib/mod.rs
```diff
@@ -1,265 +1,177 @@
...
+                *self.config.authority_secret_key(),
+                self.config.cert_validity_sec(),
+                *self.config.listen_address(),
+                task_manager.clone(),
+                notify_shutdown.clone(),
+                status_sender.clone(),
+                downstream_to_channel_manager_sender.clone(),
+                channel_manager_to_downstream_sender.clone(),
+            )
+            .await?;
```
plebhash, line 130: why are `start` and `start_downstream_server` separate calls?

and why does the second one need a clone?

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+        Ok(())
+    }
+
+    /// The central orchestrator of the Channel Manager.  
+    ///  
+    /// Responsible for receiving messages from all subsystems, processing them,  
+    /// and either forwarding them to the appropriate subsystem or updating  
+    /// the internal state of the Channel Manager as needed.
+    pub async fn start(
+        mut self,
```
plebhash: why do we take ownership here?

roles/pool/src/lib/mod.rs
```diff
@@ -1,265 +1,177 @@
...
-                info!("Pool is not running.");
-            }
-        } else {
-            error!("Failed to access Pool status lock");
         }
+
+        warn!("Graceful shutdown");
+        task_manager.abort_all().await;
+        info!("Joining remaining tasks...");
+        task_manager.join_all().await;
```
plebhash, line 166: why do we do `join_all` after `abort_all`?

bit-aloo: When you call abort_all(), it sends cancellation signals to the tasks, but it doesn’t wait for them to actually finish unwinding. Those tasks might still be in the middle of dropping resources or running Drop implementations. And join_all(), actually wait until every task finishes unwinding.

roles/pool/Cargo.toml
bit-aloo: Given that we’ve practically rewritten the entire application, wouldn’t it make sense to start fresh?

roles/pool/src/lib/mod.rs
bit-aloo: Do you think it’s better to run `downstream_server` inside `ChannelManager::start`? I’d like to make the subsystem boundaries in `mod.rs` more explicit for clarity.

As for the second question, the reason is that `start_downstream_server` takes ownership.

roles/pool/src/lib/channel_manager/mod.rs
bit-aloo: The reason for that is later down the line, when we gonna call the handlers they expect to be called on a mutable reference, so passing the share reference would restrict that.

roles/pool/Cargo.toml
plebhash: https://crates.io/crates/pool_sv2/versions

roles/pool/src/lib/channel_manager/mod.rs
plebhash: 72abf236a347d7bbd0a3925650d181387ccdd158 this compiles without issues here

no hard opinions, but I find it a bit weird pattern to take ownership of `self`

roles/pool/src/lib/task_manager.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// (most recently spawned first).
+    pub async fn join_all(&self) {
+        let handles = {
+            let mut tasks = self.tasks.lock().unwrap();
+            std::mem::take(&mut *tasks)
+        };
+
+        for handle in handles {
+            let _ = handle.await;
+        }
```
plebhash, line 59: why not `tokio::join!`?

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,378 @@
...
+pub struct TemplateReceiverChannel {
+    channel_manager_sender: Sender<TemplateDistribution<'static>>,
+    channel_manager_receiver: Receiver<TemplateDistribution<'static>>,
+    tp_sender: Sender<SV2Frame>,
+    tp_receiver: Receiver<SV2Frame>,
 }
 
-impl TemplateRx {
-    //// Establishes a connection with the template provider and sets up communication channels.
+pub struct TemplateReceiverData;
```
plebhash: what's the purpose of this?

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,378 @@
...
+
+#[allow(warnings)]
+#[derive(Clone)]
+pub struct TemplateReceiver {
+    template_receiver_channel: TemplateReceiverChannel,
+    template_receiver_data: Arc<Mutex<TemplateReceiverData>>,
```
plebhash: what's the purpose of this?

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,378 @@
...
+    },
+};
+
+#[derive(Clone)]
+pub struct TemplateReceiverChannel {
+    channel_manager_sender: Sender<TemplateDistribution<'static>>,
+    channel_manager_receiver: Receiver<TemplateDistribution<'static>>,
+    tp_sender: Sender<SV2Frame>,
+    tp_receiver: Receiver<SV2Frame>,
 }
```
plebhash, line 37: I don't have 100% clarity on my mind yet, but I feel that we should start brainstorming how `bitcoin_core_sv2` will plug into Pool

the current implementation of `bitcoin_core_sv2` has IO in the shape of:
- dedicated `async_channel` for outgoing `TemplateDistribution` messages
- dedicated `async_channel` for incoming `RequestTransactionData`
- dedicated `async_channel` for incoming `SubmitSolution`
- dedicated `async_channel` for incoming `CoinbaseOutputConstraints`

I'm considering unifying the three incoming channels into one of generic `TemplateDistribution` messages, and ignoring messages that aren't of the expected kind

but it makes me wonder about where the proper boundary of `TemplateDistribution` <-> `SV2Frame` should be, given that `bitcoin_core_sv2` doesn't deal with `SV2Frame`

plebhash: or maybe another way to frame this is how can we modularize the interactions with TP such that they can happen either under both scenarios (via `bitcoin_core_sv2` or encrypted frames over TCP) without creating too much spaghetti

plebhash: also worth pointing out that `SetupConnection` will not happen with `bitcoin_core_sv2`

roles/pool/src/lib/template_receiver/mod.rs
plebhash: I couldn't find any meaningful usage of it.

Assuming it would be removed, we would be left with `TemplateReceiverChannel` nested inside `TemplateReceiver`, which tbh feels a bit redundant?

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+        if let Ok((downstream_id, message)) = self
+            .channel_manager_channel
+            .downstream_receiver
+            .recv()
+            .await
+        {
+            let msg = match message {
+                Mining::OpenExtendedMiningChannel(mut m) => {
+                    let user_identity =
+                        format!("{}#{}", m.user_identity.as_utf8_or_hex(), downstream_id);
```
plebhash: why are we doing this?

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+        }
+
+        Ok(())
+    }
+
+    // Runs the vardiff on extended channel.
+    fn run_vardiff_on_extended_channel(
+        downstream_id: u32,
+        channel_id: u32,
+        channel_state: &mut roles_logic_sv2::channels_sv2::server::extended::ExtendedChannel<
```
plebhash: can't we just import on the top?

roles/pool/src/lib/downstream/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/// Represents a downstream client connected to this node.
+#[derive(Clone)]
+pub struct Downstream {
+    pub downstream_data: Arc<Mutex<DownstreamData>>,
+    downstream_channel: DownstreamChannel,
+    pub downstream_id: u32,
+    pub requires_standard_jobs: Arc<AtomicBool>,
+    pub requires_custom_work: Arc<AtomicBool>,
```
plebhash, line 80: why `Arc<AtomicBool>`?

we expect to set these at initialization and never change them again

plebhash: ok nevermind, I see it's actually initialized at `handle_setup_connection`

roles/pool/config-examples/pool-config-hosted-tp-example.toml
```diff
@@ -29,4 +29,4 @@ pool_signature = "Stratum V2 SRI Pool"
 tp_address = "75.119.150.111:8442"
 tp_authority_public_key = "9bwHCYnjhbHm4AS3pWg9MtAH83mzWohoJJJDELYBqZhDNqszDLc"
 shares_per_minute = 1.0
```
plebhash: let's take the opportunity of this PR and increase this value

1 share per minute is too low for most testing scenarios

```suggestion
shares_per_minute = 6.0
```

roles/pool/config-examples/pool-config-local-tp-example.toml
```diff
@@ -27,4 +27,4 @@ pool_signature = "Stratum V2 SRI Pool"
 # Local TP (this is pointing to localhost so you must run a TP locally for this configuration to work)
 tp_address = "127.0.0.1:8442"
 shares_per_minute = 1.0
```
plebhash: let's take the opportunity of this PR and increase this value

1 share per minute is too low for most testing scenarios

```suggestion
shares_per_minute = 6.0
```

roles/pool/src/lib/task_manager.rs
bit-aloo: Join macro can't take join handles only plain future.

roles/pool/src/lib/template_receiver/mod.rs
bit-aloo: Aligns TemplateReceiver with other subsystems by adding channel and data counterparts.

roles/pool/src/lib/channel_manager/mod.rs
bit-aloo: because channel manager won't know from which downstream the channel opening message came from.

roles/pool/src/lib/template_receiver/mod.rs
bit-aloo: Makes sense to use a single channel for these three messages. I’d also suggest adding a dedicated module in sv2-translation to handle the flow: sv2-frame → template-distribution message → Bitcoin Core–specific message.

roles/pool/src/lib/template_receiver/mod.rs
plebhash: how? it's an empty struct

roles/pool/src/lib/template_receiver/mod.rs
plebhash: > I’d also suggest adding a dedicated module in sv2-translation to handle the flow: sv2-frame → template-distribution message → Bitcoin Core–specific message.

tbh that's the entire scope of `bitcoin_core_sv2` (with exception of `SV2Frame`, which it's agnostic to).

"translating" Sv2 TDP to Bitcoin Core IPC is a highly stateful process, which requires spawning tasks and keeping track of different kinds of IPC clients (thread, mining, template) across the spawned tasks. I don't think it makes sense to have parts of it split across more than one crate.

if the suggestion here is to expand the scope of `stratum-translation` (which I assume is what you're referring as `sv2-translation`), that would pretty much render the idea of `bitcoin_core_sv2` as a standalone crate obsolete.

also we would be moving the scope to something broader than simply Sv1 / Sv2 translation (which was the original purpose of the crate)

cc @GitGab19

roles/pool/src/lib/channel_manager/mod.rs
plebhash: I had a bad feeling about this on JDC and couldn't put my finger on it, but now I feel I have more clarity.

it feels really awkward to be forced to modify the contents of a message just because we're lacking context across API boundaries

tbh this feels like a limitation/flaw in the interfaces established by `handlers_sv2`

when I was designing `pub trait Sv2MiningServerHandler` for `sv2-services` (f.k.a `tower-stratum`) my rationale was:

"the entity implementing this method is managing the state of multiple clients, and it cannot infer which client the message is coming from by simply looking at the message, therefore the method must also inform the `client_id`"

and in the end, almost every method on this trait ended up with a `client_id` parameter:
https://github.com/plebhash/sv2-services/blob/2be0a6834e1e2e3398f646bf89c8b06dcc889f8b/src/server/service/subprotocols/mining/handler.rs#L33

maybe there's other solutions to this problem, and I'm not gonna block this PR with this comment, since my suggestion would require changes to `handlers_sv2` and JDC

but I think it's still worth pointing out, especially because whoever is using that crate is inevitably going to face the same obstacles and be forced to come up with their own hacks to bypass it

roles/pool/src/lib/mod.rs
plebhash: no hard opinions on this

roles/pool/src/lib/channel_manager/mod.rs
plebhash: cc @GitGab19

bit-aloo: I agree, I don’t like it either. I’m currently working on removing it and finding a better way to provide context to the channel manager. For now, though, I only have some workarounds no solid solution yet.

GitGab19: Mm I guess we are encountering this issue just because we decided to implement handlers on the ChannelManager and not on the Downstream struct.

So I would say it's more because of our implementation decisions than a flaw in `handlers_sv2`.

That said, I would take the opportunity to improve our handlers trait, maybe by introducing an optional `downstream/client_id` as @plebhash did in the `sv2-services` example mentioned above.

bit-aloo: Can be done, and should be the cleanest among all the solution I have been trying locally.

roles/pool/src/lib/template_receiver/mod.rs
GitGab19: I don't get why `stratum-translation` should include what `bitcoin_core_sv2` is gonna do.

I would proceed with the direction we decided, by having a dedicated crate for that.

plebhash: I'm also not a fan of mixing things with `stratum-translation`.

just to clarify what I meant here:

> or maybe another way to frame this is how can we modularize the interactions with TP such that they can happen either under both scenarios (via `bitcoin_core_sv2` or encrypted frames over TCP) without creating too much spaghetti

when I said modularize, I meant within Pool code, and not by changing things on other crates.

I was just referring to the fact that `TemplateReceiverChannel.{tp_sender, tp_receiver}` assumes that the only possible way to interact with a TP is via `SV2Frame`, while ideally there should be no `SV2Frame` when communicating with Bitcoin Core via IPC.

as stated above, there are dedicated channels for `TemplateDistribution` message IO for interacting with `BitcoinCoreSv2` struct

plebhash: so we need to think about modularization inside Pool code in terms of optionality:

Pool could interact with a TP either via:
- IPC
- TCP

where TCP implies a few things that don't really happen on IPC scenario, namely:
- serde of `SV2Frame`
- noise encryption
- `SetupConnection` exchange

roles/pool/config-examples/pool-config-hosted-tp-example.toml
```diff
@@ -28,5 +28,5 @@ pool_signature = "Stratum V2 SRI Pool"
 # Hosted testnet TP 
 tp_address = "75.119.150.111:8442"
 tp_authority_public_key = "9bwHCYnjhbHm4AS3pWg9MtAH83mzWohoJJJDELYBqZhDNqszDLc"
-shares_per_minute = 1.0
-share_batch_size = 10
+shares_per_minute = 6.0
```
GitGab19, line 31: A side note: I would increase the `shares_per_minute` in JDC and tProxy config files at this point.

roles/pool/src/lib/template_receiver/mod.rs
plebhash: the comments above are non-blocking for this PR, perhaps it's best to address them in the near future on other PRs

however I wanted to highlight that this part of the code will likely undergo some changes in the near future, motivated by #1885

roles/pool/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -0,0 +1,1013 @@
...
+            };
+
+
+            let pool_coinbase_output = TxOut {
+                value: Amount::from_sat(last_future_template.coinbase_tx_value_remaining),
+                script_pubkey: self.coinbase_reward_script.script_pubkey(),
+            };
+
+            downstream.downstream_data.super_safe_lock(|downstream_data| {
+                if !downstream.requires_standard_jobs.load(Ordering::SeqCst) && downstream_data.group_channels.is_none() {
```
GitGab19, line 117: Why are we checking again `downstream.requires_standard_jobs.load(Ordering::SeqCst)`?

It's already being checked at line 90.

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+            let range_1 = 0..POOL_SEARCH_SPACE_BYTES;
+            (
+                0..range_1.start,
+                range_1.clone(),
+                range_1.end..MAX_EXTRANONCE_LEN,
+            )
+        };
+
+        let make_extranonce_factory = || {
+            ExtendedExtranonce::new(range_0.clone(), range_1.clone(), range_2.clone(), None)
```
GitGab19: Why aren't we using the `server_id` from config here?

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+                .expect("Failed to create ExtendedExtranonce with valid ranges")
+        };
+
+        let extranonce_prefix_factory_extended = make_extranonce_factory();
+        let extranonce_prefix_factory_standard = make_extranonce_factory();
```
GitGab19, line 129: Also, aren't we gonna create two identical factories here?

How are we avoiding the search space overlapping here?

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+    // - If the frame contains any unsupported message type, an error is returned.
+    async fn handle_template_provider_message(&mut self) -> PoolResult<()> {
+        if let Ok(message) = self.channel_manager_channel.tp_receiver.recv().await {
+            self.handle_template_distribution_message_from_server(None, message)
+                .await?;
+        }
+        Ok(())
+    }
+
+    async fn handle_downstream_message(&mut self) -> PoolResult<()> {
```
GitGab19:
```suggestion
    async fn handle_downstream_mining_message(&mut self) -> PoolResult<()> {
```

roles/pool/src/lib/downstream/common_message_handler.rs
GitGab19, line 1: I would call this file `common_message_handler`.

roles/pool/src/lib/channel_manager/template_distribution_message_handler.rs
GitGab19, line 1: i would call this file `template_distribution_message_handler`.

roles/pool/src/lib/channel_manager/mining_message_handler.rs
GitGab19, line 1: I would call this file `mining_message_handler`.

roles/pool/src/lib/downstream/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    // Performs the initial handshake with a downstream peer.
+    async fn setup_connection_with_downstream(&mut self) -> PoolResult<()> {
+        let mut frame = self.downstream_channel.downstream_receiver.recv().await?;
+
+        let Some(message_type) = frame.get_header().map(|m| m.msg_type()) else {
+            return Err(PoolError::UnexpectedMessage(0));
+        };
+        if message_type == MESSAGE_TYPE_SETUP_CONNECTION {
```
GitGab19, line 209: Add a comment here to make it clear the first ever message received by the client must always be the `SetupConnection`.

roles/pool/src/lib/downstream/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+            .map_err(|e| {
+                error!(?e, "Downstream send failed");
+                PoolError::Noise(Error::ExpectedIncomingHandshakeMessage)
+            })?;
+
+        Ok(())
+    }
+
+    // Handles incoming messages from the downstream peer.
+    async fn handle_downstream_message(self) -> PoolResult<()> {
```
GitGab19:
```suggestion
    async fn handle_downstream_mining_message(self) -> PoolResult<()> {
```

roles/pool/src/lib/template_receiver/common_message_handler.rs
```diff
@@ -1,77 +1,58 @@
...
         info!(
-            "Received RequestTransactionDataSuccess for template: {}",
-            m.template_id
+            "Received ChannelEndpointChanged with channel id: {}",
+            msg.channel_id
         );
-        debug!("RequestTransactionDataSuccess: {}", m);
-        // Just ignore tx data messages this are meant for the declarators
-        Ok(SendTo::None(None))
+        Ok(())
```
GitGab19: I think we should return an error here, I don't see why a TP should ever send a `ChannelEndpointChanged` to the Pool.

roles/pool/src/lib/template_receiver/common_message_handler.rs
GitGab19, line 1: I would call this file `common_message_handler`.

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,375 @@
...
-            match TcpStream::connect(address).await {
-                Ok(stream) => break stream,
-                Err(err) => {
-                    warn!("Failed to connect to {}: {}. Retrying...", address, err);
-                    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    /// - Opens a TCP connection
+    /// - Performs Noise handshake
+    /// - Spawns IO tasks for inbound/outbound frames
+    ///
+    /// Retries up to 3 times before returning [`JDCError::Shutdown`].
```
GitGab19:
```suggestion
    /// Retries up to 3 times before returning [`PoolError::Shutdown`].
```

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,375 @@
...
-                Ok(frame) => {
-                    handle_result!(status_tx, Self::send(self_.clone(), frame).await);
-                }
-                Err(_e) => {
-                    // return submit error
-                    todo!()
-                }
-            };
-        }
+    /// Build and send [`CoinbaseOutputConstraints`] upstream TP.
```
GitGab19:
```suggestion
    /// Build and send [`CoinbaseOutputConstraints`] to the TP.
```

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,375 @@
...
+        let socket: SocketAddr = addr.parse().map_err(|_| {
+            error!(%addr, "Invalid socket address");
+            PoolError::InvalidSocketAddress(addr.clone())
+        })?;
+
+        info!(%socket, "Building setup connection message for upstream");
+        let setup_msg = get_setup_connection_message_tp(socket);
+        let frame: StdFrame = Message::Common(setup_msg.into()).try_into()?;
+
+        info!("Sending setup connection message to upstream");
```
GitGab19:
```suggestion
        info!("Sending SetupConnection message to the Template Provider");
```

roles/pool/src/lib/template_receiver/mod.rs
```diff
@@ -1,239 +1,375 @@
...
+    }
+
+    // Performs the initial handshake with template provider.
+    pub async fn setup_connection(&mut self, addr: String) -> PoolResult<()> {
+        let socket: SocketAddr = addr.parse().map_err(|_| {
+            error!(%addr, "Invalid socket address");
+            PoolError::InvalidSocketAddress(addr.clone())
+        })?;
+
+        info!(%socket, "Building setup connection message for upstream");
```
GitGab19:
```suggestion
        debug!(%socket, "Building SetupConnection message to the Template Provider");
```

roles/pool/src/lib/channel_manager/mod.rs
plebhash: this pattern comes from the old pool implementation and it was a deliberate decision when I introduced it there

theoretically, we could use just one factory for both, but it would lead to wasted search space, because the internal increments for extended prefixes incur in increments for standard prefixes as well

for example, if there's only one single `extranonce_prefix_factory`:
- call `extranonce_prefix_factory.next_prefix_standard`: `0000000000000001`
- call `extranonce_prefix_factory.next_prefix_standard`: `0000000000000002`
- call `extranonce_prefix_factory.next_prefix_extended`: `00000001`
- call `extranonce_prefix_factory.next_prefix_standard`: `0000000100000003`

but if we have two separate `extranonce_prefix_factory_standard` + `extranonce_prefix_factory_extended`:
- call `extranonce_prefix_factory_standard.next_prefix_standard`: `0000000000000001`
- call `extranonce_prefix_factory_standard.next_prefix_standard`: `0000000000000002`
- call `extranonce_prefix_factory_extended.next_prefix_extended`: `00000001`
- call `extranonce_prefix_factory_standard.next_prefix_standard`: `0000000000000003`

as long as we have discipline to always call `next_prefix_standard` and `next_prefix_extended` on the appropriate factory, we avoid wasting search space

if we fuck up and mix these calls, then yes, we're giving room for duplicated search spaces

it's not ideal, and it's one of the reasons I wanted to completely refactor `ExtendedExtranonce` for some time now

---

one could argue it's actually one of the issues that #1924 would optimize for

plebhash: > How are we avoiding the search space overlapping here?

with our current `ExtendedExtranonce` implementation, prefixes for standard channels will always be bigger in length

I mean, yes... as the extended channel rolls its extranonce, there will be potential collisions with standard channels for specific values

for example:
- standard channel prefix: `0000000100000001` (which could happen regardless of having one or two factories)
- extended channel prefix: `00000001`

when the rolled extranonce is `00000001`, assuming all the other rollable variables (ntime, version, nonce) are also identical, then these two channels have a collision

but that's different than saying two channels have identical search spaces, because in a fraction of a second later, the rolled extranonce is already going to be different

so the collision will happen for very specific combinations of values (few amongst zillions)

plebhash: and btw the collision scenario described above cannot be avoided by using one single factory

GitGab19: Crystal clear @plebhash , thanks!

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -61,8 +61,6 @@ pub struct ChannelManagerData {
     // Mapping of `(downstream_id, channel_id)` → vardiff controller.
     // Each entry manages variable difficulty for a specific downstream channel.
     vardiff: HashMap<(u32, u32), VardiffState>,
-    // Channel_id to downstream_id map
-    channel_id_to_downstream_id: HashMap<u32, u32>,
```
plebhash: why are we removing this now?

bit-aloo: Not really needed anymore, it was originally used to map downstreams to channels so the channel manager could identify each downstream. I’ve already removed its usage from the mining_message_handler, just forgot to clean it up from the main struct.

PR #1928, opened 2025-10-01
Cleanup roles logic sv2 utils

### LINKED ISSUE #1448: `roles_logic_sv2`: refactor `utils` module

lucasbalieiro:
closes #1448**

This PR diverges a bit from the solution originally proposed in the issue:
* Dropped `roles_logic_sv2::utils::Id` in favor of using `std::sync::AtomicU32` (see https://github.com/stratum-mining/stratum/pull/1928#discussion_r2398643402)
* Removed `utils::GroupId`, since it was only referenced in its own unit tests.
* Cleaned up other unused functions in `utils`.
* moved `roles_logic_sv2::vardiff` to `channels_sv2::vardiff` (see https://github.com/stratum-mining/stratum/pull/1928#discussion_r2405407767)

A major bump was required  for this crate since we are changing the visibility of functions and moving some of them to other crates. 
`roles_logic_sv2` from `4.0.0` to `5.0.0`

### DIFF
--- a/protocols/v2/roles-logic-sv2/src/errors.rs
+++ b/protocols/v2/roles-logic-sv2/src/errors.rs
@@ -4,9 +4,11 @@
 //! module. It includes the [`Error`] enum for representing various errors.
 
 use bitcoin::hashes::FromSliceError;
-use channels_sv2::server::error::{ExtendedChannelError, GroupChannelError, StandardChannelError};
-use channels_sv2::target::InputError;
-use channels_sv2::vardiff::error::VardiffError;
+use channels_sv2::{
+    server::error::{ExtendedChannelError, GroupChannelError, StandardChannelError},
+    target::InputError,
+    vardiff::error::VardiffError,
+};
 use codec_sv2::binary_sv2::Error as BinarySv2Error;
 use mining_sv2::ExtendedExtranonceError;
 use parsers_sv2::AnyMessage as AllMessages;
--- a/roles/jd-client/src/lib/channel_manager/mod.rs
+++ b/roles/jd-client/src/lib/channel_manager/mod.rs
@@ -11,7 +11,6 @@ use async_channel::{Receiver, Sender};
 use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
 use stratum_common::{
     network_helpers_sv2::noise_stream::NoiseTcpStream,
-    roles_logic_sv2::channels_sv2::{Vardiff, VardiffState},
     roles_logic_sv2::{
         self,
         channels_sv2::{
@@ -23,6 +22,7 @@ use stratum_common::{
                 },
                 standard::StandardChannel,
             },
+            Vardiff, VardiffState,
         },
         codec_sv2::{Responder, Sv2Frame},
         handlers_sv2::{
--- a/roles/pool/src/lib/mining_pool/message_handler.rs
+++ b/roles/pool/src/lib/mining_pool/message_handler.rs
@@ -12,15 +12,17 @@ use std::{
 };
 use stratum_common::roles_logic_sv2::{
     bitcoin::{consensus::Decodable, transaction::TxOut, Amount},
-    channels_sv2::server::{
-        error::{ExtendedChannelError, StandardChannelError},
-        extended::ExtendedChannel,
-        group::GroupChannel,
-        jobs::job_store::DefaultJobStore,
-        share_accounting::{ShareValidationError, ShareValidationResult},
-        standard::StandardChannel,
+    channels_sv2::{
+        server::{
+            error::{ExtendedChannelError, StandardChannelError},
+            extended::ExtendedChannel,
+            group::GroupChannel,
+            jobs::job_store::DefaultJobStore,
+            share_accounting::{ShareValidationError, ShareValidationResult},
+            standard::StandardChannel,
+        },
+        Vardiff, VardiffState,
     },
-    channels_sv2::{Vardiff, VardiffState},
     codec_sv2::binary_sv2::Str0255,
     errors::Error,
     handlers::mining::{ParseMiningMessagesFromDownstream, SendTo, SupportedChannelTypes},
--- a/roles/pool/src/lib/mining_pool/mod.rs
+++ b/roles/pool/src/lib/mining_pool/mod.rs
@@ -45,13 +45,15 @@ use stratum_common::{
     roles_logic_sv2::{
         self,
         bitcoin::{Amount, TxOut},
-        channels_sv2::server::{
-            extended::ExtendedChannel,
-            group::GroupChannel,
-            jobs::{extended::ExtendedJob, job_store::DefaultJobStore, standard::StandardJob},
-            standard::StandardChannel,
+        channels_sv2::{
+            server::{
+                extended::ExtendedChannel,
+                group::GroupChannel,
+                jobs::{extended::ExtendedJob, job_store::DefaultJobStore, standard::StandardJob},
+                standard::StandardChannel,
+            },
+            VardiffState,
         },
-        channels_sv2::VardiffState,
         codec_sv2::{
             self, binary_sv2::U256, HandshakeRole, Responder, StandardEitherFrame, StandardSv2Frame,
         },
--- a/roles/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
+++ b/roles/translator/src/lib/sv1/sv1_server/difficulty_manager.rs
@@ -5,8 +5,7 @@ use crate::{
 use async_channel::Sender;
 use std::{collections::HashMap, sync::Arc, time::Duration};
 use stratum_common::roles_logic_sv2::{
-    channels_sv2::target::hash_rate_to_target,
-    channels_sv2::Vardiff,
+    channels_sv2::{target::hash_rate_to_target, Vardiff},
     mining_sv2::{SetTarget, Target, UpdateChannel},
     parsers_sv2::Mining,
     utils::Mutex,
--- a/roles/translator/src/lib/sv1/sv1_server/sv1_server.rs
+++ b/roles/translator/src/lib/sv1/sv1_server/sv1_server.rs
@@ -23,8 +23,7 @@ use std::{
     },
 };
 use stratum_common::roles_logic_sv2::{
-    channels_sv2::target::hash_rate_to_target,
-    channels_sv2::{Vardiff, VardiffState},
+    channels_sv2::{target::hash_rate_to_target, Vardiff, VardiffState},
     mining_sv2::{CloseChannel, SetTarget, Target},
     parsers_sv2::Mining,
     utils::Mutex,


protocols/v2/channels-sv2/src/id_factory.rs
```diff
@@ -0,0 +1,27 @@
+/// Generator of unique IDs for channels and groups.
+///
+/// It keeps an internal counter, which is incremented every time a new unique id is requested.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct IdFactory {
```
plebhash: I wonder if we really need to keep this around?

can you try to replace it with `AtomicU32`? on the places where it's used?

lucasbalieiro: @plebhash I tested your idea in this commit on the PR: [https://github.com/stratum-mining/stratum/pull/1928/commits/0878ee6add498ca97b248e9340d6502fae895da1](https://github.com/stratum-mining/stratum/pull/1928/commits/0878ee6add498ca97b248e9340d6502fae895da1)

If we decide to go this route, I’ll restructure the commit history accordingly. For now, I used `SeqCst` as the default since I wasn’t entirely sure which ordering would be most appropriate. [`Relaxed`](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Relaxed) might be sufficient here, but I’m not 100% confident.

plebhash: `Relaxed` is more than fine.

the only important thing to guarantee is atomicity, which `AtomicU32` enforces by default

lucasbalieiro: https://github.com/stratum-mining/stratum/pull/1928/commits/f14c8712772cb61c8c98d57db91cdf47d6253f99

I switched to `Relaxed` and fixed a few issues that were causing the integration tests to fail.

As you’ll see in the changes, this approach requires some discipline when using `AtomicU32`. For example, this pattern will come up often:

```rust
channel_manager_data.channel_id_factory.fetch_add(1, Ordering::Relaxed);
let group_channel_id = channel_manager_data.channel_id_factory.load(Ordering::Relaxed);
```

This is because `fetch_add` returns the value *BEFORE* incrementing, while the old `IdFactory` was doing the opposite, which is what caused some of the integration test failures.

I considered encapsulating this logic in an `impl` for `AtomicU32` with a `next()` function, but that would basically recreate the `IdFactory`, which we’d still have to maintain inside `channels_sv2`. So for this first approach, I went with handling it inline in each specific case.


What do you think?

plebhash: > This is because fetch_add returns the value BEFORE incrementing, while the old IdFactory was doing the opposite, which is what caused some of the integration test failures.

what was failing on Integration Tests?

plebhash: tbh `fetch_add` then `load` indeed feels a bit counter-intuitive and error-prone, so I would keep using only `fetch_add` (without `load` afterwards)

overall, as long as it's consistent, I don't think it really matters if we're fetching the value before or after the incrementing

I mean, what difference does it make if we start `channel_ids` at 0 or 1?

that's why I'm curious on what kind of Integration Test failures this caused

lucasbalieiro: > I mean, what difference does it make if we start channel_ids at 0 or 1?

I had the same thought. In fact, you can see in this commit [https://github.com/stratum-mining/stratum/commit/0878ee6add498ca97b248e9340d6502fae895da1](https://github.com/stratum-mining/stratum/commit/0878ee6add498ca97b248e9340d6502fae895da1) that I tried exactly that, since I was expecting it to work as long as consistency was maintained. And at first it looked fine, the CI was green for that commit without the `load`:
[https://github.com/stratum-mining/stratum/actions/runs/18198821078/job/51812194633#step:2:86](https://github.com/stratum-mining/stratum/actions/runs/18198821078/job/51812194633#step:2:86)

**But… I think that was just luck.**

After another PR was merged, I had to rebase, and then the CI started failing on this test:
`jds_wont_exit_upon_receiving_unexpected_txids_in_provide_missing_transaction_success`

It was failing because it exceeded the `cargo nextest` retries. I amended the commit a few times just to retrigger CI:

* [https://github.com/stratum-mining/stratum/actions/runs/18203514992/job/51828059585#step:2:86](https://github.com/stratum-mining/stratum/actions/runs/18203514992/job/51828059585#step:2:86)
* [https://github.com/stratum-mining/stratum/actions/runs/18202583966/job/51824932625#step:2:86](https://github.com/stratum-mining/stratum/actions/runs/18202583966/job/51824932625#step:2:86)

Both failed again with the same error.

When I checked locally, I saw the same thing: sometimes the test would pass, but most of the time it would consume all retries and fail. In many runs it would even get stuck indefinitely trying to find shares:

image  

To rule out other recent merges, I ran a `git bisect`. It pointed straight to the commit without the `load` ([https://github.com/stratum-mining/stratum/commit/0878ee6add498ca97b248e9340d6502fae895da1](https://github.com/stratum-mining/stratum/commit/0878ee6add498ca97b248e9340d6502fae895da1)) as the cause.

I reverted it, and locally the test was green again.

Then I revisited what `IdFactory` was doing and noticed it was returning the value *after* incrementing. I modified the code locally to match that behavior, and the test passed deterministically.

I committed that change here (using the `fetch_add` + `load` pattern):
[https://github.com/stratum-mining/stratum/pull/1928/commits/f14c8712772cb61c8c98d57db91cdf47d6253f99](https://github.com/stratum-mining/stratum/pull/1928/commits/f14c8712772cb61c8c98d57db91cdf47d6253f99)

And now CI is green again.

bit-aloo: I’d suggest keeping the `IdFactory`, but updating its state to use an `AtomicUsize` instead of `usize`. Initialize it with `AtomicUsize::new(1)`, and adjust the API so that `next` takes `&self` instead of `&mut self`. This avoids requiring users to manage their own atomics, which would make usage awkward and error-prone.

```rust
use std::sync::atomic::{AtomicUsize, Ordering};

#[derive(Debug, PartialEq, Eq, Clone)]
pub struct IdFactory {
    state: AtomicUsize,
}

impl IdFactory {
    /// Creates a new [`IdFactory`] with the counter starting at `1`.
    pub fn new() -> Self {
        Self { state: AtomicUsize::new(1) }
    }

    /// Atomically increments the counter and returns the new ID.
    #[allow(clippy::should_implement_trait)]
    pub fn next(&self) -> usize {
        self.state.fetch_add(1, Ordering::SeqCst)
    }
}
```

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -219,7 +168,7 @@ pub fn merkle_root_from_path<T: AsRef<[u8]>>(
 /// ## Components
 /// * `coinbase_id`: Coinbase transaction hash.
 /// * `path`: List of transaction hashes. Should be converted from [`binary_sv2::U256`].
-pub fn merkle_root_from_path_<T: AsRef<[u8]>>(coinbase_id: [u8; 32], path: &[T]) -> [u8; 32] {
+fn merkle_root_from_path_<T: AsRef<[u8]>>(coinbase_id: [u8; 32], path: &[T]) -> [u8; 32] {
```
GitGab19: Did you try to remove this?

Because we already have a copy of it in `channels_sv2`, so I don't know it this one is really used by anything.

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -445,7 +394,7 @@ pub fn target_to_difficulty(target: Target) -> f64 {
 }
 
 /// Converts a `u128` to a [`U256`].
-pub fn from_u128_to_u256(input: u128) -> U256Primitive {
+fn from_u128_to_u256(input: u128) -> U256Primitive {
```
GitGab19: Same question for this one

protocols/v2/roles-logic-sv2/src/utils.rs
GitGab19: Did you try to remove this?

Because we already have a copy of it in channels_sv2, so I don't know it this one is really used by anything.

protocols/v2/channels-sv2/src/id_factory.rs
plebhash: what worries me here is that we don't use `IdFactory` only for Channel Ids, but also for a whole range of other things... to name a few:
- Request Ids
- Token generation
- Share sequence numbers

while one could argue that share sequence numbers are directly related to Channel management, things like Request Ids and Token generation are pretty orthogonal to that, and I'm not sure it should be coming from a "repurposed" API from `channels_sv2`

so by going with this idea, we're creating an API that is not really aligned with its actual usage.

protocols/v2/roles-logic-sv2/src/utils.rs
lucasbalieiro: It is being used by the translator proxy and the mining device.

but, since theres a copy on channels_sv2, I'll delete here and replace where needed

lucasbalieiro: @GitGab19 while addressing this I noticed a few things worth mentioning:

* As you said,  we have a copy of these functions inside `channels_sv2`, but there are other public functions in `roles_logic_sv2::utils` depending on them, so I couldn’t remove them right away.
* However, after analyzing usage, I saw that the *public* functions calling these internals are only being used by `jd-server`.

Because of that, I moved the functions that are only used by `jd-server` into a file inside the `jd-server` crate. This prevents them from staying exposed through `roles_logic_sv2`, which could lead to more dependencies on it. Since we’re planning to refactor `jd-server` soon, I figured it makes sense to isolate these parts now and later reevaluate whether they should be deleted entirely.

For reference, see this commit (I also added a comment at the top of the file explaining the situation):
https://github.com/stratum-mining/stratum/pull/1928/commits/13dc60ccb9e3b43201b8ae2c4f459abd70d83416

As for the rest of the duplicated functions in `roles_logic_sv2` and `channels_sv2`:
I made `channels_sv2::merkle_root` public to remove the dependency on `roles_logic_sv2` and use it directly from `channels_sv2`.


What do you think of the idea?

plebhash: during vinteum office hours we worked on 6bca6ea36c5e0f95a4a61014ca5d1ed5956d11ba

it leveraged APIs from Rust Bitcoin that made the `block_creator` module unnecessary

this commit will be squashed later

lucasbalieiro: squashed @ f8e2f748d352f81f91cec19a3069bf20144f35aa

protocols/v2/channels-sv2/src/id_factory.rs
plebhash: after many hours debugging, it seems that `mining_device_sv1` starts hanging after the removal of `IdFactory`

why on hell that would happen is beyond me, and I don't think it would be productive to even try to figure that out

https://github.com/stratum-mining/stratum/pull/1937 deprecates it in favor of `minerd`

rebasing on top of that makes `jds_wont_exit_upon_receiving_unexpected_txids_in_provide_missing_transaction_success` pass deterministically

bit-aloo: > what worries me here is that we don't use `IdFactory` only for Channel Ids, but also for a whole range of other things... to name a few:
> 
> * Request Ids
> * Token generation
> * Share sequence numbers
> 
> while one could argue that share sequence numbers are directly related to Channel management, things like Request Ids and Token generation are pretty orthogonal to that, and I'm not sure it should be coming from a "repurposed" API from `channels_sv2`
> 
> so by going with this idea, we're creating an API that is not really aligned with its actual usage.

Based on this comment, I think it makes more sense to keep this in utils (as it was originally) rather than inside channel_sv2.

protocols/v2/roles-logic-sv2/src/utils.rs
```diff
@@ -173,59 +118,6 @@ impl<T> Mutex<T> {
...
-pub fn merkle_root_from_path_<T: AsRef<[u8]>>(coinbase_id: [u8; 32], path: &[T]) -> [u8; 32] {
-    match path.len() {
-        0 => coinbase_id,
-        _ => reduce_path(coinbase_id, path),
-    }
-}
-
 // Helper function to format bytes as hex string
 // useful for visualizing targets
 pub fn bytes_to_hex(bytes: &[u8]) -> String {
```
GitGab19: Can't we remove this?

We have a public copy of it in `channels_sv2`.

protocols/v2/roles-logic-sv2/src/utils.rs
GitGab19: Shouldn't we have this one in the vardiff module? It seems it's only used there

GitGab19: Did you try to remove this one?

It seems it's not used anywhere (except unit tests)

GitGab19: Shouldn't we have this one in the vardiff module? It seems it's only used there

protocols/v2/channels-sv2/src/id_factory.rs
plebhash: the goal of this discussion thread is to get rid of `IdFactory`, not to find the ideal place for it to continue existing

protocols/v2/roles-logic-sv2/src/utils.rs
lucasbalieiro: tproxy was using it. Adapted @ [75d299c](https://github.com/stratum-mining/stratum/pull/1928/commits/75d299c09cf2f76dfbf86c25896cc83ebf821685)

protocols/v2/roles-logic-sv2/src/utils.rs
lucasbalieiro: thanks, I forgot it there. [9b7290e](https://github.com/stratum-mining/stratum/pull/1928/commits/9b7290e5b12019949bcc297a5ba45e058f568be9)

lucasbalieiro: `channels_sv2` seems to have function of the same "domain", should I try to adapt into `channels_sv2` and call it inside vardiff?

Since `channels_sv2` already has a copy of `hash_rate_to_target`, maybe make it sense to put this one there too

https://github.com/stratum-mining/stratum/blob/abcc79dd2326d77091ce8aed97f5dcc586ce5d22/protocols/v2/channels-sv2/src/target.rs#L109

plebhash: we've been discussing making `vardiff` into a standalone crate, but maybe it could also become a module of `channels_sv2`?

at least from a high level, it feels like a conceptual fit

GitGab19: Yes, I agree with this suggestion.

In the end we run vardiff on channels, everytime.

lucasbalieiro: `vardiff` moved to `channels_sv2` and adapted where needed.

Now, `roles_logic_sv2::utils` has only things regarding our custom Mutex.


Addressed @
- [move roles_logic_sv2::vardiff to channels_sv2::vardiff](https://github.com/stratum-mining/stratum/pull/1928/commits/2b8d08a485847508e72b9bd8e73135c08ccc1a28)
- [drops helper functions used by roles_logic_sv2::vardiff since it was moved to channels_sv2::vardiff](https://github.com/stratum-mining/stratum/pull/1928/commits/c885fc3ce004319e7888e332c194d718848761ee)

lucasbalieiro: addressed while moving vardiff

protocols/v2/roles-logic-sv2/src/utils.rs
lucasbalieiro: addressed while moving `vardiff` to `channels_sv2`

protocols/v2/roles-logic-sv2/src/utils.rs
GitGab19, line 1: Now it's very clear how this module only lives because of the custom mutex.

I would consider deprecating it in a follow-up PR as we recently discussed.

cc @plebhash @Shourya742

plebhash: https://github.com/stratum-mining/stratum/issues/1940

PR #1930, opened 2025-10-02
tProxy: improve Sv1Server messages received in downstream.rs

GitGab19:
This PR improves the way messages received by the Sv1Server in downstream.rs are managed.

It fixes an issue reported by @plebhash and @lucasbalieiro for which the `jd_tproxy_integration` test was sometimes failing on macos.

### DIFF
--- a/roles/translator/src/lib/sv1/downstream/downstream.rs
+++ b/roles/translator/src/lib/sv1/downstream/downstream.rs
@@ -198,10 +198,11 @@ impl Downstream {
                         .load(std::sync::atomic::Ordering::SeqCst)
                 });
 
-                // Sv1 handshake complete - send messages immediately, or if it's a queued
-                // Sv1 handshake message response
-                if handshake_complete || is_queued_sv1_handshake_response {
-                    if let Message::Notification(notification) = &message {
+                // Handle messages based on message type and handshake state
+                if let Message::Notification(notification) = &message {
+                    // For notifications (mining.set_difficulty, mining.notify), only send if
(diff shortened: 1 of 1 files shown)


roles/translator/src/lib/sv1/downstream/downstream.rs
```diff
@@ -461,7 +479,7 @@ impl Downstream {
 
         // Send cached messages in correct order: set_difficulty first, then notify
         if let Some(set_difficulty_msg) = cached_set_difficulty {
-            debug!("Down: Sending cached mining.set_difficulty after handshake completion");
+            info!("Down: Sending cached mining.set_difficulty after handshake completion");
```
bit-aloo:
```suggestion
            debug!("Down: Sending cached mining.set_difficulty after handshake completion");
```

roles/translator/src/lib/sv1/downstream/downstream.rs
```diff
@@ -486,7 +504,7 @@ impl Downstream {
         }
 
         if let Some(notify_msg) = cached_notify {
-            debug!("Down: Sending cached mining.notify after handshake completion");
+            info!("Down: Sending cached mining.notify after handshake completion");
```
bit-aloo:
```suggestion
            debug!("Down: Sending cached mining.notify after handshake completion");
```

lucasbalieiro (approved):
Tested locally on my setup and now the test is green deterministically.

Image

PR #1931, opened 2025-10-02
Improve crates re-exports

### LINKED ISSUE #1911: reconsider strategy for re-exporting crates
we can (and should) rely on semver CI for breaking changes

unfortunately, `cargo semver-checks` is not able to pick up breaking changes on re-exported APIs

---

## context 

we had a footgun during release v1.5.0, where we didn't realize a MAJOR bump on `binary_codec_sv2` actually forced us to do a MAJOR bump on:
- `binary_sv2` (because it re-exports `binary_codec_sv2`)
- `roles_logic_sv2` (because it re-exports `binary_sv2`)
- `stratum-common` (because it re-exoprts `roles_logic_sv2`)

but CI didn't warn us, so we just bumped PATCH on `binary_sv2` (to `v3.0.1`), which was now pointing to the MAJOR bumped `binary_codec_sv2` (now `v3.0.0`)

after publishing the PATCHed version of `binary_sv2`, we realized that we actually broke compilation of the already published `stratum-common` `v3.0.0`, because it lists `binary_sv2` as `"^3.0.0"`, which started mixing old `mining_sv2`, `job_declaration_sv2` and `template_distribution_sv2` with the new (incompatible with new APIs of `binary_codec_sv2`)

the solution was simple: publish another PATCH bumped `binary_sv2` (`v3.0.2`) that pointed to the previous `binary_codec_sv2` (`v2.0.0`)

then `stratum-common` `v3.0.0` was buildable again

---

the ugly details described above shouldn't matter, and that's the point

by doing so many nested layers of re-exported crates, we are leaving way too much complexity to be managed by humans, which will inevitably continue to cause issues every release

after this experience, IMHO we should aim to avoid re-exporting crates as much as possible

GitGab19:
Closes #1911


.github/workflows/integration-tests.yaml
```diff
@@ -29,4 +29,4 @@ jobs:
 
       - name: Integration Tests
         run: |
-         RUST_BACKTRACE=1 RUST_LOG=debug cargo nextest run --manifest-path=test/integration-tests/Cargo.toml --features sv1 --nocapture
+         RUST_BACKTRACE=1 RUST_LOG=debug cargo nextest run --manifest-path=test/integration-tests/Cargo.toml --nocapture
```
plebhash, line 32: is this intentional?

GitGab19: Yes, I removed that feature from the `integration_tests_sv2` crate, since I thought there's no real need for that.

If you think differently, I can easily reintroduce it.

protocols/v2/codec-sv2/src/lib.rs
GitGab19, line 1: Because of the re-exports removal, I think we should bump MAJOR for `codec_sv2`, wdyt @plebhash ?

protocols/v2/roles-logic-sv2/src/lib.rs
GitGab19, line 1: Because of the re-exports removal, I think we should bump MAJOR for roles_logic_sv2, wdyt @plebhash ?

plebhash: I agree.

protocols/v2/codec-sv2/src/lib.rs
plebhash: I agree.

.github/workflows/integration-tests.yaml
plebhash: no, it's a good idea.

protocols/v2/codec-sv2/src/lib.rs
GitGab19: Done here: 0df8a1d56856af6987c4f86544a14aa607755d00

protocols/v2/roles-logic-sv2/src/lib.rs
GitGab19: Done here: 4d48be9b61930f637ad6ff3fb7123ff6c385c565

common/src/lib.rs
```diff
@@ -1,8 +1,27 @@
...
+//! `stratum_common` is the central hub for all Stratum protocol crates.
+//! It re-exports all the low-level v1 and v2 crates to provide a single entry point
+//! for accessing the Stratum protocol implementations.
+//!
+//! ## Features
+//!
+//! - `with_buffer_pool`: Enables buffer pooling for improved memory management and performance in
+//!   the binary serialization, framing, and codec layers.
 
-#[cfg(feature = "with_network_helpers")]
```
bit-aloo, line 6: why are we no longer importing the network-helpers from the stratum common?

roles/jd-client/Cargo.toml
```diff
@@ -18,7 +18,8 @@ path = "src/lib/mod.rs"
 [dependencies]
 async-channel = "1.5.1"
 buffer_sv2 = { path = "../../utils/buffer" }
```
bit-aloo: this buffer_sv2 import can be removed, we can directly use it from stratum-common

roles/jd-server/Cargo.toml
```diff
@@ -17,7 +17,8 @@ name = "jd_server"
 path = "src/lib/mod.rs"
 
 [dependencies]
-stratum-common = { path = "../../common", features = ["with_network_helpers"] }
+stratum-common = { path = "../../common" }
+network_helpers_sv2 = { path = "../roles-utils/network-helpers", features = ["with_buffer_pool"] }
 async-channel = "1.5.1"
 buffer_sv2 = { path = "../../utils/buffer" }
```
bit-aloo: This can be removed in favor of stratum common

roles/pool/Cargo.toml
```diff
@@ -18,7 +18,8 @@ path = "src/lib/mod.rs"
 
 [dependencies]
 async-channel = "1.5.1"
-stratum-common = { path = "../../common", features = ["with_network_helpers"] }
+stratum-common = { path = "../../common" }
+network_helpers_sv2 = { path = "../roles-utils/network-helpers", features = ["with_buffer_pool"] }
 buffer_sv2 = { path = "../../utils/buffer" }
```
bit-aloo: This can be removed in favor of stratum common

roles/test-utils/mining-device/Cargo.toml
```diff
@@ -20,7 +20,8 @@ path = "src/lib/mod.rs"
 
 
 [dependencies]
-stratum-common = { path = "../../../common", features = ["with_network_helpers"] }
+stratum-common = { path = "../../../common" }
+network_helpers_sv2 = { path = "../../roles-utils/network-helpers", features = ["with_buffer_pool"] }
 async-channel = "1.5.1"
 buffer_sv2 = { path = "../../../utils/buffer" }
```
bit-aloo: We can remove buffer_sv2 in favor of stratum-common

roles/roles-utils/rpc/Cargo.toml
```diff
@@ -14,7 +14,7 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-stratum-common = { path = "../../../common", version = "4.0.1" }
+stratum-common = { path = "../../../common", version = "5.0.0" }
```
bit-aloo: I forgot, what was the reason for keeping the version field along with the import, instead of just using:

 ```rust
 stratum-common = { path = "../../../common" }
 ```

common/src/lib.rs
GitGab19: because we decided to have `network_helpers_sv2` using `stratum-common`, instead of `stratum-common` re-exporting `network_helpers_sv2`

roles/roles-utils/rpc/Cargo.toml
GitGab19: IIRC we need to specify versions for every dependency on crates we publish, otherwise `cargo publish` fails.

common/src/lib.rs
bit-aloo: Hmmm, in that case, the dependency tree for a role would look something like this. Do you think it makes sense to move network-helpers into stratum-common as a module? Making one stop solution for all sv2 needs.

Screenshot from 2025-10-06 17-49-34

GitGab19: If we move `network_helpers_sv2` into `stratum-common`, we should do the same for every other higher-level crate (for example the new `bitcoin_core_sv2` crate).

cc @plebhash

plebhash: at this point do we still have `apps-utils`? and in which repo does it live?

GitGab19: What about having a new crate called `stratum-apps` (or similar) re-exporting `stratum-common`, `network_helpers_sv2`, `bitcoin_core_sv2`, and all the other `apps-utils` crates?

This new crate would be the umbrella crate used by sv2-apps, and it would live in the new `sv2-apps` repo.

bit-aloo: Will network-helpers and bitcoin_core_sv2 be part of this crate as modules, or will they be re-exported? If they’re included as modules, it should work fine, the dependency tree will be much leaner.

GitGab19: They could probably be modules of this new crate.

bit-aloo: Its green on my end then.

GitGab19: In https://github.com/stratum-mining/stratum/commit/27d24f349d6a3d4a82586a65df0804c22927ceb8 I created a new crate called `stratum-apps`, placed inside `roles/`, which contains three modules:
- `network_helpers`
- `config_helpers`
- `rpc`

It also re-exports stratum-common and it's the main entry-point for apps.

What about starting to use `stratum-core` (0.1.0) instead of the current `stratum-common`?
cc @plebhash @Shourya742

GitGab19: I renamed `stratum-common` to `stratum-core` in https://github.com/stratum-mining/stratum/commit/c477f8671ed5beaf0fc8ab6a88ff6e68941ce2f2, as we discussed during today's meeting.

.github/workflows/semver-check.yaml
```diff
@@ -113,19 +109,16 @@ jobs:
...
+        working-directory: protocols/stratum-translation
         run: cargo semver-checks
 
-      - name: Run semver checks for roles/roles-utils/rpc
-        working-directory: roles/roles-utils/rpc
-        run: cargo semver-checks
+      # TODO: Uncomment this when the stratum-core crate is published to crates.io
+      # - name: Run semver checks for stratum-core
+      #  working-directory: stratum-core
+      #  run: cargo semver-checks
```
plebhash, line 123: what happens to CI if we run it while these lines are not commented out?

lucasbalieiro: cargo semver will throw an error because it computes the versioning based on the version published to crates.io.

Something like this:
```bash
error: handlers_sv2 not found in registry (crates.io). For workarounds check https://github.com/obi1kenobi/cargo-semver-checks#does-the-crate-im-checking-have-to-be-published-on-cratesio
```

https://github.com/stratum-mining/stratum/actions/runs/17951039589/job/51050097621#step:8:362

plebhash: but `stratum-core` and `stratum-apps` already exist on `crates.io`

they're only on version `0.0.0`

lucasbalieiro:
```bash
    Building stratum-core v0.1.0 (current)
       Built [   8.321s] (current)
     Parsing stratum-core v0.1.0 (current)
      Parsed [   0.002s] (current)
    Building stratum-core v0.0.0 (baseline)
error: running cargo-doc on crate 'stratum-core' failed with output:
-----
warning: placeholder v0.0.0 (/Users/lucasbalieiro/Projects/stratum2/stratum-core/target/semver-checks/registry-stratum_core-0_0_0-01666ec060466c14) ignoring invalid dependency `stratum-core` which is missing a lib target
error: no library targets found in package `stratum-core`

-----

error: failed to build rustdoc for crate stratum-core v0.0.0
note: this is usually due to a compilation error in the crate,
      and is unlikely to be a bug in cargo-semver-checks
note: the following command can be used to reproduce the compilation error:
      cargo new --lib example &&
          cd example &&
          echo '[workspace]' >> Cargo.toml &&
          cargo add stratum-core@=0.0.0 &&
          cargo check

error: aborting due to failure to build rustdoc for crate stratum-core v0.0.0
```

GitGab19: Yeah this is the same error we got on CI before commenting those lines: https://github.com/stratum-mining/stratum/actions/runs/18351984191/job/52274159686

plebhash: I see!

should we log an issue so we don't forget about this? or is it already covered under the scope of some issue that already exists?

GitGab19: Just created the issue here: https://github.com/stratum-mining/stratum/issues/1943

stratum-core/Cargo.lock
lucasbalieiro, line 1: should we keep this lockfile?

Also, the .gitignore is poiting to the old common/Cargo.lock

GitGab19: Good catch, thanks!

I deleted it for both `stratum-core` and `stratum-apps`. Also added them to the `.gitignore`.

roles/stratum-apps/README.md
```diff
@@ -0,0 +1,97 @@
...
+Complete Stratum V2 application development kit - all utilities in one crate.
+
+## Overview
+
+`stratum-apps` consolidates all the role utilities into a single, well-organized crate with modules instead of separate crates. This approach provides:
+
+- **Better dependency management** - No duplicate dependencies
+- **Faster compilation** - Single crate compilation unit
+- **Easier maintenance** - All utilities in one place
+- **Cleaner imports** - Everything available through one crate
```
plebhash: these sentences are framing things in relationship to the past approach: "instead of", "better", "faster", "easier", "cleaner"

that's a description of what's going on in this PR, but I don't think we should use that kind of language on the `README.md` itself.

a `README.md` file should describe the crate as it is now, not in comparison to what it once was.

GitGab19: Good catch, fixed.

stratum-core/Cargo.lock
lucasbalieiro: The file is back again

roles/stratum-apps/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+//! - `mining_device` - Everything needed for mining device applications
+//!
+//! ## Modules
+//!
+//! - [`network_helpers`] - High-level networking utilities for SV2 connections
+//! - [`config_helpers`] - Configuration management and parsing utilities
+//! - [`rpc`] - RPC utilities with custom serializable types (`Hash`, `BlockHash`, `Amount`)
+
+/// Re-export all the modules from `stratum_core`
+pub use stratum_core;
```
lucasbalieiro, line 28: Since we are going to make the apps use the low-level crates from crates.io

Should we re-export this? Because the `stratum-core` is re-exporting the crates from path

plebhash: I guess the answer is:
- `stratum-apps` points to `crates.io`
- apps point to local `stratum-apps`

GitGab19: It's exactly how @plebhash explained 💯

stratum-core/Cargo.lock
GitGab19: I might have re-introduced it while solving conflicts, thanks for spotting it.

PR #1934, opened 2025-10-02
`channels_sv2::outputs::deserialize_outputs`

plebhash:
with new roles refactoring we're falling into a pattern of copy-pasting (and effectively duplicating) a lot of "utils" code

#1843 introduced a `jd_client::utils::deserialize_coinbase_outputs` that is now being copy-pasted into a redundant `pool_sv2::utils::deserialize_coinbase_outputs` on #1926

this PR introduces a new `channels_sv2::outputs::deserialize_outputs` to avoid this

the old `channels_sv2::template` module was renamed into `channels_sv2::outputs`

### DIFF
--- a/common/Cargo.lock
+++ b/common/Cargo.lock
@@ -279,7 +279,7 @@ dependencies = [
 
 [[package]]
 name = "channels_sv2"
-version = "1.0.2"
+version = "2.0.0"
 dependencies = [
  "binary_sv2",
  "bitcoin",
--- a/protocols/v2/channels-sv2/Cargo.toml
+++ b/protocols/v2/channels-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "channels_sv2"
-version = "1.0.2"
+version = "2.0.0"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"
--- a/protocols/v2/roles-logic-sv2/Cargo.toml
+++ b/protocols/v2/roles-logic-sv2/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 [dependencies]
 bitcoin = { version = "0.32.5" }
-channels_sv2 = { path = "../channels-sv2", version = "^1.0.0" }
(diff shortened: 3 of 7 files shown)


roles/jd-client/src/lib/utils.rs
```diff
@@ -490,7 +484,7 @@ impl PendingChannelRequest {
 ///
 /// The `msg` is converted into a [`Str0255`] reason code.  
 /// If conversion fails, this function will panic.
-pub(crate) fn create_close_channel_msg<'a>(channel_id: u32, msg: &'a str) -> CloseChannel<'a> {
+pub(crate) fn create_close_channel_msg(channel_id: u32, msg: &str) -> CloseChannel<'_> {
```
bit-aloo, line 487: why this?

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -148,7 +148,8 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
                         data.last_new_prev_hash.clone(),
                     ) {
                         let request_id = data.request_id_factory.next();
-                        let outputs = deserialize_coinbase_outputs(&data.coinbase_outputs);
+                        let outputs = deserialize_outputs(data.coinbase_outputs.clone())
```
bit-aloo: I see we have introduced new error types, can't we bubble up this error?

bit-aloo: Same for all the place we using this.

roles/jd-client/src/lib/utils.rs
plebhash: when I ran clippy for my changes, it did this

I isolated it into a standalone commit to avoid getting it mixed up with the other changes

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
plebhash: done, but for many cases I had to move these into codeblocks in the beginning of methods

otherwise it would conflict with the local context

PR #1937, opened 2025-10-03
deprecate `mining-device-sv1` in favor of `minerd`

plebhash:
discussions in https://github.com/stratum-mining/stratum/pull/1928#discussion_r2398643402 started drifting into really puzzling directions

the simple removal of `IdFactory` on apps started making `mining-device-sv1` hang during execution of `jds_wont_exit_upon_receiving_unexpected_txids_in_provide_missing_transaction_success`

I have no idea why that would happen, and even theorized unrelated things that turned out to be true
 
but most importantly, this is not the first, second, or third time that `mining-device-sv1` introduces puzzling behavior on Integration Tests that leads to many wasted hours debugging random things

this PR replaces it with `minerd`, which is a way more mature and stable codebase


test/integration-tests/lib/minerd/process.rs
```diff
@@ -0,0 +1,670 @@
...
+        let minerd_dir = current_dir.join("minerd");
+
+        let os = std::env::consts::OS;
+        let arch = std::env::consts::ARCH;
+        let download_filename = get_minerd_filename(os, arch)?;
+
+        if !minerd_dir.exists() {
+            fs::create_dir_all(minerd_dir.clone()).expect("failed to create minerd directory");
+            let download_endpoint = format!(
+                "https://github.com/plebhash/cpuminer/releases/download/v{VERSION_MINERD}/"
```
plebhash: ideally this should be
```suggestion
                "https://github.com/stratum-mining/cpuminer/releases/download/v{VERSION_MINERD}/"
```

test/integration-tests/lib/minerd/process.rs
```diff
@@ -0,0 +1,670 @@
...
+    match (os, arch) {
+        ("macos", "aarch64") => Ok(format!(
+            "pooler-cpuminer-{VERSION_MINERD}-arm64-apple-darwin.tar.gz"
+        )),
+        // todo
+        // ("macos", "x86_64") =>
+        // "pooler-cpuminer-{VERSION_MINERD}-x86_64-apple-darwin.tar.gz",
+        ("linux", "x86_64") => Ok(format!(
+            "pooler-cpuminer-{VERSION_MINERD}-x86_64-linux-gnu.tar.gz"
+        )),
```
plebhash: @GitGab19 can you create a `pooler-cpuminer-2.5.1-x86_64-apple-darwin.tar.gz`?

you can simply do a manual build of https://github.com/pooler/cpuminer/tree/v2.5.1 and then compress the executable into the tarball

so we can host here: https://github.com/plebhash/cpuminer/releases/tag/v2.5.1

(ideally transfering repo to `stratum-mining` org instead of my personal account)

test/integration-tests/Cargo.toml
```diff
@@ -19,14 +19,14 @@ once_cell = { version = "1.19.0", default-features = false }
 rand = { version = "0.9.0", default-features = false, features = ["thread_rng"] }
 tar = { version = "0.4.41", default-features = false }
 tokio = { version="1.44.1", default-features = false,  features = ["tracing"] }
+tokio-util = { version = "0.7", default-features = false }
```
bit-aloo, line 22: why are we adding tokio-utils?

test/integration-tests/lib/sv1_minerd/process.rs
```diff
@@ -0,0 +1,670 @@
...
+                            Ok((downstream_stream, _)) => {
+                                info!("New connection from minerd");
+
+                                // Connect to upstream server
+                                match TcpStream::connect(upstream_address).await {
+                                    Ok(upstream_stream) => {
+                                        // Split streams for bidirectional communication
+                                        let (downstream_read, downstream_write) = downstream_stream.into_split();
+                                        let (upstream_read, upstream_write) = upstream_stream.into_split();
+
```
bit-aloo, line 201: lets use buffered readers and writers.

test/integration-tests/lib/minerd/process.rs
```diff
@@ -0,0 +1,670 @@
...
+            Err(_) => {
+                error!("Mutex poisoned in Drop implementation, cannot kill process cleanly");
+            }
+        }
+    }
+}
+
+/// Measures the hashrate of the local minerd binary in benchmark mode
+/// Returns the hashrate in hashes per second
+pub async fn measure_hashrate() -> Result<f64, MinerdError> {
```
bit-aloo: This is defined above as well?

test/integration-tests/lib/minerd/process.rs
GitGab19: Just built and sent to you through DMs.

test/integration-tests/Cargo.toml
plebhash: for `CancellationToken`

test/integration-tests/lib/sv1_minerd/process.rs
plebhash: what would be the motivation?

here's an analysis by `claude-4-sonnet`:

> ## Benefits of Buffering
> - Better performance for small, frequent reads/writes
> - Reduced system calls by batching I/O operations
> - More efficient for line-based or message-based protocols
> ## Considerations for This Use Case
> - However, in this specific proxy scenario:
> - The current code reads in 4KB chunks - already reasonably efficient
> - It's doing binary proxying - not line-based reading where BufReader shines most
> - You'd need to call flush() on BufWriter after each write to ensure data is sent immediately (important for real-time > proxying)
> - The proxy methods would need signature changes to accept the buffered types
> ## Recommendation
> - For a TCP proxy like this, the current approach is actually quite reasonable.

test/integration-tests/lib/minerd/process.rs
plebhash: cleaned up

test/integration-tests/lib/minerd/process.rs
plebhash: great, thanks! just added to the code

can you fetch this branch and assert that it's working? just run the integration tests once and check if nothing breaks

GitGab19: I had some issues for which minerd was extracted in `minerd/GNUSparseFile.0/` directory when running tests.

That made tests fail because they were unable to find the binary.

I fixed it with `claude-4.5-sonnet` and pushed a new commit on a new branch called `fix-minerd-macos-x86_64` of my fork: eabdd4562741e1dedb24898b2f58ea9d8da34861

Feel free to cherry-pick it and test on your machine.

plebhash: done

lucasbalieiro (approved):
AMEN!

I have been running the tests in:
```bash
Darwin 25.0.0 Darwin Kernel Version 25.0.0: Wed Sep 17 21:42:08 PDT 2025; root:xnu-12377.1.9~141/RELEASE_ARM64_T8132 arm64
```

Tests green deterministically at the most recent [commit](https://github.com/stratum-mining/stratum/pull/1937/commits/a41321ec9a5e082768281773b0c716552cc9012b).

Image

plebhash:
> Tests green deterministically at the most recent [commit](https://github.com/stratum-mining/stratum/pull/1937/commits/a41321ec9a5e082768281773b0c716552cc9012b).

`jds_should_not_panic_if_jdc_shutsdown` is still flaky on my setup, the assertion on `TcpListener::bind` fails sometimes

but that's totally unrelated to this PR, given that this test doesn't even have a CPU miner

also I assume this could be potentially improved with #1778

test/integration-tests/lib/sv1_minerd/mod.rs
GitGab19, line 1: I would call the module `minerd_sv1`/`sv1_minerd` or something similar to make it very clear it's a Sv1 miner.

plebhash: done

PR #1944, opened 2025-10-10
refine extranonce assumptions

plebhash:
close stratum-mining/stratum#1861

summary of changes:
- simplified `try_strip_bip141` for better maintainability
- validation of extended shares now takes `rollable_extranonce_size` into consideration
- `full_extranonce_size` of extended channels (both server and client sides) is always calculated based on `extranonce_prefix` plus `rollable_extranonce_size`, and no longer assumed to be `MAX_EXTRANONCE_LEN`
- `rollable_extranonce_size` never changes across the lifespan of a channel (even after `SetExtranoncePrefix`)
- `JobFactory` takes `full_extranonce_size` as a parameter on a few methods, which is used to set appropriate `scriptSig_len` values inside `coinbase_tx_prefix`
- `GroupChannel` takes `full_extranonce_size` as a parameter on its constructor, since `JobFactory` now needs it
- adapts JDC with fixed extranonce assumptions
- adapts Pool with fixed extranonce assumptions

### DIFF
--- a/protocols/v2/channels-sv2/src/client/extended.rs
+++ b/protocols/v2/channels-sv2/src/client/extended.rs
@@ -14,6 +14,7 @@ use crate::{
     },
     merkle_root::merkle_root_from_path,
     target::{bytes_to_hex, target_to_difficulty, u256_to_block_hash},
+    MAX_EXTRANONCE_PREFIX_LEN,
 };
 use alloc::{format, string::String, vec, vec::Vec};
 use binary_sv2::{self, Sv2Option};
@@ -27,7 +28,7 @@ use bitcoin::{
 };
 use mining_sv2::{
     NewExtendedMiningJob, SetCustomMiningJob, SetCustomMiningJobSuccess,
-    SetNewPrevHash as SetNewPrevHashMp, SubmitSharesExtended, Target, MAX_EXTRANONCE_LEN,
+    SetNewPrevHash as SetNewPrevHashMp, SubmitSharesExtended, Target,
 };
 use tracing::debug;
 
@@ -146,7 +147,7 @@ impl<'a> ExtendedChannel<'a> {
         &mut self,
         new_extranonce_prefix: Vec<u8>,
     ) -> Result<(), ExtendedChannelError> {
-        if new_extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
+        if new_extranonce_prefix.len() > MAX_EXTRANONCE_PREFIX_LEN {
             return Err(ExtendedChannelError::NewExtranoncePrefixTooLarge);
         }
 
--- a/protocols/v2/channels-sv2/src/server/extended.rs
+++ b/protocols/v2/channels-sv2/src/server/extended.rs
@@ -48,6 +48,7 @@ use crate::{
         share_accounting::{ShareAccounting, ShareValidationError, ShareValidationResult},
     },
     target::{bytes_to_hex, hash_rate_to_target, target_to_difficulty, u256_to_block_hash},
+    MAX_EXTRANONCE_PREFIX_LEN,
 };
 use binary_sv2::{self};
 use bitcoin::{
@@ -56,7 +57,7 @@ use bitcoin::{
     transaction::TxOut,
     CompactTarget, Target as BitcoinTarget,
 };
-use mining_sv2::{SetCustomMiningJob, SubmitSharesExtended, Target, MAX_EXTRANONCE_LEN};
+use mining_sv2::{SetCustomMiningJob, SubmitSharesExtended, Target};
 use std::{collections::HashMap, convert::TryInto, marker::PhantomData};
 use template_distribution_sv2::{NewTemplate, SetNewPrevHash as SetNewPrevHashTdp};
 use tracing::debug;
@@ -217,7 +218,7 @@ where
             return Err(ExtendedChannelError::RequestedMaxTargetOutOfRange);
         }
 
-        if extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
+        if extranonce_prefix.len() > MAX_EXTRANONCE_PREFIX_LEN {
             return Err(ExtendedChannelError::ExtranoncePrefixTooLarge);
         }
 
@@ -293,7 +294,7 @@ where
         &mut self,
         extranonce_prefix: Vec<u8>,
     ) -> Result<(), ExtendedChannelError> {
-        if extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
+        if extranonce_prefix.len() > MAX_EXTRANONCE_PREFIX_LEN {
             return Err(ExtendedChannelError::ExtranoncePrefixTooLarge);
         }
 
--- a/protocols/v2/channels-sv2/src/client/standard.rs
+++ b/protocols/v2/channels-sv2/src/client/standard.rs
@@ -14,6 +14,7 @@ use crate::{
     },
     merkle_root::merkle_root_from_path,
     target::{bytes_to_hex, target_to_difficulty, u256_to_block_hash},
+    MAX_EXTRANONCE_PREFIX_LEN,
 };
 use alloc::{format, string::String, vec::Vec};
 use binary_sv2::{self, Sv2Option};
@@ -24,7 +25,7 @@ use bitcoin::{
 };
 use mining_sv2::{
     NewExtendedMiningJob, NewMiningJob, SetNewPrevHash as SetNewPrevHashMp, SubmitSharesStandard,
-    Target, MAX_EXTRANONCE_LEN,
+    Target,
 };
 use tracing::debug;
 
@@ -110,7 +111,7 @@ impl<'a> StandardChannel<'a> {
         &mut self,
         extranonce_prefix: Vec<u8>,
     ) -> Result<(), StandardChannelError> {
-        if extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
+        if extranonce_prefix.len() > MAX_EXTRANONCE_PREFIX_LEN {
             return Err(StandardChannelError::NewExtranoncePrefixTooLarge);
         }
 
--- a/protocols/v2/channels-sv2/src/lib.rs
+++ b/protocols/v2/channels-sv2/src/lib.rs
@@ -15,6 +15,9 @@
 //! - [`client`] module is `no_std` compatible. To enable it build the crate with `no_std` feature.
 #![cfg_attr(feature = "no_std", no_std)]
 
+/// Maximum length for extranonce prefixes in bytes
+const MAX_EXTRANONCE_PREFIX_LEN: usize = 32;
+
 #[cfg(not(feature = "no_std"))]
 pub mod server;
 
--- a/protocols/v2/channels-sv2/src/server/standard.rs
+++ b/protocols/v2/channels-sv2/src/server/standard.rs
@@ -43,6 +43,7 @@ use crate::{
         share_accounting::{ShareAccounting, ShareValidationError, ShareValidationResult},
     },
     target::{bytes_to_hex, hash_rate_to_target, target_to_difficulty, u256_to_block_hash},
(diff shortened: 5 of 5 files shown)


protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -141,28 +141,25 @@ impl<'a> ExtendedChannel<'a> {
...
         new_extranonce_prefix: Vec<u8>,
     ) -> Result<(), ExtendedChannelError> {
-        let new_rollable_extranonce_size =
-            MAX_EXTRANONCE_LEN as u16 - new_extranonce_prefix.len() as u16;
-
-        // we return an error if the new extranonce_prefix would violate
-        // min_rollable_extranonce_size that was already established with the client when the
-        // channel was created
-        if new_rollable_extranonce_size < self.rollable_extranonce_size {
+        if new_extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
```
GitGab19: Can you remind me why we need to check this limit here?

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -217,17 +217,11 @@ where
             return Err(ExtendedChannelError::RequestedMaxTargetOutOfRange);
         }
 
-        let available_rollable_extranonce_size =
-            (MAX_EXTRANONCE_LEN - extranonce_prefix.len()) as u16;
-        if requested_min_rollable_extranonce_size > available_rollable_extranonce_size {
-            return Err(ExtendedChannelError::RequestedMinExtranonceSizeTooLarge);
-        }
```
GitGab19, line 224: Shouldn't we have a check here to ensure we don't exceed the max available bytes of `scriptSig` (100 bytes) ?

protocols/v2/channels-sv2/src/server/extended.rs
```diff
@@ -276,23 +270,17 @@ where
...
         extranonce_prefix: Vec<u8>,
     ) -> Result<(), ExtendedChannelError> {
-        let new_rollable_extranonce_size =
-            MAX_EXTRANONCE_LEN as u16 - extranonce_prefix.len() as u16;
-
-        // we return an error if the new extranonce_prefix would violate
-        // min_rollable_extranonce_size that was already established with the client when the
-        // channel was created
-        if new_rollable_extranonce_size < self.rollable_extranonce_size {
+        if extranonce_prefix.len() > MAX_EXTRANONCE_LEN {
```
GitGab19: Why are we enforcing this here?

roles/jd-client/src/lib/channel_manager/template_message_handler.rs
```diff
@@ -337,10 +338,17 @@ impl HandleTemplateDistributionMessagesFromServerAsync for ChannelManager {
         let declare_job = self.channel_manager_data.super_safe_lock(|data| {
             let job_factory = data.job_factory.as_mut()?;
 
+            let full_extranonce_size = data
+                .upstream_channel
+                .as_ref()
+                .map(|channel| channel.get_full_extranonce_size())
+                .unwrap_or(32);
```
GitGab19, line 345: Why do we have the `unwrap_or` here?

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
```diff
@@ -346,18 +349,21 @@ impl HandleMiningMessagesFromServerAsync for ChannelManager {
...
                     }
 
-                    let prefix_len = msg.extranonce_prefix.len();
-                    let extranonce_size = MAX_EXTRANONCE_LEN - prefix_len;
+                    let new_prefix_len = msg.extranonce_prefix.len();
+                    let rollable_extranonce_size = upstream_channel.get_rollable_extranonce_size();
+                    let full_extranonce_size = new_prefix_len + rollable_extranonce_size as usize;
+                    if full_extranonce_size > MAX_EXTRANONCE_LEN {
+                        return Err(JDCError::ExtranonceSizeTooLarge);
+                    }
```
GitGab19, line 357: Why are we enforcing this here?

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
GitGab19, line 344: Shouldn't we call `set_extranonce_prefix()` on the channel after the checks?

protocols/v2/channels-sv2/src/client/extended.rs
plebhash: previously, we were assuming that the full extranonce was always 32 bytes

that made the `rollable_extranonce_size` dynamic across the lifespan of the channel, because if the size of `new_extranonce_prefix` is different, then `rollable_extranonce_size` also needs to change

but that would violate the value that's already been established via `OpenExtendedMiningChannel.Success.extranonce_size`, so it's arguably wrong.

but that's not what you asked, that's just the context for why the check on line 156 was removed.

after removing it, I wondered if I should put something in its place (given that the method already returns `Result` and that's how our application layer currently deals with it)... one could argue that's not necessary, given that this is the client side of the channel and therefore `new_extranonce_prefix` must always be coming from a `SetExtranoncePrefix` message, where it could only have been possibly encoded as `B0_32`, and therefore it's impossible for it to be bigger than `MAX_EXTRANONCE_LEN`

but I thought it wouldn't hurt to leave this extra check there (avoiding adaptations on application layer), and perhaps it could also protect the user against some crazy ideas that I cannot really imagine, but the API still allows for

protocols/v2/channels-sv2/src/server/extended.rs
plebhash: because a `Vec<u8>` could be any size, but any `extranonce_prefix` can only be 32 bytes max (because it's encoded as `B0_32`)

and this is the server side, so the argument that the user is "protected" from API misusage is not as strong as on the client side, since the message hasn't been crafted/sent yet

I guess one thing that could be pointed out is that this check should also be done on the constructor, for consistency

roles/jd-client/src/lib/channel_manager/template_message_handler.rs
plebhash: `self.channel_manager_data` has `upstream_channel` as `Option<ExtendedChannel<'static>>`

so we need to deal with scenarios where it's `None`

IIRC this was a suggestion by @Shourya742 while we were pair-programming some of the JDC adaptations for this PR a few weeks ago

but tbh I also don't like these "default" values, it's probably better to return errors here

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
plebhash: in theory there's nothing preventing `full_extranonce_size` from being bigger than 32 bytes (as long as it doesn't violate `scriptSig` size together with everything else that goes into `scriptSig`), but  `ExtendedExtranonce` will fail if we use `range_2` that's bigger than that

I guess that's yet another limitation of our current implementation, and something we should keep in mind for #1924

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
plebhash: `upstream_channel.set_extranonce_prefix()` will fail iff `msg.extranonce_prefix` is bigger than 32 bytes

the checks below are meant to assert sanity for the creation of the new factories under the new `extranonce_prefix`, for the reasons explained in https://github.com/stratum-mining/stratum/pull/1944#discussion_r2427413386

so the potential edge cases for each lines of code are actually different

roles/jd-client/src/lib/channel_manager/upstream_message_handler.rs
plebhash: https://github.com/stratum-mining/stratum/issues/1924#issuecomment-3399235235

protocols/v2/channels-sv2/src/server/extended.rs
plebhash: a8ee3a840c3d6589bd810f139bfe8a4633ce85c9

protocols/v2/channels-sv2/src/server/extended.rs
plebhash: 55978ed55a022c3935cd07407ac5d425ac890c60

protocols/v2/channels-sv2/src/client/extended.rs
GitGab19: Alright, now it's super clear, and I would keep the check you added 👍 

Maybe I would add a simple comment to clarify that, and I would also rename `MAX_EXTRANONCE_LEN` into `MAX_EXTRANONCE_PREFIX_LEN`, if it makes more sense across all the other places where that's used.

protocols/v2/channels-sv2/src/server/extended.rs
GitGab19: As stated in [this comment](https://github.com/stratum-mining/stratum/pull/1944#discussion_r2428581234), I would rename the constant and I would add a comment on top of this check to make it very clear.

roles/jd-client/src/lib/channel_manager/template_message_handler.rs
GitGab19: Maybe we can improve later this point as part of #1875 ? 

cc @Shourya742

protocols/v2/channels-sv2/src/client/extended.rs
plebhash: I added a new private constant to `channels_sv2`, instead of renaming the old one in `mining_sv2` 2549a5d391c4318f8bc2dd7eedf14750aadeef01

that's because the implementations of `mining_sv2::{Extranonce,ExtendedExtranonce}` are mixing these concepts, so simply renaming it in `mining_sv2` would cause semantic problems on the crate

plebhash: there's no perfect solution here unfortunately

until we get a better implementation of `mining_sv2::{Extranonce,ExtendedExtranonce}` (not necessarily with the same names, and definitely without the same bad assumptions) we have to manage this technical debt in ways the cause the least amount of damage

but having a new constant on `channels_sv2` feels like a reasonable compromise, especially because as I replaced the old for the new one, it felt like this crate is now even more agnostic to the bad assumptions on `mining_sv2::{Extranonce,ExtendedExtranonce}`

so I would probably keep this constant on `channels_sv2`, even after #1924

protocols/v2/channels-sv2/src/client/extended.rs
```diff
@@ -296,7 +294,10 @@ impl<'a> ExtendedChannel<'a> {
 
         let mut script_sig = vec![];
         script_sig.extend_from_slice(set_custom_mining_job.coinbase_prefix.inner_as_ref());
-        script_sig.extend_from_slice(&[0; MAX_EXTRANONCE_LEN]);
+        let full_extranonce_size =
+            self.extranonce_prefix.len() + self.rollable_extranonce_size as usize;
```
bit-aloo:
```suggestion
        let full_extranonce_size = self.get_full_extranonce_size()
```

protocols/v2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -324,10 +338,16 @@ impl JobFactory {
...
-        let coinbase_tx_suffix =
-            self.coinbase_tx_suffix(template.clone(), additional_coinbase_outputs.clone())?;
+        let coinbase_tx_prefix = self.coinbase_tx_prefix(
+            template.clone(),
+            additional_coinbase_outputs.clone(),
+            full_extranonce_size,
+        )?;
+        let coinbase_tx_suffix = self.coinbase_tx_suffix(
+            template.clone(),
+            additional_coinbase_outputs.clone(),
```
bit-aloo:
```suggestion
        let coinbase_tx_suffix = self.coinbase_tx_suffix(
            template,
            additional_coinbase_outputs,
```

PR #1945, opened 2025-10-13
Fix CI build failure by pinning generic-array to 0.14.7

Shourya742:
The CI build was failing due to generic-array v0.14.8 using the cargo::rustc-check-cfg directive, which is only supported on Rust 1.79 and above.
Pinned generic-array to version 0.14.7 to maintain MSRV.

Build failure can be observed here: https://github.com/stratum-mining/stratum/actions/runs/18458666773/job/52584886703?pr=1926


test/integration-tests/minerd/minerd
GitGab19, line 1: Why do we have this change here?

bit-aloo: Must have slipped, let me add this to gitignore.

PR #1860, opened 2025-09-08
New stratum-core structure

lucasbalieiro:
address https://github.com/stratum-mining/stratum/issues/1799#issuecomment-3247001218;

- remove roles, since they moved to `sv2-apps` repo
- move `protocols/v2` -->  `sv2/`
- move `protocols/v2` -->  `sv1/`
- move `protocols/stratum-translation` --> `stratum-core/stratum-translation`
- move `utils/buffer` --> `sv2/buffer-sv2`
- adapted scripts and workflows where needed

PR #1947, opened 2025-10-15
deprecate `roles_logic_sv2`, `benches`, `error_handling`, `utils::bip32_derivation`

### LINKED ISSUE #1940: deprecate `roles_logic_sv2`
after #1928, this crate is only being consumed in SRI for its custom `Mutex`

this primitive is not a fundamental requirement for building Sv2 apps, and it doesn't make sense for SRI to maintain it

lucasbalieiro:
closes #1940
**To be merged AFTER #1931**

* drops `benches`
* drops `error_handling`
* drops `utils::bip32_derivation`

Drops `roles_logic_sv2` by:

* Refactoring `roles::jd-client`, `roles::pool`, and `roles::translator` to no longer depend on `roles_logic_sv2::error`.
* Updating `jd-server` and `roles::test-utils::mining-device` to use crates from crates.io.

  * These crates still implement `roles_logic_sv2::handlers`, and their refactor is already planned.
  * This avoids keeping `roles_logic_sv2` in the codebase just for them.
* Moving `roles_logic_sv2::utils::Mutex` to `roles::stratum-apps::custom_mutex::Mutex`.


roles/jd-client/src/lib/error.rs
```diff
@@ -82,6 +83,18 @@ pub enum JDCError {
...
+    FailedToUpdateExtendedChannel(ServerExtendedChannelError),
+    ExtranoncePrefixFactoryError(ExtendedExtranonceError),
+    FailedToProcessSetNewPrevHashExtendedChannel(ServerExtendedChannelError),
+    FailedToProcessSetNewPrevHashStandardChannel(ServerStandardChannelError),
+    FailedToProcessNewTemplateExtendedChannel(ServerExtendedChannelError),
+    FailedToProcessNewTemplateStandardChannel(ServerStandardChannelError),
+    FailedToCreateExtendedChannel(ServerExtendedChannelError),
+    FailedToCreateStandardChannel(ServerStandardChannelError),
+    FailedToProcessNewTemplateGroupChannel(GroupChannelError),
+    FailedToProcessSetNewPrevHashGroupChannel(GroupChannelError),
```
bit-aloo: Since all of these are just channel-related errors, and the error variants are already defined in `channel_sv2`, we don’t need to redefine them here. You can check this commit for reference: https://github.com/stratum-mining/stratum/pull/1921/commits/1d365c5ca54d888d88825067c3855c0588161b43
 What I did there was simply encapsulate the `channel_sv2` error and reuse it

roles/jd-server/Cargo.toml
```diff
@@ -17,9 +17,11 @@ name = "jd_server"
 path = "src/lib/mod.rs"
 
 [dependencies]
-stratum-apps = { path = "../stratum-apps", features = ["jd_server"] }
-error_handling = { path = "../../utils/error-handling" }
-key-utils = { path = "../../utils/key-utils" }
+stratum-common = {version = "4.0.1", features = ["with_network_helpers"]}
+config_helpers_sv2 = "0.1.0"
```
plebhash: I guess we're doing this because `stratum-apps` hasn't been published to crates.io yet?

roles/stratum-apps/src/custom_mutex.rs
plebhash, line 1: we should definitely `git mv` this file, but perhaps it would be good to also remove or modify the comments on the top of the file, because they're no longer accurate

and it will be pretty easy to overlook in the future

roles/jd-client/src/lib/error.rs
lucasbalieiro: Applied

roles/jd-server/Cargo.toml
lucasbalieiro: Yes, exactly

test/integration-tests/Cargo.toml
```diff
@@ -30,5 +31,8 @@ tokio-util = { version = "0.7", default-features = false }
 tracing = { version = "0.1.41", default-features = false }
 tracing-subscriber = { version = "0.3.19", default-features = false }
 
+[patch.crates-io]
+key-utils = { path = "../../utils/key-utils" }
```
lucasbalieiro: @GitGab19 I found a way to make the IT work without too many drastic changes.

Check this commit:
https://github.com/stratum-mining/stratum/pull/1947/commits/05583593ede97f1cf8a6ed05b15627b1d0587f98

I tested the approach you shared via discord and hit the same crate conflicts inside the IT: mainly with **key-utils** and **config_helpers**.

The main difference is that your solution drops `stratum-apps` and imports each crate independently from crates.io.
What I did instead was to keep using `stratum-common` and handle only the exceptions for those two crates:

* **key-utils:** added a patch to `test/integration-tests/Cargo.toml` so that, instead of pulling from crates.io, the compiler falls back to the local path. Which will affect all the dependency tree to use just one version of the crate.
* **config_helpers:** removed the re-export from `stratum-apps` and made all roles depend directly on the crates.io version.

I would’ve needed to make these same adjustments if I had applied your approach anyway.

With these changes, everything works fine now, no need to deal with conflicts between `roles_logic_sv2` and `stratum-apps`.

I’m still a bit unsure whether it’s better to keep using `stratum-common` for `jd-server` and `mining-device`, or to import each crate independently like in your proposal: https://github.com/GitGab19/stratum/commit/26874c3b3e1220ca54ef228be1f62dc5430764dc

What do you prefer? I can easily apply your strategy since we now know where the problem lies

roles/stratum-apps/src/custom_mutex.rs
lucasbalieiro: thanks, comments adapted

test/integration-tests/Cargo.toml
GitGab19: I opened https://github.com/stratum-mining/stratum/pull/1951 which makes `key-utils` and `error-handling` modules of `stratum-apps`. 

If you rebase your PR here on top of that one, you should be able to simply fix the JDS and the mining device by importing `roles_logic_sv2` version `5.0.0` from crates.io there.

Without having to use `stratum-common`, or add any patch to ITF.

Let me know how it goes :)

plebhash: #1951 has some CI issues.

is it meant to be reviewed as a standalone PR, or is it only to give inspiration and perhaps cherry-pick commits for this PR?

lucasbalieiro: Oh sorry, @GitGab19 and I were talking via Discord and we are trying a few other hypothesis. To solve the problem.

I'm just trying a few more things and if it works I'll get some commits from that PR. Because the initial solution was not working.

GitGab19: The CI issues there are due to the last commit I pushed (c5bc73a1794fbb1a3fde94c9e6247752757bc8f6, which is just to be cherry-picked by @lucasbalieiro on this PR in case it helps him here).

I'll drop it once he tells how it went.

lucasbalieiro: I was able to make it work without applying this patch. 

But, @GitGab19 gave some more ideas to make it better using feature flags. I'll draft this PR again until I apply these changes and adjust the commit history accordingly.

roles/jd-server/Cargo.toml
lucasbalieiro: We’ve changed the strategy a bit, now we’re using stratum-apps, and for `jd-server` and the `mining-device`, we’re pulling the other crates (like binary_sv2, codec_sv2, etc.) directly from crates.io.

check c938db59ee6c63e69fd21f01805038afc5966abf

test/integration-tests/Cargo.toml
lucasbalieiro: >  I'll draft this PR again until I apply these changes and adjust the commit history accordingly.

We are ready to review again, now using the strategy of using the feature flags for `stratum-apps` re-exports.

And with some commits from: https://github.com/stratum-mining/stratum/pull/1951#issuecomment-3414700213

.github/workflows/release-libs.yaml
```diff
@@ -129,4 +121,4 @@ jobs:
       # stratum-apps (depends on stratum-core and external crates)
       - name: Publish crate stratum-apps
         run: |
-          ./scripts/release-libs.sh roles/stratum-apps
\ No newline at end of file
+          ./scripts/release-libs.sh roles/stratum-apps
```
bit-aloo, line 113: L101 - L104, publishing of roles-logic crate can be removed.

.github/workflows/rust-msrv.yaml
```diff
@@ -22,8 +22,6 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           override: true
```
bit-aloo, line 24: We can add build for stratum-core here.

roles/jd-server/Cargo.toml
```diff
@@ -18,8 +18,18 @@ path = "src/lib/mod.rs"
...
+codec_sv2 = "3.0.1"
+framing_sv2 = "5.0.1"
+mining_sv2 = "5.0.1"
+noise_sv2 = "1.4.0"
+parsers_sv2 = "0.1.1"
+job_declaration_sv2 = "5.0.1"
+common_messages_sv2 = "6.0.1"
+network_helpers_sv2 = "4.0.1"
+rpc_sv2 = "1.1.1"
+bitcoin = "0.32.5"
```
bit-aloo, line 33: why these?

roles/stratum-apps/src/error_handling/mod.rs
```diff
@@ -29,8 +29,8 @@ macro_rules! handle_result {
                 // handle error
                 let res = crate::status::handle_error(&$sender, e.into()).await;
                 match res {
-                    error_handling::ErrorBranch::Break => break,
-                    error_handling::ErrorBranch::Continue => continue,
+                    $crate::error_handling::ErrorBranch::Break => break,
+                    $crate::error_handling::ErrorBranch::Continue => continue,
```
bit-aloo: Do we even need this, as we not longer are following these error handling ergonomics?

roles/stratum-apps/src/lib.rs
```diff
@@ -25,6 +25,7 @@
 //! - [`rpc`] - RPC utilities with custom serializable types (`Hash`, `BlockHash`, `Amount`)
 
 /// Re-export all the modules from `stratum_core`
+#[cfg(feature = "core")]
```
bit-aloo, line 28: why do we need to have core feature?

utils/Cargo.toml
```diff
@@ -3,8 +3,6 @@ resolver="2"
 
 members = [
     "buffer",
-    "error-handling",
-    "key-utils",
     "bip32-key-derivation",
```
bit-aloo: lets remove `bip32-key-derivation` crate as well, we don't use it at all.

.github/workflows/release-libs.yaml
lucasbalieiro: thanks, fixed here 24ac9d5726d7f0bdc5b5f4be3bb98287a3ea339c and adapted other CI and scripts that I missed

.github/workflows/rust-msrv.yaml
lucasbalieiro: added: 92da9753d65ef08491d6573ab2a3293f991e595e

utils/Cargo.toml
lucasbalieiro: dropped: 1b8603269fe585f60e937806c8cd7127bf695e67

roles/jd-server/Cargo.toml
lucasbalieiro: well, that’s a long story but basically:

When @GitGab19 and I were trying to get this working, we ran into several crate conflicts and compiler complaints.
Crates like `binary_sv2`, `codec_sv2`, etc., were conflicting with the versions re-exported by `roles_logic_sv2` and `stratum-core`.

To work around that, we decided to pull everything from crates.io **ONLY** for `jd-server` and `mining-device`.
These two haven’t been refactored yet, they still rely on `roles_logic_sv2::handlers` and related modules.

Also, as you questioned in another section of this PR, this is one of the reasons why we feature-gated `stratum-core`.
`stratum-core` also re-exports these low-level crates, so for `jd-server` and `mining-device`, we disable that re-export.
For the other roles that have already been refactored, everything continues as before, they get their dependencies only from `stratum-apps`.

All of this is to avoid conflicts like this: 

image

roles/stratum-apps/src/lib.rs
lucasbalieiro: we feature-gated it due to crate conflicts between the dependencies required by roles that are **yet to be refactored** and those that are **already refactored**.
A bit detailed explanation can be found in this PR comment:
https://github.com/stratum-mining/stratum/pull/1947#discussion_r2440045972

This will probably be removed once we refactor them.

roles/stratum-apps/src/error_handling/mod.rs
plebhash: I agree that it feels a bit unnecessary to move `error_handling` to `stratum-apps`, given that:
- [it's already been published to `crates.io`](https://crates.io/crates/error_handling)
- the crate is unlikely to be used again, aside from this legacy dependency

however, it would be good to clean that up from Pool, because it's still being imported as a dependency, despite not being used anymore

https://github.com/stratum-mining/stratum/blob/1223d1437ac9c15f04af92016b85eb37e4246fe1/roles/pool/Cargo.toml#L21

utils/Cargo.toml
bit-aloo: @plebhash @lucasbalieiro  now utils only contain buffer crate after removal of the rest, and we are moving buffer to stratum-core aren't we? Should we get rid of utils?

plebhash: sure, but I guess it doesn't really matter if that happens here or in #1860

bit-aloo: Ah!! we have another PR, yup make sense.

roles/jd-server/Cargo.toml
bit-aloo: Cool, sounds reasonable.

roles/stratum-apps/src/error_handling/mod.rs
bit-aloo: Should we do this in this PR? or the other one?

roles/stratum-apps/src/key_utils/mod.rs
plebhash, line 1: I'm lacking a lot of context of the long journey that led to this decision, so maybe this is a dumb question:

but given that [`key-utils` is available on `crates.io`](https://crates.io/crates/key-utils), does it really need to become a module of `stratum-apps`?

roles/stratum-apps/src/error_handling/mod.rs
lucasbalieiro: I'm already removing it. It will take just a few more minutes. Just testing somethings locally. To see If haven't made anything stupid

roles/stratum-apps/src/key_utils/mod.rs
GitGab19: `key-utils` is used by all the apps we have, that's why I put it as a module of `stratum-apps`.

In this way any app simply have to import `stratum-apps`. 

It's the same thing we did for `network_helpers_sv2`, `config_helpers_sv2`, etc.

roles/stratum-apps/src/error_handling/mod.rs
bit-aloo: Awesome!!!

lucasbalieiro: done [32513af](https://github.com/stratum-mining/stratum/pull/1947/commits/32513af3279ce538f30f46f9c1762033e7240d96)

Shourya742 (approved):

GitGab19 (approved):
utACK

Issue #1834, opened 2025-08-13
replace `mining_sv2::Target` with `bitcoin::Target` on `channels_sv2`


plebhash:
https://github.com/stratum-mining/stratum/blob/c3b15a0046bb6ab75d14620a3a272dc08bf26fa0/protocols/v2/channels-sv2/src/server/extended.rs#L53

this is just one example where I left a comment as a reminder, there's usage on other places where it should be replaced too

xTwo56:
@Sansh2356 yes.

plebhash:
sorry for the late reply @icarusmaximal 

I gave a quick look at #1950, and despite the current conflicts, it seems to be going in a good direction

> should `hash_rate_to_target` from `roles-logic-sv2::src::utils` be modified to return `Target` instead of `U256`, or renamed to `hash_rate_to_target_U256`?

I see you went ahead and made `channels_sv2::target::hash_rate_to_target` return `Target`, and I think that was the right decision, given how this function is used across the crate.

it's important to note that `roles_logic_sv2` is on the process of being deprecated, so I wouldn't worry about the duplicates of this function there.

> @plebhash should i stick to migrate `mining_sv2::Target` only, or convert other types like `U256` as well wherever possible e.g. `pub max_target: U256<'decoder>` to `pub max_target: Target`
>
> here, `max_target` is an attribute of `OpenExtendedChannel`

we need every message definition to conform to the [Sv2 Spec](https://github.com/stratum-mining/sv2-spec/), and changing `OpenExtendedMiningChannel.max_target` to something other than `U256` would violate that

PR #1950, opened 2025-10-16
migrate `mining_sv2::Target` to `bitcoin::Target`

icarusmaximal:
addresses #1834


plebhash (approved):

plebhash:
thanks for your contribution @icarusmaximal 

moving forward, please be careful when committing `Cargo.lock` files

the ones we have committed on the repo have `version = 3`, which abides by MSRV 1.75

but if you build with a recent compiler, that will get changed into `version = 4` and break [MSRV CI](https://github.com/stratum-mining/stratum/blob/main/.github/workflows/rust-msrv.yaml)

PR #1951, opened 2025-10-16
move `key-utils` and `error-handling` as modules of `stratum-apps`

GitGab19:
Follow-up to PR #1931.

This PR moves `key-utils` and `error-handling` into `stratum-apps`, dropping their crates.

This should also help the work by @lucasbalieiro on #1947.


GitGab19:
I had to add https://github.com/stratum-mining/stratum/pull/1951/commits/efc78bf214931e4afa47ebea1f85b0367ed69c60 to unblock the semver check on CI: https://github.com/stratum-mining/stratum/actions/runs/18557093752/job/52897061182?pr=1951.

The error there is due to a manual publish of `roles_logic_sv2` which I did yesterday, to unblock the work on #1947. 

But given we are gonna deprecate `roles_logic_sv2` in that PR, we can safely delete it from the CI here.

PR #1952, opened 2025-10-16
Remove binary codec sv2

Shourya742:
closes: #1462

### DIFF
--- a/test/integration-tests/Cargo.lock
+++ b/test/integration-tests/Cargo.lock
@@ -247,11 +247,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d965446196e3b7decd44aa7ee49e31d630118f90ef12f97900f262eb915c951d"
 
 [[package]]
+name = "binary_codec_sv2"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8ad24342e0abdcc463ad6ad4ac7b0ec606122c11eddf92de186a657df0114eb7"
+
+[[package]]
+name = "binary_codec_sv2"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d16415a0a9ccee2f71820da352c1f2a7f16d9f8e3ae6fb5e97834c6d732e98cd"
+dependencies = [
+ "buffer_sv2 2.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "binary_sv2"
+version = "4.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba8295945d51b707f3a49e17810dddef858549e2b52383c7f2c4dd036f6bc1e6"
+dependencies = [
+ "binary_codec_sv2 3.0.0",
+ "derive_codec_sv2 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "binary_sv2"
 version = "5.0.0"
 dependencies = [
- "buffer_sv2",
- "derive_codec_sv2",
+ "buffer_sv2 2.0.0",
+ "derive_codec_sv2 1.1.1",
 ]
 
 [[package]]
@@ -463,20 +488,6 @@ version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "11ac02b93b5bd92a7dda2bc4b8c9d1f087e1fffc8b1018b532109135629051fc"
 dependencies = [
- "binary_sv2 4.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "bitcoin",
- "common_messages_sv2 6.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "job_declaration_sv2 5.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "mining_sv2 5.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "primitive-types",
- "template_distribution_sv2 4.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "tracing",
-]
-
-[[package]]
-name = "channels_sv2"
-version = "2.0.0"
-dependencies = [
  "binary_sv2 4.0.0",
  "bitcoin",
  "common_messages_sv2 6.0.1",
@@ -484,6 +495,20 @@ dependencies = [
  "mining_sv2 5.0.1",
  "primitive-types",
  "template_distribution_sv2 4.0.1",
+ "tracing",
+]
+
+[[package]]
+name = "channels_sv2"
+version = "2.0.1"
+dependencies = [
+ "binary_sv2 5.0.0",
+ "bitcoin",
+ "common_messages_sv2 6.0.2",
+ "job_declaration_sv2 5.0.2",
+ "mining_sv2 5.0.2",
+ "primitive-types",
+ "template_distribution_sv2 4.0.2",
  "tracing",
 ]
 
@@ -544,9 +569,9 @@ version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "01e6d43e79e66d0f98038922157db8b6101594921be87ac2cca3754d669f2a05"
 dependencies = [
- "binary_sv2 4.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "binary_sv2 4.0.0",
  "buffer_sv2 2.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "framing_sv2 5.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "framing_sv2 5.0.1",
  "noise_sv2 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.8.5",
  "tracing",
@@ -554,11 +579,11 @@ dependencies = [
 
 [[package]]
 name = "codec_sv2"
-version = "4.0.0"
+version = "4.0.1"
 dependencies = [
- "binary_sv2 4.0.0",
+ "binary_sv2 5.0.0",
  "buffer_sv2 2.0.0",
- "framing_sv2 5.0.1",
+ "framing_sv2 5.0.2",
  "noise_sv2 1.4.0",
  "rand 0.8.5",
  "tracing",
@@ -572,25 +597,18 @@ checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 
 [[package]]
 name = "common_messages_sv2"
-version = "6.0.2"
-dependencies = [
- "binary_sv2 4.0.0",
-]
-
-[[package]]
-name = "common_messages_sv2"
-version = "6.0.1"
-dependencies = [
- "binary_sv2 4.0.0",
-]
-
-[[package]]
-name = "common_messages_sv2"
 version = "6.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2e6ec6ab527aeebf8ead273d6ab712ff181c050ee5e1082f3f6a2c65c0a10bf6"
 dependencies = [
- "binary_sv2 4.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "binary_sv2 4.0.0",
+]
+
+[[package]]
+name = "common_messages_sv2"
+version = "6.0.2"
+dependencies = [
+ "binary_sv2 5.0.0",
 ]
 
 [[package]]
@@ -780,6 +798,15 @@ name = "derive_codec_sv2"
 version = "1.1.1"
 
 [[package]]
+name = "derive_codec_sv2"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "924f288d967a5cd37956b195269ee7f710999169895cf670a736e1b2267d6137"
+dependencies = [
+ "binary_codec_sv2 1.2.0",
+]
+
+[[package]]
 name = "digest"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


protocols/v2/binary-sv2/src/codec/encodable.rs
plebhash, line 1: if these files are simply being moved to a different directory (despite being a different crate), that should be done via `git mv`

otherwise we'll lose git history

plebhash: that goes for everything going on in commits 9aad6accf3471eb4fe811b263d7eae7d07366c98 and cf785d327f66fe5c6acdbc47fa2dbbf753cbd7c1

protocols/v2/binary-sv2/Cargo.toml
plebhash, line 1: we'll need to bump MAJOR

```
$ cargo +stable semver-checks
    Building binary_sv2 v4.0.0 (current)
       Built [   2.407s] (current)
     Parsing binary_sv2 v4.0.0 (current)
      Parsed [   0.006s] (current)
    Building binary_sv2 v4.0.0 (baseline)
       Built [   2.129s] (baseline)
     Parsing binary_sv2 v4.0.0 (baseline)
      Parsed [   0.002s] (baseline)
    Checking binary_sv2 v4.0.0 -> v4.0.0 (no change)
     Checked [   0.003s] 95 checks: 94 pass, 1 fail, 0 warn, 0 skip

--- failure function_missing: pub fn removed or renamed ---

Description:
A publicly-visible function cannot be imported by its prior path. A `pub use` may have been removed, or the function itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.37.0/src/lints/function_missing.ron

Failed in:
  function binary_sv2::clone_message, previously in file /Users/plebhash/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/binary_sv2-4.0.0/src/lib.rs:25

     Summary semver requires new major version: 1 major and 0 minor checks failed
```

protocols/v2/binary-sv2/src/codec/encodable.rs
bit-aloo: Ah! I thought Git would automatically detect when files are moved without major changes, but I guess not. I’ve rewritten the history and moved everything properly now, so it should be fine.

examples/ping-pong-encrypted/src/messages.rs
GitGab19, line 1: As said by @plebhash [here](https://github.com/stratum-mining/stratum/pull/1952#issuecomment-3412197761), we're gonna deprecated `ping-pong` examples soon.

I still don't know the implications of keeping them in this PR, but I just want to underline that we should not care about them anymore.

examples/ping-pong/src/messages.rs
GitGab19, line 1: As said by @plebhash [here](https://github.com/stratum-mining/stratum/pull/1952#issuecomment-3412197761), we're gonna deprecated `ping-pong` examples soon.

I still don't know the implications of keeping them in this PR, but I just want to underline that we should not care about them anymore.

protocols/v2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -556,10 +555,10 @@ fn parse_struct_fields(group: Vec<TokenTree>) -> Vec<ParsedField> {
 /// This generated code enables `Test` to be decoded from a binary stream, defines how each
 /// field should be parsed, and provides `into_static` and `as_static` methods to facilitate
 /// ownership and lifetime management of decoded fields in the struct.
-#[proc_macro_derive(Decodable)]
+#[proc_macro_derive(Decodable, attributes(codec))]
```
GitGab19: What is this?

protocols/v2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -556,10 +555,10 @@ fn parse_struct_fields(group: Vec<TokenTree>) -> Vec<ParsedField> {
...
 pub fn decodable(item: TokenStream) -> TokenStream {
+    let crate_path = extract_crate_path(item.clone()).unwrap_or_else(|| "binary_sv2".to_string());
```
GitGab19: Why do we need this?

Can you explain me the entire idea about this `extract_crate_path` function?

protocols/v2/channels-sv2/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/codec-sv2/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/framing-sv2/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/handlers-sv2/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/parsers-sv2/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/roles-logic-sv2/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/subprotocols/common-messages/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/subprotocols/job-declaration/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/subprotocols/mining/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/subprotocols/template-distribution/Cargo.toml
GitGab19: Shouldn't we bump PATCH here?

cc @plebhash

protocols/v2/binary-sv2/derive_codec/src/lib.rs
bit-aloo: As I mentioned in the PR description, the proc macro definitions `Encodable` and `Decodable` in `derive-codec-sv2` are tightly coupled to whichever module provides the corresponding traits. These traits are what the macros rely on for serialization and deserialization logic.

Currently, the generated (expanded) code produced by the proc macro includes direct imports to that dependency. For example, during macro expansion, it generates something like:

```rust
mod impl_parse_encodable_test {
    use binary_sv2;
    // ...
}
```

This encapsulated scope still explicitly depends on `binary_sv2`.

Now, the issue arises with crates such as `ping-pong` or `ping-pong-encrypted`. These crates don’t directly depend on `binary_sv2`, yet the expanded macro code tries to access APIs from it (e.g., for serialization or deserialization). Because the generated code contains a hard reference to `use binary_sv2;`, it fails unless the crate explicitly includes that dependency.

Previously, this wasn’t a problem because `derive-codec-sv2` itself imported `binary_codec_sv2`, so the reference was always valid. However, after decoupling and moving to `binary_sv2`, that implicit linkage no longer exists.

The reason for introducing the extra `codec` attribute is to handle this exact case, it allows crates like `ping-pong`, which don’t directly import `binary_sv2`, to explicitly specify the path to it.

There’s an alternative way to remove the need for this attribute, but that would require a full refactor of the `binary_sv2` crate itself.

protocols/v2/channels-sv2/Cargo.toml
bit-aloo: Patch would be required as we updated its dependency, good point.

bit-aloo: As API's are not changing, so we don't need to update major and minor versions.

plebhash: hmm idk

from the perspective of someone who's using `channels_sv2`, does it really matter where the internal implementation details are coming from?

are we changing the behavior of anything, or just moving things around?

maybe I have some blindspot, but my first instinct here would not to bump anything

protocols/v2/binary-sv2/derive_codec/src/lib.rs
GitGab19: Got it, now it's more clear to me.

So, is this being done just for `examples/` crates?

Because in that case, I would log an issue to remove this once we deprecate `examples/` in https://github.com/stratum-mining/stratum/pull/1860.

Or I think we could also simplify this PR by having `examples/` using `binary_sv2` from crates.io, instead of over-complicating the crate here.

protocols/v2/channels-sv2/Cargo.toml
GitGab19: If we don't bump PATCH here we won't update crates published to crates.io during release.

So we would keep having the old `binary_sv2` on those published crates.

plebhash: I find it hard to reason about all the potential second-order implications of these things

so if this would be on the safer side, I guess there's no problem

protocols/v2/binary-sv2/derive_codec/src/lib.rs
bit-aloo: Getting it from crates.io gonna work, I will remove the codec attribute patch and update the PR.

protocols/v2/channels-sv2/Cargo.toml
bit-aloo: I will bump the patch for them, in the same PR.

examples/ping-pong-encrypted/Cargo.toml
```diff
@@ -7,8 +7,10 @@ authors = [ "SRI Community" ]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-stratum-apps = { path = "../../roles/stratum-apps" }
 key-utils = { version = "^1.0.0", path = "../../utils/key-utils" }
 rand = "0.8"
 tokio = { version = "1.44.1", features = [ "full" ] }
 async-channel = "1.5.1"
+binary_sv2 = "1.0.0"
```
GitGab19: Why 1.0.0? 

Last published version is 3.0.2.

examples/ping-pong-encrypted/Cargo.toml
```diff
@@ -7,8 +7,10 @@ authors = [ "SRI Community" ]
...
+network_helpers_sv2 = { version = "2.0.0", features = ["with_tokio"] }
```
GitGab19: Why 2.0.0?

Last published version is 4.0.1.

examples/ping-pong-encrypted/Cargo.toml
```diff
@@ -7,8 +7,10 @@ authors = [ "SRI Community" ]
...
+codec_sv2 = { version = "1.0.0", features = ["noise_sv2"] }
```
GitGab19: Why 1.0.0?

Last published version is 3.0.1.

protocols/v2/binary-sv2/derive_codec/src/lib.rs
GitGab19: Do we still need this?

protocols/v2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -643,7 +642,7 @@ pub fn decodable(item: TokenStream) -> TokenStream {
     let result = format!(
         "mod impl_parse_decodable_{} {{
 
-    use super::binary_codec_sv2::{{decodable::DecodableField, decodable::FieldMarker, Decodable, Error, SizeHint}};
+    use {crate_path}::{{decodable::DecodableField, decodable::FieldMarker, Decodable, Error, SizeHint}};
```
GitGab19: Why not having `binary_sv2` directly here?

GitGab19:
I would squash a891dcc3786cec2da61463f9fa4f39e7e416e357, since it's reverting something you did on this PR.

Please reword commit 25f3417a1d82c00157b3f1058a29e5bc5307fd9b which doesn't make sense (there's no stratum-common) and it's not descriptive at all.

Usually, I tend to always squash fmt and clippy commits with the related changes, to not pollute git history too much.

Shourya742:
> I would squash [a891dcc](https://github.com/stratum-mining/stratum/commit/a891dcc3786cec2da61463f9fa4f39e7e416e357), since it's reverting something you did on this PR.
> 
> Please reword commit [25f3417](https://github.com/stratum-mining/stratum/commit/25f3417a1d82c00157b3f1058a29e5bc5307fd9b) which doesn't make sense (there's no stratum-common) and it's not descriptive at all.
> 
> Usually, I tend to always squash fmt and clippy commits with the related changes, to not pollute git history too much.

Should be ok now.

examples/ping-pong-encrypted/Cargo.toml
bit-aloo: I was facing some compatibility issue, but let me have a deeper look.

GitGab19 (approved):

protocols/v2/channels-sv2/Cargo.toml
plebhash: @GitGab19 @Shourya742 we should be careful in extrapolating this into an umbrella rule where we always bump PATCH just because some dependencies were updated

`handlers_sv2` is now at `v0.2.1` on this repo, while the published version is still at `v0.1.0`... so `v0.2.0` will never be published

it doesn't make sense to bump PATCH for something that hasn't been published yet

plebhash: `channels_sv2` is also at `v2.0.1`, while the last published version is at `v1.0.2`, so `v2.0.0` will never be published

plebhash: `codec_sv2` is also at `v4.0.1` while the last published version is at `v3.0.1`, so `v4.0.0` will never be published

bit-aloo: Shoot, I didn't looked at the crates.io versions for them. Thanks for pointing it out.

PR #1962, opened 2025-10-23
Remove unreachable macro and add share validation error variant in Pool

Shourya742:
We had unreachable macro in pool, this PR removes that and adds additional and improved logs.

Part of: stratum-mining/sv2-apps#42

### DIFF
--- a/roles/pool/src/lib/channel_manager/mining_message_handler.rs
+++ b/roles/pool/src/lib/channel_manager/mining_message_handler.rs
@@ -636,7 +636,7 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         messages.push((downstream_id, Mining::SubmitSharesError(error)).into());
                     }
                     Err(e) => {
-                        return Err(PoolError::ShareValidationError(e));
+                        return Err(e)?;
                     }
                 }
 
@@ -805,7 +805,7 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         messages.push((downstream_id, Mining::SubmitSharesError(error)).into());
                     }
                     Err(e) => {
-                        return Err(PoolError::ShareValidationError(e));
+                        return Err(e)?;
                     }
                 }
 
--- a/roles/pool/src/lib/error.rs
+++ b/roles/pool/src/lib/error.rs
@@ -28,6 +28,7 @@ pub enum ChannelSv2Error {
     StandardChannelServerSide(StandardChannelError),
     GroupChannelServerSide(GroupChannelError),
     ExtranonceError(ExtendedExtranonceError),
+    ShareValidationError(ShareValidationError),
 }
 
 /// Represents various errors that can occur in the pool implementation.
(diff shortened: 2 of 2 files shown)


roles/pool/src/lib/error.rs
```diff
@@ -85,6 +88,8 @@ pub enum PoolError {
     ParseInt(std::num::ParseIntError),
     /// Failed to create group channel
     FailedToCreateGroupChannel(GroupChannelError),
+    /// Share validation failed
+    ShareValidationError(ShareValidationError),
```
GitGab19: Shouldn't this be a `ChannelSv2Error` variant?

roles/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -380,6 +380,7 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                                 extranonce_size: extended_channel.get_rollable_extranonce_size(),
                             }
                             .into_static();
+                        info!("Sending OEMCS message: {open_extended_mining_channel_success}");
```
GitGab19:
```suggestion
                        info!("Sending OpenExtendedMiningChannel.Success: {open_extended_mining_channel_success}");
```

GitGab19: If possible, I would also add the `downstream_id` here.

roles/pool/src/lib/error.rs
bit-aloo: Done here: ed27c9a

roles/pool/src/lib/channel_manager/mining_message_handler.rs
bit-aloo: Added, let me know if it is verbose

GitGab19 (approved):

roles/pool/src/lib/channel_manager/mining_message_handler.rs
plebhash: please let's always avoid introducing acronyms for messages, I don't think that's ever a good idea

how would OEMCS make any sense in the eyes of who's reading this?

laziness for a few keystrokes while writing code comes with the cost of permanent damage to who needs to understand and make sense of it

bit-aloo: This is removed now, my bad.

roles/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -634,9 +635,20 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
+                            sequence_number: msg.sequence_number,
+                            error_code: "bad-extranonce-size"
+                                .to_string()
+                                .try_into()
+                                .expect("error code must be valid string"),
+                        };
+                        messages.push((downstream_id, Mining::SubmitSharesError(error)).into());
+                    }
+                    Err(e) => {
+                        return Err(e)?;
```
plebhash: `SubmitSharesStandard` have no `extranonce`

this is still `unreachable`

plebhash: or the generic `Err(e)` arm, doesn't matter

the point is that we don't really expect this to ever happen

bit-aloo: This has now been removed, thanks for pointing out. Regarding the error arm, it will disconnect the downstream cleint conneciton  gracefully, instead of panicking.

PR #1965, opened 2025-10-24
refactor(channels-sv2): Update share validation result with share metadata

average-gary:
- Modify `ShareValidationResult` enum to include `share_work` and `share_hash` for both `Valid` and `BlockFound` variants
- Update share validation logic in `extended.rs`, `standard.rs` to pass share metadata

Minimal returned data from `validate_share()` to enable application side handling. i.e. Persistence.

PR #1966, opened 2025-10-24
improve share validation

plebhash:
@Shourya742 raised some concerns about the direction #1902 is going https://github.com/stratum-mining/stratum/pull/1902#issuecomment-3442428460

a proposed alternative is to do all the persistence at the application layer, which is currently missing some information

this PR improves `ShareValidationResult` so that the application layer has access to the share hash

every other information relevant for some kind of persistence on the application layer can be retrieved from:
- the current channel state (e.g.: `share_work` is simply the current channel's `Target`, or `user_identity` can be retrieved via `get_user_identity`)
- the context where we're calling `validate_share()` (e.g.: `channel_id` and `sequence_number` is already available on the share message itself)
- the channel's internal `ShareAccounting`, which can be retrieved via `get_share_accounting()`

this also avoids some redundancies on the proposed solution of #1965

### DIFF
--- a/roles/pool/src/lib/channel_manager/mining_message_handler.rs
+++ b/roles/pool/src/lib/channel_manager/mining_message_handler.rs
@@ -541,9 +541,10 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                             info!("SubmitSharesStandard: {} ✅", success);
                             messages.push((downstream_id, Mining::SubmitSharesSuccess(success)).into());
                         } else {
+                            let share_work = standard_channel.get_target().difficulty_float();
                             info!(
-                                "SubmitSharesStandard: valid share | downstream_id: {}, channel_id: {}, sequence_number: {}, share_hash: {} ✅",
-                                downstream_id, channel_id, msg.sequence_number, share_hash
+                                "SubmitSharesStandard: valid share | downstream_id: {}, channel_id: {}, sequence_number: {}, share_hash: {}, share_work: {} ✅",
+                                downstream_id, channel_id, msg.sequence_number, share_hash, share_work
                             );
                         }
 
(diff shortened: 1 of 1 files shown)


roles/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -529,28 +530,28 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
...
+                            let success = SubmitSharesSuccess {
+                                channel_id,
+                                last_sequence_number: share_accounting.get_last_share_sequence_number(),
+                                new_submits_accepted_count: share_accounting.get_shares_accepted(),
+                                new_shares_sum: share_accounting.get_share_work_sum() as u64,
+                            };
+                            info!("SubmitSharesStandard: {} ✅", success);
+                            messages.push((downstream_id, Mining::SubmitSharesSuccess(success)).into());
+                        } else {
+                            let share_work = standard_channel.get_target().difficulty_float();
```
average-gary, line 544: wouldn't `share_work` be something related to the share and not the channel difficulty?

plebhash: a share should be rewarded because it meets the difficulty target established by the server, if the hash has more zeros than necessary, that doesn't make that share more valuable

that's the rationale we have always followed to update `ShareAccounting::share_work_sum`, which I got from an interaction with Braiins a while ago

btw, that's also the same rationale you're following on #1965, the only difference is that it's being returned inside `ShareValidationResult::Valid`, which as I pointed out is unnecessary, since that information is already available via `get_target`

https://github.com/stratum-mining/stratum/pull/1965/files#r2461063936

plebhash: please see https://github.com/stratum-mining/sv2-spec/pull/126

for some reason the rendering of images on my fork is messed up, but I think this one summarizes things:

image

PR #1969, opened 2025-10-26
Remove Plain connection from network helpers

Shourya742:
We have an unmaintained plain connection module under the network helpers that's not used anywhere. This PR removes it, as SRI has no use for unencrypted connections (and I don’t see that changing anytime soon). If we ever get a feature request for it in the future, we can implement a cleaner and more suitable version then.


Shourya742:
Closing in favor of: https://github.com/stratum-mining/sv2-apps/pull/49

Issue #1933, opened 2025-10-02
version bumps for `v1.6.0`


GitGab19:
Similar to #1801

plebhash:
#1934 is bumping `channels_sv2` to `2.0.0`

plebhash:
#1931 is bumping `codec_sv2` from `3.0.0` to `4.0.0`

PR #1970, opened 2025-10-27
Add output GAT to handlers

Shourya742:
Currently, our handlers are quite restrictive, only allowing a method output of `()`. This works for our current use case, where no further processing happens after a handler completes. However, users may want to define their own execution flow.

This PR introduces an `Output` type, allowing trait implementors to decide what their handler methods should return. This makes the handler trait more flexible and extensible for different use cases.

Version bump is not included, as it was already done for this milestone: https://github.com/stratum-mining/stratum/issues/1933


Shourya742:
Can be added later, when we gonna make a dedicated change to handlers.

PR #1971, opened 2025-10-27
Correct Vardiff insertions

Shourya742:
Previously, the pool was using an incorrect key (downstream_id, channel_id) for inserting vardiff state. This has been corrected to the proper order (channel_id, downstream_id). Additionally, this PR updates the type of downstream_id from u32 to usize, since it serves as a local counter for downstreams and should support the maximum range allowed by the system.

### DIFF
--- a/roles/pool/src/lib/channel_manager/mod.rs
+++ b/roles/pool/src/lib/channel_manager/mod.rs
@@ -39,7 +39,7 @@ use crate::{
     error::PoolResult,
     status::{handle_error, Status, StatusSender},
     task_manager::TaskManager,
-    utils::{Message, ShutdownMessage},
+    utils::{Message, ShutdownMessage, VardiffKey},
 };
 
 mod mining_message_handler;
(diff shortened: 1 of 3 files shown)


Shourya742:
> Are you sure this is the right way to go?
> 
> The `client_id/downstream_id` MUST be unique, so it seems the perfect key.
> 
> In addition to this, different clients could theoretically have the same `channel_id`, but I don't remember now if the Pool allows for this or not.

Yes, the Pool allows different downstreams to share the same channel ID. Therefore, the combination of (downstream_id, channel_id) serves as the primary key to uniquely identify the vardiff state at the channel manager level.

Shourya742:
> There's no need for other keys IMO.
> 
> But the order which makes more sense is (`client_id`, `channel_id`), given that different clients can have the same `channel_id`.
> 
> So I would rather adapt JDC than changing the Pool in a "non-optimal" way.

I meant as an abstraction, rather than these vanilla `usize` types.

roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -187,7 +188,7 @@ impl ChannelManager {
                             Ok(ShutdownMessage::ShutdownAll) => {
                                 info!("Channel Manager: received shutdown signal");
                                 break;
-                            }
+                            },
```
GitGab19: Why this?

bit-aloo: are you referring to `,`, then might be because of fmt.

GitGab19: I just checked out your branch locally, removed it, and everything works.

PR #1972, opened 2025-10-27
Add contextual error to Pool

Shourya742:
This PR adds contextual error handling for issues originating in the channel manager. With these changes, errors caused by a single downstream no longer bring down the entire Pool only the affected downstream connection is gracefully disconnected.

### DIFF
--- a/roles/pool/src/lib/channel_manager/mod.rs
+++ b/roles/pool/src/lib/channel_manager/mod.rs
@@ -36,7 +36,7 @@ use tracing::{debug, error, info, warn};
 use crate::{
     config::PoolConfig,
     downstream::Downstream,
-    error::PoolResult,
+    error::{ContextualError, ContextualPoolResult, PoolResult, WithContext},
     status::{handle_error, Status, StatusSender},
     task_manager::TaskManager,
     utils::{Message, ShutdownMessage, VardiffKey},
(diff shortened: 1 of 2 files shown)


roles/pool/src/lib/error.rs
```diff
@@ -143,6 +144,47 @@ impl std::fmt::Display for PoolError {
...
+    TemplateReceiver {
+        error: Box<PoolError>,
+    },
+}
+
+pub trait WithContext<T> {
+    fn with_downstream_context(self, downstream_id: usize) -> Result<T, ContextualError>;
+    fn with_template_receiver_context(self) -> Result<T, ContextualError>;
+    fn with_channel_manager_context(self) -> Result<T, ContextualError>;
+}
```
plebhash, line 165: why add a trait here?

plebhash:
> errors caused by a single downstream no longer bring down the entire Pool only the affected downstream connection is gracefully disconnected.

how?

roles/pool/src/lib/error.rs
bit-aloo: Very astute observation.

My idea here is to couple the context with the status handler we already use. Currently, we provide the subsystem context directly to the status handler instead, I want the error itself to carry that context as it’s propagated. This trait is an initial step in that direction. Eventually, it will be moved into a common utilities module or crate that all roles can use.

plebhash: yeah overall I'm not a fan of adding traits that will only have one single local `impl`, because that's not only overengineered, but also arguably misleading

bit-aloo: As I mentioned, I plan to expand on this further. We’ll be adding more implementations across the roles over time. Without it, we can't get the .with_*_context kinda abstraction.

Shourya742:
Closed in favor of: https://github.com/stratum-mining/sv2-apps/pull/48

PR #1973, opened 2025-10-27
refactor: improve fuzz test suite with targeted harnesses

erickcestari:
Replace monolithic fuzz target with focused, isolated test harnesses for better coverage and debugging. This restructuring improves fuzzing efficiency by allowing the fuzzer to explore each code path independently.

Some targets are easily crashing by panic which indicates that we already found bugs that need to be fixed.

Current targets that are crashing:

**deserialize_seq0255**: crash due to attempt to add with overflow

`cargo fuzz run deserialize_seq0255`

```rust
panicked at /stratum/protocols/v2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs:338:1:
impl_into_encodable_field_for_seq!(bool);
```

**deserialize_setup_connection**: crash due to range start index 217 out of range for slice of length 103

`cargo fuzz run deserialize_setup_connection`

```rust
panicked at /stratum/protocols/v2/subprotocols/common-messages/src/setup_connection.rs:26:21:
#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)] // <-- comes from `Deserialize` stratum/protocols/v2/binary-sv2/derive_codec/src/lib.rs
pub struct SetupConnectionError<'decoder> {
```


fuzz/Cargo.toml
```diff
@@ -21,10 +21,47 @@ binary_sv2 = { path = "../sv2/binary-sv2"}
 parsers_sv2 = { path = "../sv2/parsers-sv2" }
 framing_sv2 = { path = "../sv2/framing-sv2" }
 codec_sv2 = { path = "../sv2/codec-sv2", features = ["noise_sv2"]}
+common_messages_sv2 = { path = "../sv2/subprotocols/common-messages" }
 affinity = "0.1.1"
```
lucasbalieiro: we should take this opportunity to update these dependencies.
this one isn’t used in the crate and is preventing builds on macOS.

erickcestari: Updated!

fuzz/Cargo.toml
plebhash, line 1: can we `git mv` this file instead of deleting and adding a new one?

plebhash:
`cargo fuzz run deserialize_seq0255` should be fixed with #1985

fuzz/fuzz_targets/deserialize_setup_connection.rs
```diff
@@ -0,0 +1,20 @@
...
+    let mut data1 = data.clone();
+    if let Ok(setup_msg) = SetupConnection::from_bytes(&mut data1) {
+        let dst = Vec::new();
+        setup_msg.to_bytes(&mut dst.clone()).unwrap();
+
+        let mut data2 = dst.clone();
+
+        let setup_msg2 = SetupConnection::from_bytes(&mut data2).unwrap();
+        let mut dst2 = Vec::new();
+        setup_msg2.to_bytes(&mut dst2).unwrap();
```
lucasbalieiro, line 16:
```rust
    let mut data1 = data.clone();
    let len = data1.len(); // <- get the len
    if let Ok(setup_msg) = SetupConnection::from_bytes(&mut data1) {
        let dst = vec![0u8; len]; // <-- pre-allocate it with the length
```
Shouldn't these `Vec`s be pre-allocated? Like you did for the other targets?

Also, while reviewing this I think I found some sort `off by one` error in the `GetSize` impl for the `SetupConnection`. This is why I'm suggesting using   `let len = data1.len();` here instead of `setup_msg.get_size()` because when using this the fuzzer catches some error related to the array length.

I'm still trying to find the problem with `GetSize` but you guys might find the answer first.

cc @plebhash

lucasbalieiro: Also, this `GetSize` error that I mentioned previously is unrelated to the error mentioned in the PR description: `range start index 217 out of range for slice of length 103`.

I think I was able to solve this one but I'm still running a few local tests to see if I haven't done anything stupid.

erickcestari: I have updated to use the `setup_msg.get_size()` that's probably the correct way of pre-allocate the buffer? Similar to the serialization of the others stratum types.

fuzz/Cargo.toml
erickcestari: Yes, done!

fuzz/Cargo.toml
```diff
@@ -0,0 +1,63 @@
...
+arbitrary = { version = "1.4.2", features = ["derive"] }
+binary_sv2 = { path = "../sv2/binary-sv2"}
+parsers_sv2 = { path = "../sv2/parsers-sv2" }
+framing_sv2 = { path = "../sv2/framing-sv2" }
+codec_sv2 = { path = "../sv2/codec-sv2", features = ["noise_sv2"]}
+common_messages_sv2 = { path = "../sv2/subprotocols/common-messages" }
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
```
bit-aloo: You can add fuzz inside the top-level workspace exclude field.

fuzz/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+/artifacts
+/corpus
```
bit-aloo, line 3: Can you add coverage here

fuzz/Cargo.toml
erickcestari: Sorry for the late reply, I've been very busy this week. Done!

fuzz/.gitignore
erickcestari: Added!

marshalokos (commented):
Approved

marshalokos:
Thank you.


Sent from Yahoo Mail for iPhone


On Monday, November 10, 2025, 3:28 PM, Erick λ ***@***.***> wrote:

erickcestari left a comment (stratum-mining/stratum#1973)

Hey, you might have generated a coverage HTML, can you share that.


Yes, you can see it here:

https://erickcestari.github.io/stratum/

PR #1975, opened 2025-10-27
drops closed channels from vardiff

lucasbalieiro:
closes stratum-mining/sv2-apps#43 

This makes the pool and JDC drop channels from the vardiff hashmap:
- when receive `CloseChannel` message.
- if the downstream drops without closing channels

### DIFF
--- a/roles/jd-client/src/lib/channel_manager/downstream_message_handler.rs
+++ b/roles/jd-client/src/lib/channel_manager/downstream_message_handler.rs
@@ -172,6 +172,9 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                     data.extended_channels.remove(&msg.channel_id);
                     data.standard_channels.remove(&msg.channel_id);
                 });
+                channel_manager_data
+                    .vardiff
+                    .remove(&(msg.channel_id, *downstream_id));
                 Ok(())
             })
     }
--- a/roles/jd-client/src/lib/channel_manager/mod.rs
+++ b/roles/jd-client/src/lib/channel_manager/mod.rs
@@ -143,7 +143,7 @@ pub struct ChannelManagerData {
     pending_downstream_requests: VecDeque<PendingChannelRequest>,
     // Factory for creating **custom mining jobs**, if available.
     job_factory: Option<JobFactory>,
-    // Mapping of `(downstream_id, channel_id)` → vardiff controller.
+    // Mapping of `(channel_id, downstream_id)` → vardiff controller.
     // Each entry manages variable difficulty for a specific downstream channel.
     vardiff: HashMap<(u32, u32), VardiffState>,
 }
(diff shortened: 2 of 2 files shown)


roles/pool/src/lib/channel_manager/mining_message_handler.rs
```diff
@@ -66,6 +67,11 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                         downstream_data.standard_channels.remove(&msg.channel_id);
                         downstream_data.extended_channels.remove(&msg.channel_id);
                     });
+                let vardiff_key = VardiffKey {
+                    downstream_id,
+                    channel_id: msg.channel_id,
+                };
+                channel_manager_data.vardiff.remove(&vardiff_key);
```
bit-aloo:
```suggestion
                let vardiff_key = VardiffKey {
                    downstream_id,
                    channel_id: msg.channel_id,
                };
                channel_manager_data.vardiff.remove(&(downstream_id, msg.channel_id).into());
```

roles/jd-client/src/lib/channel_manager/downstream_message_handler.rs
```diff
@@ -172,6 +172,9 @@ impl HandleMiningMessagesFromClientAsync for ChannelManager {
                     data.extended_channels.remove(&msg.channel_id);
                     data.standard_channels.remove(&msg.channel_id);
                 });
+                channel_manager_data
+                    .vardiff
+                    .remove(&(*downstream_id, msg.channel_id));
```
bit-aloo: This should be other way around, I have a PR upcoming for this soonish to remove this confusion.
```suggestion
                    .remove(&(msg.channel_id, *downstream_id));
```

roles/jd-client/src/lib/channel_manager/mod.rs
```diff
@@ -567,6 +567,7 @@ impl ChannelManager {
                     }
                 });
             }
+            cm_data.vardiff.retain(|key, _| key.0 != downstream_id);
```
bit-aloo:
```suggestion
            cm_data.vardiff.retain(|key, _| key.1 != downstream_id);
```

lucasbalieiro: huum, the comments above vardiff says its the other way around:
https://github.com/stratum-mining/stratum/blob/0b7f8257ffe6f05bdc025e8619045dab77afbd08/roles/jd-client/src/lib/channel_manager/mod.rs#L146-L148

roles/jd-client/src/lib/channel_manager/downstream_message_handler.rs
lucasbalieiro: ah ok, thanks

roles/jd-client/src/lib/channel_manager/mod.rs
bit-aloo: That's a mistake, I used the other convention for some reason. Misleading my bad.

bit-aloo: Can you change the doc as well, to remove the confusion.

PR #1977, opened 2025-10-27
make Pool's `CLIENT_SEARCH_SPACE_BYTES` more permissive

plebhash:
close stratum-mining/sv2-apps#44

since tProxy will likely operate with `downstream_extranonce2_size` at most 8, allowing total of 16 bytes for Pool's clients should be sufficient for most cases

### DIFF
--- a/test/integration-tests/tests/jd_tproxy_integration.rs
+++ b/test/integration-tests/tests/jd_tproxy_integration.rs
@@ -2,7 +2,7 @@ use integration_tests_sv2::{interceptor::MessageDirection, template_provider::Di
 use stratum_apps::stratum_core::{common_messages_sv2::*, mining_sv2::*};
 
 #[tokio::test]
-async fn jd_tproxy_integration() {
+async fn jd_non_aggregated_tproxy_integration() {
     start_tracing();
     let (tp, tp_addr) = start_template_provider(None, DifficultyLevel::Low);
     let (_pool, pool_addr) = start_pool(Some(tp_addr)).await;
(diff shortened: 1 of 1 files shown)


roles/pool/src/lib/channel_manager/mod.rs
```diff
@@ -46,7 +46,7 @@ mod mining_message_handler;
 mod template_distribution_message_handler;
 
 const POOL_ALLOCATION_BYTES: usize = 4;
-const CLIENT_SEARCH_SPACE_BYTES: usize = 8;
+const CLIENT_SEARCH_SPACE_BYTES: usize = 16;
```
bit-aloo, line 49: Should we make it 28?

plebhash: there's no fundamental reason preventing us from doing that

however, that's highly suggestive of the misconception that full extranonces must be fixed at 32 bytes, which in turn makes us highly susceptible to blindspots

in fact, I foresee some challenges around this topic when we start making adaptations coming from https://github.com/stratum-mining/sv2-spec/pull/162

so I'd like to avoid that

bit-aloo: Ah cool!!! let me check the spec stuff. Consider it non-blocking.

test/integration-tests/tests/jd_tproxy_integration.rs
```diff
@@ -2,15 +2,61 @@ use integration_tests_sv2::{interceptor::MessageDirection, template_provider::Di
...
+            MessageDirection::ToDownstream,
+            MESSAGE_TYPE_OPEN_EXTENDED_MINING_CHANNEL_SUCCESS,
+        )
+        .await;
+    jdc_pool_sniffer
+        .wait_for_message_type(
+            MessageDirection::ToUpstream,
+            MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED,
+        )
+        .await;
```
bit-aloo, line 62: This test doesn’t actually verify the behavior in non-aggregated mode. Ideally, we should check whether multiple mining instance connections result in multiple OpenExtendedMiningChannel messages.

plebhash: how can we achieve that?

unfortunately asserting for two identical messages is not trivial...

if we check the inner workings of `wait_for_message_type`, we'll see that it's redundant to call it 2x for the same message type

I tried doing `wait_for_message_type_and_clean_queue` but the test started timing out

---

I see how the test is asserting less than ideal

but overall, my goal with adding this test was just to make sure the original edge case that motivated this PR would be caught on CI

if we keep the test and revert the change on pool, it should crash

plebhash: wait, just realized I was doing some dumb stuff

plebhash: done

PR #1979, opened 2025-10-28
update code coverage workflow

### LINKED ISSUE #1978: `coverage_protocols.sh` has references of `roles_logic_sv2`
After merging #1860 the usual tarpaulin workflow has been executed, but it failed: https://github.com/stratum-mining/stratum/actions/runs/18876385849/job/53867175993

The reason is that the `coverage_protocols.sh` script still has some references to `roles_logic_sv2`, so it has to be fixed.

cc @lucasbalieiro

lucasbalieiro:
closes #1978 
as follow-up of #1860

this removes `roles_logic_sv2` from the code coverage script, and adapt where needed

### DIFF
--- a/.github/workflows/coverage-protocols.yaml
+++ b/.github/workflows/coverage-protocols.yaml
@@ -29,6 +29,22 @@ jobs:
           flags: protocols
           token: ${{ secrets.CODECOV_TOKEN }}
 
+      - name: Upload stratum_translation-coverage to codecov.io
+        uses: codecov/codecov-action@v4
+        with:
+          directory: ./target/tarpaulin-reports/stratum_translation-coverage
+          file: ./target/tarpaulin-reports/stratum-translation-coverage/cobertura.xml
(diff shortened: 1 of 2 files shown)


plebhash (approved):

.github/workflows/coverage-protocols.yaml
```diff
@@ -4,7 +4,9 @@ on:
   push:
     branches:
       - main
-
+  pull_request:
+    branches:
+      - main
```
GitGab19: remove this, we want to run this workflow only when we merge a PR to `main`.

lucasbalieiro: Sorry, I was just seeing if the workflow would work with these changes to the script. Commit removed.

Here is a workflow session with the buffer_sv2 and stratum_translation: https://github.com/stratum-mining/stratum/actions/runs/18880273971/job/53881119132?pr=1979

PR #1980, opened 2025-10-28
Fix clippy warnings by replacing `.is_multiple_of()` calls with idiomatic patterns

BRIGHTON-ASUMANI:
### Problem
The initial approach used `.is_multiple_of()`, but this method is only available in Rust 1.77+, while the project maintains MSRV 1.75.

### Changes
- `protocols/v2/framing-sv2/src/header.rs`: Use ceiling division for chunking
- `protocols/v1/src/utils.rs`: Use bitwise AND for oddness check  
- `protocols/v2/channels-sv2/src/server/share_accounting.rs`: Add allow attribute


BRIGHTON-ASUMANI:
> NACK. Please use MRSV

Thank you for the feedback. Switched to ceiling division and bitwise AND to maintain MSRV 1.75; clippy warnings are resolved.

PR #1982, opened 2025-10-30
fix channel_msg bit check in sv2 frame header

vnprc:
- correct CHANNEL_MSG_MASK from 0b0000_0000_0000_0001 (bit 0) to 0b1000_0000_0000_0000 (bit 15) to check MSB per SV2 spec
- update comparison logic from equality check to != 0 for correct bit masking semantics


GitGab19 (commented):
ACK, thanks for reporting and fixing this!

Could you please add an assertion on this function to the `sv2_frame` example that we have at `sv2/framing_sv2/examples/`?

GitGab19:
The `fmt` check is failing @vnprc, fix it by running `cargo +nightly fmt` locally.

PR #1983, opened 2025-10-30
Add support for extensions negotiations, `TLV`, and `Worker-Specific Hashrate Tracking` extension

### LINKED ISSUE #1742: Implement extension n.1 (`extensions-negotiation`) on SRI
Since we merged https://github.com/stratum-mining/sv2-spec/pull/112 on specs, we need to implement the needed messages to allow extensions negotiation.

This is a prerequisites for every other future extensions SRI wants to support.

### LINKED ISSUE #1743: Implement extension n.2 (`worker-specific-hashrate-tracking`) on SRI
Since we merged https://github.com/stratum-mining/sv2-spec/pull/113 on specs, and we have adopters needing this, we should implement extension n.2 (`worker-specific-hashrate-tracking`)  on SRI.

More details about it here: https://github.com/stratum-mining/sv2-spec/blob/main/extensions/worker-specific-hashrate-tracking.md

Blocked by #1742, since we need to implement the extensions negotiation messages first.

GitGab19:
This PR adds support for extensions negotiations, `TLV`, and `Worker-Specific Hashrate Tracking` extension.

A new crate called `extensions_sv2` has been created, with modules defining [Extension n.1 message types](https://github.com/stratum-mining/sv2-spec/blob/main/extensions/extensions-negotiation.md#2-messages-defined-by-this-extension), TLV encoding (as defined [here](https://github.com/stratum-mining/sv2-spec/blob/main/03-Protocol-Overview.md#341-stratum-v2-tlv-encoding-model)), and [Extension n.2](https://github.com/stratum-mining/sv2-spec/blob/main/extensions/worker-specific-hashrate-tracking.md) helper functions.

To better support and handle extension messages and TLV fields appended to existing messages, some breaking changes on `handlers_sv2` have been done:
- `Extensions negotiation`:
  - created new handler traits in `extensions.rs` to manage incoming new messages defined by extensions (for now we only have the ones defined in [Extension n.1](https://github.com/stratum-mining/sv2-spec/blob/main/extensions/extensions-negotiation.md))
- `TLV` and `Extension n.2`:
  - modified all the current handlers to accept a new parameter for optional TLV fields appended to the current messages
  - added a new method to all current handlers called `get_negotiated_extensions_with_server()` which is used by the trait methods `handle_*_message_frame_from_*` internally
 
I still need to fully validate these new changes by implementing it on some apps.

Closes #1742 
Closes #1743 

PS: integration tests are failing on CI, because this PR is introducing breaking changes in handlers, so that is perfectly expected.

### DIFF
--- a/sv2/handlers-sv2/Cargo.toml
+++ b/sv2/handlers-sv2/Cargo.toml
@@ -16,7 +16,7 @@ trait-variant = "0.1.2"
 parsers_sv2 = { path = "../parsers-sv2", version = "^0.2.0"}
 binary_sv2 = { path = "../binary-sv2", version = "^5.0.0" }
 common_messages_sv2 = { path = "../subprotocols/common-messages", version = "^6.0.0" }
-framing_sv2 = { path = "../framing-sv2", version = "^5.0.0" }
+framing_sv2 = { path = "../framing-sv2", version = "^6.0.0" }
 mining_sv2 = { path = "../subprotocols/mining", version = "^6.0.0" }
 template_distribution_sv2 = { path = "../subprotocols/template-distribution", version = "^4.0.0" }
 job_declaration_sv2 = { path = "../subprotocols/job-declaration", version = "^5.0.0" }
--- a/stratum-core/Cargo.toml
+++ b/stratum-core/Cargo.toml
@@ -17,7 +17,7 @@ bitcoin = "0.32.5"
 binary_sv2 = { path = "../sv2/binary-sv2", version = "^5.0.0" }
 codec_sv2 = { path = "../sv2/codec-sv2", version = "^4.0.0", features = ["noise_sv2"]}
 extensions_sv2 = { path = "../sv2/extensions-sv2", version = "^0.1.0" }
-framing_sv2 = { path = "../sv2/framing-sv2", version = "^5.0.0" }
+framing_sv2 = { path = "../sv2/framing-sv2", version = "^6.0.0" }
 noise_sv2 = { path = "../sv2/noise-sv2", version = "^1.0.0" }
 parsers_sv2 = { path = "../sv2/parsers-sv2", version = "^0.2.0" }
 handlers_sv2 = { path = "../sv2/handlers-sv2", version = "^0.2.0" }
--- a/sv2/codec-sv2/Cargo.toml
+++ b/sv2/codec-sv2/Cargo.toml
@@ -12,7 +12,7 @@ homepage = "https://stratumprotocol.org"
 keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 [dependencies]
-framing_sv2 = { path = "../framing-sv2", version = "^5.0.0" }
+framing_sv2 = { path = "../framing-sv2", version = "^6.0.0" }
 noise_sv2 = { path = "../noise-sv2", default-features = false, optional = true, version = "^1.0.0" }
 binary_sv2 = { path = "../binary-sv2", version = "^5.0.0" }
 buffer_sv2 = { path = "../buffer-sv2", version = "^2.0.0" }
--- a/sv2/extensions-sv2/Cargo.toml
+++ b/sv2/extensions-sv2/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["stratum", "mining", "bitcoin", "protocol", "extensions"]
 [dependencies]
 binary_sv2 = { path = "../binary-sv2", version = "^5.0.0" }
 codec_sv2 = { path = "../codec-sv2", version = "^4.0.0" }
-framing_sv2 = { path = "../framing-sv2", version = "^5.0.2" }
+framing_sv2 = { path = "../framing-sv2", version = "^6.0.0" }
 mining_sv2 = { path = "../subprotocols/mining", version = "^6.0.0" }
 
 [dev-dependencies]
--- a/sv2/framing-sv2/Cargo.toml
+++ b/sv2/framing-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "framing_sv2"
-version = "5.0.2"
+version = "6.0.0"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"
--- a/sv2/parsers-sv2/Cargo.toml
+++ b/sv2/parsers-sv2/Cargo.toml
@@ -13,7 +13,7 @@ keywords = ["stratum", "mining", "bitcoin", "protocol"]
 
 [dependencies]
 binary_sv2 = { path = "../binary-sv2", version = "^5.0.0" }
-framing_sv2 = { path = "../framing-sv2", version = "^5.0.0" }
+framing_sv2 = { path = "../framing-sv2", version = "^6.0.0" }
 common_messages_sv2 = { path = "../subprotocols/common-messages", version = "^6.0.0" }
 mining_sv2 = { path = "../subprotocols/mining", version = "^6.0.0" }
 template_distribution_sv2 = { path = "../subprotocols/template-distribution", version = "^4.0.0" }


sv2/extensions-sv2/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Worker-Specific Hashrate Tracking (0x0002)
+
+This extension enables tracking of individual worker hashrates within extended channels by adding TLV (Type-Length-Value) fields to `SubmitSharesExtended` messages. It allows pools to monitor per-worker performance while maintaining a single extended channel for multiple workers.
+
+**Types:**
+- `UserIdentity` - Worker identifier (max 32 bytes UTF-8)
+- `build_submit_shares_extended_with_user_identity_frame()` - Helper to create frames with optional worker identity
+- `encode_user_identity_tlv()` - Encode worker identity as TLV
+- `decode_user_identity_tlv()` - Decode worker identity from TLV
```
bit-aloo: I won't put this here. Or maybe under a different header name.

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+/// are negotiated. The format is:
+/// - Type (U16 | U8): Extension type (U16) + Field type (U8) = 3 bytes
+/// - Length (U16): Length of the value in bytes = 2 bytes  
+/// - Value: The actual field data = N bytes
+///
+/// Total overhead: 5 bytes + value length
+///
+/// TLV header size in bytes (3 bytes Type + 2 bytes Length)
+pub const TLV_HEADER_SIZE: usize = 5;
+
```
bit-aloo: For this, we should introduce a Tlv struct, something along these lines:

```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct Tlv {
    pub extension_type: u16,
    pub field_type: u8,
    pub value: Vec<u8>,
}
```

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+/// # Example
+///
+/// ```ignore
+/// use extensions_sv2::encode_tlv;
+///
+/// let worker_name = b"Worker_001";
+/// let tlv = encode_tlv(0x0002, 0x01, worker_name);
+/// // tlv = [0x00, 0x02, 0x01, 0x00, 0x0a, b'W', b'o', ...]
+/// ```
+pub fn encode_tlv(extension_type: u16, field_type: u8, value: &[u8]) -> Vec<u8> {
```
bit-aloo: Rather than vanilla &[u8], we can have AsRef<[u8]>, Something like this:

```rust
pub fn encode_tlv(extension_type: u16, field_type: u8, value: AsRef<[u8]>)
```

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+///
+/// ```ignore
+/// use extensions_sv2::decode_tlv;
+///
+/// let tlv_data = &[0x00, 0x02, 0x01, 0x00, 0x0a, ...];
+/// let (ext_type, field_type, value, consumed) = decode_tlv(tlv_data)?;
+/// assert_eq!(ext_type, 0x0002);
+/// assert_eq!(field_type, 0x01);
+/// ```
+pub fn decode_tlv(data: &[u8]) -> Result<(u16, u8, Vec<u8>), TlvError> {
```
bit-aloo: Same for this, we can make it AsRef<[u8]>

Shourya742 (commented):
Just a basic review on tlv side of definitions. Will be having deeper look soon.

sv2/extensions-sv2/src/tlv/codec.rs
bit-aloo: All the encode, decode and other helper methods can be part of its impl.

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
+use alloc::vec;
+use alloc::vec::Vec;
+use binary_sv2::{Deserialize, Encodable};
```
bit-aloo:
```suggestion
use binary_sv2::{Deserialize, Serialize};
```

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
...
+
+/// Generic TLV (Type-Length-Value) encoder and decoder.
+///
+/// TLV fields are appended to the end of base protocol messages when extensions
+/// are negotiated. The format is:
+/// - Type (U16 | U8): Extension type (U16) + Field type (U8) = 3 bytes
+/// - Length (U16): Length of the value in bytes = 2 bytes  
+/// - Value: The actual field data = N bytes
+///
+/// Total overhead: 5 bytes + value length
```
bit-aloo: Total overhead can be changed.

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
...
+/// Structure matches the Stratum V2 TLV specification:
+/// - Type (3 bytes): extension_type (U16) + field_type (U8)
+/// - Length (2 bytes): U16 indicating value size
+/// - Value (N bytes): The actual data
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Tlv {
+    /// The extension type (U16) - identifies which extension this TLV belongs to
+    pub extension_type: u16,
+    /// The field type (U8) - identifies the specific field within the extension
+    pub field_type: u8,
```
bit-aloo: We can group these field under the struct `Type`.

```rust
struct Type {
  pub extension_type: u16,
  pub field_type: u8
}
```

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
...
+        Ok(Self {
+            extension_type,
+            field_type,
+            length,
+            value,
+        })
+    }
+
+    /// Returns the total size of this TLV when encoded (header + value).
+    pub fn encoded_size(&self) -> usize {
```
bit-aloo:
```suggestion
   #[inline]
    pub fn encoded_size(&self) -> usize {
```

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
...
+        TLV_HEADER_SIZE + self.value.len()
+    }
+
+    /// Validates that the length field matches the actual value length.
+    ///
+    /// This is useful for ensuring TLV integrity after deserialization.
+    pub fn is_valid(&self) -> bool {
```
bit-aloo:
```suggestion
   #[inline]
    pub fn is_valid(&self) -> bool {
```

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
...
+
+        self.length
+            .to_bytes(&mut tlv[offset..offset + 2])
+            .map_err(|_| TlvError::EncodingError)?;
+        offset += 2;
+
+        tlv[offset..].copy_from_slice(&self.value);
+
+        Ok(tlv)
+    }
```
Shourya742:
```suggestion
pub fn encode(&self) -> Result<Vec<u8>, TlvError> {
    let mut tlv = Vec::with_capacity(TLV_HEADER_SIZE + self.value.len());

    tlv.extend_from_slice(&self.extension_type.to_le_bytes());
    tlv.push(self.field_type);
    tlv.extend_from_slice(&self.length.to_le_bytes());
    tlv.extend_from_slice(&self.value);

    Ok(tlv)
}
```

bit-aloo: We can create a `From` implementation for the encoding error now.

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,324 @@
...
+        offset += 2;
+
+        let mut field_type_bytes = [data[offset]];
+        let field_type =
+            u8::from_bytes(&mut field_type_bytes).map_err(|_| TlvError::DecodingError)?;
+        offset += 1;
+
+        let mut length_bytes = [data[offset], data[offset + 1]];
+        let length = u16::from_bytes(&mut length_bytes).map_err(|_| TlvError::DecodingError)?;
+        offset += 2;
```
bit-aloo: We can create a From implementation for the decoding error.

sv2/extensions-sv2/src/tlv/error.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+}
+
+impl From<binary_sv2::Error> for TlvError {
+    fn from(err: binary_sv2::Error) -> Self {
+        TlvError::DeserializationFailed(err)
+    }
+}
+
```
bit-aloo: We can add from implementation here for `from_bytes` and `to_bytes` error?

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+    ///
+    /// Each u16 value corresponds to a specific extension identifier.
+    /// For example: 0x0001 for Extensions Negotiation (this extension itself)
+    pub requested_extensions: Seq064K<'decoder, u16>,
+}
+
+impl fmt::Display for RequestExtensions<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Clone to get the inner vec for display
+        let extensions_list = format!("{:?}", self.requested_extensions.clone().into_inner());
```
bit-aloo: We implemented Display Trait for Seq064k a while ago, cloning can be removed.
```suggestion
        let extensions_list = format!("{}", self.requested_extensions);
```

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// This must be a subset of the extensions requested in the RequestExtensions
+    /// message. Extensions not listed here are not supported and MUST NOT be used
+    /// by the client.
+    pub supported_extensions: Seq064K<'decoder, u16>,
+}
+
+impl fmt::Display for RequestExtensionsSuccess<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Clone to get the inner vec for display
+        let extensions_list = format!("{:?}", self.supported_extensions.clone().into_inner());
```
bit-aloo: Same for this, cloning can be removed
```suggestion
        let extensions_list = format!("{}", self.supported_extensions);
```

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Contains extension identifiers that the server requires but were not included
+    /// in the client's request. If the client does not retry with these required
+    /// extensions, the server MUST disconnect the client.
+    pub required_extensions: Seq064K<'decoder, u16>,
+}
+
+impl fmt::Display for RequestExtensionsError<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Clone to get the inner vecs for display
+        let unsupported_list = format!("{:?}", self.unsupported_extensions.clone().into_inner());
```
bit-aloo: cloning can be removed.
```suggestion
        let unsupported_list = format!("{}", self.unsupported_extensions);
```

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+        let required_list = format!("{:?}", self.required_extensions.clone().into_inner());
```
bit-aloo: Cloning can be removed.
```suggestion
        let required_list = format!("{}", self.required_extensions);
```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/user_identity.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Returns the raw bytes of the worker identity.
+    ///
+    /// This is useful when you need to serialize the identity or work with
+    /// the raw bytes directly.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.user_identity
+    }
+
+    /// Returns the length of the worker identity in bytes.
+    pub fn len(&self) -> usize {
```
bit-aloo, line 82:
```suggestion
   #[inline]
    pub fn len(&self) -> usize {
```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/user_identity.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.user_identity.len()
+    }
+
+    /// Returns true if the worker identity is empty.
+    pub fn is_empty(&self) -> bool {
```
bit-aloo, line 88:
```suggestion
   #[inline]
    pub fn is_empty(&self) -> bool {
```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
```diff
@@ -0,0 +1,445 @@
...
+        0
+    };
+
+    // Construct complete frame bytes
+    let mut frame_bytes = Vec::with_capacity(Header::SIZE + payload.len());
+    frame_bytes.extend_from_slice(&extension_type.to_le_bytes());
+    frame_bytes.push(MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED);
+    frame_bytes.push((msg_length & 0xFF) as u8);
+    frame_bytes.push(((msg_length >> 8) & 0xFF) as u8);
+    frame_bytes.push(((msg_length >> 16) & 0xFF) as u8);
```
bit-aloo: This should be easier to understand
```suggestion
    frame_bytes.extend_from_slice(&msg_length.to_le_bytes()[..3]);
```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
```diff
@@ -0,0 +1,445 @@
...
+    frame_bytes.extend_from_slice(&payload);
+
+    // Convert to StandardSv2Frame
+    StandardSv2Frame::from_bytes(frame_bytes.into())
+        .map_err(|_| TlvError::SerializationFailed(binary_sv2::Error::WriteError(0, 0)))
```
bit-aloo: Why `WriterError(0,0)` ?

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
```diff
@@ -0,0 +1,445 @@
...
+// /// // Extract worker identity if present
+// /// match extract_worker_identity_from_submit_shares(&base_msg, payload)? {
+// ///     Some(worker_id) => {
+// ///         println!("Share from worker: {}", worker_id.as_string_or_hex());
+// ///     }
+// ///     None => {
+// ///         println!("Share without worker identity");
+// ///     }
+// /// }
+// /// ```
```
bit-aloo: Should this be uncommented?

GitGab19: I already deleted it.

sv2/extensions-sv2/src/tlv/codec.rs
GitGab19: What do you mean?

bit-aloo: Shouldn't this be something like total size?

sv2/extensions-sv2/src/tlv/codec.rs
GitGab19: Why?

sv2/extensions-sv2/src/tlv/codec.rs
GitGab19: Why?

sv2/extensions-sv2/src/tlv/codec.rs
bit-aloo: Because these methods are simple one-liners, adding an additional stack frame for each would be unnecessary overhead.

sv2/extensions-sv2/src/tlv/codec.rs
bit-aloo: Same as above

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
GitGab19: Changed.

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// field from the TLV data.
+    pub fn find_field(tlvs: &[Self], extension_type: u16, field_type: u8) -> Option<&Self> {
+        tlvs.iter().find(|tlv| {
+            tlv.type_.extension_type == extension_type && tlv.type_.field_type == field_type
+        })
+    }
+}
+
+/// Checks if data contains a TLV field related to the specified extension type.
+pub fn has_tlv_for_extension<T: AsRef<[u8]>>(data: T, expected_extension_type: u16) -> bool {
```
bit-aloo: I think the Tlv struct shouldn’t be responsible for parsing entire TLV buffers. Instead, we could introduce a TlvList type to represent a collection of TLVs, and provide an iterator-based interface to stream over them, something like this:

```rust

pub struct TlvIter<'a> {
    data: &'a [u8],
    offset: usize,
}

impl<'a> TlvIter<'a> {
    pub fn new(data: &'a [u8]) -> Self {
        Self { data, offset: 0 }
    }
}

impl<'a> Iterator for TlvIter<'a> {
    type Item = Result<Tlv, TlvError>;

    fn next(&mut self) -> Option<Self::Item> {
        if self.offset + TLV_HEADER_SIZE > self.data.len() {
            return None;
        }

        match Tlv::decode(&self.data[self.offset..]) {
            Ok(tlv) => {
                let size = tlv.encoded_size();
                self.offset += size;
                Some(Ok(tlv))
            }
            Err(e) => {
                self.offset = self.data.len();
                Some(Err(e))
            }
        }
    }
}

#[derive(Debug, Clone)]
pub struct TlvList<'a> {
    data: &'a [u8],
}

impl<'a> TlvList<'a> {
    pub fn new(data: &'a [u8]) -> Self {
        Self { data }
    }

    pub fn iter(&self) -> TlvIter<'a> {
        TlvIter::new(self.data)
    }

    pub fn to_vec(&self) -> Vec<Tlv> {
        self.iter()
            .filter_map(|r| r.ok())
            .collect()
    }

    pub fn for_extensions(&self, negotiated: &[u16]) -> Vec<Tlv> {
        self.iter()
            .filter_map(|r| match r {
                Ok(tlv) if negotiated.contains(&tlv.type_.extension_type) => Some(tlv),
                _ => None,
            })
            .collect()
    }

    pub fn find(&self, ext: u16, field: u8) -> Option<Tlv> {
        self.iter().find_map(|r| match r {
            Ok(tlv) if tlv.type_.extension_type == ext && tlv.type_.field_type == field => Some(tlv),
            _ => None,
        })
    }
}

```

Let me know what you think of this?

GitGab19: I've done it, cleaning up the previous similar functions we had implemented on `Tlv`.

sv2/parsers-sv2/src/lib.rs
```diff
@@ -422,10 +499,137 @@ impl AnyMessage<'_> {
...
+        if !tlvs.is_empty() {
+            Some(tlvs)
+        } else {
+            None
+        }
+    } else {
+        None
+    };
+
+    Ok((message.into_static(), tlv_fields))
```
bit-aloo: We can consolidate all these parsing functions into a single entry point. Since the message type already indicates which message variant it belongs to, determining the correct decoder is straightforward. A unified abstraction, for example, a deserialize_frame abstraction with a decode method and supporting helper routines, can handle this cleanly and reduce duplication.

sv2/extensions-sv2/src/tlv/error.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Failed to encode TLV data using binary_sv2.
+    ///
+    /// Contains the underlying binary_sv2::Error.
+    EncodingError(binary_sv2::Error),
+
+    /// Failed to decode TLV data using binary_sv2.
+    ///
+    /// Contains the underlying binary_sv2::Error.
+    DecodingError(binary_sv2::Error),
```
bit-aloo: Aren't these pairs the same?

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,513 @@
...
+
+/// Finds a specific TLV field by extension type and field type from a slice of TLVs.
+///
+/// This is a convenience function for searching through already-parsed TLV fields.
+/// For searching in a byte buffer, use `TlvList::find()` instead.
+pub fn find_tlv_field(tlvs: &[Tlv], extension_type: u16, field_type: u8) -> Option<&Tlv> {
+    tlvs.iter().find(|tlv| {
+        tlv.type_.extension_type == extension_type && tlv.type_.field_type == field_type
+    })
+}
```
bit-aloo: We can remove this, and let caller do iter find

sv2/extensions-sv2/src/tlv/error.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/// Wrapper type for encoding context to enable `From` trait implementation.
+#[derive(Debug)]
+pub struct EncodingContext(pub binary_sv2::Error);
+
+/// Wrapper type for decoding context to enable `From` trait implementation.
+#[derive(Debug)]
+pub struct DecodingContext(pub binary_sv2::Error);
+
```
bit-aloo: We dont need these

sv2/extensions-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,513 @@
...
+/// let tlvs = list.for_extensions(&negotiated);
+/// ```
+#[derive(Debug, Clone)]
+pub struct TlvList<'a> {
+    data: &'a [u8],
+}
+
+impl<'a> TlvList<'a> {
+    /// Creates a new TLV list from a byte buffer.
+    pub fn new(data: &'a [u8]) -> Self {
```
bit-aloo: AsRef can  be used here

sv2/extensions-sv2/src/tlv/codec.rs
GitGab19: Why? Isn't it simpler to just call this function passing all data needed?

sv2/parsers-sv2/src/lib.rs
GitGab19: The `message_type` is not enough (a new message defined in an extension can have the same `message_type` of other protocol messages).

So the real key is made by both `extension_type` and `message_type` together.

That said, I think to remove duplication we can have a unified function which extracts the tlvs and return them with the parsed message.

But this function is called internally by all the other functions.

I'll push very soon these updates.

GitGab19: Done here: https://github.com/stratum-mining/stratum/pull/1983/commits/853ee5236faedfb4ae0e64bd30b53998070abac0, let me know what you think about it.

sv2/extensions-sv2/src/tlv/error.rs
GitGab19: Thx, I left only `EncodingError` and `DecodingError` variants.

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
```diff
@@ -0,0 +1,638 @@
...
+    // Construct complete frame bytes (header + payload)
+    let header_size = 6; // 2 bytes ext_type + 1 byte msg_type + 3 bytes length
+    let mut frame_bytes = Vec::with_capacity(header_size + payload.len());
+    frame_bytes.extend_from_slice(&extension_type.to_le_bytes());
+    frame_bytes.push(MESSAGE_TYPE_SUBMIT_SHARES_EXTENDED);
+    frame_bytes.extend_from_slice(&msg_length.to_le_bytes()[..3]);
+    frame_bytes.extend_from_slice(&payload);
+
+    Ok(frame_bytes)
+}
```
bit-aloo: Instead of this we can go with something like, it gonna work for all the messages SRI supports : 

```rust
pub fn build_frame_bytes<T>(&self, message: T) -> Result<Vec<u8>, TlvError>
where
    T: Serialize + GetSize + IsSv2Message,
{
    let extension_type: u16 = (message.channel_bit() as u16) << 15;
    let message_type = message.message_type();

    let mut payload = to_bytes(message).map_err(TlvError::EncodingError)?;

    for tlv in self.iter().flatten() {
        payload.extend_from_slice(&tlv.encode()?);
    }

    let msg_length = payload.len() as u32;
    let mut frame_bytes = Vec::with_capacity(6 + payload.len());

    frame_bytes.extend_from_slice(&extension_type.to_le_bytes());
    frame_bytes.push(message_type);  
    frame_bytes.extend_from_slice(&msg_length.to_le_bytes()[..3]); 
    frame_bytes.extend_from_slice(&payload);       

    Ok(frame_bytes)
}
```

bit-aloo: This method is part of TlvList

sv2/extensions-sv2/src/worker_specific_hashrate_tracking.rs
```diff
@@ -0,0 +1,638 @@
...
+/// The worker identity is stored as raw UTF-8 bytes (max 32 bytes) to match
+/// the TLV specification exactly.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UserIdentity {
+    /// Worker identifier as raw UTF-8 bytes (max 32 bytes).
+    ///
+    /// The TLV Value field contains these raw bytes directly.
+    pub(crate) user_identity: Vec<u8>,
+}
+
```
bit-aloo, line 36: I am now thinking if we can introduce a trait to represent all extension types:
```rust

pub trait TlvField: Sized {

    const EXTENSION_TYPE: u16;
    const FIELD_TYPE: u8;

    fn from_bytes(bytes: &[u8]) -> Result<Tlv, TlvError>;

    fn to_bytes(&self) -> Result<Vec<u8>, TlvError>;

    fn from_tlv(tlv: &Tlv) -> Result<Self, TlvError>;

    fn to_tlv(&self) -> Result<Tlv, TlvError>;
}

```
Which encapsulate all the extensions, that would be much better for extensibility. And also the methods defined in the trait, are the one's which are important. Rest can go into the implementor impl.

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
```diff
@@ -0,0 +1,638 @@
...
+        Ok(vec![self.to_tlv()?])
+    }
+
+    /// Creates a UserIdentity from a string and converts it to Vec<Tlv> in one step.
+    pub fn from_str_to_tlv_vec(user_identity: &str) -> Result<Vec<Tlv>, TlvError> {
+        let identity =
+            Self::new(user_identity).map_err(|_| TlvError::WorkerIdTooLong(user_identity.len()))?;
+        identity.to_tlv_vec()
+    }
+
```
bit-aloo: These are utility methods not sure if we should have them inside impl

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
```diff
@@ -0,0 +1,638 @@
...
+        tlv::find_tlv_field(tlv_fields, EXTENSION_TYPE, FIELD_TYPE_USER_IDENTITY).and_then(|tlv| {
+            // Validate length
+            if tlv.value.len() > MAX_WORKER_ID_LENGTH {
+                return None;
+            }
+
+            // Convert to UTF-8 string
+            core::str::from_utf8(&tlv.value).ok().map(|s| s.into())
+        })
+    }
```
bit-aloo: Same for this, more like a utility method. Having something like `from_tlv` would be more useful

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
GitGab19: I find this idea very interesting, I'll try to add it soon.

GitGab19:
> As for the user-specific hashrate part, we can define a TlvField trait within parser-sv2, which UserIdentity can implement to provide TLV-specific methods.

In this case the if we keep the `UserIdentity` definition in `extensions_sv2`, we have to import `parsers_sv2` to get the trait --> circular dependency

The only alternative is to have the `UserIndentity` TLV (together with all the other eventual future TLVs) in `parsers_sv2`, but I don't like too much this idea.

sv2/extensions-sv2/src/tlv/codec.rs
GitGab19: I'll remove it.

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
GitGab19: I refactored this following your suggestion, let me know what you think about it.

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
GitGab19: I removed them for now. 

I'll adapt again all the apps tomorrow and I'll see if it was really needed or not.

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlv.rs
GitGab19: I removed them for now. 

I'll adapt again all the apps tomorrow and I'll see if it was really needed or not.

sv2/extensions-sv2/src/worker_specific_hashrate_tracking.rs
GitGab19: I added the trait as you suggested here, with the implementation on the `UserIdentity` in `parsers_sv2` as we discussed.

sv2/parsers-sv2/src/tlv_extensions/error.rs
```diff
@@ -0,0 +1,68 @@
...
+    InvalidExtensionType(u16),
+
+    /// Field type in TLV does not match expected value (0x01).
+    ///
+    /// Contains the actual field_type found.
+    InvalidFieldType(u8),
+
+    /// Invalid UTF-8 in UserIdentity.
+    InvalidUtf8(String),
+}
```
bit-aloo, line 51: I think we can structure the extension_sv2 module similar to the subprotocol module, just define the messages here and expose its error type through the parser-sv2 error variant.

sv2/parsers-sv2/src/tlv/error.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Failed to decode/deserialize TLV data using binary_sv2.
+    ///
+    /// Contains the underlying binary_sv2::Error.
+    DecodingError(binary_sv2::Error),
+
+    /// Failed to construct a StandardSv2Frame from bytes.
+    ///
+    /// Contains the error code returned by `Sv2Frame::from_bytes`.
+    FrameConstructionFailed(isize),
+}
```
bit-aloo, line 25: We can make this a variant of Parsers-sv2

sv2/parsers-sv2/src/tlv_extensions/mod.rs
```diff
@@ -0,0 +1,77 @@
+//! TLV field implementations for known SV2 extensions.
+//!
+//! This module provides `TlvField` trait implementations for extension types
```
bit-aloo, line 3: I think this module should be outside of tlv module.

sv2/parsers-sv2/src/tlv/extensions.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Maximum length for worker identifiers in bytes as per the spec.
+const MAX_WORKER_ID_LENGTH: usize = 32;
+
+/// Implementation of TlvField trait for UserIdentity.
+///
+/// This provides the standard interface for encoding/decoding UserIdentity
+/// as a TLV field in the Worker-Specific Hashrate Tracking extension.
+impl TlvField for UserIdentity {
+    type Error = UserIdentityError;
```
bit-aloo: If we start using Parser-sv2 error variant, we won't be needing this associated type.

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,552 @@
...
+use alloc::{boxed::Box, string::String, vec::Vec};
+use core::fmt::{self, Write};
+
+extern crate alloc;
+
+/// TLV header size in bytes (3 bytes Type + 2 bytes Length)
+pub const TLV_HEADER_SIZE: usize = 5;
+
+/// Trait for types that can be encoded/decoded as TLV fields.
+///
```
bit-aloo: @GitGab19 what you think of modularizing this file. This one currently, contains all the traits, objects and iterators. Maybe we can structure them inside a separate submodule.

sv2/parsers-sv2/src/tlv_extensions/mod.rs
GitGab19: I'll create a new module called `tlv_extensions` for this though.

sv2/parsers-sv2/src/tlv/codec.rs
GitGab19: I restructured the module into different submodules: list, iter, and error.

sv2/parsers-sv2/src/tlv/extensions.rs
GitGab19: Removed it.

sv2/parsers-sv2/src/tlv/error.rs
GitGab19: Done.

sv2/parsers-sv2/src/tlv_extensions/error.rs
GitGab19: Moved this into `parser_sv2::tlv_extensions::error`, wrapped it in a new `ExtensionError`, and added to the `ParserError` variants.

sv2/parsers-sv2/src/lib.rs
GitGab19: I added a new unique entry point for handlers called `parse_message_frame_with_tlvs` in https://github.com/stratum-mining/stratum/pull/1983/commits/b4cd94489fc33c23e007b1c6fb1ea788bb9f62e3.

That's gonna be useful for apps as well, let me know what you think about it.

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,554 @@
...
+/// # Structure
+/// - Type (3 bytes): extension_type (U16) + field_type (U8)
+/// - Length (2 bytes): U16 length of value in bytes
+/// - Value (variable): Raw byte data
+///
+/// Total size: 5 bytes (TLV header) + value length
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Tlv {
+    /// TLV type (extension_type + field_type)
+    pub type_: Type,
```
bit-aloo:
```suggestion
     pub r#type: Type
```

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// Creates a new TLV field.
+    ///
+    /// # Arguments
+    /// * `extension_type` - The extension identifier
+    /// * `field_type` - The field type within the extension
+    /// * `value` - The raw value bytes
+    pub fn new(extension_type: u16, field_type: u8, value: Vec<u8>) -> Self {
+        let length = value.len() as u16;
+        Self {
+            type_: Type::new(extension_type, field_type),
```
bit-aloo:
```suggestion
            r#type: Type::new(extension_type, field_type),
```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking/tlvs.rs
```diff
@@ -0,0 +1,169 @@
...
+impl UserIdentity {
+    /// Creates a new UserIdentity from a string.
+    pub fn new(user_identity: &str) -> Result<Self, &'static str> {
+        let bytes = user_identity.as_bytes();
+        if bytes.len() > MAX_USER_IDENTITY_LENGTH {
+            return Err("UserIdentity exceeds 32 bytes");
+        }
+        Ok(Self {
+            user_identity: bytes.to_vec(),
+        })
```
bit-aloo:
```suggestion
     Self::from_bytes(user_identity.as_bytes())
```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking.rs
```diff
@@ -0,0 +1,169 @@
...
+                for byte in &self.user_identity {
+                    use core::fmt::Write;
+                    let _ = write!(&mut hex, "{:02x}", byte);
+                }
+                hex
+            }
+        }
+    }
+
+    /// Returns the raw bytes of the UserIdentity.
```
bit-aloo, line 74: we can make this inline

sv2/extensions-sv2/src/worker_specific_hashrate_tracking.rs
```diff
@@ -0,0 +1,169 @@
...
+        if bytes.len() > MAX_USER_IDENTITY_LENGTH {
+            return Err("UserIdentity exceeds 32 bytes");
+        }
+        Ok(Self {
+            user_identity: bytes.to_vec(),
+        })
+    }
+
+    /// Returns the UserIdentity as a string slice (if valid UTF-8).
+    pub fn as_str(&self) -> Option<&str> {
```
bit-aloo, line 55: We can make this inline

sv2/extensions-sv2/src/worker_specific_hashrate_tracking.rs
```diff
@@ -0,0 +1,169 @@
+use alloc::{
```
bit-aloo, line 6: Can we change name of this file?, We don't need this module, we can just have a `worker_specific_hashrate_tracking.rs` file

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+impl fmt::Display for RequestExtensions<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Clone to get the inner vec for display
+        let extensions_list = format!("{}", self.requested_extensions);
+        write!(
+            f,
+            "RequestExtensions(request_id: {}, requested_extensions: {})",
+            self.request_id, extensions_list
+        )
+    }
```
bit-aloo:
```suggestion
        write!(
            f,
            "RequestExtensions(request_id: {}, requested_extensions: {})",
            self.request_id, self.requested_extensions
        )
    }
```

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl fmt::Display for RequestExtensionsSuccess<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Clone to get the inner vec for display
+        let extensions_list = format!("{}", self.supported_extensions);
+        write!(
+            f,
+            "RequestExtensions.Success(request_id: {}, supported_extensions: {})",
+            self.request_id, extensions_list
+        )
```
bit-aloo:
```suggestion
        write!(
            f,
            "RequestExtensions.Success(request_id: {}, supported_extensions: {})",
            self.request_id,  self.supported_extensions
        )
```

sv2/extensions-sv2/src/extensions_negotiation/messages.rs
```diff
@@ -0,0 +1,159 @@
...
+impl fmt::Display for RequestExtensionsError<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Clone to get the inner vecs for display
+        let unsupported_list = format!("{}", self.unsupported_extensions);
+        let required_list = format!("{}", self.required_extensions);
+        write!(
+            f,
+            "RequestExtensions.Error(request_id: {}, unsupported: {}, required: {})",
+            self.request_id, unsupported_list, required_list
+        )
```
bit-aloo:
```suggestion
        write!(
            f,
            "RequestExtensions.Error(request_id: {}, unsupported: {}, required: {})",
            self.request_id, self.unsupported_extensions, self.required_extensions
        )
```

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,554 @@
...
+                Some(Ok(tlv))
+            }
+            Err(e) => {
+                // Stop iteration on error by advancing to end
+                self.offset = self.data.len();
+                Some(Err(e))
+            }
+        }
+    }
+}
```
bit-aloo: This can be removed now.

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// Creates a new TLV list from a slice of TLVs (building mode).
+    pub fn from_slice(tlvs: &[Tlv]) -> Self {
+        Self::Owned(tlvs.to_vec())
+    }
+
+    /// Returns an iterator over the TLV fields in this list.
+    ///
+    /// For byte-backed lists, this lazily decodes TLVs.
+    /// For owned lists, this iterates over the stored TLVs.
+    pub fn iter(&self) -> Box<dyn Iterator<Item = Result<Tlv, TlvError>> + '_> {
```
bit-aloo:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item = Result<Tlv, TlvError>> + '_  {
```

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,554 @@
...
+        bytes.extend_from_slice(&self.length.to_le_bytes());
+
+        // Append Value
+        bytes.extend_from_slice(&self.value);
+
+        Ok(bytes)
+    }
+
+    /// Returns the total encoded size of this TLV in bytes.
+    pub fn encoded_size(&self) -> usize {
```
bit-aloo, line 125:
```suggestion
   #[inline]
    pub fn encoded_size(&self) -> usize {
```

sv2/parsers-sv2/src/tlv/codec.rs
bit-aloo: the iter stuff

sv2/parsers-sv2/src/tlv/codec.rs
```diff
@@ -0,0 +1,554 @@
...
+        Self::Bytes(data.as_ref())
+    }
+
+    /// Creates a new TLV list from a slice of TLVs (building mode).
+    pub fn from_slice(tlvs: &[Tlv]) -> Self {
+        Self::Owned(tlvs.to_vec())
+    }
+
+    /// Returns an iterator over the TLV fields in this list.
+    ///
```
bit-aloo: We can remove TlvList stuff from here.

sv2/parsers-sv2/src/tlv/list.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// Creates a new TLV list from a slice of TLVs (building mode).
+    pub fn from_slice(tlvs: &[Tlv]) -> Self {
+        Self::Owned(tlvs.to_vec())
+    }
+
+    /// Returns an iterator over the TLV fields in this list.
+    ///
+    /// For byte-backed lists, this lazily decodes TLVs.
+    /// For owned lists, this iterates over the stored TLVs.
+    pub fn iter(&self) -> Box<dyn Iterator<Item = Result<Tlv, TlvError>> + '_> {
```
bit-aloo:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item = Result<Tlv, TlvError>> + '_  {
```

sv2/parsers-sv2/src/tlv/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+impl Type {
+    /// Creates a new TLV Type.
+    pub fn new(extension_type: u16, field_type: u8) -> Self {
+        Self {
+            extension_type,
+            field_type,
+        }
+    }
+}
```
bit-aloo: Wait don't we have `Type` and `Tlv` already defined in codec? We can keep TlvField trait here, and rest to codec

sv2/handlers-sv2/src/template_distribution.rs
```diff
@@ -87,52 +137,100 @@ pub trait HandleTemplateDistributionMessagesFromServerSync {
...
-            self.handle_template_distribution_message_from_server(server_id, parsed)
-                .await
+            let negotiated_extensions = self.get_negotiated_extensions_with_server(server_id);
+            if negotiated_extensions.is_empty() {
+                let parsed: TemplateDistribution<'_> = (header.msg_type(), payload)
+                    .try_into()
+                    .map_err(Self::Error::parse_error)?;
+                self.handle_template_distribution_message_from_server(server_id, parsed, None)
+                    .await
+            } else {
```
bit-aloo: Lets return early, and remove this else.

```suggestion
                return self.handle_template_distribution_message_from_server(server_id, parsed, None)
                    .await;
            } 
```

sv2/handlers-sv2/src/template_distribution.rs
```diff
@@ -169,46 +271,92 @@ pub trait HandleTemplateDistributionMessagesFromServerAsync {
...
-            .try_into()
-            .map_err(Self::Error::parse_error)?;
-        self.handle_template_distribution_message_from_client(client_id, parsed)
+        let negotiated_extensions = self.get_negotiated_extensions_with_client(client_id);
+        if negotiated_extensions.is_empty() {
+            let parsed: TemplateDistribution<'_> = (header.msg_type(), payload)
+                .try_into()
+                .map_err(Self::Error::parse_error)?;
+            self.handle_template_distribution_message_from_client(client_id, parsed, None)
+        } else {
```
bit-aloo: Lets return early, and remove this else.
```suggestion
           return self.handle_template_distribution_message_from_client(client_id, parsed, None);
        } 
```

sv2/handlers-sv2/src/template_distribution.rs
```diff
@@ -237,53 +388,103 @@ pub trait HandleTemplateDistributionMessagesFromClientSync {
...
-            self.handle_template_distribution_message_from_client(client_id, parsed)
-                .await
+            let negotiated_extensions = self.get_negotiated_extensions_with_client(client_id);
+            if negotiated_extensions.is_empty() {
+                let parsed: TemplateDistribution<'_> = (header.msg_type(), payload)
+                    .try_into()
+                    .map_err(Self::Error::parse_error)?;
+                self.handle_template_distribution_message_from_client(client_id, parsed, None)
+                    .await
+            } else {
```
bit-aloo: Lets return early, and remove this else.
```suggestion
                return self.handle_template_distribution_message_from_client(client_id, parsed, None)
                    .await;
            }
```

sv2/handlers-sv2/src/mining.rs
```diff
@@ -26,28 +26,65 @@ pub enum SupportedChannelTypes {
...
-            .try_into()
-            .map_err(Self::Error::parse_error)?;
-        self.handle_mining_message_from_server(server_id, parsed)
+        let negotiated_extensions = self.get_negotiated_extensions_with_server(server_id);
+        if negotiated_extensions.is_empty() {
+            let parsed: Mining<'_> = (header.msg_type(), payload)
+                .try_into()
+                .map_err(Self::Error::parse_error)?;
+            self.handle_mining_message_from_server(server_id, parsed, None)
+        } else {
```
bit-aloo: Lets return early, and remove this else.
```suggestion
           return self.handle_mining_message_from_server(server_id, parsed, None)
        } 
```

sv2/handlers-sv2/src/mining.rs
```diff
@@ -240,32 +318,74 @@ pub trait HandleMiningMessagesFromServerSync {
...
-                .await
+            let negotiated_extensions = self.get_negotiated_extensions_with_server(server_id);
+            if negotiated_extensions.is_empty() {
+                let parsed: Mining<'_> = (header.msg_type(), payload)
+                    .try_into()
+                    .map_err(Self::Error::parse_error)?;
+                self.handle_mining_message_from_server(server_id, parsed, None)
+                    .await
+            } else {
+                let (parsed, tlv_fields) =
```
bit-aloo: Lets return early, and remove this else.
```suggestion
                return self.handle_mining_message_from_server(server_id, parsed, None)
                    .await;
            } 
                let (parsed, tlv_fields) =
```

sv2/handlers-sv2/src/mining.rs
```diff
@@ -479,22 +657,55 @@ pub trait HandleMiningMessagesFromClientSync {
...
-            .try_into()
-            .map_err(Self::Error::parse_error)?;
-        self.handle_mining_message_from_client(client_id, parsed)
+        let negotiated_extensions = self.get_negotiated_extensions_with_client(client_id);
+        if negotiated_extensions.is_empty() {
+            let parsed: Mining<'_> = (header.msg_type(), payload)
+                .try_into()
+                .map_err(Self::Error::parse_error)?;
+            self.handle_mining_message_from_client(client_id, parsed, None)
+        } else {
```
bit-aloo:
```suggestion
            return self.handle_mining_message_from_client(client_id, parsed, None)
        } 
 ```

File: sv2/handlers-sv2/src/mining.rs [hunk 58]
```diff
@@ -652,25 +895,63 @@ pub trait HandleMiningMessagesFromClientAsync {
...
-            self.handle_mining_message_from_client(client_id, parsed)
-                .await
+            let negotiated_extensions = self.get_negotiated_extensions_with_client(client_id);
+            if negotiated_extensions.is_empty() {
+                let parsed: Mining<'_> = (header.msg_type(), payload)
+                    .try_into()
+                    .map_err(Self::Error::parse_error)?;
+                self.handle_mining_message_from_client(client_id, parsed, None)
+                    .await
+            } else {
```
bit-aloo:
```suggestion
                return self.handle_mining_message_from_client(client_id, parsed, None)
                    .await
            } 
```

sv2/parsers-sv2/src/lib.rs
```diff
@@ -422,7 +506,40 @@ impl AnyMessage<'_> {
...
+    if raw_payload.len() > message_size {
+        let remaining = &raw_payload[message_size..];
+        let tlvs = TlvList::from_bytes(remaining).for_extensions(negotiated_extensions);
+        if !tlvs.is_empty() {
+            Some(tlvs)
+        } else {
+            None
         }
+    } else {
+        None
```
bit-aloo:
```suggestion
fn extract_tlvs_fields(
    payload: &[u8],
    message_size: usize,
    negotiated_extensions: &[u16],
) -> Option<Vec<Tlv>> {
    let remainder = payload.get(message_size..)?;

    let tlvs = TlvList::from_bytes(remainder).for_extensions(negotiated_extensions);
    (!tlvs.is_empty()).then_some(tlvs)
}

```

sv2/extensions-sv2/src/worker_specific_hashrate_tracking.rs
GitGab19: Removed the module and added the `worker_specific_hashrate_tracking.rs` file.

Did the same for the other extension.

sv2/parsers-sv2/src/tlv/codec.rs
bit-aloo: Ah, now I see. `TlvList` currently mixes two fundamentally different representations:
* `Bytes(&[u8])` — raw undecoded TLV bytes
* `Owned(Vec<Tlv>)` — fully decoded TLVs
This means the enum contains two different logical types, which makes the abstraction inconsistent.
```rust
#[derive(Debug, Clone)]
pub enum TlvList<'a> {
    Bytes(&'a [u8]),
    Owned(Vec<u8>),
}
```

With both variants representing raw TLV-encoded bytes, the type is now consistent. Our custom iterator only performs lazy decoding, operating over the byte slice rather than over `Tlv` values.

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -261,6 +261,11 @@ where
         &self.user_identity
     }
 
+    /// Sets the user identity string for this channel.
+    pub fn set_user_identity(&mut self, user_identity: String) {
```
plebhash: why are we introducing this?

`user_identity` field never changes across the lifespan of a channel, and AFAIU extension `0x0002` doesn't change that

I worry that introducing this can mislead API users towards violating spec compliance

plebhash: after some research on how this is used on https://github.com/stratum-mining/sv2-apps/pull/65, it seems that we're dynamically changing the channel `user_identity` field every time a new share arrives?

but then what?

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -116,6 +116,11 @@ impl<'a> ExtendedChannel<'a> {
         &self.user_identity
     }
 
+    /// Sets the `user_identity` for this channel.
+    pub fn set_user_identity(&mut self, user_identity: String) {
```
plebhash: why are we introducing this?

`user_identity` field never changes across the lifespan of a channel, and AFAIU extension `0x0002` doesn't change that

I worry that introducing this can mislead API users towards violating spec compliance

plebhash: after some research on how this is used on https://github.com/stratum-mining/sv2-apps/pull/65, differently from the server counterpart, I don't see this being used anywhere?

sv2/handlers-sv2/Cargo.toml
plebhash, line 3: don't we need to bump MAJOR here?

plebhash: I mean, CI is not complaining, but at this point I don't fully trust it anymore

sv2/handlers-sv2/src/common.rs
```diff
@@ -12,38 +12,78 @@ use crate::error::HandlerErrorType;
...
+///
+/// The `tlv_data` parameter in message handlers contains validated TLV fields if the message has
+/// extension data appended. TLV fields are only passed if they match negotiated extensions
+/// returned by `get_negotiated_extensions_with_server()`.
 pub trait HandleCommonMessagesFromServerSync {
     type Error: HandlerErrorType;
+
+    /// Returns the list of negotiated extension_types with a server.
+    ///
+    /// Used to validate TLV fields appended to messages. Return an empty Vec if no
```
plebhash: > Used to validate TLV fields appended to messages.

is usage of this method restricted to TLV extensions only?

sv2/extensions-sv2/examples/worker_specific_hashrate_tracking.rs
```diff
@@ -0,0 +1,229 @@
...
+        println!("   TLV Size: {} bytes", tlv_bytes.len());
+        println!(
+            "   TLV Format: [Type: 0x{:04x}|0x{:02x}] [Length: 0x{:04x}] [Value: \"{}\"]",
+            EXTENSION_TYPE_WORKER_HASHRATE_TRACKING,
+            TLV_FIELD_TYPE_USER_IDENTITY,
+            worker_name.len(),
+            worker_name
+        );
+
+        // Show the binary representation conceptually
```
plebhash: is this comment in the right place?

I have the impression it's describing what already happened a few lines above?

sv2/handlers-sv2/Cargo.toml
bit-aloo: Since handlers are still in the 0.x line, the major version stays at 0. Per semver guidelines, 0.x versions can introduce breaking changes as we continue stabilizing the API.

sv2/channels-sv2/src/server/extended.rs
GitGab19: > I worry that introducing this can mislead API users towards violating spec compliance

Is there any real constraint on the specs about this? Last time I checked there wasn't.

> we're dynamically changing the channel user_identity field every time a new share arrives?

Yeah this is what currently happens on the server side, and once the `user_identity` of the channel is updated, we have the correct `worker_name` stored in it (consider that we only use extension n.2 when there's no channel aggregation in tProxy, so when every client gets its own channel).

GitGab19: > but then what?

With the correct `user_identity` stored in the channel, every possible monitoring API, or persistence system based on `channel_sv2` can retrieve the correct `user_identity`.

At least this was my idea when doing that, but I'm also happy to remove the new API from `channels_sv2` and leave a comment on the applications which says something like "here you can extract the user_identity set by the client, to be later used for monitoring/persistence purposes, or in any external share accounting system".

sv2/channels-sv2/src/client/extended.rs
GitGab19: I just checked and I confirm we are not using it, but my idea was to use it in the tProxy, more specifically here: https://github.com/GitGab19/sv2-apps/blob/3ce3a292038f74fd92d6ce25c7637ad9185899ba/miner-apps/translator/src/lib/sv2/channel_manager/channel_manager.rs#L584-L592

sv2/handlers-sv2/src/common.rs
GitGab19: It can be called in different cases and contexts actually, so I'm gonna rephrase this for every handler.

sv2/extensions-sv2/examples/worker_specific_hashrate_tracking.rs
GitGab19: Removing it.

sv2/handlers-sv2/Cargo.toml
plebhash: that's true I keep forgetting about it

sv2/channels-sv2/src/server/extended.rs
plebhash: I was missing context, my original assumption here was that this extension was going to be used under aggregated mode

sv2/channels-sv2/src/client/extended.rs
GitGab19: https://github.com/stratum-mining/stratum/pull/1983#issuecomment-3571380212

sv2/channels-sv2/src/server/extended.rs
GitGab19: https://github.com/stratum-mining/stratum/pull/1983#issuecomment-3571380212

sv2/handlers-sv2/src/extensions.rs
```diff
@@ -0,0 +1,412 @@
...
+            },
+        }
+    }
+
+    fn handle_request_extensions(
+        &mut self,
+        client_id: Option<usize>,
+        msg: RequestExtensions,
+        tlv_fields: Option<&[Tlv]>,
+    ) -> Result<(), Self::Error>;
```
plebhash, line 314: this is empty and has no doc comments

is that intentional?

sv2/handlers-sv2/src/extensions.rs
```diff
@@ -0,0 +1,412 @@
...
+            }
+        }
+    }
+
+    async fn handle_request_extensions(
+        &mut self,
+        client_id: Option<usize>,
+        msg: RequestExtensions,
+        tlv_fields: Option<&[Tlv]>,
+    ) -> Result<(), Self::Error>;
```
plebhash, line 423: this is empty and has no doc comments, is that intentional?

sv2/handlers-sv2/src/extensions.rs
GitGab19: This is simply the method definition, which is then implemented in the apps.

sv2/handlers-sv2/src/extensions.rs
GitGab19: This is simply the method definition, which is then implemented in the apps.

sv2/parsers-sv2/src/lib.rs
```diff
@@ -23,6 +23,13 @@
 //!   submission).
 
 pub mod error;
+pub mod tlv;
+pub mod tlv_extensions;
```
bit-aloo: We are already exposing public facing API from these modules.
```suggestion
mod tlv;
mod tlv_extensions;
```

sv2/parsers-sv2/src/lib.rs
```diff
@@ -23,6 +23,13 @@
...
+
+// Re-export TLV types for public API
+pub use tlv::{Tlv, TlvError, TlvField, TlvList};
```
bit-aloo:
```suggestion
pub use tlv::{Tlv, TlvError, TlvField, TlvList, TLV_HEADER_SIZE};
```

sv2/handlers-sv2/src/extensions.rs
plebhash: ok that explains it being empty, but doesn't it deserve docs like the other ones?

sv2/handlers-sv2/src/extensions.rs
plebhash: ok that explains it being empty, but doesn't it deserve docs like the other ones?

sv2/handlers-sv2/src/extensions.rs
GitGab19: We don't have docs in any other method besides of the `handle_*_message_frame_from_*` and `handle_*_message_from_*` ones, for all the handler traits we have.

If you think we should add docs to every single method, I think this is probably something to do on all the traits we have in a follow-up PR.

sv2/handlers-sv2/src/extensions.rs
GitGab19: We don't have docs in any other method besides of the `handle_*_message_frame_from_*` and `handle_*_message_from_*` ones, for all the handler traits we have.

If you think we should add docs to every single method, I think this is probably something to do on all the traits we have in a follow-up PR.

sv2/handlers-sv2/src/extensions.rs
plebhash: ok

PR #1986, opened 2025-11-02
correct typos and grammar across codebase

BRIGHTON-ASUMANI:
Fix typosand correct grammar on comments and read me


GitGab19:
PRs touching the old `roles/` should be opened against the new `sv2-apps` repo: https://github.com/stratum-mining/sv2-apps, while changes to the old `protocols/` should be opened here.

PR #1987, opened 2025-11-02
Fix: Prevent JD Server panic on invalid token length

BRIGHTON-ASUMANI:
### Problem

The `verify_job()` function used `.unwrap()` to convert `B0255` (0-255 bytes) to a 4-byte array. If a client sends a token with length ≠ 4 bytes, the server panics and crashes.

**Impact**: High - DoS vulnerability. Malicious clients can crash the JD Server by sending malformed `DeclareMiningJob` messages.

### Solution

Added length validation before conversion:
- Returns `false` (instead of panicking) for invalid token lengths
- Maintains existing error handling (triggers `DeclareMiningJobError`)
- Uses safe array indexing after validation

### Changes

- Added length check in `verify_job()` function (`roles/jd-server/src/lib/job_declarator/message_handler.rs`)
- Replaced unsafe `unwrap()` with length validation + safe array indexing

### DIFF
--- a/roles/jd-server/src/lib/job_declarator/message_handler.rs
+++ b/roles/jd-server/src/lib/job_declarator/message_handler.rs
@@ -27,13 +27,20 @@ use super::JobDeclaratorDownstream;
 impl JobDeclaratorDownstream {
     fn verify_job(&mut self, message: &DeclareMiningJob) -> bool {
         // Convert token from B0255 to u32
-        let four_byte_array: [u8; 4] = message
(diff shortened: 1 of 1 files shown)


lucasbalieiro:
Hello, @BRIGHTON-ASUMANI.

I know this PR is still in draft but we had a recent split on our repos. 
Now this changes for the apps should be done in the https://github.com/stratum-mining/sv2-apps

This repo here will only contain the low-level crates for sv2

PR #1997, opened 2025-11-13
associate `Target` to `Job` and use it for share validation (instead of channel `Target`)

### LINKED ISSUE #1981: `channels_sv2` should update channel target only on job updates
as noted on https://github.com/stratum-mining/sv2-apps/issues/20#issuecomment-3462378568, `channels_sv2` is currently overlooking the fact that the spec says:

> Maximum target is valid until the next SetTarget message is sent and is applicable for all jobs received on the channel in the future or already received with an empty min_ntime. The message is not applicable for already received jobs with min_ntime=nTime, as their maximum target remains stable.

plebhash:
close #1981

note: public APIs of `channels_sv2` are being changed, but [we already bumped MAJOR within this release cycle](https://github.com/stratum-mining/stratum/issues/1933#issuecomment-3363346970)

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -526,52 +526,55 @@ impl<'a> ExtendedChannel<'a> {
         };
 
         // convert the header hash to a target type for easy comparison
-        let hash = header.block_hash();
-        let raw_hash: [u8; 32] = *hash.to_raw_hash().as_ref();
-        let block_hash_target = Target::from_le_bytes(raw_hash);
-        let hash_as_diff = block_hash_target.difficulty_float();
+        let share_hash = header.block_hash();
+        let share_raw_hash: [u8; 32] = *share_hash.to_raw_hash().as_ref();
+        let share_hash_target = Target::from_le_bytes(share_raw_hash);
+        let share_hash_as_diff = share_hash_target.difficulty_float();
 
         let network_target = Target::from_compact(nbits);
 
         let job_target = job.2;
 
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
         let job_target_bytes = job.2.to_be_bytes();
 
         debug!(
             "share validation \nshare:\t\t{}\nchannel target:\t{}\nnetwork target:\t{}",
-            bytes_to_hex(&block_hash_target_bytes),
+            bytes_to_hex(&share_hash_target_bytes),
             bytes_to_hex(&job_target_bytes),
             format!("{:x}", network_target)
         );
 
         // check if a block was found
-        if network_target.is_met_by(hash) {
+        if network_target.is_met_by(share_hash) {
             self.share_accounting.update_share_accounting(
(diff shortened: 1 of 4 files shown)


plebhash:
cc @coleFD @colerob1776 this might be relevant to you

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -38,7 +38,8 @@ use tracing::debug;
 /// Each job includes:
 /// - A [`NewExtendedMiningJob`] message
 /// - The `extranonce_prefix` in use when the job was created
-pub type ExtendedJob<'a> = (NewExtendedMiningJob<'a>, Vec<u8>);
+/// - The target for the job
+pub type ExtendedJob<'a> = (NewExtendedMiningJob<'a>, Vec<u8>, Target);
```
colerob1776, line 42: this forces us to duplicate jobs for each connection right? Why not have a job_id -> target map for the extended channel in the case of a central job store?

colerob1776: For example, i believe right now. I could broadcast the same job across all clients without being forced to clone the job state to each individual client state, and when solutions arrive, I could lookup by job_id from a central job_store and check the hash against the channel's local extranonce_prefix and target (where target could be fetched by job_id, or if job_id is monotonic it could be optimized a bit)

plebhash: > this forces us to duplicate jobs for each connection right?

I don't understand the question.

This is on the client side. There's only one connection.

colerob1776: my bad, i had tunnel vision on the pool side

sv2/channels-sv2/src/server/jobs/extended.rs
```diff
@@ -32,6 +32,7 @@ pub struct ExtendedJob<'a> {
     coinbase_tx_prefix_with_bip141: Vec<u8>,
     coinbase_tx_suffix_with_bip141: Vec<u8>,
     job_message: NewExtendedMiningJob<'a>,
+    target: Option<Target>, // Group Channels don't have a target... however Extended Channels MUST have Some(target) here
```
colerob1776, line 35: I think this is where i should have put my comment from above, but considering it's optional, it doesnt seem like it's limiting

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -652,30 +655,32 @@ where
...
-        let block_hash_target = Target::from_le_bytes(raw_hash);
-        let hash_as_diff = block_hash_target.difficulty_float();
+        let share_hash = header.block_hash();
+        let share_raw_hash: [u8; 32] = *share_hash.to_raw_hash().as_ref();
+        let share_hash_target = Target::from_le_bytes(share_raw_hash);
+        let share_hash_as_diff = share_hash_target.difficulty_float();
 
         let network_target = Target::from_compact(nbits);
 
+        let job_target = job.get_target().expect("extended job must have a target");
```
colerob1776, line 665: This would fall inline with my initial comment. There is no way to pass in a target so target's are forced 1:1 with a standard/extended channel even though jobs can be 1:many (optimally)

PR #2000, opened 2025-11-14
associate `Target` to `Job` and use it for share validation (instead of channel `Target`)

### LINKED ISSUE #1981: `channels_sv2` should update channel target only on job updates

plebhash:
close #1981

replaces #1997

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -525,51 +525,54 @@ impl<'a> ExtendedChannel<'a> {
         };
 
         // convert the header hash to a target type for easy comparison
-        let hash = header.block_hash();
-        let raw_hash: [u8; 32] = *hash.to_raw_hash().as_ref();
-        let block_hash_target = Target::from_le_bytes(raw_hash);
-        let hash_as_diff = block_hash_target.difficulty_float();
+        let share_hash = header.block_hash();
+        let raw_share_hash: [u8; 32] = *share_hash.to_raw_hash().as_ref();
+        let share_hash_target = Target::from_le_bytes(raw_share_hash);
+        let share_hash_as_diff = share_hash_target.difficulty_float();
 
         let network_target = Target::from_compact(nbits);
         let job_target = job.2;
 
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
         let job_target_bytes = job_target.to_be_bytes();
 
         debug!(
             "share validation \nshare:\t\t{}\njob target:\t{}\nnetwork target:\t{}",
-            bytes_to_hex(&block_hash_target_bytes),
+            bytes_to_hex(&share_hash_target_bytes),
             bytes_to_hex(&job_target_bytes),
             format!("{:x}", network_target)
         );
 
         // check if a block was found
-        if network_target.is_met_by(hash) {
+        if network_target.is_met_by(share_hash) {
             self.share_accounting.update_share_accounting(
                 job_target.difficulty_float(),
                 share.sequence_number,
-                hash.to_raw_hash(),
+                share_hash.to_raw_hash(),
             );
-            return Ok(ShareValidationResult::BlockFound(hash.to_raw_hash()));
+            return Ok(ShareValidationResult::BlockFound(share_hash.to_raw_hash()));
         }
 
         // check if the share hash meets the job target
-        if block_hash_target < job_target {
-            if self.share_accounting.is_share_seen(hash.to_raw_hash()) {
+        if share_hash_target < job_target {
+            if self
+                .share_accounting
+                .is_share_seen(share_hash.to_raw_hash())
+            {
                 return Err(ShareValidationError::DuplicateShare);
             }
 
             self.share_accounting.update_share_accounting(
                 job_target.difficulty_float(),
                 share.sequence_number,
-                hash.to_raw_hash(),
+                share_hash.to_raw_hash(),
             );
 
             // update the best diff
-            self.share_accounting.update_best_diff(hash_as_diff);
+            self.share_accounting.update_best_diff(share_hash_as_diff);
 
-            return Ok(ShareValidationResult::Valid(hash.to_raw_hash()));
+            return Ok(ShareValidationResult::Valid(share_hash.to_raw_hash()));
         }
 
         Err(ShareValidationError::DoesNotMeetTarget)
--- a/sv2/channels-sv2/src/client/standard.rs
+++ b/sv2/channels-sv2/src/client/standard.rs
@@ -315,51 +315,54 @@ impl<'a> StandardChannel<'a> {
         };
 
         // convert the header hash to a target type for easy comparison
-        let hash = header.block_hash();
-        let raw_hash: [u8; 32] = *hash.to_raw_hash().as_ref();
-        let block_hash_target = Target::from_le_bytes(raw_hash);
-        let hash_as_diff = block_hash_target.difficulty_float();
+        let share_hash = header.block_hash();
+        let raw_share_hash: [u8; 32] = *share_hash.to_raw_hash().as_ref();
+        let share_hash_target = Target::from_le_bytes(raw_share_hash);
+        let share_hash_as_diff = share_hash_target.difficulty_float();
         let network_target = Target::from_compact(nbits);
 
         let job_target = job.1;
 
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
         let job_target_bytes = job_target.to_be_bytes();
 
         debug!(
             "share validation \nshare:\t\t{}\njob target:\t{}\nnetwork target:\t{}",
-            bytes_to_hex(&block_hash_target_bytes),
+            bytes_to_hex(&share_hash_target_bytes),
             bytes_to_hex(&job_target_bytes),
             format!("{:x}", network_target)
         );
 
         // check if a block was found
-        if network_target.is_met_by(hash) {
+        if network_target.is_met_by(share_hash) {
             self.share_accounting.update_share_accounting(
                 job_target.difficulty_float(),
                 share.sequence_number,
-                hash.to_raw_hash(),
+                share_hash.to_raw_hash(),
             );
-            return Ok(ShareValidationResult::BlockFound(hash.to_raw_hash()));
+            return Ok(ShareValidationResult::BlockFound(share_hash.to_raw_hash()));
         }
 
         // check if the share hash meets the job target
-        if block_hash_target < job_target {
-            if self.share_accounting.is_share_seen(hash.to_raw_hash()) {
+        if share_hash_target < job_target {
+            if self
+                .share_accounting
+                .is_share_seen(share_hash.to_raw_hash())
+            {
                 return Err(ShareValidationError::DuplicateShare);
             }
 
             self.share_accounting.update_share_accounting(
                 job_target.difficulty_float(),
                 share.sequence_number,
-                hash.to_raw_hash(),
+                share_hash.to_raw_hash(),
             );
 
             // update the best diff
-            self.share_accounting.update_best_diff(hash_as_diff);
+            self.share_accounting.update_best_diff(share_hash_as_diff);
 
-            return Ok(ShareValidationResult::Valid(hash.to_raw_hash()));
+            return Ok(ShareValidationResult::Valid(share_hash.to_raw_hash()));
         }
 
         Err(ShareValidationError::DoesNotMeetTarget)
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -686,30 +686,30 @@ where
         };
 
         // convert the header hash to a target type for easy comparison
-        let hash = header.block_hash();
-        let raw_hash: [u8; 32] = *hash.to_raw_hash().as_ref();
-        let block_hash_target = Target::from_le_bytes(raw_hash);
-        let hash_as_diff = block_hash_target.difficulty_float();
+        let share_hash = header.block_hash();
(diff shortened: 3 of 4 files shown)


sv2/channels-sv2/src/server/extended.rs
```diff
@@ -92,7 +88,8 @@ where
     extranonce_prefix: Vec<u8>,
     rollable_extranonce_size: u16,
     requested_max_target: Target,
-    target: Target, // todo: try to use Target from rust-bitcoin
+    target: Target,
+    job_id_to_target: HashMap<u32, Target>,
```
colerob1776, line 92: what do you think about making this a generic with trait bounds `get(), insert()`. If job_ids are monotonic, you wouldnt necessarily have to insert ever job. You would just have to insert the job_id at every target change and any job_id in between would be defined by either the upper or lower bound depending on implementation

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -240,12 +241,20 @@ impl<'a> ExtendedChannel<'a> {
                 if let Some(active_job) = self.active_job.clone() {
                     self.past_jobs.insert(active_job.0.job_id, active_job);
                 }
-                self.active_job = Some((new_extended_mining_job, self.extranonce_prefix.clone()));
+                self.active_job = Some((
+                    new_extended_mining_job,
+                    self.extranonce_prefix.clone(),
+                    self.target,
```
colerob1776, line 247: If sending a SetTarget after a newJob message, would the target not get updated until a new job arrived? Meaning SetTargets should always be sent directly before a new job?

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -32,13 +32,14 @@ use mining_sv2::{
 };
 use tracing::debug;
 
-/// A type alias representing an extended mining job tied to a specific `extranonce_prefix`.
+/// A type alias representing an extended mining job tied to a specific `extranonce_prefix` and target.
 ///
 /// Extended jobs allow Merkle root rolling, providing broader control over the search space.
 /// Each job includes:
 /// - A [`NewExtendedMiningJob`] message
 /// - The `extranonce_prefix` in use when the job was created
```
colerob1776, line 40: need to add Target to the docstring. Makes me wonder if this should be a struct instead..

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -501,6 +512,14 @@ where
                 ) {
                     return Err(ExtendedChannelError::TemplateIdNotFound);
                 }
+
+                // associate the new active job with the current target
+                let job_id = self
+                    .job_store
+                    .get_active_job()
+                    .expect("active job must exist")
```
colerob1776, line 529: can you add in the message that this is safe because the job was activated above

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -652,30 +674,35 @@ where
...
+        let raw_share_hash: [u8; 32] = *share_hash.to_raw_hash().as_ref();
+        let share_hash_target = Target::from_le_bytes(raw_share_hash);
+        let share_hash_as_diff = share_hash_target.difficulty_float();
 
         let network_target = Target::from_compact(nbits);
 
+        let job_target = self
+            .job_id_to_target
+            .get(&job_id)
+            .expect("job target must exist");
```
colerob1776: can you move this under where the job is defined. It add's context so the .expect() is a little less sketchy. in general i would log errors instead of panicking though

sv2/channels-sv2/src/client/extended.rs
plebhash: > If sending a SetTarget after a newJob message, would the target not get updated until a new job arrived?

client's channel target gets updated immediately upon arrival of `SetTarget`, and that's enforced for:
- new jobs that arrive after `SetTarget`
- future jobs that get activated after `SetTarget` arrival

jobs received or activated prior to `SetTarget` arrival remain associated with whatever was the channel target at the time

> Meaning SetTargets should always be sent directly before a new job?

`SetTarget` messages should be sent whenever the server deems [channel target vs share submission rate] is not properly calibrated

but the message shouldn't be sent just because there's a new job on the pipeline... it's sent because it needs to be sent, and share submission will adjust whenever there's a new job

sv2/channels-sv2/src/client/extended.rs
plebhash: it's on line 41, below the red line

I agree a struct could improve readability... but I tried switching and it's breaking a lot of stuff, so I'd prefer to leave that as an improvement for the future

sv2/channels-sv2/src/client/extended.rs
colerob1776: This means the client should update it's active_job's target when receiving a SetTarget then right, so set_target needs to update in the job as well i think

sv2/channels-sv2/src/server/extended.rs
plebhash: about to push this:

```diff
-                    .expect("active job must exist")
+                    .expect("active job must exist, it was just activated")
```

sv2/channels-sv2/src/client/extended.rs
plebhash: not really... target for `active_job` remains the same

when a new job arrives and we update `active_job`, we'll associate it with `self.target` that was set by the latest `SetTarget`

here's what the spec says:

> Maximum target is valid until the next SetTarget message is sent and is applicable for all jobs received on the channel in the future or already received with an empty min_ntime. The message is not applicable for already received jobs with min_ntime=nTime, as their maximum target remains stable.

sv2/channels-sv2/src/server/extended.rs
plebhash: how can we guarantee that `job_id`s will be monotonic on every implementation?

I guess we know that about the standard `JobFactory`, and I assume that whatever you're planning to use as a replacement will also follow this assumption

but maybe some alternative implementation could choose to follow different assumptions?

moreover, we're flushing the `HashMap` every time there's a chain tip update, so we're not a risk of unbounded memory growth

colerob1776: Yep, that was my idea. Not saying job_ids have to be monotonic, but if an implementation was monotonic, a custom implementation could require less writes to the job_id_to_target store. Our job_id implementation is monotonic, so I think we'd prefer to use the optimized approach

plebhash: I'm a bit swamped so I'm not going to have time to implement this optimization now

if you feel this is a hard pre-requisite or a blocker on your side, please create an issue

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -595,6 +617,11 @@ where
                 .expect("stale job must exist")
         };
 
+        let job_target = self
+            .job_id_to_target
+            .get(&job_id)
+            .expect("job target must exist");
```
GitGab19, line 633: Aren't we gonna panic here in case of a stale share?

Given you're clearing the hashmap on every `SetNewPrevHash`, the `job_id` won't be there anymore.

Or is it safe to assume it's impossible to make this happen because of some detail I don't remember now?

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -555,6 +585,11 @@ where
                 .expect("stale job must exist")
         };
 
+        let job_target = self
+            .job_id_to_target
+            .get(&job_id)
+            .expect("job target must exist");
```
GitGab19, line 593: Same as https://github.com/stratum-mining/stratum/pull/2000/changes#r2613781464

GitGab19:
In addition to the above comments, don't we need to bump MAJOR on `channels_sv2` given the changes you made?

sv2/channels-sv2/src/server/extended.rs
plebhash: a few lines above we're doing this:

```rust
        // check if job_id is stale job
        let is_stale_job = self.job_store.get_stale_jobs().contains_key(&job_id);
        if is_stale_job {
            return Err(ShareValidationError::Stale);
        }
```

so if the share belongs to a stale job, we return the function execution much earlier than reaching this

plebhash:
> In addition to the above comments, don't we need to bump MAJOR on `channels_sv2` given the changes you made?

that's a good point, and yet another episode where `cargo semver-checks` lets me very disappointed.

I asked Claude Sonnet 4.5, here's what it said:

> Why cargo-semver-checks Doesn't Catch It
>
> According to the search results, this is a known limitation of cargo-semver-checks:
>
> "The tool does not yet check for changes in type information, including type aliases, which can lead to undetected breaking changes."
>
> The problem is that `cargo-semver-checks` analyzes function signatures at the API level. When it sees:
> `pub fn get_active_job(&self) -> Option<&ExtendedJob<'a>>`
> It sees the type alias name `(ExtendedJob)` staying the same and considers it non-breaking, but it doesn't expand the alias to see that the underlying tuple changed from 2 elements to 3.
> This is described as the "final boss" of SemVer checking - detecting type changes requires understanding complex Rust semantics like type coercions, lifetimes, and variance.
>
> Recommendation
> Bump `channels_sv2` to version 3.0.0 before merging PR #2000. Any downstream code that unpacks these tuples (e.g., let `(job, extranonce) = channel.get_active_job().unwrap())` will break with this change.

just cherry-picked b584347f366f68093eddb5045e9fa04ea103b859 from #2011

sv2/channels-sv2/src/server/extended.rs
GitGab19: Oh I missed it, thanks.

GitGab19:
I just got my BMM running against the changes introduced here, and everything is working smoothly.

Not a single share rejected so far.

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -582,29 +617,29 @@ where
...
+        let share_raw_hash: [u8; 32] = *share_hash.to_raw_hash().as_ref();
+        let share_hash_target = Target::from_le_bytes(share_raw_hash);
+        let share_hash_as_diff = share_hash_target.difficulty_float();
         let network_target = Target::from_compact(nbits);
 
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
-        let target_bytes = self.target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
+        let job_target_bytes = self.target.to_be_bytes();
```
GitGab19: I'm pretty sure the issue which I reported on https://github.com/stratum-mining/stratum/pull/2000#issuecomment-3665054150  is due to the fact that here we are still getting the channel's target instead of the job one.


```suggestion
        let job_target_bytes = job_target.to_be_bytes();
```

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -582,29 +617,29 @@ where
...
 
         debug!(
             "share validation \nshare:\t\t{}\nchannel target:\t{}\nnetwork target:\t{}",
```
GitGab19:
```suggestion
            "share validation \nshare:\t\t{}\njob target:\t{}\nnetwork target:\t{}",
```

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -659,30 +686,30 @@ where
...
         let network_target = Target::from_compact(nbits);
 
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
-        let target_bytes = self.target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
+        let job_target_bytes = job_target.to_be_bytes();
 
         debug!(
             "share validation \nshare:\t\t{}\nchannel target:\t{}\nnetwork target:\t{}",
```
GitGab19:
```suggestion
            "share validation \nshare:\t\t{}\njob target:\t{}\nnetwork target:\t{}",
```

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -512,50 +525,54 @@ impl<'a> ExtendedChannel<'a> {
...
+        let job_target = job.2;
 
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
-        let target_bytes = self.target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
+        let job_target_bytes = job_target.to_be_bytes();
 
         debug!(
             "share validation \nshare:\t\t{}\nchannel target:\t{}\nnetwork target:\t{}",
```
```suggestion
            "share validation \nshare:\t\t{}\njob target:\t{}\nnetwork target:\t{}",
```

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -310,49 +315,54 @@ impl<'a> StandardChannel<'a> {
...
+        let job_target = job.1;
+
         // print hash_as_target and self.target as human readable hex
-        let block_hash_target_bytes = block_hash_target.to_be_bytes();
-        let target_bytes = self.target.to_be_bytes();
+        let share_hash_target_bytes = share_hash_target.to_be_bytes();
+        let job_target_bytes = job_target.to_be_bytes();
 
         debug!(
             "share validation \nshare:\t\t{}\nchannel target:\t{}\nnetwork target:\t{}",
```
GitGab19:
```suggestion
            "share validation \nshare:\t\t{}\njob target:\t{}\nnetwork target:\t{}",
```

sv2/channels-sv2/src/server/standard.rs
GitGab19: I can confirm this is the issue, I fixed it locally and now everything works perfectly.

plebhash: thanks for spotting this

pushed the fix

PR #2002, opened 2025-11-14
properly mark past jobs as stale on server channels

### LINKED ISSUE #2001: `channels_sv2` server channels are not properly marking past jobs as stale
server channels are supposed to mark past jobs as stale during `on_set_new_prev_hash`

`StandardChannel` currently does not do it:

https://github.com/stratum-mining/stratum/blob/bdd60eb3d587d3c8ee6b38b0f90f78bcef76e9f0/sv2/channels-sv2/src/server/standard.rs#L481-L503

`ExtendedChannel` only does it if there's no future jobs:

https://github.com/stratum-mining/stratum/blob/bdd60eb3d587d3c8ee6b38b0f90f78bcef76e9f0/sv2/channels-sv2/src/server/extended.rs#L487-L514

plebhash:
close #2001

### DIFF
--- a/sv2/channels-sv2/src/server/standard.rs
+++ b/sv2/channels-sv2/src/server/standard.rs
@@ -499,6 +499,9 @@ where
         // mark past jobs as stale
         self.job_store.mark_past_jobs_as_stale();
 
+        // clear seen shares, as shares for past chain tip will be rejected as stale
+        self.share_accounting.flush_seen_shares();
+
         // update the chain tip
         self.chain_tip = Some(set_new_prev_hash.into());
 


sv2/channels-sv2/src/server/standard.rs
```diff
@@ -496,6 +496,12 @@ where
             }
         }
```
GitGab19, line 497: Can you change this in the same way you did for the extended channel?

plebhash: I expected some confusion around here, so let me try to clarify.

First, we need to remember that extended channels can potentially cater exclusively for custom work, in which case their `JobStore` will never have future jobs.

So while checking `match self.job_store.get_future_jobs().is_empty()`, standard channels must return `TemplateIdNotFound` in the `true` arm (because that's a strong indication that something went wrong somewhere), while extended channels don't need to return an error.

When it comes to the faulty application of `self.job_store.mark_past_jobs_as_stale()`, extended and standard channels were both buggy, but in different ways:
- extended channel was doing it inside the `true` arm of the `match`, while this needs to happen regardless of whether the job store has future jobs or not
- standard channel simply didn't have it

so on the extended channel, this line was moved out of the `match`, leaving the `true` arm empty, which then motivated the change towards `if` clauses (which were intentionally more verbose but clippy clubbed together into a `if !X && !Y` patten)

that's why we ended up with different patterns for each channel kind

if we're aiming for some kind of "symmetry" on the standard channel side, I guess in theory we could also replace `match` with `if`/`else`:

```rust
        if self.job_store.get_future_jobs().is_empty() {
            return Err(StandardChannelError::TemplateIdNotFound);
        } else {
            if !self.job_store.activate_future_job(
                set_new_prev_hash.template_id,
                set_new_prev_hash.header_timestamp,
            ) {
                return Err(StandardChannelError::TemplateIdNotFound);
            }
        }
```

or maybe a cleaner version:
```rust
        if self.job_store.get_future_jobs().is_empty() {
            return Err(StandardChannelError::TemplateIdNotFound);
        }
        
        if !self.job_store.activate_future_job(
            set_new_prev_hash.template_id,
            set_new_prev_hash.header_timestamp,
        ) {
            return Err(StandardChannelError::TemplateIdNotFound);
        }
```

but overall, I don't think it's possible to have an identical pattern due to differences in nature of each channel kind

GitGab19: Crystal clear, thanks for the explanation.

It looks good as it is 👍

PR #2011, opened 2025-11-28
`JobStore` trait methods return owned types instead of refs

### LINKED ISSUE #1927: Allow JobStore trait to support Send/Sync implementations
As mentioned here https://github.com/stratum-mining/stratum/pull/1884#issuecomment-3308549867, The current implementation of JobStore has methods that return references which prevent `JobStore` from from storing Jobs in Send/Sync types. For example, if jobs are stored behind an `Arc<Mutex<Job>>`, there is no way to return `Option<&Job>` in the `get_active_job()` trait method.

plebhash:
close #1927

companion https://github.com/stratum-mining/sv2-apps/pull/122

since we're no longer returning references, we're no longer providing getter methods that return the entire `HashMap`s, because cloning those could become computationally expensive, especially in hot execution paths such as `validate_share`

so for example, the getter `get_future_jobs(&self) -> &HashMap(u32, T)` was replaced with:
- `get_future_job(&self, job_id: u32) -> T`: which performs the lookup and returns an owned copy of the actual type `T`
- `has_future_jobs(&self) -> bool`: which allows for quickly verifying whether the `HashMap` is empty without having to clone it

---

additionally to the scope of #1927, we're also removing some inefficients from `DefaultJobStore::mark_past_jobs_as_stale`

### DIFF
--- a/sv2/channels-sv2/src/server/jobs/job_store.rs
+++ b/sv2/channels-sv2/src/server/jobs/job_store.rs
@@ -150,12 +150,8 @@ impl<T: Job + Clone + Debug> JobStore<T> for DefaultJobStore<T> {
     }
 
     fn mark_past_jobs_as_stale(&mut self) {
-        // Mark all past jobs as stale, so that shares can be rejected with the appropriate error
-        // code
-        self.stale_jobs = self.past_jobs.clone();
-
-        // Clear past jobs, as we're no longer going to validate shares for them
-        self.past_jobs.clear();
+        // Transfer past jobs to stale jobs collection and reset past jobs to empty
+        self.stale_jobs = std::mem::take(&mut self.past_jobs);
     }
 
     fn get_future_job_id_from_template_id(&self, template_id: u64) -> Option<u32> {


sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -41,19 +44,19 @@ pub trait JobStore<T: Job>: Send + Sync + Debug {
...
-    fn get_future_jobs(&self) -> &HashMap<u32, T>;
+    fn get_future_jobs(&self) -> HashMap<u32, T>;
 
     /// Returns all past jobs (previously active jobs), indexed by job ID.
-    fn get_past_jobs(&self) -> &HashMap<u32, T>;
+    fn get_past_jobs(&self) -> HashMap<u32, T>;
 
     /// Returns all stale jobs (jobs from previous chain tip), indexed by job ID.
-    fn get_stale_jobs(&self) -> &HashMap<u32, T>;
+    fn get_stale_jobs(&self) -> HashMap<u32, T>;
```
bit-aloo: We should rethink how we’re exposing these HashMaps. Returning fully cloned maps is pretty expensive, especially since this code runs in the hot path. Most of the time those cloned maps are short-lived anyway.

We could do these instead:

1. Return an iterator over the jobs instead of cloning the whole map. That only clones what we actually touch and avoids the extra allocations.

2. If we only ever fetch jobs by id, then we can just expose a method like:

```rust
get_future_job_by_id(&self, id: u64) -> Option<T>
```

I checked all the call sites and we ever use these methods to fetch a job by id. So option 2 makes the most sense. And if we ever need the full list of jobs, we can still expose an iterator for that.

Shourya742 (commented):
We need to major bump channel sv2 for this.

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: I think these are valid points. 

I'll take some time to consider the suggestions and try to come up with a more optimal approach.

plebhash:
> We need to major bump channel sv2 for this.

that's a good reminder!

and also makes me wonder why CI is not breaking... `cargo semver` is so disappointing :(

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -146,23 +158,35 @@ impl<T: Job + Clone + Debug> JobStore<T> for DefaultJobStore<T> {
         self.past_jobs.clear();
     }
 
-    fn get_future_template_to_job_id(&self) -> &HashMap<u64, u32> {
-        &self.future_template_to_job_id
+    fn get_future_job_id_from_template_id(&self, template_id: u64) -> Option<u32> {
```
colerob1776, line 157: i couldnt add a comment on the correct line for some reason, but i noticed `mark_past_jobs_as_stale()` clones past jobs into stale jobs, which is fine for reference counted types, but it may be better to just move the past jobs into the stale jobs

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -405,17 +406,23 @@ where
...
+    pub fn get_active_job(&self) -> Option<ExtendedJob<'a>> {
+        // cloning happens inside the job store
         self.job_store.get_active_job()
     }
-    /// Returns all future jobs for this channel.
-    pub fn get_future_jobs(&self) -> &HashMap<u32, ExtendedJob<'a>> {
-        self.job_store.get_future_jobs()
+
+    /// Returns an owned copy of a future job from its job ID, if any.
+    pub fn get_future_job(&self, job_id: u32) -> Option<ExtendedJob<'a>> {
```
colerob1776, line 417: These are returning `ExtendedJob` which locks us into this type. Wouldnt we need this to only be bound by the Job trait

colerob1776: Seems like something for another story though. As far as JobStore goes, everything looks valid

bit-aloo: We do have more associated methods with ExtendedJob, if we make the return type as a trait object that won't cover all the required expected methods.
@plebhash might have more insights on this.

plebhash: > These are returning ExtendedJob which locks us into this type.

well, `ExtendedChannel` is already locked into `ExtendedJob` because of this:
https://github.com/stratum-mining/stratum/blob/7c0c883f6e5529dd298085ef1a73e82c7308e1b9/sv2/channels-sv2/src/server/extended.rs#L105-L108

so if you believe this is a blocker, could you please create a new issue for a feature request? because it would require more aggressive refactoring and we're not going to be able to tackle this now

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: thanks for pointing it out, that should be fixed on 1d452c6158e21d10027cabd7e2747e23d3eaa5e5

plebhash:
so I guess I need to create a companion PR on `sv2-apps` and (temporarily) use its branch for the local patching of `stratum-apps`?

that would be aiming to run CI ITF to validate these changes on both layers

cc @GitGab19

PR #2012, opened 2025-12-01
update `run-integration-tests.sh` scripts to patch the sources from crates.io

### LINKED ISSUE #2008: Update `run-integration-tests.sh` script to make it patch crates.io crates
In https://github.com/stratum-mining/sv2-apps/pull/96 we are finally making `sv2-apps` get `stratum-core` from crates.io, with two exceptions: JDS and mining-device.

Those two crates are now excluded from the main workspaces, and they are still getting the protocol crates from this github repository (v1.5.0 tag):

- https://github.com/GitGab19/sv2-apps/blob/49550d782bf7762bd8a44c0c558399687f238655/pool-apps/jd-server/Cargo.toml#L20
- https://github.com/GitGab19/sv2-apps/blob/49550d782bf7762bd8a44c0c558399687f238655/miner-apps/mining-device/Cargo.toml#L23

I guess we would simply need to add another line for crates.io around here:
https://github.com/stratum-mining/stratum/blob/41f56ee133e4dcc6ea28e77e8242845ae11a7804/scripts/run-integration-tests.sh#L54-L58

cc @lucasbalieiro

lucasbalieiro:
closes #2008 ;

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -38,23 +38,22 @@ fi
 cd "$INTEGRATION_TESTS_DIR"
 
 # # Add patch section to override all git dependencies with local paths
-echo "🔧 Adding patch section to override git dependencies..."
+echo "🔧 Adding patch section to override dependencies..."
 
 # Remove any existing patch section first
 if [[ "$OSTYPE" == "darwin"* ]]; then
-    sed -i '' '/^# Override git dependencies with local paths/,/^$/d' Cargo.toml
+    sed -i '' '/^# Override dependencies with local paths/,/^$/d' Cargo.toml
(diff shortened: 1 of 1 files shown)


scripts/run-integration-tests.sh
```diff
@@ -52,9 +52,8 @@ fi
 cat >> Cargo.toml << 'EOF'
 
 # Override git dependencies with local paths to avoid version conflicts
```
plebhash: shouldn't we also update this comment? in a way, it's no longer describing what's actually going on here

and the same goes for the `sed` command we do against this string a few lines above

GitGab19 (approved):

PR #2017, opened 2025-12-03
noise-sv2: add 10s clock skew tolerance to signature verification

### LINKED ISSUE #2015: should `verify_with_now`have a time leeway?
following the context discussed in https://github.com/stratum-mining/sv2-apps/issues/110, I think it would be helpful to add a small clock-skew tolerance to the certificate time validation. This would prevent handshakes from failing when the local machine is only a few seconds off from the server’s clock. Something like a 10-second buffer should be enough to handle minor drift without weakening the security model.

The relevant section of the code is here:
[https://github.com/stratum-mining/stratum/blob/5faf1193dff75c682363ba45bc488a7c9bd524cb/sv2/noise-sv2/src/signature_message.rs#L91-L115](https://github.com/stratum-mining/stratum/blob/5faf1193dff75c682363ba45bc488a7c9bd524cb/sv2/noise-sv2/src/signature_message.rs#L91-L115)

Here’s a sketch of what this could look like:

```rust
    #[inline]
    pub fn verify_with_now(
        self,
        pk: &XOnlyPublicKey,
        authority_pk: &Option<XOnlyPublicKey>,
        now: u32,
    ) -> bool {
        // Allow the local clock to drift up to 10 seconds ahead or behind.
        const TIME_LEEWAY: u32 = 10;

        if let Some(authority_pk) = authority_pk {
            if self.valid_from.saturating_sub(TIME_LEEWAY) <= now
                && self.not_valid_after.saturating_add(TIME_LEEWAY) >= now
            {
                let secp = Secp256k1::verification_only();
                let (m, s) = self.split();
                // m = SHA-256(version || valid_from || not_valid_after || server_static_key)
                let m = [&m[0..10], &pk.serialize()].concat();
                let m = Message::from_hashed_data::<sha256::Hash>(&m);
                let s = match Signature::from_slice(&s) {
                    Ok(s) => s,
                    _ => return false,
                };
                secp.verify_schnorr(&s, &m, authority_pk).is_ok()
            } else {
                false
            }
        } else {
            true
        }
    }
```

This adds just enough tolerance to avoid false negatives caused by tiny NTP differences, while preserving the intended validation boundary.

lucasbalieiro:
closes #2015 

`SignatureNoiseMessage:: verify_with_now` verification now accepts messages whose `valid_from`/`not_valid_after` bounds fall within a 10-second leeway.

impl Display trait for `SignatureNoiseMessage`.

Bump `noise-sv2` from `1.4.0` to `1.4.1`

### DIFF
--- a/sv2/noise-sv2/src/signature_message.rs
+++ b/sv2/noise-sv2/src/signature_message.rs
@@ -22,6 +22,7 @@
 // validity period.
 
 use core::convert::TryInto;
+use core::fmt;
 
 use secp256k1::{hashes::sha256, schnorr::Signature, Keypair, Message, Secp256k1, XOnlyPublicKey};
 
@@ -32,6 +33,7 @@ use secp256k1::{hashes::sha256, schnorr::Signature, Keypair, Message, Secp256k1,
 ///
 /// This structure ensures that messages are authenticated and valid only within
 /// a specified time window, using Schnorr signatures over the `secp256k1` elliptic curve.
+#[derive(Debug, PartialEq, Eq)]
 pub struct SignatureNoiseMessage {
     // Version of the protocol being used.
     pub version: u16,
@@ -43,6 +45,23 @@ pub struct SignatureNoiseMessage {
     pub signature: [u8; 64],
 }
 
+impl fmt::Display for SignatureNoiseMessage {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // write header
+        write!(
+            f,
+            "SignatureNoiseMessage {{ version: {}, valid_from: {}, not_valid_after: {}, signature: ",
+            self.version, self.valid_from, self.not_valid_after
+        )?;
+
+        // write signature as hex manually
+        for byte in &self.signature {
+            write!(f, "{:02x}", byte)?;
+        }
+
+        write!(f, " }}")
+    }
+}
 impl From<[u8; 74]> for SignatureNoiseMessage {
     // Converts a 74-byte array into a [`SignatureNoiseMessage`].
     //
--- a/sv2/codec-sv2/src/error.rs
+++ b/sv2/codec-sv2/src/error.rs
@@ -78,7 +78,14 @@ impl fmt::Display for Error {
             ),
             MissingBytes(u) => write!(f, "Missing `{u}` Noise bytes"),
             #[cfg(feature = "noise_sv2")]
-            NoiseSv2Error(e) => write!(f, "Noise SV2 Error: `{e:?}`"),
+            NoiseSv2Error(e) => match e {
+                NoiseError::InvalidCertificate(msg) => {
+                    write!(f, "Invalid Certificate: {}", msg)
+                }
+                other => {
+                    write!(f, "Noise SV2 Error: {:?}", other)
+                }
+            },
             #[cfg(feature = "noise_sv2")]
             NotInHandShakeState => write!(
                 f,
--- a/sv2/noise-sv2/src/error.rs
+++ b/sv2/noise-sv2/src/error.rs
@@ -6,6 +6,8 @@ use alloc::vec::Vec;
 
 use aes_gcm::Error as AesGcm;
 
+use crate::signature_message::SignatureNoiseMessage;
+
 /// Noise protocol error handling.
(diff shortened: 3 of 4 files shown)


lucasbalieiro:
added [841641d](https://github.com/stratum-mining/stratum/pull/2017/commits/841641d9accb678399cf1ee9ae23fec5f481b979) which implements Display for SignatureNoiseMessage, allowing sv2-apps to show clear, human-readable info instead of raw byte dumps.


Before display trait:
```bash
2025-12-02T14:16:24.691962Z ERROR translator_sv2::sv2::upstream::upstream: Failed Noise handshake with 188.40.233.17:34254: CodecError(NoiseSv2Error(InvalidCertificate([0, 0, 190, 244, 46, 105, 206, 2, 47, 105, 146, 164, 238, 218, 246, 86, 50, 46, 132, 248, 120, 123, 118, 153, 4, 114, 57, 4, 38, 69, 151, 151, 236, 168, 65, 8, 171, 78, 155, 197, 49, 22, 124, 242, 209, 235, 58, 247, 35, 255, 50, 197, 166, 89, 214, 131, 117, 189, 48, 206, 159, 209, 220, 99, 110, 212, 215, 210, 225, 183, 35, 226, 102, 46]))). Retrying...
```

after display trait:
```bash
2025-12-03T23:56:47.689861Z ERROR translator_sv2::sv2::upstream::upstream: Failed Noise handshake with 188.40.233.17:34254: Invalid Certificate: SignatureNoiseMessage { version: 0, valid_from: 1764806224, not_valid_after: 1764809824, signature: 679a439dedcd34e2524a0812237575f3734fb69ccdf8b03c3019d286b63fd7c72acbbab2df28ff8541a06ad0b850e37bc0f282621ff0255a3e1b33212542f600 }. Retrying...
```

sv2/noise-sv2/src/signature_message.rs
```diff
@@ -94,8 +114,13 @@ impl SignatureNoiseMessage {
         authority_pk: &Option<XOnlyPublicKey>,
         now: u32,
     ) -> bool {
+        // Allow the local clock to drift up to 10 seconds ahead or behind.
```
GitGab19, line 117: I would add a reference to the issue you opened here (https://github.com/stratum-mining/stratum/issues/2015).

lucasbalieiro: added

sv2/noise-sv2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "noise_sv2"
-version = "1.4.0"
+version = "1.4.1"
```
lucasbalieiro, line 3: @GitGab19, noise-sv2 bumped

sv2/noise-sv2/src/signature_message.rs
```diff
@@ -94,8 +114,14 @@ impl SignatureNoiseMessage {
...
+        // See https://github.com/stratum-mining/stratum/issues/2015
+        const TIME_LEEWAY: u32 = 10;
+
         if let Some(authority_pk) = authority_pk {
-            if self.valid_from <= now && self.not_valid_after >= now {
+            if self.valid_from.saturating_sub(TIME_LEEWAY) <= now
+                && self.not_valid_after.saturating_add(TIME_LEEWAY) >= now
```
plebhash, line 125: what's the rationale behind the choice of `saturating_add` and `saturating_sub` here?

lucasbalieiro: Because the timestamp is a `u32` from an untrusted peer.
`saturating_add`/`sub` avoids wrap-around or panics if the value is near 0 or `u32::MAX`, preventing subtle bugs.

plebhash: ok but it seems to me that we're deliberately choosing that on edge cases we are:
- capping `valid_from` to `0`
- capping `not_valid_after` to `u32::MAX`

I don't have anything against the rationale behind this design decision, but it should be properly documented for future devs who encounter issues around this part of the implementation

lucasbalieiro: added a comment there

plebhash (approved):

Issue #1994, opened 2025-11-12
cover all Sv2 messages with fuzz targets


plebhash:
currently we only cover `SetupConnection`

lucasbalieiro:
I want to take this one, after I solve the docker setup in https://github.com/stratum-mining/sv2-apps/issues/50

PR #2020, opened 2025-12-08
introduce fuzz coverage for all `subprotocol::common_messages_sv2`

lucasbalieiro:
closes https://github.com/stratum-mining/stratum/issues/2021
partially address #1994 

add fuzz targets to cover serialization <->  deserialization round-trips for all the `subprotocol::common_messages_sv2`

make the `fuzz_tests.yaml` skip targets that do not have a corpus in https://github.com/stratum-mining/stratum-fuzzing-corpus

Added a README to the fuzz crate with initial instructions on how to get started with our targets

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -47,6 +47,30 @@ test = false
 doc = false
 
 [[bin]]
+name = "deserialize_setup_connection_success"
+path = "fuzz_targets/deserialize_setup_connection_success.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "deserialize_setup_connection_error"
+path = "fuzz_targets/deserialize_setup_connection_error.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "deserialize_reconnect"
+path = "fuzz_targets/deserialize_reconnect.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "deserialize_channel_endpoint_changed"
+path = "fuzz_targets/deserialize_channel_endpoint_changed.rs"
+test = false
+doc = false
+
+[[bin]]
 name = "deserialize_stdframe"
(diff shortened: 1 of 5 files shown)


.github/workflows/fuzz_tests.yaml
```diff
@@ -29,7 +29,16 @@ jobs:
       - name: Run fuzz regressions
         run: |
           for TARGET in $(cargo fuzz list); do
+            CORPUS_DIR="fuzz/repo-corpus/corpus/$TARGET"
+
+            echo "==> Checking corpus for: $TARGET"
+
+            if [ ! -d "$CORPUS_DIR" ]; then
```
plebhash, line 36: so this is the actual solution to #2021 ?

lucasbalieiro: yes, it is in a separate commit: 566f56b7dbda09e9b68ae5f4dcc2a7afb02c98e6

GitGab19 (approved):
ACK

fuzz/README.md
```diff
@@ -0,0 +1,77 @@
...
+Before running, install LLVM tools. Instructions:
+[https://doc.rust-lang.org/stable/rustc/instrument-coverage.html#installing-llvm-coverage-tool](https://doc.rust-lang.org/stable/rustc/instrument-coverage.html#installing-llvm-coverage-tool)
+
+Then install `cargo-fuzz`:
+
+```sh
+cargo install cargo-fuzz
+```
+
+Also, make sure you are on the nightly toolchain.
```
plebhash, line 16: isn't it simpler to just add `+nightly` on the command below?

speaking from experience, I don't really switch toolchains back and forth when I need to use `nightly`, I just do `+nightly`

lucasbalieiro: done

PR #2025, opened 2025-12-11
avoid calling `mark_past_jobs_as_stale` 2x

### LINKED ISSUE #2013: Share rejection on pool based on invalid job id.
While testing the SRI stack, we observed that pool was rejecting most of shares even the one's which were creating the blocks. Adding logs: 

Image

plebhash:
close #2013

### DIFF
--- a/sv2/channels-sv2/src/server/standard.rs
+++ b/sv2/channels-sv2/src/server/standard.rs
@@ -487,6 +487,7 @@ where
                 return Err(StandardChannelError::TemplateIdNotFound);
             }
             false => {
+                // try to activate the future job, and also mark past jobs as stale
                 if !self.job_store.activate_future_job(
                     set_new_prev_hash.template_id,
                     set_new_prev_hash.header_timestamp,
@@ -495,9 +496,6 @@ where
                 }
             }
         }
-
-        // mark past jobs as stale
-        self.job_store.mark_past_jobs_as_stale();
 
         // clear seen shares, as shares for past chain tip will be rejected as stale
         self.share_accounting.flush_seen_shares();
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -489,17 +489,22 @@ where
         set_new_prev_hash: SetNewPrevHashTdp<'a>,
     ) -> Result<(), ExtendedChannelError> {
         // extended channels dedicated to custom work don't need to keep track of future jobs
-        if !self.job_store.get_future_jobs().is_empty()
-            && !self.job_store.activate_future_job(
-                set_new_prev_hash.template_id,
-                set_new_prev_hash.header_timestamp,
-            )
-        {
-            return Err(ExtendedChannelError::TemplateIdNotFound);
+        match self.job_store.get_future_jobs().is_empty() {
+            true => {
+                // explicitly mark past jobs as stale, because we're not going to
+                // do it implicitly via activate_future_job in case this extended channel is doing custom work
+                self.job_store.mark_past_jobs_as_stale();
+            }
+            false => {
+                // try to activate the future job, and also mark past jobs as stale
+                if !self.job_store.activate_future_job(
+                    set_new_prev_hash.template_id,
+                    set_new_prev_hash.header_timestamp,
+                ) {
+                    return Err(ExtendedChannelError::TemplateIdNotFound);
+                }
+            }
         }
-
-        // mark past jobs as stale
-        self.job_store.mark_past_jobs_as_stale();
 
         // clear seen shares, as shares for past chain tip will be rejected as stale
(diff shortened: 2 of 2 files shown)


lucasbalieiro:
> @lucasbalieiro can you please reproduce the same low-diff-`signet`-based testing workflow you did on #2013 to confirm this solution?

Done! 

using sv2-apps@main + changes of this PR @ c849b6aef60d374b7fa3819ffcf921717d88efb6

Full suite of apps + bitcoin-v30 + minerd

On signet with low difficulty. Mined so far 32 blocks and the error: `invalid-job-id` does not appear on the logs.

Shourya742 (approved):
tACK. Super subtle

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -487,6 +487,7 @@ where
                 return Err(StandardChannelError::TemplateIdNotFound);
```
GitGab19, line 487: why are we returning the error here while on the extended channel case we use `self.job_store.mark_past_jobs_as_stale();`?

plebhash: Extended Channels could be potentially catering for a JD connection, where maybe the application layer never calls `on_new_template`, leaving `self.job_store.get_future_jobs()` to always return empty

in that case, we don't want to return `TemplateIdNotFound`, but we still need to mark past jobs as stale (because this is not going to be done implicitly via `activate_future_job`)

on Standard Channels, we should always expect a mining server to call `on_new_template`, leaving no possibility of `self.job_store.get_future_jobs()` being empty during `on_set_new_prev_hash`... and if that case we return an error

with that said, if it helps in making both channel implementations more symmetrical and intuitive, I guess we could also make Standard Channels a bit more flexible, and avoid returning an error if there's no matching template

---

this is somehow related to the discussions happening on https://github.com/stratum-mining/sv2-apps/pull/131/, so it makes sense that you would ask this quesiton

and it also made me realize that my comment https://github.com/stratum-mining/sv2-apps/pull/131/files#r2613560222 is actually innacurate

GitGab19: Ok, now it's very clear to me.

There's no need to change anything on standard channels at this point.

I would probably add a simple comment on the extended channel case to explain why we mark those jobs as stale (trying to make a quick point about the possible JD scenario).

plebhash: the first line of `on_set_new_prev_hash` for the Extended Channel is:
```
        // extended channels dedicated to custom work don't need to keep track of future jobs
```

then the following code is:
```rust
        match self.job_store.get_future_jobs().is_empty() {
            true => {
                // explicitly mark past jobs as stale, because we're not going to
                // do it implicitly via activate_future_job
                self.job_store.mark_past_jobs_as_stale();
            }
            false => {
                // try to activate the future job, and also mark past jobs as stale
                if !self.job_store.activate_future_job(
                    set_new_prev_hash.template_id,
                    set_new_prev_hash.header_timestamp,
                ) {
                    return Err(ExtendedChannelError::TemplateIdNotFound);
                }
            }
        }
```

do you feel this could be further expanded?

plebhash: ok maybe:
```diff
        match self.job_store.get_future_jobs().is_empty() {
            true => {
                // explicitly mark past jobs as stale, because we're not going to
-               // do it implicitly via activate_future_job
+               // do it implicitly via activate_future_job in case this extended channel is doing custom work
                self.job_store.mark_past_jobs_as_stale();
            }
            false => {
                // try to activate the future job, and also mark past jobs as stale
                if !self.job_store.activate_future_job(
                    set_new_prev_hash.template_id,
                    set_new_prev_hash.header_timestamp,
                ) {
                    return Err(ExtendedChannelError::TemplateIdNotFound);
                }
            }
        }
```

GitGab19: I did miss the first line which mentions JD, my bad.

GitGab19: > ok maybe:
> 
> ```diff
>         match self.job_store.get_future_jobs().is_empty() {
>             true => {
>                 // explicitly mark past jobs as stale, because we're not going to
> -               // do it implicitly via activate_future_job
> +               // do it implicitly via activate_future_job in case this extended channel is doing custom work
>                 self.job_store.mark_past_jobs_as_stale();
>             }
>             false => {
>                 // try to activate the future job, and also mark past jobs as stale
>                 if !self.job_store.activate_future_job(
>                     set_new_prev_hash.template_id,
>                     set_new_prev_hash.header_timestamp,
>                 ) {
>                     return Err(ExtendedChannelError::TemplateIdNotFound);
>                 }
>             }
>         }
> ```

Better :+1:

PR #2026, opened 2025-12-12
ci: detect companion PR changes and conditionally trigger integration tests

### LINKED ISSUE #2022: need to refine automation around Integration Tests
## problem

while working on #2011, I found myself on a chicken-and-egg situation

PR #2011 has changes to `channels_sv2`, which require a companion https://github.com/stratum-mining/sv2-apps/pull/122 on `sv2-apps` introducing the corresponding apaptations to the application layer

but the changes of https://github.com/stratum-mining/sv2-apps/pull/122 are also a pre-requisite for Integration Tests CI on #2011, because it cannot be merged before we know for sure adaptations are validated on the application layer

I could naively modify `stratum/scripts/run-integration-tests.sh` so it checks out `sv2-apps` on the companion branch from https://github.com/stratum-mining/sv2-apps/pull/122, but that means the script would be permanently changed once #2011 is merged, and every subsequent execution of the script would be pointing to this development branch, which is not desirable at all

## proposed solution

we modify `stratum/scripts/run-integration-tests.sh` so that when it's being executed in CI for a PR on `stratum` (e.g.: #2011), the script checks out the same fork+branch from the companion PR (e.g.: https://github.com/stratum-mining/sv2-apps/pull/122) on the local clone of `sv2-apps` that it's about to patch

the fork and branch names are encoded as environment variables:
- `$SV2_APPS_COMPANION_FORK`
- `$SV2_APPS_COMPANION_BRANCH`

`run-integration-tests.sh` reads these variables from its execution environment

we also modify the `integration-tests.yaml` workflow of `stratum` repo so we can trigger execution of `run-integration-tests.sh` by writing a line on the `stratum` PR description that sets these variables for a targeted execution of `stratum/scripts/run-integration-tests.sh`

for example, the author of #2011 writes this line on the PR description:
> companion https://github.com/stratum-mining/sv2-apps/pull/ 122

(doesn't have to be exactly this)

which triggers `integration-tests.yaml` to start a new execution of `run-integration-tests.sh` while populating `$SV2_APPS_COMPANION_FORK`+`$SV2_APPS_COMPANION_BRANCH` as environment variables

if `$SV2_APPS_COMPANION_FORK`+`$SV2_APPS_COMPANION_BRANCH` are empty, then most likely there's no need for adaptations on the application layer, for which case `run-integration-tests.sh` should simply run against upstream `main` without any explicit intervetions from the PR author on `stratum`

essentially, this should be the default/initial behavior of CI immediately after every PR is opened on `stratum` repo... then Integration Test failure becomes a clear indication that a companion PR is needed on `sv2-apps` before proceeding

as usual, PRs should not be mergeable without a successful execution of `integration-tests.yaml`

## TLDR

we should be able to dynamically point `run-integration-tests.sh` to specific fork+branch of companion PRs at `sv2-apps`

the ideal trigger should be simply writing comments on PRs of `stratum` with the URL of the companion PR

## steps

- [ ] modify `run-integration-tests.sh`
- [ ] modify `integration-tests.yaml`
- [ ] modify `CONTRIBUTING.md`
- [ ] validate solution against #2011 + https://github.com/stratum-mining/sv2-apps/pull/122 and further iterate if the behavior is not the expected one

lucasbalieiro:
closes #2022.

This PR updates `./scripts/run-integration-tests.sh` and `./.github/workflows/integration-tests.yaml` to support the feature described in the issue #2022.

From now on, whenever you open a PR in this repo and realize the changes should also be reflected in the `https://github.com/stratum-mining/sv2-apps` repo, you can simply add a line to this PR’s description.

Example (as shown in the image below):
```
<keyword> <pr_link>
```
image

This will re-trigger the integration tests, but this time they’ll run against the specified PR in `sv2-apps`.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,33 +1,68 @@
+name: Integration Tests
+
 on:
   pull_request:
+    types: [opened, synchronize, reopened, edited]
     branches:
       - main
-
-name: Integration Tests
+  workflow_dispatch:
 
 jobs:
   ci:
-    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-        include:
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-gnu
+        os: [ubuntu-latest, macos-latest]
+
+    runs-on: ${{ matrix.os }}
 
     steps:
-      - name: Use stable toolchain
-        uses: actions/checkout@v4
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-
-      - name: Install capnproto
+      - name: Detect companion line
+        id: detect
+        env:
+          BODY_NEW: ${{ github.event.pull_request.body }}
+          BODY_OLD: ${{ github.event.changes.body.from || '' }}
+          ACTION: ${{ github.event.action }}
         run: |
-          sudo apt-get update && sudo apt-get install -y capnproto libcapnp-dev
+          OLD_COMPANION=$(printf "%s" "$BODY_OLD" \
+            | grep -Eo 'companion https://github\.com/stratum-mining/sv2-apps/pull/[0-9]+' \
+            || true)
+
+          NEW_COMPANION=$(printf "%s" "$BODY_NEW" \
+            | grep -Eo 'companion https://github\.com/stratum-mining/sv2-apps/pull/[0-9]+' \
+            || true)
+
+          if [ "$ACTION" = "edited" ]; then
+            if [ "$OLD_COMPANION" != "$NEW_COMPANION" ]; then
+              echo "should_run_integration_test=true" >> "$GITHUB_OUTPUT"
+            else
+              echo "should_run_integration_test=false" >> "$GITHUB_OUTPUT"
+            fi
+          else
+            echo "should_run_integration_test=true" >> "$GITHUB_OUTPUT"
+          fi
+
+          COMPANION_NUM=$(printf "%s" "$NEW_COMPANION" | grep -Eo '[0-9]+$' || true)
+          echo "COMPANION_PR_NUMBER=${COMPANION_NUM:-}" >> "$GITHUB_ENV"
+
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@1.85
+
+      - name: Install capnproto (Ubuntu)
+        if: steps.detect.outputs.should_run_integration_test == 'true' && matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt-get install -y capnproto libcapnp-dev
+
+      - name: Install capnproto (macOS)
+        if: steps.detect.outputs.should_run_integration_test == 'true' && matrix.os == 'macos-latest'
+        run: |
+          brew install capnp
+
+      - name: Install cargo-nextest
+        run: cargo install cargo-nextest --version 0.9.100 --locked
 
       - name: Run Integration Tests Script
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -1,33 +1,56 @@
...
-            target: x86_64-unknown-linux-gnu
+    runs-on: ubuntu-latest
 
     steps:
-      - name: Use stable toolchain
-        uses: actions/checkout@v4
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
```
plebhash, line 25: are we intentionally removing these?

.github/workflows/integration-tests.yaml
```diff
@@ -1,33 +1,56 @@
...
 jobs:
   ci:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os:
-          - ubuntu-latest
-        include:
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-gnu
```
plebhash, line 17: are we intentionally removing these?

.github/workflows/integration-tests.yaml
```diff
@@ -1,33 +1,56 @@
...
       - name: Install capnproto
-        run: |
-          sudo apt-get update && sudo apt-get install -y capnproto libcapnp-dev
+        if: steps.detect.outputs.should_run_integration_test == 'true'
+        run: sudo apt-get update && sudo apt-get install -y capnproto libcapnp-dev
 
       - name: Run Integration Tests Script
-        run: |
-          ./scripts/run-integration-tests.sh
+        if: steps.detect.outputs.should_run_integration_test == 'true'
```
plebhash, line 66: the only conditions in which this will be `false` is if we edit the PR description but the PR link remains the same

is that correct?

.github/workflows/integration-tests.yaml
lucasbalieiro: Yeah, I’m not sure why those args were there to begin with. They don’t exist in the `actions/checkout` action.

Docs: https://github.com/actions/checkout?tab=readme-ov-file#usage

My guess is that it was just an oversight. Those kinds of arguments do exist in the `actions-rs` ecosystem (https://github.com/actions-rs), but that’s a completely different set of actions.

.github/workflows/integration-tests.yaml
lucasbalieiro: yeah, its pointless to have a matrix with just one OS.

.github/workflows/integration-tests.yaml
lucasbalieiro: Yes, exactly.

The idea is that not every edit to the PR description is related to the companion PR. Sometimes you’re just fixing a typo, and there’s no reason to trigger the integration tests again for that.

.github/workflows/integration-tests.yaml
lucasbalieiro: Now that I’m looking at this, it makes me wonder: shouldn’t the integration tests here also run on macOS? That would actually give the matrix some real purpose, since we’d be testing across multiple OSes, similar to what we already do in sv2-apps.

Reference: https://github.com/stratum-mining/sv2-apps/blob/07b594b6cc22e5ac4adb616931b1d59e0ed288da/.github/workflows/integration-tests.yaml#L11-L21

GitGab19: Yeah I would keep testing also on macos.

lucasbalieiro: Done, repurposed the matrix, now we run on mac and ubuntu

PR #2028, opened 2025-12-14
fuzz: refactor SV2 message fuzzing and expand coverage

### LINKED ISSUE #1994: cover all Sv2 messages with fuzz targets

lucasbalieiro:
closes #1994;

this PR adds end-to-end serialization fuzz targets for all SV2 subprotocol messages and refactors the fuzzing setup to eliminate duplicated logic.

A shared round-trip macro is introduced, with one fuzz entry point per subprotocol. Legacy targets made redundant by this refactor are removed.


Shourya742 (commented):
These fuzz targets are largely cosmetic and almost identical. I think we should generalize them and define the logic in a single place instead of duplicating it per message type. A small generic helper like the following should be sufficient to cover all of these end-to-end deserialize → serialize round-trip cases:

 ```rust
 #![no_main]
 use binary_sv2::{Deserialize, Encodable, GetSize};
 use libfuzzer_sys::fuzz_target;
 use mining_sv2::{OpenStandardMiningChannel, OpenStandardMiningChannelSuccess};

 fn end_to_end<'a, T>(data: &'a mut [u8])
 where
     T: Deserialize<'a> + Encodable + GetSize + std::fmt::Display + Clone,
 {
     if let Ok(msg) = T::from_bytes(data) {
         let mut dst = vec![0u8; msg.get_size()];
         msg.to_bytes(&mut dst).unwrap();
     }
 }

 fuzz_target!(|data: Vec<u8>| {
     end_to_end::<OpenStandardMiningChannelSuccess>(&mut data.clone());
     end_to_end::<OpenStandardMiningChannel>(&mut data.clone());
     ...
 });
 ```
This should be fine

fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,61 @@
...
+/// test_roundtrip!(MyMessage, input_bytes);
+/// ```
+#[macro_export]
+macro_rules! test_roundtrip {
+    ($msg_type:ty, $data:expr) => {{
+        // Step 1: Try to parse the input bytes.
+        // Invalid inputs are expected in fuzzing, so we silently ignore failures.
+        let mut input = $data.clone();
+        let Ok(parsed) = <$msg_type>::from_bytes(&mut input) else {
+            return;
```
bit-aloo: This return will terminate the fuzz target early.

lucasbalieiro: uuuu, thanks. adjusted. and also rebased with main

fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,59 @@
...
+            //
+            // Not all message types implement `Eq`, so we compare their `Display`
+            // output instead. If both messages can be parsed successfully and
+            // represent the same data, their formatted output should match.
+            assert_eq!(
+                parsed.to_string(),
+                reparsed.to_string(),
+                "Display output mismatch"
+            );
+        };
```
bit-aloo, line 57: Shouldn't asserting data and encoding is sufficient?

lucasbalieiro: No, doing `assert_eq!(data, encoded_1)` will fail most of the time.

Because our macro doesn’t preserve the original byte stream. When we call `from_bytes` directly on the fuzz input, it *normalizes* the data by trimming bytes it considers useless or redundant. So the round-trip is **parse → canonicalize → re-encode**, not a raw byte-for-byte echo.

For example, I tweaked the macro to include this assertion:

```rust
#[macro_export]
macro_rules! test_roundtrip {
    ($msg_type:ty, $data:expr) => {{
        // Step 1: Try to parse the input bytes.
        // Invalid inputs are expected during fuzzing, so failures are ignored.
        let mut input = $data.clone();
        if let Ok(parsed) = <$msg_type>::from_bytes(&mut input) {
            // Step 2: Serialize the successfully parsed message.
            let mut encoded_1 = vec![0u8; parsed.get_size()];
            parsed
                .clone()
                .to_bytes(&mut encoded_1)
                .expect("Encoding failed after a successful parse");

            assert_eq!($data, encoded_1); //assert the fuzz input with the output of to_bytes
        }
    }};
}
```

What you get most of the time is something like this:

```bash
assertion `left == right` failed
  left:  [0, 255, 255, 0]
 right: [0, 255, 255]
```

The parser drops trailing or otherwise meaningless bytes, so the encoder emits a *canonical* representation, not the original fuzz input.

bit-aloo: `assert_eq!(data[..encoded.len()], encoded)`? We can do this, right? The reason why I am mentioning this, is because this extra round trip is not helping much.

lucasbalieiro: For some cases, yes.

But other times, the macro remove things between  the byte array itself, not only at the end:

```bash
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1310975141
INFO: Loaded 1 modules   (17475 inline 8-bit counters): 17475 [0x1024a17a0, 0x1024a5be3), 
INFO: Loaded 1 PC tables (17475 PCs): 17475 [0x1024a5be8,0x1024ea018), 
fuzz/target/aarch64-apple-darwin/release/end_to_end_serialization_for_mining_messages: Running 1 inputs 1 time(s) each.
Running: fuzz/artifacts/end_to_end_serialization_for_mining_messages/crash-d07dc60ba8083aee3632a81fd60a708dee857142

thread '<unnamed>' (9571877) panicked at fuzz_targets/end_to_end_serialization_for_mining_messages.rs:12:5:
assertion `left == right` failed
  left: [255, 231, 255, 0, 50, 197, 5, 5, 1, 255, 255, 5, 5, 255, 5, 5, 1, 255, 0, 7, 0, 255, 255, 255, 255, 173, 0, 197, 0, 0]
 right: [255, 231, 255, 0, 50, 197, 5, 5, 1, 255, 255, 5, 5, 255, 5, 5, 1, 1, 0, 7, 0, 255, 255, 255, 255, 173, 0, 197, 0, 0]
```

lucasbalieiro: but at the same time, makes me question: should it be changing this bytes? It is changing things from 255 to 1

bit-aloo: There is something for sure, happening here and we should investigate. The serialization and deserialization should be lossless, and here that's not the case.

lucasbalieiro: I adapted the macro and saved the crashing input so we can deal with it in another issue; so we don't block this PR;

bit-aloo: > For some cases, yes.
> 
> But other times, the macro remove things between the byte array itself, not only at the end:
> 
> ```shell
> INFO: Running with entropic power schedule (0xFF, 100).
> INFO: Seed: 1310975141
> INFO: Loaded 1 modules   (17475 inline 8-bit counters): 17475 [0x1024a17a0, 0x1024a5be3), 
> INFO: Loaded 1 PC tables (17475 PCs): 17475 [0x1024a5be8,0x1024ea018), 
> fuzz/target/aarch64-apple-darwin/release/end_to_end_serialization_for_mining_messages: Running 1 inputs 1 time(s) each.
> Running: fuzz/artifacts/end_to_end_serialization_for_mining_messages/crash-d07dc60ba8083aee3632a81fd60a708dee857142
> 
> thread '<unnamed>' (9571877) panicked at fuzz_targets/end_to_end_serialization_for_mining_messages.rs:12:5:
> assertion `left == right` failed
>   left: [255, 231, 255, 0, 50, 197, 5, 5, 1, 255, 255, 5, 5, 255, 5, 5, 1, 255, 0, 7, 0, 255, 255, 255, 255, 173, 0, 197, 0, 0]
>  right: [255, 231, 255, 0, 50, 197, 5, 5, 1, 255, 255, 5, 5, 255, 5, 5, 1, 1, 0, 7, 0, 255, 255, 255, 255, 173, 0, 197, 0, 0]
> ```

Considering this, and trimming we did currently, won't it cause test to fail? As the bytes mismatch?

lucasbalieiro: Yes, the test did fail as expected at the `assert_eq!(data[..encoded.len()], encoded`

lucasbalieiro: @Shourya742, I reverted to the original intent as we discussed during the dev call and opened an issue to track the problem found: https://github.com/stratum-mining/stratum/issues/2043

plebhash (approved):

Issue #2019, opened 2025-12-08
Allow Channels structs should use generic jobs to support generic JobStore


colerob1776:
Now that we support a generic JobStore which is bound by the `Job` trait, we should allow the Channels structs in channels_sv2 to utilize trait bounds as well in order to fully support the JobStore.

For more info see https://github.com/stratum-mining/stratum/pull/2011#discussion_r2596426319

GitGab19:
Should we close this issue? cc @colerob1776 @plebhash

Issue #2010, opened 2025-11-28
version bumps for `v1.7.0`


GitGab19:
Similar to #1933

plebhash:
https://github.com/stratum-mining/stratum/pull/2017 is bumping `noise_sv2` from `1.4.0` to `1.4.1`

plebhash:
#2030 is bumping `binary_sv2` from `5.0.0` to `5.1.0`

plebhash:
#2044 is bumping `mining_sv2` from `6.0.0` to `7.0.0`

PR #2030, opened 2025-12-15
Draft: Refactor channels with generic jobs

coleFD:


sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -47,72 +66,57 @@ pub trait JobStore<T: Job>: Send + Sync + Debug {
...
-    fn get_past_job(&self, job_id: u32) -> Option<T>;
+    /// Removes an owned copy of a future job from its job ID, if any.
+    fn remove_future_job(&mut self, job_id: u32) -> Option<EitherJob<'a>>;
 
-    /// Returns true if there are any stale jobs, false otherwise.
-    fn has_stale_jobs(&self) -> bool;
-
-    /// Returns an owned copy of a stale job from its job ID, if any.
-    fn get_stale_job(&self, job_id: u32) -> Option<T>;
+    fn try_validate_job(
```
coleFD, line 77: I went this route because the current trait kinda nudges you toward having the same structure as DefaultJobStore. In it's most basic form I beleive the JobStore should allow you to:
1. add an active job
2. add and activate a future job
3. make jobs stale
4. retrieve jobs for validation

The extra methods are bloat imo. 
In our case we use timestamp based job ids, and it could be possible---albeit a very low chance---that we have duplicate job_ids. Also, assuming there is access to the underlying JobStore, the user could override share validation to allow for merged mining

sv2/channels-sv2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "channels_sv2"
-version = "3.0.0"
```
plebhash, line 3: version `3.0.0` hasn't been published yet, so no need to bump this

see https://github.com/stratum-mining/stratum/issues/2010#issuecomment-3593147843

sv2/channels-sv2/src/server/jobs/either_job.rs
```diff
@@ -24,31 +32,108 @@ use template_distribution_sv2::NewTemplate;
...
+
+        let mut coinbase_outputs = vec![];
+        coinbase_outputs.extend(additional_coinbase_outputs);
+        coinbase_outputs.extend(template_coinbase_outputs);
+
+        Ok(Self {
+            origin: JobOrigin::NewTemplate(template),
+            extranonce_prefix,
+            coinbase_outputs,
+            merkle_root: None,
```
plebhash: None?

a fundamental characteristic of a Standard Job is a fixed merkle root

coleFD: copy/paste and didnt copy in the merkle root from the job message. Updated

sv2/channels-sv2/src/server/jobs/factory.rs
```diff
@@ -64,6 +64,8 @@ impl JobIdFactory {
 /// Enables creation of new Extended Jobs from NewTemplate and SetCustomMiningJob messages.
 ///
 /// Enables creation of new Standard Jobs from NewTemplate messages.
+///
+/// The `JobContainer` generic parameter allows for smart pointers that wrap `<dyn Job<'a>>` (ie. `Box<dyn Job<'a>>`, `Arc<dyn Job<'a>>`, etc.)
```
plebhash, line 68: I don't see any `JobContainer`?

sv2/channels-sv2/src/server/jobs/either_job.rs
```diff
@@ -0,0 +1,432 @@
...
+/// Please note that `coinbase_tx_prefix` and `coinbase_tx_suffix` are stored in memory with bip141
+/// data (marker, flag and witness). That makes it easy to reconstruct the segwit coinbase while
+/// trying to propagate a block.
+///
+/// However, the `coinbase_tx_prefix` and `coinbase_tx_suffix` contained in the
+/// `NewExtendedMiningJob` message to be sent across the wire DO NOT contain bip141 data.
+/// That makes it easy to calculate the coinbase `txid` (instead of `wtxid`) for merkle root
+/// calculation.
+#[derive(Debug, Clone, PartialEq)]
+pub struct EitherJob<'a> {
```
plebhash, line 47: so why are we going with a concrete type `EitherJob`?

from the description of #2019 and discussions on https://github.com/stratum-mining/stratum/pull/2011#discussion_r2585391628, I was more under the impression that we were going to introduce new traits for Extended and Standard jobs?

coleFD: I started down that route but it was painful. I realized the only difference between the jobs once they are stored in the channel is a pre-calcuated merkle root so it's easier to just make a single job struct with an optional merkle_root field

coleFD: Granted, JobStore does still allow Generic jobs that dont have to be EitherJob. We will need to implement our own jobs for merged mining so this is still possible. The EitherJob just makes compatibility easier with this set up

sv2/channels-sv2/src/server/jobs/factory.rs
coleFD: outdated comment. Will remove

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -94,17 +95,19 @@ where
...
+    _phantom: PhantomData<(&'a JobIn, &'a JobOut)>,
 }
 
-impl<'a, J> ExtendedChannel<'a, J>
+impl<'a, JobIn, JobOut, Store> ExtendedChannel<'a, JobIn, JobOut, Store>
 where
-    J: JobStore<ExtendedJob<'a>>,
+    Store: JobStore<'a, JobIn, JobOut>,
+    JobIn: Job<'a> + From<EitherJob<'a>> + Clone,
+    JobOut: Job<'a> + Clone,
```
plebhash, line 110: what's the rationale behind `JobIn` vs `JobOut`?

sv2/channels-sv2/src/server/jobs/either_job.rs
plebhash: > but it was painful.

can you elaborate? what kinds of challenges did you encounter?

sv2/channels-sv2/src/server/extended.rs
coleFD: future jobs need to be mutable to activate `min_ntime`, active jobs are immutable. So you can add a future job of type `T` and then your implementation can wrap that as `Arc<T>` or `Rc<T>` if you want when fetching jobs to avoid cloning the full job

sv2/channels-sv2/src/server/jobs/either_job.rs
coleFD: Casting from the strong typed result of the JobFactory::new_extended_job or new_standard_job to the trait. Sure we could box the types, but EitherJob seems more elegant to me considering the only difference is a cached merkle_root. I even noticed coinbase_prefix and suffix are calculated for the original calculation of the merkle_root for standard jobs in the JobFactory and then they just get thrown away, which requires rebuilding and serializing the coinbase on every share considering `ShareValidationResult::BlockFound` requires a serialized coinbase transaction. Seems like a ways to not keep the prefix/suffix and do the same thing as an extended share

PR #2029, opened 2025-12-15
rename `DeclareMiningJob.tx_ids_list` to `wtxid_list`

plebhash:
close #2018 

companion https://github.com/stratum-mining/sv2-apps/pull/134

PR #2031, opened 2025-12-16
make `run-integration-tests.sh` update the integration tests Cargo.lock.

lucasbalieiro:
as follow-up of #2026 

to unblock #2029 

This avoids "patch was not used in the crate graph" warnings when the version of a patched dependency changes and the lockfile still pins the old version.

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -122,14 +122,20 @@ fi
 
 # Add the patch section at the end of the file
 cat >> Cargo.toml << 'EOF'
-
 # Override dependencies with local paths
 [patch.crates-io] 
 stratum-core = {path = "../../../stratum-core"}
-
 [patch."https://github.com/stratum-mining/stratum"]
 stratum-core = {path = "../../../stratum-core"}
 EOF
+
+# Force a refresh of Cargo.lock so the patched stratum-core is not ignored after
+# a crate version change. When stratum-core's version changes, the lockfile may still
+# pin an older version, causing Cargo to ignore the patch entirely
+# ("patch was not used in the crate graph"). Running `cargo update -p stratum-core`
+# updates the lockfile to match the patched crate, ensuring the local override
+# actually takes effect.
+cargo update -p stratum-core
 
 echo "✅ Updated Cargo.toml to use local dependencies"
 echo "🏃 Running integration tests..."


scripts/run-integration-tests.sh
```diff
@@ -122,15 +122,16 @@ fi
 
 # Add the patch section at the end of the file
 cat >> Cargo.toml << 'EOF'
-
 # Override dependencies with local paths
 [patch.crates-io] 
 stratum-core = {path = "../../../stratum-core"}
-
```
lucasbalieiro, line 129: These lines were removed because a preceding sed command deletes content up to the first blank line.
The extra blank line triggered incorrect behavior in local sessions.
CI did not catch this because each CI run fetches `sv2-apps` from a clean state.

scripts/run-integration-tests.sh
```diff
@@ -122,15 +122,16 @@ fi
...
 [patch."https://github.com/stratum-mining/stratum"]
 stratum-core = {path = "../../../stratum-core"}
 EOF
 
+# we are inside the INTEGRATION_TESTS_DIR at this point
+cargo update -p stratum-core
```
plebhash, line 138: why do we need to run `cargo update`?

we should add a comment explaining the motivations

lucasbalieiro: `ammend`ed a comment expanding the explanation of the PR description

```diff
diff --git a/scripts/run-integration-tests.sh b/scripts/run-integration-tests.sh
index 10362dacfb..f14aac4c28 100755
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -122,15 +122,21 @@ fi
 
 # Add the patch section at the end of the file
 cat >> Cargo.toml << 'EOF'
-
 # Override dependencies with local paths
 [patch.crates-io] 
 stratum-core = {path = "../../../stratum-core"}
-
 [patch."https://github.com/stratum-mining/stratum"]
 stratum-core = {path = "../../../stratum-core"}
 EOF
 
+# Force a refresh of Cargo.lock so the patched stratum-core is not ignored after
+# a crate version change. When stratum-core's version changes, the lockfile may still
+# pin an older version, causing Cargo to ignore the patch entirely
+# ("patch was not used in the crate graph"). Running `cargo update -p stratum-core`
+# updates the lockfile to match the patched crate, ensuring the local override
+# actually takes effect.
+cargo update -p stratum-core
+
 echo "✅ Updated Cargo.toml to use local dependencies"
 echo "🏃 Running integration tests..."
 
```

PR #2034, opened 2025-12-19
Add rust-toolchain.toml 

Shourya742:
This PR adds rust-toolchain pinning the version to 1.75

### DIFF
--- a/.github/workflows/clippy-lint.yaml
+++ b/.github/workflows/clippy-lint.yaml
@@ -28,6 +28,9 @@ jobs:
         with:
           components: clippy
       
+      # NOTE: A repo-local `rust-toolchain.toml` is still respected by rustup.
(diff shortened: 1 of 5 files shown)


.github/workflows/clippy-lint.yaml
```diff
@@ -24,9 +24,14 @@ jobs:
 
       - name: Install Rust (1.89) + Clippy
         uses: dtolnay/rust-toolchain@1.89
+        id: toolchain
         with:
           components: clippy
-          
+
```
GitGab19, line 30: Let's add a comment here to explain why we need this.

The same for all the other workflows modified by this PR.

bit-aloo: Done

PR #2035, opened 2025-12-19
Arrange buffer-sv2 fuzz and bench correctly

Shourya742:
part of #2042 

This PR removes the buffer_sv2 fuzz targets. It also makes criterion a dev-dependency; it was previously optional, which caused cargo bench to break. In addition, the benchmarks were not compiling correctly this PR applies minimal fixes to get them building and running. Overall, this brings the benchmarking suite back to a working state.

### DIFF
--- a/sv2/codec-sv2/Cargo.toml
+++ b/sv2/codec-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "codec_sv2"
-version = "4.0.0"
+version = "4.0.1"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"
--- a/sv2/framing-sv2/Cargo.toml
+++ b/sv2/framing-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "framing_sv2"
-version = "6.0.0"
+version = "6.0.1"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"


fuzz/Cargo.toml
```diff
@@ -81,3 +86,15 @@ name = "deserialize_datatypes"
...
+[[bin]]
+name = "slower"
+path = "fuzz_targets/slower.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "faster"
+path = "fuzz_targets/faster.rs"
+test = false
```
lucasbalieiro: I would recommend to change the name of this targets to a more descriptive one.

fuzz/fuzz_targets/faster.rs
lucasbalieiro, line 1: Since you have more context of the `buffer-sv2` crate itself, do you think these targets make sense?

lucasbalieiro: running the `faster` one I get a local execution of ~= exec/s: 29, which is extremely slow and not recommended.

In less then 30 seconds of execution we get an OOM:
`SUMMARY: libFuzzer: out-of-memory`

fuzz/Cargo.toml
```diff
@@ -16,11 +16,16 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = "0.4.10"
 arbitrary = { version = "1.4.2", features = ["derive"] }
+buffer_sv2 = {path = "../sv2/buffer-sv2", feature = ["fuzz"]}
 binary_sv2 = { path = "../sv2/binary-sv2"}
 parsers_sv2 = { path = "../sv2/parsers-sv2" }
 framing_sv2 = { path = "../sv2/framing-sv2" }
 codec_sv2 = { path = "../sv2/codec-sv2", features = ["noise_sv2"]}
 common_messages_sv2 = { path = "../sv2/subprotocols/common-messages" }
+affinity = "0.1.1"
```
lucasbalieiro: Just a note: this crate blocks execution of all the targets on macos. This is used by the `slower.rs`

lucasbalieiro: If we are going to keep this fuzz targets, maybe we could put this behind  `features`?

fuzz/fuzz_targets/faster.rs
bit-aloo: TBH, this isn’t really a fuzz target. It’s only using the fuzz harness, but what it’s actually doing is a concurrency stress test with some randomness and timing mixed in. The goal here seems to be to scheck for race condition or memory corruption, not to fuzz inputs in the usual sense.. I am removing this.

fuzz/random
lucasbalieiro, line 1: We can drop this file, since we are dropping the targets

bit-aloo: oopsie done

sv2/buffer-sv2/benches/control_struct.rs
lucasbalieiro, line 1: random question: on the root of the `buffer-sv2/` we have a `BENCHES.md` with what seems to be some reports generated by the `cargo bench`.  Do we need to update it?

sv2/buffer-sv2/benches/pool_benchmark.rs
```diff
@@ -1,15 +1,11 @@
...
 use buffer_sv2::{Buffer, BufferFromSystemMemory as BufferFromMemory, BufferPool as Pool, Slice};
 use core::time::Duration;
 use rand::Rng;
 
 mod control_struct;
-use control_struct::{
-    add_random_bytes, bench_no_thread, keep_slice, Load, MaxESlice, MaxEfficeincy, PPool, SSlice,
-    DATA,
-};
+use control_struct::{add_random_bytes, bench_no_thread, keep_slice, MaxEfficeincy, PPool, DATA};
```
lucasbalieiro:
```suggestion
use control_struct::{add_random_bytes, bench_no_thread, keep_slice, MaxEfficiency, PPool, DATA};
```
we could take the opportunity to address this typo.

I was trying to drop that weird `random` bin, but `pool_iai` is using it

sv2/buffer-sv2/benches/control_struct.rs
bit-aloo: I don’t think that’s the case. We didn’t modify any tests there, only fixed the setup to make it compile. And as far as I am aware, buffer_sv2 hasn’t seen any significant changes for a while.

sv2/buffer-sv2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "buffer_sv2"
-version = "2.0.0"
+version = "3.0.0"
```
plebhash, line 3: what was the criteria for this? did `semver-check` CI break?

sv2/buffer-sv2/Cargo.toml
```diff
@@ -19,6 +19,7 @@ generic-array = "=0.14.7"
 rand = "0.8.3"
 iai = "0.1"
 criterion = "0.3"
+rayon-core = "=1.12.1"
```
plebhash, line 22: why pinning?

sv2/buffer-sv2/Cargo.toml
bit-aloo: We removed criterion as the feature flag.

sv2/buffer-sv2/Cargo.toml
bit-aloo: criterion was pulling rayon-core version, which was not compatible with 1.75. So, had to pin it manually.

sv2/buffer-sv2/Cargo.toml
plebhash: I'm still wondering whether this constitutes a major bump.

ultimately, that flag didn't really have an impact on public APIs, did it?

I'm not hardly opinionated on this, but I think it's worth reflecting about.

bit-aloo: The lines get blurry here when you consider semantic versioning. If a crate depends on a feature that gets removed and we only bump the minor or patch version, Cargo may automatically upgrade to that version. Since the feature is no longer available, this would result in a compilation error for downstream users. Just for this major bumping make sense, though we could argue that we are not changing any public API's. Very blurry.

GitGab19: Which feature has been removed?

bit-aloo: Criterion was optional initially, I moved it to dev-dependency. Optionality is exposed as feature flag.

GitGab19: And why did you move it to `dev-dependencies`?

bit-aloo: I don’t think that’s quite right. Benches run in a separate context from normal builds, and they rely on tooling that isn’t part of the regular build path. Because of that, benches related dependencies really shouldn’t be pulled in during normal compilation. In practice, Criterion is just benchmarking infrastructure, so it makes more sense for it to live in dev-dependencies. That way it’s only around when running benches, and doesn’t affect regular builds of the crate. I was also surprised to see that our buffer_sv2 benches didn’t work with cargo bench unless extra flags were passed. Since there’s nothing in buffer_sv2 itself that depends on Criterion.

GitGab19: Alright, moving it to `dev-dependencies` makes sense.

But from my perspective this doesn't require a MAJOR bump, since the crate public APIs are not touched in any way.

Benches, as you said, are something external to the normal lib usage, so I would simply bump PATCH here.

sv2/buffer-sv2/README.md
GitGab19, line 1: Shouldn't we update the numbers here and on `BENCHES.md`?

sv2/buffer-sv2/benches/control_struct.rs
GitGab19: Did you run again the benches, comparing the results with the ones in the current `BENCHES.md` and `README.md`?

If they are different, we need to update those files.

sv2/buffer-sv2/benches/pool_benchmark.rs
```diff
@@ -265,12 +262,12 @@ fn criterion_benchmark(c: &mut Criterion) {
...
-    //});
-    //c.bench_function("with pool threaded test 2", |b| {
-    //    b.iter(|| with_pool_trreaded_test_2(&input[..]))
-    //});
+    c.bench_function("with pool threaded test", |b| {
+        b.iter(|| with_pool_trreaded_test_1(&input[..]))
+    });
+    c.bench_function("with pool threaded test 2", |b| {
+        b.iter(|| with_pool_trreaded_test_2(&input[..]))
+    });
```
GitGab19, line 270: Do you know why these lines were commented out?

bit-aloo: Updated the Benches.md with benchmark for these.

sv2/buffer-sv2/README.md
bit-aloo: Numbers were very similar to one in `BENCHES.md`,

sv2/buffer-sv2/benches/control_struct.rs
bit-aloo: They differed slightly but were largely comparable. I have trimmed the README to reflect the latest command changes and the removal of the fuzz target.

Shourya742:
> > This PR moves the buffer_sv2 fuzz targets to the top-level fuzz module.
> 
> [70276bd](https://github.com/stratum-mining/stratum/commit/70276bd760dd6ed1cf55ea95a24058edd9574e35) does this, but then [e815715](https://github.com/stratum-mining/stratum/commit/e815715962679ec068ab94000848a8609ba71078) deletes those files?

History should be ok now.

sv2/buffer-sv2/benches/pool_benchmark.rs
```diff
@@ -265,12 +262,12 @@ fn criterion_benchmark(c: &mut Criterion) {
...
 
     c.sample_size(500);
-    //c.bench_function("with pool threaded test", |b| {
-    //    b.iter(|| with_pool_trreaded_test_1(&input[..]))
-    //});
-    //c.bench_function("with pool threaded test 2", |b| {
-    //    b.iter(|| with_pool_trreaded_test_2(&input[..]))
-    //});
+    c.bench_function("with pool threaded test", |b| {
+        b.iter(|| with_pool_trreaded_test_1(&input[..]))
```
GitGab19:
```suggestion
        b.iter(|| with_pool_threaded_test_1(&input[..]))
```

sv2/buffer-sv2/benches/pool_benchmark.rs
```diff
@@ -265,12 +262,12 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+    });
+    c.bench_function("with pool threaded test 2", |b| {
+        b.iter(|| with_pool_trreaded_test_2(&input[..]))
```
GitGab19:
```suggestion
        b.iter(|| with_pool_threaded_test_2(&input[..]))
```

sv2/codec-sv2/Cargo.toml
GitGab19, line 1: We need to bump PATCH of  `codec_sv2` to have this new version published on release.

sv2/framing-sv2/Cargo.toml
GitGab19, line 1: We need to bump PATCH of  `framing_sv2` to have this new version published on release.

sv2/codec-sv2/Cargo.toml
bit-aloo: Nice catch, bumped PATCH.

sv2/framing-sv2/Cargo.toml
bit-aloo: Nice catch, bumped PATCH.

GitGab19 (approved):

PR #2037, opened 2025-12-19
Add noise bench

Shourya742:
Closes: #2041 

This PR adds noise benches, targeting the handshake and post-handshake scenario's.

### DIFF
--- a/sv2/noise-sv2/BENCHES.md
+++ b/sv2/noise-sv2/BENCHES.md
@@ -1,0 +1,145 @@
+# Benchmarks
+
+This document records the output of `cargo bench` for the `noise_sv2` crate.
+The benchmarks measure the performance characteristics of the Noise protocol
+implementation used by Stratum V2.
+
+The results are intended to:
+- Quantify the cost of the Noise handshake
+- Measure post-handshake encryption and decryption overhead
+- Serve as a reference point for future comparison and regression detection
+
+All measurements are environment-dependent and should not be interpreted as
+absolute performance guarantees.
+
+## Benchmark Output
+
+The following section contains the **raw output** of `cargo bench` captured
+during a single run.
+
+### Handshake Benchmarks
+
+```text
+Benchmarking handshake/step_0_initiator
+Benchmarking handshake/step_0_initiator: Warming up for 3.0000 s
+Benchmarking handshake/step_0_initiator: Collecting 100 samples in estimated 5.3787 s (56k iterations)
+Benchmarking handshake/step_0_initiator: Analyzing
+handshake/step_0_initiator
+                        time:   [18.560 µs 18.737 µs 18.981 µs]
+                        change: [-50.218% -49.023% -47.697%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 17 outliers among 100 measurements (17.00%)
+  11 (11.00%) high mild
+  6 (6.00%) high severe
+
+Benchmarking handshake/step_1_responder
+Benchmarking handshake/step_1_responder: Warming up for 3.0000 s
+Benchmarking handshake/step_1_responder: Collecting 100 samples in estimated 7.4286 s (15k iterations)
+Benchmarking handshake/step_1_responder: Analyzing
+handshake/step_1_responder
+                        time:   [177.39 µs 178.07 µs 178.93 µs]
+                        change: [-54.298% -53.548% -52.733%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 10 outliers among 100 measurements (10.00%)
+  1 (1.00%) low mild
+  2 (2.00%) high mild
+  7 (7.00%) high severe
+
+Benchmarking handshake/step_2_initiator
+Benchmarking handshake/step_2_initiator: Warming up for 3.0000 s
+Benchmarking handshake/step_2_initiator: Collecting 100 samples in estimated 6.1452 s (10k iterations)
+Benchmarking handshake/step_2_initiator: Analyzing
+handshake/step_2_initiator
+                        time:   [120.55 µs 120.91 µs 121.32 µs]
+                        change: [-44.862% -43.680% -42.592%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 10 outliers among 100 measurements (10.00%)
+  1 (1.00%) low mild
+  6 (6.00%) high mild
+  3 (3.00%) high severe
+
+Benchmarking handshake/handshake
+Benchmarking handshake/handshake: Warming up for 3.0000 s
+Benchmarking handshake/handshake: Collecting 100 samples in estimated 6.1256 s (10k iterations)
+Benchmarking handshake/handshake: Analyzing
+handshake/handshake     time:   [316.58 µs 317.41 µs 318.34 µs]
+                        change: [-43.167% -39.954% -36.598%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 7 outliers among 100 measurements (7.00%)
+  1 (1.00%) low mild
+  5 (5.00%) high mild
+  1 (1.00%) high severe
+````
+
+---
+
+### Transport Roundtrip Benchmarks
+
+```text
+Gnuplot not found, using plotters backend
+
+Benchmarking transport/roundtrip/64B
+Benchmarking transport/roundtrip/64B: Warming up for 3.0000 s
+Benchmarking transport/roundtrip/64B: Collecting 100 samples in estimated 5.0026 s (914k iterations)
+Benchmarking transport/roundtrip/64B: Analyzing
+transport/roundtrip/64B time:   [5.1966 µs 5.2037 µs 5.2126 µs]
+                        change: [-34.332% -31.654% -28.918%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 15 outliers among 100 measurements (15.00%)
+  3 (3.00%) high mild
+  12 (12.00%) high severe
+
+Benchmarking transport/roundtrip/256B
+Benchmarking transport/roundtrip/256B: Warming up for 3.0000 s
+Benchmarking transport/roundtrip/256B: Collecting 100 samples in estimated 5.0128 s (858k iterations)
+Benchmarking transport/roundtrip/256B: Analyzing
+transport/roundtrip/256B
+                        time:   [5.4840 µs 5.5004 µs 5.5190 µs]
+                        change: [-37.044% -34.403% -31.675%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 10 outliers among 100 measurements (10.00%)
+  4 (4.00%) high mild
+  6 (6.00%) high severe
+
+Benchmarking transport/roundtrip/1024B
+Benchmarking transport/roundtrip/1024B: Warming up for 3.0000 s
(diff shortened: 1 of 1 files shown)


sv2/noise-sv2/benches/handshake.rs
```diff
@@ -0,0 +1,43 @@
...
+use noise_sv2::{Initiator, Responder};
+use rand::{rngs::StdRng, SeedableRng};
+
+pub fn rng() -> StdRng {
+    StdRng::seed_from_u64(0xdead_beef)
+}
+
+pub fn payload(len: usize) -> Vec<u8> {
+    vec![0u8; len]
+}
```
lucasbalieiro: this seems to be unused

sv2/noise-sv2/benches/handshake.rs
```diff
@@ -0,0 +1,43 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use noise_sv2::{Initiator, Responder};
+use rand::{rngs::StdRng, SeedableRng};
+
+pub fn rng() -> StdRng {
+    StdRng::seed_from_u64(0xdead_beef)
+}
```
lucasbalieiro: we are doing this to add predictability to the bench measurement? Like trying to generate the same random number every time?

The docs of Rand states:

> Note: seeding does not imply reproducibility of results. For that you need to use a named RNG with a fixed algorithm (e.g. ChaCha12Rng not StdRng). See also [Reproducibility](https://rust-random.github.io/book/crate-reprod.html).

sv2/noise-sv2/benches/handshake.rs
```diff
@@ -0,0 +1,43 @@
...
+            },
+            |(mut initiator, mut responder)| {
+                let msg_0 = initiator.step_0().unwrap();
+                let (msg_2, _) = responder
+                    .step_1_with_now_rng(msg_0, 10, &mut rng())
+                    .unwrap();
+                let _ = initiator.step_2_with_now(msg_2, 10).unwrap();
+            },
+            criterion::BatchSize::SmallInput,
+        );
```
lucasbalieiro, line 73: I saw in the `criterion` docs that we have a bunch of `iter*` variants. 

The reason we are using the `iter_batched` here is because it requires this little setup closure at the beginning?

And if that assumption is right we could make the `rng` function return once during the setup and use it like this(?):
```suggestion
        b.iter_batched(
            || {
               let rng = rng();
                let initiator = Initiator::without_pk().unwrap();
                let responder = Responder::new(
                    secp256k1::Keypair::from_secret_key(
                        &secp256k1::Secp256k1::new(),
                        &secp256k1::SecretKey::from_slice(&[1u8; 32]).unwrap(),
                    ),
                    60,
                );
                (initiator, responder, rng)
            },
            |(mut initiator, mut responder, mut rng)| {
                let msg_0 = initiator.step_0().unwrap();
                let (msg_2, _) = responder
                    .step_1_with_now_rng(msg_0, 10, &mut rng)
                    .unwrap();
                let _ = initiator.step_2_with_now(msg_2, 10).unwrap();
            },
            criterion::BatchSize::SmallInput,
        );
```

sv2/noise-sv2/benches/handshake.rs
```diff
@@ -0,0 +1,43 @@
...
+                (initiator, responder)
+            },
+            |(mut initiator, mut responder)| {
+                let msg_0 = initiator.step_0().unwrap();
+                let (msg_2, _) = responder
+                    .step_1_with_now_rng(msg_0, 10, &mut rng())
+                    .unwrap();
+                let _ = initiator.step_2_with_now(msg_2, 10).unwrap();
+            },
+            criterion::BatchSize::SmallInput,
```
lucasbalieiro: the reason we are choosing `SmallInput` is because we can hold the input generated by the setup entirely in memory?

sv2/noise-sv2/benches/codec.rs
plebhash, line 1: I don't think the term `codec` is appropriate here.

We're not encoding/decoding data. We're encrypting/decrypting data.

sv2/noise-sv2/benches/codec.rs
```diff
@@ -0,0 +1,50 @@
...
+            return keypair;
+        }
+    }
+}
+
+pub fn payload(len: usize) -> Vec<u8> {
+    vec![0u8; len]
+}
+
+fn bench_codec_roundtrip(c: &mut Criterion) {
```
plebhash:
```suggestion
fn bench_encryption_roundtrip(c: &mut Criterion) {
```

plebhash: this is just one suggestion, but the intent is to replace the term `codec` everywhere... be it variables, functions or filenames

sv2/noise-sv2/benches/codec.rs
plebhash: I see this is a pattern that's already been established on the `noise_sv2` codebase, and there's multiple instances of the term `codec` being used inappropriately there.

Nevertheless, past mistakes shouldn't justify new ones.

It's definitely outside of the scope of this PR and arguably low prio, but IMHO we should eventually fix those other instances too.

sv2/noise-sv2/benches/codec.rs
```diff
@@ -0,0 +1,50 @@
...
+    let secp = Secp256k1::new();
+    let mut rng = thread_rng();
+
+    loop {
+        let (secret_key, _) = secp.generate_keypair(&mut rng);
+        let keypair = Keypair::from_secret_key(&secp, &secret_key);
+
+        if keypair.x_only_public_key().1 == Parity::Even {
+            return keypair;
+        }
```
plebhash: I would add a comment here explaining why this is needed

sv2/noise-sv2/benches/handshake.rs
plebhash: I think it would be good to add comments explaining every non-trivial thing we're doing.

If we just answer questions on this PR and move on, we're going to have the same doubts in the future.

sv2/noise-sv2/benches/handshake.rs
```diff
@@ -0,0 +1,43 @@
...
+            || {
+                let initiator = Initiator::without_pk().unwrap();
+                let responder = Responder::new(
+                    secp256k1::Keypair::from_secret_key(
+                        &secp256k1::Secp256k1::new(),
+                        &secp256k1::SecretKey::from_slice(&[1u8; 32]).unwrap(),
+                    ),
+                    60,
+                );
+                (initiator, responder)
```
plebhash: why aren't we doing the same check against `Parity::Even` that we did inside `generate_key` on the other file?

sv2/noise-sv2/benches/codec.rs
plebhash: it seems we already took note of this a while ago: https://github.com/stratum-mining/stratum/issues/1130#issuecomment-2356725083

personally speaking, I would avoid further propagating the `codec` terminology across the codebase

but I'll leave it up to @Shourya742 to decide how to proceed here, and worst case we can rename it whenever #1130 is addressed

resolving

sv2/noise-sv2/benches/handshake.rs
bit-aloo: moved to common module. Thanks for pointing out.

sv2/noise-sv2/benches/handshake.rs
bit-aloo: this makes sense, thanks

sv2/noise-sv2/benches/handshake.rs
bit-aloo: Yup

sv2/noise-sv2/benches/codec.rs
bit-aloo: Removed it, moved all common methods inside a common module

sv2/noise-sv2/benches/handshake.rs
bit-aloo: Removed it, moved all common methods inside a common module, And kept behavior consistent across benches.

sv2/noise-sv2/benches/codec.rs
plebhash: ok but still no comments?

plebhash: my point here was that it's not trivial why we're checking for the parity of the pubkey

sv2/noise-sv2/benches/handshake.rs
bit-aloo: Isn't this comment on diffferent versions of rand, and not specifically on a single version of rand.?

sv2/noise-sv2/benches/codec.rs
bit-aloo: Added: 6b7ec1d

sv2/noise-sv2/benches/handshake.rs
```diff
@@ -0,0 +1,79 @@
...
+            || {
+                let mut rng = rng();
+                let responder_key = generate_key_with_rng(&mut rng);
+                let initiator = Initiator::new(None);
+                let responder = Responder::new(responder_key, 60);
+                (initiator, responder, rng)
+            },
+            |(mut initiator, mut responder, mut rng)| {
+                let msg_0 = initiator.step_0().unwrap();
+                let (msg_2, _) = responder.step_1_with_now_rng(msg_0, 10, &mut rng).unwrap();
```
lucasbalieiro, line 69: these values for `now` and `cert_validity` seem reasonable for the benchmark. 

was there anything specific you had in mind when choosing this particular value?

bit-aloo: Nothing really, the actual motivation here was just to execute the encryption and decryption and see the latency incurred. That's it like how much time its taking for a step to execute and are we seeing quite few outliers or not.

sv2/noise-sv2/benches/common.rs
```diff
@@ -0,0 +1,22 @@
+use rand::{rngs::StdRng, SeedableRng};
+use secp256k1::{Keypair, Secp256k1};
+
+pub fn rng() -> StdRng {
+    // Fixed seed for deterministic benchmark runs.
+    StdRng::seed_from_u64(0xdead_beef)
+}
+
+// Generates BIP340 complaint keypair
```
plebhash:
```suggestion
// Generates BIP340 compliant keypair
```

sv2/noise-sv2/benches/handshake.rs
lucasbalieiro: > Isn't this comment on diffferent versions of rand, and not specifically on a single version of rand.?

I’m not sure I fully understand your question.

That said, the doc explains that if reproducibility is the goal, it’s best achieved by using a fixed algorithm with well-defined behavior. `StdRng` relies on the system’s implementation, which can vary across platforms and may therefore produce different results.

However, after a second pass over this PR, I think that for the purposes of this benchmark, `StdRng` is perfectly adequate. This is worth calling out for correctness, but it’s not a blocker for me.

refs:
https://docs.rs/rand/0.9.2/rand/rngs/struct.StdRng.html
https://rust-random.github.io/book/guide-seeding.html

plebhash:
it would be good to add some words about this to the `noise_sv2` crate's `README.md`

also a `BENCHES.md` with results

PR #2044, opened 2025-12-24
add Group Channel adaptations on `channels_sv2` + `mining_sv2`

### LINKED ISSUE #2014: Group Channel adaptations
introduce spec changes from https://github.com/stratum-mining/sv2-spec/pull/162

plebhash:
close #2014 

companion https://github.com/stratum-mining/sv2-apps/pull/156

---

important note: this PR is introducing a BREAKING SPEC CHANGE on `OpenExtendedMiningChannel.Success` message, which now has an extra field `group_channel_id`.

### DIFF
--- a/sv2/channels-sv2/src/server/group.rs
+++ b/sv2/channels-sv2/src/server/group.rs
@@ -157,9 +157,21 @@ where
         })
     }
 
-    /// Adds a channel ID to this group channel.
-    pub fn add_channel_id(&mut self, channel_id: u32) {
+    /// Adds a channel ID to this group channel. Also takes the `full_extranonce_size` of the channel to be added.
+    ///
+    /// Returns an error if the provided `full_extranonce_size` doesn't match the group channel's `full_extranonce_size`.
+    pub fn add_channel_id(
+        &mut self,
+        channel_id: u32,
+        full_extranonce_size: usize,
+    ) -> Result<(), GroupChannelError> {
         self.channel_ids.insert(channel_id);
+
+        if self.full_extranonce_size != full_extranonce_size {
+            return Err(GroupChannelError::FullExtranonceSizeMismatch);
+        }
+
+        Ok(())
     }
 
     /// Removes a channel ID from this group channel.
@@ -173,7 +185,12 @@ where
     }
 
     /// Set the full extranonce size for this group channel.
+    /// Also clears all channel IDs, as no channels can belong to the same group while having different `full_extranonce_size`s.
     pub fn set_full_extranonce_size(&mut self, full_extranonce_size: usize) {
+        if self.full_extranonce_size != full_extranonce_size {
+            self.channel_ids.clear();
+        }
+
         self.full_extranonce_size = full_extranonce_size;
     }
 
@@ -309,7 +326,7 @@ mod tests {
     use binary_sv2::Sv2Option;
     use bitcoin::{transaction::TxOut, Amount, ScriptBuf};
     use mining_sv2::NewExtendedMiningJob;
-    use std::convert::TryInto;
+    use std::{collections::HashSet, convert::TryInto};
     use template_distribution_sv2::{NewTemplate, SetNewPrevHash};
 
     const SATS_AVAILABLE_IN_TEMPLATE: u64 = 5000000000;
@@ -599,5 +616,66 @@ mod tests {
             .is_err());
 
         assert!(!group_channel.job_store.has_future_jobs());
+    }
+
+    #[test]
+    fn test_add_channel_id() {
+        let group_channel_id = 1;
+        let job_store = DefaultJobStore::new();
+        let full_extranonce_size = 32;
+        let mut group_channel = GroupChannel::new(
+            group_channel_id,
+            job_store,
+            full_extranonce_size,
+            None,
+            None,
+        )
+        .unwrap();
+
+        // add a first channel with the correct full extranonce size
+        group_channel
+            .add_channel_id(1, full_extranonce_size)
+            .unwrap();
+        assert_eq!(group_channel.get_channel_ids(), &HashSet::from([1]));
+        assert_eq!(
+            group_channel.get_full_extranonce_size(),
+            full_extranonce_size
+        );
+
+        // add a second channel with the correct full extranonce size
+        group_channel
+            .add_channel_id(2, full_extranonce_size)
+            .unwrap();
+        assert_eq!(group_channel.get_channel_ids(), &HashSet::from([1, 2]));
+        assert_eq!(
+            group_channel.get_full_extranonce_size(),
+            full_extranonce_size
+        );
+
+        // add a third channel with a different full extranonce size
+        // this should return an error
+        let new_full_extranonce_size = 24;
+        assert!(group_channel
+            .add_channel_id(3, new_full_extranonce_size)
+            .is_err());
+
+        // set the full extranonce size to a new value
+        group_channel.set_full_extranonce_size(new_full_extranonce_size);
+        assert_eq!(
+            group_channel.get_full_extranonce_size(),
+            new_full_extranonce_size
+        );
+        // all channel IDs should be cleared
+        assert_eq!(group_channel.get_channel_ids(), &HashSet::new());
+
+        // add a fourth channel with the correct full extranonce size
+        group_channel
+            .add_channel_id(4, new_full_extranonce_size)
+            .unwrap();
+        assert_eq!(group_channel.get_channel_ids(), &HashSet::from([4]));
+
+        // add a fifth channel with the old full extranonce size
+        // this should return an error because the full extranonce size is now set to 24
+        assert!(group_channel.add_channel_id(5, 32).is_err());
     }
 }
--- a/sv2/channels-sv2/src/server/error.rs
+++ b/sv2/channels-sv2/src/server/error.rs
@@ -18,6 +18,7 @@ pub enum ExtendedChannelError {
 
 #[derive(Debug)]
 pub enum GroupChannelError {
+    FullExtranonceSizeMismatch,
     ChainTipNotSet,
     TemplateIdNotFound,
     JobFactoryError(JobFactoryError),


sv2/channels-sv2/src/client/group.rs
```diff
@@ -2,8 +2,8 @@
 //!
 //! This module provides the [`GroupChannel`] struct, which acts as a mining client's
 //! abstraction over the state of a Sv2 group channel. It tracks group-level job state
-//! and associated standard channels, but delegates share validation and job lifecycle
-//! to standard channels.
+//! and associated standard and extended channels, but delegates share validation and job lifecycle
+//! to channels the channels themselves.
```
GitGab19:
```suggestion
//! to the channels themselves.
```

sv2/channels-sv2/src/client/group.rs
```diff
@@ -38,32 +38,32 @@ impl<'a> GroupChannel<'a> {
...
         }
     }
 
-    /// Adds a [`StandardChannel`](crate::client::standard::StandardChannel) to the group channel
+    /// Adds a channel to the group channel
     /// by referencing its `channel_id`.
-    pub fn add_standard_channel_id(&mut self, standard_channel_id: u32) {
-        self.standard_channel_ids.insert(standard_channel_id);
+    pub fn add_channel_id(&mut self, channel_id: u32) {
+        self.channel_ids.insert(channel_id);
```
GitGab19, line 65: Shouldn't we make some kind of check on the full extranonce size before adding the channel to the group here?

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -489,6 +489,38 @@ where
...
+        mut extended_job: ExtendedJob<'a>,
+    ) -> Result<(), ExtendedChannelError> {
+        // make sure the extranonce prefix is associated to the channel's extranonce prefix
+        extended_job.set_extranonce_prefix(self.extranonce_prefix.clone());
+
+        let template_id = match extended_job.get_origin() {
+            JobOrigin::NewTemplate(template) => template.template_id,
+            JobOrigin::SetCustomMiningJob(_) => {
+                return Err(ExtendedChannelError::InvalidJobOrigin);
+            }
```
GitGab19, line 513: Does it mean we are not going to allow clients to set custom jobs to a single group channel instead of many individual ones?

Maybe it's not related to that at all, but I wanted to clarify it.

sv2/channels-sv2/src/server/group.rs
```diff
@@ -3,9 +3,9 @@
 //! This module defines the [`GroupChannel`] struct, which provides an abstraction of a Stratum V2
 //! (SV2) group channel as maintained by a mining server.
 //!
-//! A group channel represents a logical grouping of standard channels, allowing multiple mining
+//! A group channel represents a logical grouping of standard and extended channels, allowing multiple mining
 //! entities to share jobs. It manages job distribution and activation for all
-//! associated standard channels, but delegates share validation and accounting to those standard
+//! associated channels, but delegates share validation and accounting to those channels.
 //! channels.
```
GitGab19:
```suggestion
```

sv2/channels-sv2/src/server/group.rs
```diff
@@ -157,28 +158,33 @@ where
         })
     }
 
-    /// Adds a standard channel ID to this group channel.
-    pub fn add_standard_channel_id(&mut self, standard_channel_id: u32) {
-        self.standard_channel_ids.insert(standard_channel_id);
+    /// Adds a channel ID to this group channel.
+    pub fn add_channel_id(&mut self, channel_id: u32) {
+        self.channel_ids.insert(channel_id);
```
GitGab19, line 168: Shouldn't we make some kind of check on the full extranonce size before adding the channel to the group here?

sv2/channels-sv2/src/server/group.rs
```diff
@@ -222,14 +228,13 @@ where
     ) -> Result<(), GroupChannelError> {
         match template.future_template {
             true => {
-                let new_job = self
+                let new_job: ExtendedJob<'_> = self
```
GitGab19: Why do we need this change?

plebhash: we don't, probably some autocomplete, removed

sv2/channels-sv2/src/server/extended.rs
plebhash: we can (and should) allow clients to set custom jobs to a single group channel... tbh I didn't consider this on https://github.com/stratum-mining/sv2-apps/pull/156, but it's actually a good idea, so I'll add a new commit there changing `handle_set_custom_mining_job`

this is a really good idea because now we unlock the use-case of a non-aggregating JDC that only sends one single pair of `DeclareMiningJob` + `SetCustomMiningJob`

---

but coming back to the original question, `on_group_channel_job` remains reserved for non-JD uses only... I'm adding the following comment to `on_new_template` and `on_group_channel_job` to make this idea explicitly clear:
```rust
/// Only meant to be used if REQUIRES_CUSTOM_WORK is not set on the connection this channel exists on.
```

if a `SetCustomMiningJob` arrives, and it's directed to a Group Channel ID, then the Pool application layer must iterate over all Extended Channels and call `on_set_custom_mining_job` on each one of them 

---

to be clear: a non-aggregating JDC is not going to be implemented on https://github.com/stratum-mining/sv2-apps/pull/156

only an adaptation to `handle_set_custom_mining_job` on Pool

sv2/channels-sv2/src/client/group.rs
plebhash: d7128e3adf5b273f58472a23c04cf637d92dea1e

sv2/channels-sv2/src/server/group.rs
plebhash: 6f0ed1437a0e032b8d9f00c7639f3d19852b03eb

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -438,6 +438,8 @@ where
     ///
     /// Only meant for usage on a Sv2 Pool Server or a Sv2 Job Declaration Client,
     /// but not on mining clients such as Mining Devices or Proxies.
+    ///
+    /// Only meant to be used if REQUIRES_CUSTOM_WORK is not set on the connection this channel exists on.
```
GitGab19:
```suggestion
    /// Only meant to be used if REQUIRES_CUSTOM_WORK is NOT set on the connection this channel exists on.
```

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -489,6 +491,40 @@ where
         Ok(())
     }
 
+    /// Used as an alternative to `on_new_template` when an extended job is meant to be broadcast to the group channel,
+    /// instead of multiple extended jobs to different extended channels.
+    ///
+    /// We use this method to update the channel state, so it can validate shares from the job that was broadcast to the group channel.
+    ///
+    /// Only meant to be used if REQUIRES_CUSTOM_WORK is not set on the connection this channel exists on.
```
GitGab19:
```suggestion
    /// Only meant to be used if REQUIRES_CUSTOM_WORK is NOT set on the connection this channel exists on.
```

sv2/channels-sv2/src/server/extended.rs
GitGab19: Sounds good!

I would just add a short phrase to indicate the usage of `on_set_custom_mining_job` in those cases.

sv2/channels-sv2/src/server/group.rs
```diff
@@ -157,28 +157,50 @@ where
...
     /// Returns the unique group channel ID for this group channel.
     pub fn get_group_channel_id(&self) -> u32 {
         self.group_channel_id
     }
 
+    /// Set the full extranonce size for this group channel.
+    /// Also clears all channel IDs, as no channels can belong to the same group while having different `full_extranonce_size`s.
+    pub fn set_full_extranonce_size(&mut self, full_extranonce_size: usize) {
+        if self.full_extranonce_size != full_extranonce_size {
+            self.channel_ids.clear();
```
GitGab19, line 191: When we do you expect to call this method?

Given the channels cleanup, I think timing of operations can become crucial here.

sv2/channels-sv2/src/server/extended.rs
GitGab19: Sounds good!

I would just add a short phrase to indicate the usage of `on_set_custom_mining_job` in those cases.

sv2/channels-sv2/src/server/group.rs
plebhash: this is only used on JDC

the group channel is instantiated during execution of every `Downstream::new`, which takes a `full_extranonce_size` parameter from the context of `ChannelManager::start_downstream_server`, [where everytime a new downstream client connects, we instantiate `full_extranonce_size` as:](https://github.com/plebhash/sv2-apps/blob/42837ee722a07a7186bf4a909719d4667515e6c5/miner-apps/jd-client/src/lib/channel_manager/mod.rs#L431)
```rust
let full_extranonce_size = self.channel_manager_data
    .super_safe_lock(|data| {
        data.upstream_channel.as_ref().map(|channel| channel
        get_full_extranonce_size())
        .unwrap_or(FULL_EXTRANONCE_SIZE) // Default to FULL_EXTRANONCE_SIZE if upstream channel is not present (solo mining mode)
});
```

if `data.upstream_channel` is `None`, we use `FULL_EXTRANONCE_SIZE`

but while we are processing `OpenExtendedMiningChannel.Success` for the channel established with the upstream server (which happens AFTER `Downstream::new` has already executed, and therefore the group channel already exists), we need to take the full extranonce size established there and adapt the group channel for that, [which is where `group_channel.set_full_extranonce_size` comes into play](https://github.com/plebhash/sv2-apps/blob/42837ee722a07a7186bf4a909719d4667515e6c5/miner-apps/jd-client/src/lib/channel_manager/upstream_message_handler.rs#L263
)

but this happens BEFORE we dispatch the pending open channel requests, so at this point, `handle_open_extended_mining_channel` or `handle_open_standard_mining_channel` execution hasn't started yet, and the group channel still doesn't contain any channel ids, and this cleanup operation has no effect

so we end up with a coherent full extranonce size for all channels and there's no timing issues

yet, the `channels_sv2` API remains with a safeguard against footguns, which is the goal of this cleanup operation

sv2/channels-sv2/src/server/extended.rs
plebhash: switched to

```
/// Only meant to be used if REQUIRES_CUSTOM_WORK is NOT set on the connection this channel exists on.
/// If this flag is set, on_set_custom_mining_job should be used instead.
```

sv2/channels-sv2/src/server/group.rs
GitGab19: I see, thanks for clarifying.

GitGab19 (approved):

Shourya742 (approved):

PR #2049, opened 2026-01-09
optimized extranonce prefix generation

coleFD:
PR https://github.com/stratum-mining/stratum/issues/1924

EDIT: This is just a proof of concept to show a better way to manage extranonce prefixes. Not looking to get this merged. I am preoccupied with other things atm

### DIFF
--- a/sv2/channels-sv2/src/extranonce_manager.rs
+++ b/sv2/channels-sv2/src/extranonce_manager.rs
@@ -1,0 +1,232 @@
+use bitvec::prelude::*;
+
+/// Represents a unique extranonce prefix for a client connection.
+#[derive(Eq, Hash, PartialEq, Clone, Debug)]
+pub struct ExtranoncePrefix {
+    bit_index: u16,
+    id: [u8; 4],
+}
+impl PartialOrd for ExtranoncePrefix {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+impl Ord for ExtranoncePrefix {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.bit_index.cmp(&other.bit_index)
+    }
+}
+
+/// Extranonce prefixes are generated using a bit vector where each prefix represents a bit
+/// in the vector. The prefix is then composed of the index of the assigned bit
+/// and the server ID. When a channel ends, the bit is cleared, allowing it to be reused.
+///
+/// # Generic:
+/// `NUM_BIT_CONTAINERS`: The number of u64 containers allocated for generating unique extranonce prefixes (ex. `ExtranoncePrefixManager<625>` results in `40k` bits for generating extranonce prefixes)
+/// The extranonce prefix component of the extranonce1 is capped at 2 bytes, so the maximum useful value for NUM_BIT_CONTAINERS is 1024 (65536 bits).
+pub struct ExtranoncePrefixManager<const NUM_BIT_CONTAINERS: usize> {
+    server_id: Option<u8>,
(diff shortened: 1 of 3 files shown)


sv2/channels-sv2/src/extranonce_manager.rs
```diff
@@ -0,0 +1,232 @@
+use bitvec::prelude::*;
+
+/// Represents a unique extranonce prefix for a client connection.
+#[derive(Eq, Hash, PartialEq, Clone, Debug)]
+pub struct ExtranoncePrefix {
+    bit_index: u16,
+    id: [u8; 4],
```
coleFD, line 7: id is hardcoded to 4 bytes because it gives 2 bytes (65,536 connections) which is more than enough, and another 2 bytes for the server_id. This could be changed to support 3 or 4 as well but not sure if it's worth the extra byte. I dont see a reason to support more than 4 bytes

plebhash: > it gives 2 bytes (65,536 connections)

are we assuming that each connection will only consume one single `extranonce_prefix`? what if the same connection has multiple channels?

also, what's the rationale behind the number 65,536?

coleFD: rather than connections, i should say channels or "unique ids". 

65,536 is a u16::max (2 bytes). It's fairly easy to make this size configurable to whatever size the user wants with generics.

jbesraa:
The code is exposing two structs with the same name and now an `ExtranoncePrefix/Manager` is added that is NOT using any of them. I think this is too confusing. If the intention is to keep the three of them(two of them is already confusing enough IMO) - a good documentation explaining when to use what is crucial.

```
stratum_apps::stratum_core::mining_sv2::Extranonce
stratum_apps::stratum_core::sv1_api::utils::Extranonce
  ```

sv2/channels-sv2/src/extranonce_manager.rs
coleFD: But an added benefit of restricting the total extranonce-prefix to 4 bytes is that it could be used as a channel id as well

GitGab19: @coleFD what do you think about using 1 byte for the `server_id` (which would support 256 different servers), and 3 bytes for the `extranonce_prefix(es)` (which would support up to 16.777.216 different channels)?

coleFD: that works too. You could allow this to be "configurable" with generics. Or maybe the optimal is somewhere in between like 12 bits for server ID (4096 servers) and 20 bits (1.05M) for the channel component. I think it's just preference. One thing to note is that as the channel component increases, finding a free bit gets slower since it's O(n). at 2 bytes, i benchmarked somewhere around 1-2 micros for finding available bits which is fine considering this is only called once per channel at origination. There may be a better way to start at the last freed bit index as opposed to `.skip_while()` though. Either way, performance should be considered because it is not uncommon to drop/connect a bunch of miners as fast as possible (ie. curtailment and non-agg proxies) and the connection time across thousands of clients add up.

coleFD: Another thing to note which may point out the benefit to extra serverID bits is it can be used to distinguish between clusters in different regions, like `1 byte serverID + 1 byte cluster ID + 2 bytes channel component`. In this case the api can recognize this as `2 bytes server ID + 2 bytes channel component` where it is up to the user to ensure the server ID is unique with the cluster ID. This is the route we are going

GitGab19:
Closing this in favor of https://github.com/stratum-mining/stratum/pull/2098

PR #2051, opened 2026-01-19
Add framing benches

Shourya742:
closes: #2059

### DIFF
--- a/sv2/binary-sv2/src/lib.rs
+++ b/sv2/binary-sv2/src/lib.rs
@@ -301,10 +301,9 @@ impl GetSize for Vec<u8> {
     }
 }
 
-// Only needed for implement encodable for Frame never called
 impl From<Vec<u8>> for EncodableField<'_> {
-    fn from(_v: Vec<u8>) -> Self {
-        unreachable!()
+    fn from(v: Vec<u8>) -> Self {
+        EncodableField::Struct(v.into_iter().map(Into::into).collect())
     }
 }
 
--- a/sv2/framing-sv2/benches/framing.rs
+++ b/sv2/framing-sv2/benches/framing.rs
@@ -70,7 +70,8 @@ fn bench_serialize(c: &mut Criterion) {
     let mut group = c.benchmark_group(format!("sv2frame::serialize_fresh::{BACKEND}"));
 
     for &size in PAYLOAD_SIZES {
-        let frame = Sv2Frame::<Vec<u8>, Slice>::from_bytes(frame_from_payload_size(size).into()).unwrap();
+        let frame =
+            Sv2Frame::<Vec<u8>, Slice>::from_bytes(frame_from_payload_size(size).into()).unwrap();
 
         let mut buf = vec![0u8; frame.encoded_length()];
 


sv2/framing-sv2/src/header.rs
```diff
@@ -120,14 +120,13 @@ impl Header {
     ///
     /// The calculated length includes the full payload length and any additional space required
     /// for the MACs.
+    #[allow(clippy::manual_div_ceil)]
```
lucasbalieiro, line 123: Why not  use `div_ceil`?

sv2/binary-sv2/src/lib.rs
```diff
@@ -303,8 +303,8 @@ impl GetSize for Vec<u8> {
 
 // Only needed for implement encodable for Frame never called
```
lucasbalieiro: We should update or remove this comment

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,149 @@
...
+#[cfg(feature = "with_buffer_pool")]
+const BACKEND: &str = "buffer_pool";
+
+#[cfg(not(feature = "with_buffer_pool"))]
+const BACKEND: &str = "vec";
+
+const PAYLOAD_SIZES: &[usize] = &[64, 1024, 16 * 1024, 60 * 1024];
+
+fn payload(size: usize) -> Vec<u8> {
+    assert!(size <= 0xFF_FFFF, "payload too large for U24");
```
lucasbalieiro: why this `assert!`? Don't we already have the sizes statically defined in `PAYLOAD_SIZES`?

sv2/framing-sv2/src/header.rs
bit-aloo: on my machine div_ceil was quite slow compared to current version.

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    group.finish();
+}
+
+fn bench_from_bytes(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!("sv2frame::from_bytes::{BACKEND}"));
+
+    for &size in PAYLOAD_SIZES {
+        group.bench_with_input(BenchmarkId::from_parameter(size), &size, |b, _| {
+            b.iter(|| Sv2Frame::<Vec<u8>, _>::from_bytes(black_box(payload(size))).unwrap())
```
lucasbalieiro: aren't we also computing the time for the `payload()` to create the `vec`? Shoudn't we compute the `payload()`outside of the `iter()`?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+#[derive(Serialize, Clone)]
+struct Tester {
+    _a: Vec<u8>,
+}
+
+fn tester(size: usize) -> Tester {
+    Tester {
+        _a: vec![2u8; size],
```
lucasbalieiro: shoudn't this be using `payload()`?

```suggestion
        _a: payload(size),
```

bit-aloo: We no longer need this struct, considering we are crafting the payload manually.

sv2/framing-sv2/benches/framing.rs
bit-aloo: almost comparable to clone, still moved out.

sv2/binary-sv2/src/lib.rs
```diff
@@ -301,10 +301,9 @@ impl GetSize for Vec<u8> {
     }
 }
 
-// Only needed for implement encodable for Frame never called
 impl From<Vec<u8>> for EncodableField<'_> {
-    fn from(_v: Vec<u8>) -> Self {
-        unreachable!()
+    fn from(v: Vec<u8>) -> Self {
```
lucasbalieiro, line 305: I remember in the dev call that you said you were aiming to bench the public apis, but why we never needed to reach this function before? via our public APIs?

bit-aloo: We can only benchmark the public APIs. I don’t think this is needed anymore earlier we had a tester struct with a Vec<u8> field, but our parser only knows how to deserialize into primitive types or spec-defined nested types, not arbitrary Rust types (other than primitives). Since we no longer use the tester struct, I woukd remove this entirely.

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        group.bench_with_input(BenchmarkId::from_parameter(size), &size, |b, _| {
+            b.iter(|| {
+                frame.clone().serialize(black_box(&mut buf)).unwrap();
+            })
+        });
+    }
+
+    group.finish();
+}
```
lucasbalieiro: I'm not following the semantic difference between this two benches: `bench_serialize_fresh` and `bench_serialize_fast`.


You are saying that the `buf` can cache something if the same message is being serialized again?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,137 @@
...
+
+fn bench_size_hint(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!("sv2frame::size_hint::{BACKEND}"));
+
+    for &size in PAYLOAD_SIZES {
+        let payload = payload(size);
+        group.bench_with_input(BenchmarkId::from_parameter(size), &size, |b, _| {
+            b.iter(|| black_box(Sv2Frame::<Vec<u8>, Slice>::size_hint(&payload)))
```
lucasbalieiro: Why `black_box` the entire `size_hint` instead of only the input `payload`?


```suggestion
            b.iter(|| Sv2Frame::<Vec<u8>, Slice>::size_hint(black_box(&payload)))
```

sv2/binary-sv2/src/lib.rs
lucasbalieiro: > We can only benchmark the public APIs.

oh yeah, thanks.

sv2/framing-sv2/benches/framing.rs
bit-aloo: Updated. Thanks.

sv2/framing-sv2/benches/framing.rs
bit-aloo: It should. I was making some changes to the framing logic, mainly around the header, to avoid serializing the initialized buffer. Looks like this slipped through during experimentation. Thanks for pointing it out, I’ve removed it.

sv2/framing-sv2/Cargo.toml
```diff
@@ -19,12 +19,18 @@ buffer_sv2 = { path = "../buffer-sv2", optional=true, version = "^3.0.0" }
 noise_sv2 = { path = "../noise-sv2", version = "^1.0.0" }
 
 [dev-dependencies]
-noise_sv2 = { path = "../noise-sv2", version = "^1.0.0" }
```
GitGab19, line 22: Why did you remove this?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+#[cfg(feature = "with_buffer_pool")]
+type Slice = buffer_sv2::Slice;
+
+#[cfg(feature = "with_buffer_pool")]
+const BACKEND: &str = "buffer_pool";
+
+#[cfg(not(feature = "with_buffer_pool"))]
+const BACKEND: &str = "vec";
+
+const PAYLOAD_SIZES: &[usize] = &[64, 1024, 16 * 1024, 60 * 1024];
```
GitGab19: What's the rationale behind these numbers?

sv2/framing-sv2/src/header.rs
```diff
@@ -120,14 +120,13 @@ impl Header {
...
-        let mut chunks = len / (SV2_FRAME_CHUNK_SIZE - AEAD_MAC_LEN);
-        if len % (SV2_FRAME_CHUNK_SIZE - AEAD_MAC_LEN) != 0 {
-            chunks += 1;
-        }
-        let mac_len = chunks * AEAD_MAC_LEN;
-        len + mac_len
+        let payload_per_chunk = SV2_FRAME_CHUNK_SIZE - AEAD_MAC_LEN;
+
+        let chunks = (len + payload_per_chunk - 1) / payload_per_chunk;
+        len + chunks * AEAD_MAC_LEN
```
GitGab19, line 129: Why this change?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+    ve.push(0);
+
+    ve.push((len & 0xFF) as u8);
+    ve.push(((len >> 8) & 0xFF) as u8);
+    ve.push(((len >> 16) & 0xFF) as u8);
+
+    ve.extend(std::iter::repeat(2u8).take(size));
+
+    ve
+}
```
GitGab19: Can you add some comments on some of these lines which help explaining what this function does?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+
+fn bench_from_message(c: &mut Criterion) {
```
GitGab19, line 55: Can you add a comment to explain what this bench does?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+                Sv2Frame::<Vec<u8>, Slice>::from_message(black_box(payload.clone()), 1, 0, false)
+                    .unwrap()
+            })
+        });
+    }
+
+    group.finish();
+}
+
+fn bench_serialize(c: &mut Criterion) {
```
GitGab19, line 69: Can you add a comment to explain what this bench does?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+            b.iter(|| {
+                frame.clone().serialize(&mut buf).unwrap();
+            })
+        });
+    }
+
+    group.finish();
+}
+
+fn bench_from_bytes(c: &mut Criterion) {
```
GitGab19, line 89: Can you add a comment to explain what this bench does?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+        let payload = payload(size);
+        group.bench_with_input(BenchmarkId::from_parameter(size), &size, |b, _| {
+            b.iter(|| Sv2Frame::<Vec<u8>, _>::from_bytes(black_box(payload.clone())).unwrap())
+        });
+    }
+
+    group.finish();
+}
+
+fn bench_size_hint(c: &mut Criterion) {
```
GitGab19, line 103: Can you add a comment to explain what this bench does?

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,117 @@
...
+        let payload = payload(size);
+        group.bench_with_input(BenchmarkId::from_parameter(size), &size, |b, _| {
+            b.iter(|| Sv2Frame::<Vec<u8>, Slice>::size_hint(black_box(&payload)))
+        });
+    }
+
+    group.finish();
+}
+
+fn bench_encrypted_len(c: &mut Criterion) {
```
GitGab19, line 117: Can you add a comment to explain what this bench does?

sv2/binary-sv2/src/lib.rs
```diff
@@ -301,10 +301,9 @@ impl GetSize for Vec<u8> {
...
+        EncodableField::Struct(v.into_iter().map(Into::into).collect())
```
GitGab19, line 306: Why do we need this change?

bit-aloo: Removed it, It was present because in an earlier iteration of benches we were using `Vec<>` inside the encodable and decodable macro.

sv2/framing-sv2/src/header.rs
bit-aloo: Previous implementation was kinda slow, as it was using `%`,  this one works better. We have a bench for this: `bench_encrypted_len`

sv2/framing-sv2/benches/framing.rs
bit-aloo: Different payload sizes, no particular rational as such.

sv2/framing-sv2/Cargo.toml
bit-aloo: We were not using it.

GitGab19: Mm, does that mean cargo-machete doesn't check dev-dependencies?

sv2/framing-sv2/benches/framing.rs
GitGab19: Shouldn't we also have the max possible size of a frame here?

GitGab19 (approved):
ACK, but I noticed we have CI failing.

Shourya742:
> @Shourya742, I ran again this benchmarks considering the recent changes and I'm consistently hitting a unreachable code:
> 
> https://github.com/stratum-mining/stratum/blob/c05fd9966b7a66ea8016b7463c859e8e35a5c77a/sv2/binary-sv2/src/lib.rs#L304-L309
> 
> My logs:
> 
> ```shell
> sv2frame::from_message::vec/61440                                                                             
>                         time:   [1.2964 µs 1.3120 µs 1.3288 µs]
>                         change: [-1.2614% -0.7263% -0.1038%] (p = 0.01 < 0.05)
>                         Change within noise threshold.
> Found 14 outliers among 100 measurements (14.00%)
>   8 (8.00%) low severe
>   1 (1.00%) low mild
>   3 (3.00%) high mild
>   2 (2.00%) high severe
> 
> Benchmarking sv2frame::serialize_fresh::vec/64: Warming up for 3.0000 sthread 'main' panicked at sv2/binary-sv2/src/lib.rs:307:9:
> internal error: entered unreachable code
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> error: bench failed, to rerun pass `--bench framing`
> ```
> 
> checkd at: [bf3440c](https://github.com/stratum-mining/stratum/commit/bf3440c1c2014e39244e52f56b90b062c83f66fc)
> 
> ```shell
> git log
> commit bf3440c1c2014e39244e52f56b90b062c83f66fc (HEAD -> 2051-add-benches-framing)
> Author: bit-aloo <sshourya17@gmail.com>
> Date:   Wed Feb 11 17:22:47 2026 +0530
> 
>     add framing benches
> ```

Should work now.

sv2/binary-sv2/src/lib.rs
lucasbalieiro: The comment got back

sv2/binary-sv2/src/lib.rs
GitGab19: I see you re-added this, why?

bit-aloo: We are serializating `Vec<u8>`, and that requires this in benches.

sv2/binary-sv2/src/lib.rs
bit-aloo: Thanks, removed

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,129 @@
...
+
+// Backend identifier for benchmark naming
+#[cfg(feature = "with_buffer_pool")]
+const BACKEND: &str = "buffer_pool";
+
+#[cfg(not(feature = "with_buffer_pool"))]
+const BACKEND: &str = "vec";
+
+// Test payload sizes
+const PAYLOAD_SIZES: &[usize] = &[64, 1024, 16 * 1024, 60 * 1024, 0xFFFFF9];
```
GitGab19: Why `0xFFFFF9`?

bit-aloo: Highest U24 value, which is type of payload length. Ideally it should 0xFFFFFF, but for some reason its 0xFFFFF9. I will look into this during property testing binary_sv2

```rust
impl TryFrom<u32> for U24 {
    type Error = Error;

    fn try_from(value: u32) -> Result<Self, Self::Error> {
        if value <= 16777215 {
            Ok(Self(value))
        } else {
            Err(Error::InvalidU24(value))
        }
    }
}
``

Comment (GitGab19) [reply sv2/framing-sv2/benches/framing.rs]: Why can't you write `0xFFFFFF`?

Comment (bit-aloo) [reply sv2/framing-sv2/benches/framing.rs]: check the try_from implementation, it gonna error out if the value is greater than 0xFFFFF9.

Comment (GitGab19) [reply sv2/framing-sv2/benches/framing.rs]: But `0xFFFFFF` = `16777215`, which is the value used in the `try_from` ?

Comment (bit-aloo) [reply sv2/framing-sv2/benches/framing.rs]: Pretty embarrassing in hindsight 😅 the from_message API expects only the payload, not the header + payload. That’s why using 0xFFFFF was erroring out: from_message expects only the payload, while I was passing the full frame. Thanks for pointing this out, it should be good now. I’ve also updated the numbers to be up to date.

Comment (GitGab19) [reply sv2/framing-sv2/benches/framing.rs]: Where are you passing only the payload now?

Comment (bit-aloo) [reply sv2/framing-sv2/benches/framing.rs]: We have two API's. `from_message`, and `from_bytes`. From message expects just the payload (like just the message), from_bytes expects header + payload.

File: sv2/framing-sv2/benches/framing.rs [hunk 20]
```diff
@@ -0,0 +1,138 @@
...
+const BACKEND: &str = "buffer_pool";
+
+#[cfg(not(feature = "with_buffer_pool"))]
+const BACKEND: &str = "vec";
+
+// Test payload sizes
+const PAYLOAD_SIZES: &[usize] = &[64, 1024, 16 * 1024, 60 * 1024, 0xFFFFFF];
+
+// Creates a mock SV2 frame: 3-byte header + 3-byte length + payload
+fn frame_payload(size: usize) -> Vec<u8> {
```
GitGab19: Why is this called `frame_payload` if it builds a full frame?

bit-aloo: Should I call it frame_with_payload? or vanilla frame? what I meant was just frame bytes.

GitGab19: The size param is internally used as payload size, right?

If so, I would call it `frame_from_payload_size`.

bit-aloo: Yup, I will change ti.

bit-aloo: done

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    group.finish();
+}
+
+// Benchmarks calculating frame size from partial data
+fn bench_size_hint(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!("sv2frame::size_hint::{BACKEND}"));
+
+    for &size in PAYLOAD_SIZES {
+        let payload = frame_from_payload_size(size);
```
GitGab19:
```suggestion
        let frame = frame_from_payload_size(size);
```

sv2/framing-sv2/benches/framing.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    group.finish();
+}
+
+// Benchmarks parsing frames from network bytes
+fn bench_from_bytes(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!("sv2frame::from_bytes::{BACKEND}"));
+
+    for &size in PAYLOAD_SIZES {
+        let payload = frame_from_payload_size(size);
```
GitGab19:
```suggestion
        let frame = frame_from_payload_size(size);
```

bit-aloo: Done

PR #2052, opened 2026-01-20
Make encoder backend agnostic

Shourya742:
This PR makes the encoder backend agnostic. Previously, the buffer backend was only available for Noise-enabled encoders, while non-encrypted encoders were restricted to a vanilla vector. This change extends the feature-gated buffer support to non-encrypted encoders as well, bringing the encoder API in line with the decoder and making it symmetric.

### DIFF
--- a/sv2/codec-sv2/src/encoder.rs
+++ b/sv2/codec-sv2/src/encoder.rs
@@ -62,13 +62,13 @@ type Buffer = BufferPool<BufferFromSystemMemory>;
 
 /// Standard Sv2 encoder with Noise protocol support.
 ///
-/// Used for decoding generic message types (`T`) encoded in Sv2 frames.
+/// Used for encoding generic message types (`T`) in Sv2 frames.
 #[cfg(feature = "noise_sv2")]
 pub type NoiseEncoder<T> = WithNoise<Buffer, T>;
 
 /// Standard Sv2 encoder without Noise protocol support.
 ///
-/// Used for decoding generic message types (`T`) encoded in Sv2 frames.
+/// Used for encoding generic message types (`T`) in Sv2 frames.
 pub type Encoder<T> = WithoutNoise<Buffer, T>;
 
 /// Encoder for Sv2 frames with Noise protocol encryption.


stratum-core/Cargo.toml
GitGab19: We need to bump this to `0.2.1` to publish it in our next release.

sv2/codec-sv2/src/encoder.rs
```diff
@@ -57,48 +57,40 @@ use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory, BufferPool};
...
-//
-// When the `with_buffer_pool` feature is disabled, [`Slice`] defaults to a `Vec<u8>`, which serves
-// as a dynamically allocated array to hold the serialized bytes of Sv2 frames. This provides
-// flexibility in managing the encoded data during transmission or further processing, though it
-// may not offer the same memory efficiency as the pool-allocated version [`BufferPool`].
-#[cfg(not(feature = "with_buffer_pool"))]
-type Slice = Vec<u8>;
+/// Standard Sv2 encoder with Noise protocol support.
+///
+/// Used for decoding generic message types (`T`) encoded in Sv2 frames.
```
GitGab19:
```suggestion
/// Used for encoding generic message types (`T`)  in Sv2 frames.
```

sv2/codec-sv2/src/encoder.rs
```diff
@@ -57,48 +57,40 @@ use buffer_sv2::{Buffer as IsBuffer, BufferFromSystemMemory, BufferPool};
...
-//
-// When the `with_buffer_pool` feature is enabled, [`Slice`] defaults to a `buffer_sv2::Slice`,
-// which serves as a slice of the `Buffer` that stores the encoded data. It holds the frame's
-// serialized bytes temporarily, ensuring the data is ready for transmission or encryption,
-// depending on whether Noise protocol support is enabled.
-#[cfg(feature = "with_buffer_pool")]
-type Slice = buffer_sv2::Slice;
+/// Standard Sv2 encoder without Noise protocol support.
+///
+/// Used for decoding generic message types (`T`) encoded in Sv2 frames.
```
GitGab19:
```suggestion
/// Used for encoding generic message types (`T`) in Sv2 frames.
```

stratum-core/Cargo.toml
bit-aloo: Its already bumped. I just rebased on current tip.

Issue #2072, opened 2026-02-03
Add cargo-machete to CI


Shourya742:
Add cargo-machete to the CI workflow to detect unused dependencies automatically. This helps keep the dependency graph clean and reduces build overhead.

PR #2073, opened 2026-02-03
Add machete to CI

Shourya742:
closes: #2072

This PR also removes unused dependency across the stratum and unify clippy, fmt and machete under ci.yaml instead of individual workflow files. We also patch crates where unused dependencies were present.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,3 +1,5 @@
+name: CI
+
 on:
   pull_request:
     branches:
@@ -18,6 +20,8 @@ jobs:
           components: clippy
 
       # NOTE: A repo-local `rust-toolchain.toml` is still respected by rustup.
+      # `dtolnay/rust-toolchain` installs the requested toolchain but does not
+      # automatically override it.
       - name: Override toolchain
         run: rustup override set ${{ steps.toolchain.outputs.name }}
 
--- a/stratum-core/Cargo.toml
+++ b/stratum-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "stratum-core"
-version = "0.2.0"
+version = "0.2.1"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"
--- a/sv2/handlers-sv2/Cargo.toml
+++ b/sv2/handlers-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "handlers_sv2"
-version = "0.2.1"
+version = "0.2.2"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"
--- a/sv2/parsers-sv2/Cargo.toml
+++ b/sv2/parsers-sv2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parsers_sv2"
-version = "0.2.1"
+version = "0.2.2"
 authors = ["The Stratum V2 Developers"]
 edition = "2021"
 readme = "README.md"


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust (1.89) + Clippy
+        uses: dtolnay/rust-toolchain@1.89
+        id: toolchain
+        with:
+          components: clippy
+
+      # NOTE: A repo-local `rust-toolchain.toml` is still respected by rustup.
```
GitGab19, line 22:
```suggestion
      # NOTE: A repo-local `rust-toolchain.toml` is still respected by rustup.
      # `dtolnay/rust-toolchain` installs the requested toolchain but does not
      # automatically override it.
```

sv2/subprotocols/template-distribution/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "template_distribution_sv2"
-version = "4.0.2"
+version = "5.0.0"
```
GitGab19, line 3: Why did you bump this?

sv2/subprotocols/common-messages/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "common_messages_sv2"
-version = "6.0.2"
+version = "7.0.0"
```
GitGab19, line 3: Why did you bump this?

sv2/subprotocols/template-distribution/Cargo.toml
bit-aloo: quickcheck macro was an optional dependency.

.github/workflows/ci.yaml
GitGab19, line 1: Can you give a name to this workflow here at the beginning?

sv2/subprotocols/template-distribution/Cargo.toml
GitGab19: Was `semver-check` complaining about this?

bit-aloo: yes

GitGab19: But how is it possible, given that the macro was not used at all in  `template_distribution_sv2` and `common_messages_sv2`?

bit-aloo: The very same reason why criterion was not used at all in buffer but we still required a major bump.

GitGab19: I asked clarifications to AI, and here the response:

---
In Cargo’s model it **is** a feature, just an *implicit* one.

### Why `quickcheck_macros` counts as a feature

You have:

```toml
[dependencies]
quickcheck = { version = "1.0.3", optional = true }
quickcheck_macros = { version = "1", optional = true }

[features]
prop_test = ["quickcheck"]
```

In Cargo:

- **Every `optional = true` dependency automatically creates a feature with the same name.**  
  So `quickcheck_macros = { optional = true, ... }` defines a feature named `quickcheck_macros`, even though it’s not listed under `[features]`.

- Downstream crates can write:

  ```toml
  [dependencies]
  common_messages_sv2 = { version = "6", features = ["quickcheck_macros"] }
  ```

If you remove that optional dependency, you also remove the **implicit feature** `quickcheck_macros`, which is why `cargo-semver-checks` reports:

> feature `quickcheck_macros` in the package’s Cargo.toml

and requires a **major version bump** if you drop it.

---

Is this analysis correct?

bit-aloo: looks correct

stratum-core/Cargo.toml
GitGab19, line 1: We need to bump PATCH of `stratum-core` here to publish this new version on release.

sv2/handlers-sv2/Cargo.toml
GitGab19, line 1: We need to bump PATCH of `handlers_sv2` here.

sv2/parsers-sv2/Cargo.toml
GitGab19, line 1: We need to bump PATCH of `parsers_sv2` here.

stratum-core/Cargo.toml
bit-aloo: done

sv2/handlers-sv2/Cargo.toml
bit-aloo: done

PR #2074, opened 2026-02-03
Add fuzz targets for sv2 datatypes

### LINKED ISSUE #2062: add fuzz coverage to `binary_sv2`
we should add fuzz coverage to the `binary_sv2`

lucasbalieiro:
closes #2062.

this PR introduces a single fuzz target that covers all Sv2 datatypes and removes the outdated per-datatype targets.

because the number of datatypes is large, maintaining one fuzz target per type would create significant duplication and overhead. Instead, this follows the same consolidated fuzzing approach we adopted for Sv2 messages in #2028.

While broad fuzz targets are generally discouraged, the same justification applies here: Sv2 datatypes share a common construction pattern, so grouping them provides good coverage without meaningful loss of effectiveness. See the rationale discussed in #2028: https://github.com/stratum-mining/stratum/pull/2028#issuecomment-3666989904

this is also interleaving the message target that were previously created

### DIFF
--- a/fuzz/fuzz_targets/end_to_end_serialization_for_common_messages.rs
+++ b/fuzz/fuzz_targets/end_to_end_serialization_for_common_messages.rs
@@ -1,15 +1,36 @@
 #![no_main]
 
 mod common;
+use arbitrary::Arbitrary;
 use binary_sv2::{Deserialize, GetSize, Serialize};
 use common_messages_sv2::*;
-
 use libfuzzer_sys::fuzz_target;
 
-fuzz_target!(|data: Vec<u8>| {
-    test_roundtrip!(SetupConnection, data);
-    test_roundtrip!(SetupConnectionError, data);
-    test_roundtrip!(SetupConnectionSuccess, data);
-    test_roundtrip!(Reconnect, data);
-    test_roundtrip!(ChannelEndpointChanged, data);
+#[derive(Arbitrary, Debug)]
+enum FuzzInput {
+    SetupConnection(Vec<u8>),
+    SetupConnectionError(Vec<u8>),
+    SetupConnectionSuccess(Vec<u8>),
+    Reconnect(Vec<u8>),
+    ChannelEndpointChanged(Vec<u8>),
+}
+
+fuzz_target!(|input: FuzzInput| {
+    match input {
+        FuzzInput::SetupConnection(data) => {
+            test_roundtrip!(SetupConnection, data)
+        }
+        FuzzInput::SetupConnectionError(data) => {
+            test_roundtrip!(SetupConnectionError, data)
+        }
+        FuzzInput::SetupConnectionSuccess(data) => {
+            test_roundtrip!(SetupConnectionSuccess, data)
+        }
+        FuzzInput::Reconnect(data) => {
+            test_roundtrip!(Reconnect, data)
+        }
+        FuzzInput::ChannelEndpointChanged(data) => {
+            test_roundtrip!(ChannelEndpointChanged, data)
+        }
+    }
 });
--- a/fuzz/fuzz_targets/end_to_end_serialization_for_job_declaration_messages.rs
+++ b/fuzz/fuzz_targets/end_to_end_serialization_for_job_declaration_messages.rs
@@ -1,18 +1,49 @@
 #![no_main]
 
 mod common;
+use arbitrary::Arbitrary;
 use binary_sv2::{Deserialize, GetSize, Serialize};
 use job_declaration_sv2::*;
-
 use libfuzzer_sys::fuzz_target;
 
-fuzz_target!(|data: Vec<u8>| {
-    test_roundtrip!(AllocateMiningJobToken, data);
-    test_roundtrip!(AllocateMiningJobTokenSuccess, data);
-    test_roundtrip!(DeclareMiningJob, data);
-    test_roundtrip!(DeclareMiningJobSuccess, data);
-    test_roundtrip!(DeclareMiningJobError, data);
-    test_roundtrip!(ProvideMissingTransactions, data);
-    test_roundtrip!(ProvideMissingTransactionsSuccess, data);
-    test_roundtrip!(PushSolution, data);
+#[derive(Arbitrary, Debug)]
+enum FuzzInput {
+    AllocateMiningJobToken(Vec<u8>),
+    AllocateMiningJobTokenSuccess(Vec<u8>),
+    DeclareMiningJob(Vec<u8>),
+    DeclareMiningJobSuccess(Vec<u8>),
+    DeclareMiningJobError(Vec<u8>),
+    ProvideMissingTransactions(Vec<u8>),
+    ProvideMissingTransactionsSuccess(Vec<u8>),
+    PushSolution(Vec<u8>),
+}
+
+fuzz_target!(|input: FuzzInput| {
+    match input {
+        FuzzInput::AllocateMiningJobToken(data) => {
+            test_roundtrip!(AllocateMiningJobToken, data)
+        }
+        FuzzInput::AllocateMiningJobTokenSuccess(data) => {
+            test_roundtrip!(AllocateMiningJobTokenSuccess, data)
+        }
+        FuzzInput::DeclareMiningJob(data) => {
+            test_roundtrip!(DeclareMiningJob, data)
+        }
+        FuzzInput::DeclareMiningJobSuccess(data) => {
+            test_roundtrip!(DeclareMiningJobSuccess, data)
+        }
+        FuzzInput::DeclareMiningJobError(data) => {
+            test_roundtrip!(DeclareMiningJobError, data)
+        }
+        FuzzInput::ProvideMissingTransactions(data) => {
+            test_roundtrip!(ProvideMissingTransactions, data)
+        }
+        FuzzInput::ProvideMissingTransactionsSuccess(data) => {
+            test_roundtrip!(ProvideMissingTransactionsSuccess, data)
+        }
+        FuzzInput::PushSolution(data) => {
+            test_roundtrip!(PushSolution, data)
+        }
+    }
 });
+
--- a/fuzz/fuzz_targets/end_to_end_serialization_for_mining_messages.rs
+++ b/fuzz/fuzz_targets/end_to_end_serialization_for_mining_messages.rs
@@ -1,31 +1,102 @@
 #![no_main]
 
 mod common;
+
+use arbitrary::Arbitrary;
 use binary_sv2::{Deserialize, GetSize, Serialize};
+use libfuzzer_sys::fuzz_target;
 use mining_sv2::*;
 
-use libfuzzer_sys::fuzz_target;
+#[derive(Arbitrary, Debug)]
+enum FuzzInput {
+    CloseChannel(Vec<u8>),
+    NewMiningJob(Vec<u8>),
+    NewExtendedMiningJob(Vec<u8>),
+    OpenStandardMiningChannel(Vec<u8>),
+    OpenStandardMiningChannelSuccess(Vec<u8>),
+    OpenExtendedMiningChannel(Vec<u8>),
+    OpenExtendedMiningChannelSuccess(Vec<u8>),
+    OpenMiningChannelError(Vec<u8>),
+    SetCustomMiningJob(Vec<u8>),
+    SetCustomMiningJobSuccess(Vec<u8>),
+    SetCustomMiningJobError(Vec<u8>),
+    SetExtranoncePrefix(Vec<u8>),
+    SetGroupChannel(Vec<u8>),
+    SetNewPrevHash(Vec<u8>),
+    SetTarget(Vec<u8>),
+    SubmitSharesStandard(Vec<u8>),
+    SubmitSharesExtended(Vec<u8>),
+    SubmitSharesSuccess(Vec<u8>),
+    SubmitSharesError(Vec<u8>),
+    UpdateChannel(Vec<u8>),
+    UpdateChannelError(Vec<u8>),
+}
 
-fuzz_target!(|data: Vec<u8>| {
-    test_roundtrip!(CloseChannel, data);
-    test_roundtrip!(NewMiningJob, data);
-    test_roundtrip!(NewExtendedMiningJob, data);
-    test_roundtrip!(OpenStandardMiningChannel, data);
-    test_roundtrip!(OpenStandardMiningChannelSuccess, data);
-    test_roundtrip!(OpenExtendedMiningChannel, data);
-    test_roundtrip!(OpenExtendedMiningChannelSuccess, data);
-    test_roundtrip!(OpenMiningChannelError, data);
-    test_roundtrip!(SetCustomMiningJob, data);
-    test_roundtrip!(SetCustomMiningJobSuccess, data);
-    test_roundtrip!(SetCustomMiningJobError, data);
-    test_roundtrip!(SetExtranoncePrefix, data);
-    test_roundtrip!(SetGroupChannel, data);
-    test_roundtrip!(SetNewPrevHash, data);
-    test_roundtrip!(SetTarget, data);
-    test_roundtrip!(SubmitSharesStandard, data);
-    test_roundtrip!(SubmitSharesExtended, data);
-    test_roundtrip!(SubmitSharesSuccess, data);
-    test_roundtrip!(SubmitSharesError, data);
-    test_roundtrip!(UpdateChannel, data);
-    test_roundtrip!(UpdateChannelError, data);
+fuzz_target!(|input: FuzzInput| {
+    match input {
+        FuzzInput::CloseChannel(data) => {
+            test_roundtrip!(CloseChannel, data)
+        }
+        FuzzInput::NewMiningJob(data) => {
+            test_roundtrip!(NewMiningJob, data)
+        }
+        FuzzInput::NewExtendedMiningJob(data) => {
+            test_roundtrip!(NewExtendedMiningJob, data)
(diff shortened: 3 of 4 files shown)


lucasbalieiro:
@erickcestari, thanks for this insight.

I also adapted the other targets that we have for the messages with this strategy.

fuzz/fuzz_targets/common.rs
```diff
@@ -57,3 +57,109 @@ macro_rules! test_roundtrip {
...
+        // Only run the roundtrip checks if parsing succeeds. Invalid inputs are ignored,
+        // because this macro validates stability of valid encodings, not rejection behavior.
+        if let Ok(parsed) = bool::from_bytes(&mut input) {
+            // Allocate exactly the number of bytes required by the parsed value.
+            // This ensures we test the canonical serialized size.
+            let mut encoded = vec![0u8; parsed.get_size()];
+
+            // A successful parse must always be serializable.
+            parsed.to_bytes(&mut encoded)
+                .expect("Bool encoding failed after a successfull pase");
```
bit-aloo: nit:
```suggestion
                .expect("Bool encoding failed after a successful parse");
```

fuzz/fuzz_targets/end_to_end_serialization_for_common_messages.rs
```diff
@@ -1,15 +1,27 @@
...
+fuzz_target!(|input: FuzzInput| {
+    let data = input.data;
+
+    match input.selector {
+        0 => test_roundtrip!(SetupConnection, data),
+        1 => test_roundtrip!(SetupConnectionError, data),
+        2 => test_roundtrip!(SetupConnectionSuccess, data),
+        3 => test_roundtrip!(Reconnect, data),
+        4 => test_roundtrip!(ChannelEndpointChanged, data),
+        _ => {}
```
bit-aloo: Lets add a NUM_SIZE here, so we don't let go of a lot of fuzz inputs, something like this:

```suggestion
    let data = input.data;
    const NUM_SIZE:u8 = 5;

    match input.selector%NUM_SIZE {
        0 => test_roundtrip!(SetupConnection, data),
        1 => test_roundtrip!(SetupConnectionError, data),
        2 => test_roundtrip!(SetupConnectionSuccess, data),
        3 => test_roundtrip!(Reconnect, data),
        4 => test_roundtrip!(ChannelEndpointChanged, data),
        _ => {}
```

fuzz/fuzz_targets/end_to_end_serialization_for_datatypes.rs
```diff
@@ -0,0 +1,82 @@
...
+struct FuzzInput {
+    selector: u8,
+    data: Vec<u8>,
+}
+
+fuzz_target!(|input: FuzzInput| {
+    let data = input.data;
+
+    match input.selector {
+        0 => test_datatype_roundtrip!(PubKey, data),
```
bit-aloo: Same for this: 
```suggestion
    let data = input.data;
    const NUM_SIZE: u8 = 52;

    match input.selector%NUM_SIZE {
        0 => test_datatype_roundtrip!(PubKey, data),
```

fuzz/fuzz_targets/end_to_end_serialization_for_job_declaration_messages.rs
```diff
@@ -1,18 +1,30 @@
...
+    match input.selector {
+        0 => test_roundtrip!(AllocateMiningJobToken, data),
+        1 => test_roundtrip!(AllocateMiningJobTokenSuccess, data),
+        2 => test_roundtrip!(DeclareMiningJob, data),
+        3 => test_roundtrip!(DeclareMiningJobSuccess, data),
+        4 => test_roundtrip!(DeclareMiningJobError, data),
+        5 => test_roundtrip!(ProvideMissingTransactions, data),
+        6 => test_roundtrip!(ProvideMissingTransactionsSuccess, data),
+        7 => test_roundtrip!(PushSolution, data),
+        _ => {}
```
bit-aloo: Same for this:

```suggestion
    let data = input.data;
    const NUM_SIZE: u8 = 8;
    match input.selector%NUM_SIZE {
        0 => test_roundtrip!(AllocateMiningJobToken, data),
        1 => test_roundtrip!(AllocateMiningJobTokenSuccess, data),
        2 => test_roundtrip!(DeclareMiningJob, data),
        3 => test_roundtrip!(DeclareMiningJobSuccess, data),
        4 => test_roundtrip!(DeclareMiningJobError, data),
        5 => test_roundtrip!(ProvideMissingTransactions, data),
        6 => test_roundtrip!(ProvideMissingTransactionsSuccess, data),
        7 => test_roundtrip!(PushSolution, data),
        _ => {}
```

fuzz/fuzz_targets/end_to_end_serialization_for_mining_messages.rs
```diff
@@ -1,31 +1,43 @@
...
+        12 => test_roundtrip!(SetGroupChannel, data),
+        13 => test_roundtrip!(SetNewPrevHash, data),
+        14 => test_roundtrip!(SetTarget, data),
+        15 => test_roundtrip!(SubmitSharesStandard, data),
+        16 => test_roundtrip!(SubmitSharesExtended, data),
+        17 => test_roundtrip!(SubmitSharesSuccess, data),
+        18 => test_roundtrip!(SubmitSharesError, data),
+        19 => test_roundtrip!(UpdateChannel, data),
+        20 => test_roundtrip!(UpdateChannelError, data),
+        _ => {}
```
bit-aloo: Same for this:


```suggestion
    let data = input.data;
    const NUM_SIZE: u8 = 21;
    match input.selector%NUM_SIZE {
        0 => test_roundtrip!(CloseChannel, data),
        1 => test_roundtrip!(NewMiningJob, data),
        2 => test_roundtrip!(NewExtendedMiningJob, data),
        3 => test_roundtrip!(OpenStandardMiningChannel, data),
        4 => test_roundtrip!(OpenStandardMiningChannelSuccess, data),
        5 => test_roundtrip!(OpenExtendedMiningChannel, data),
        6 => test_roundtrip!(OpenExtendedMiningChannelSuccess, data),
        7 => test_roundtrip!(OpenMiningChannelError, data),
        8 => test_roundtrip!(SetCustomMiningJob, data),
        9 => test_roundtrip!(SetCustomMiningJobSuccess, data),
        10 => test_roundtrip!(SetCustomMiningJobError, data),
        11 => test_roundtrip!(SetExtranoncePrefix, data),
        12 => test_roundtrip!(SetGroupChannel, data),
        13 => test_roundtrip!(SetNewPrevHash, data),
        14 => test_roundtrip!(SetTarget, data),
        15 => test_roundtrip!(SubmitSharesStandard, data),
        16 => test_roundtrip!(SubmitSharesExtended, data),
        17 => test_roundtrip!(SubmitSharesSuccess, data),
        18 => test_roundtrip!(SubmitSharesError, data),
        19 => test_roundtrip!(UpdateChannel, data),
        20 => test_roundtrip!(UpdateChannelError, data),
        _ => {}
```

fuzz/fuzz_targets/end_to_end_serialization_for_template_distribution_messages.rs
```diff
@@ -1,17 +1,29 @@
...
+
+    match input.selector {
+        0 => test_roundtrip!(CoinbaseOutputConstraints, data),
+        1 => test_roundtrip!(NewTemplate, data),
+        2 => test_roundtrip!(RequestTransactionData, data),
+        3 => test_roundtrip!(RequestTransactionDataSuccess, data),
+        4 => test_roundtrip!(RequestTransactionDataError, data),
+        5 => test_roundtrip!(SetNewPrevHash, data),
+        6 => test_roundtrip!(SubmitSolution, data),
+        _ => {}
```
bit-aloo: Same for this: 

```suggestion
    let data = input.data;
    const NUM_SIZE: u8 = 7;
    match input.selector%NUM_SIZE {
        0 => test_roundtrip!(CoinbaseOutputConstraints, data),
        1 => test_roundtrip!(NewTemplate, data),
        2 => test_roundtrip!(RequestTransactionData, data),
        3 => test_roundtrip!(RequestTransactionDataSuccess, data),
        4 => test_roundtrip!(RequestTransactionDataError, data),
        5 => test_roundtrip!(SetNewPrevHash, data),
        6 => test_roundtrip!(SubmitSolution, data),
        _ => {}
```

fuzz/fuzz_targets/end_to_end_serialization_for_datatypes.rs
```diff
@@ -0,0 +1,82 @@
...
+        15 => test_datatype_roundtrip!(u64, data),
+
+        16 => test_datatype_roundtrip!(Seq0255<u8>, data),
+        17 => test_datatype_roundtrip!(Seq0255<u16>, data),
+        18 => test_datatype_roundtrip!(Seq0255<u32>, data),
+        19 => test_datatype_roundtrip!(Seq0255<u64>, data),
+
+        20 => test_datatype_roundtrip!(Seq0255<PubKey>, data),
+        21 => test_datatype_roundtrip!(Seq0255<Signature>, data),
+        22 => test_datatype_roundtrip!(Seq0255<Str0255>, data),
```
bit-aloo: We can add variant with B032, and U32AsRef

bit-aloo: bool, can also be added.

fuzz/Cargo.toml
```diff
@@ -78,3 +60,10 @@ name = "end_to_end_serialization_for_template_distribution_messages"
 path = "fuzz_targets/end_to_end_serialization_for_template_distribution_messages.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "end_to_end_serialization_for_datatypes"
+path = "fuzz_targets/end_to_end_serialization_for_datatypes.rs"
+test = false
+doc = false
+bench = false
```
bit-aloo: why `bench = false`?

fuzz/fuzz_targets/common.rs
```diff
@@ -57,3 +57,109 @@ macro_rules! test_roundtrip {
...
+        // Only test successful parses; this macro checks roundtrip invariants.
+        if let Ok(parsed) = <$datatype>::from_bytes(&mut input) {
+            // Allocate exactly the canonical serialized size.
+            let mut encoded = vec![0u8; parsed.get_size()];
+
+            // A parsed value must always serialize successfully.
+            parsed
+                .clone()
+                .to_bytes(&mut encoded)
+                .expect("{stringify!($datatype)}: Encoding failed after a successful parse");
```
bit-aloo:
```suggestion
                .expect(concat!(stringify!($datatype), ": Encoding failed after a successful parse"))
```

fuzz/fuzz_targets/common.rs
```diff
@@ -57,3 +57,109 @@ macro_rules! test_roundtrip {
...
+
+            // Serialized bytes must be parseable again into the same datatype.
+            let reparsed = <$datatype>::from_bytes(&mut encoded)
+                .expect("{stringify!($datatype)}: The bytes generated from a valid datatype should be parseable");
```
bit-aloo:
```suggestion
                .expect(concat!(stringify!($datatype), ": The bytes generated from a valid datatype should be parseable));
```

fuzz/fuzz_targets/end_to_end_serialization_for_common_messages.rs
erickcestari: > Lets add a NUM_SIZE here, so we don't let go of a lot of fuzz inputs, something like this:

Adding the `NUM_SIZE` modulo will not help much. The fuzzer will notice that if the first byte (the selector) is greater than four, it will not reach any interesting code. Naturally, the fuzzer will learn that creating or mutating an input greater than five is useless.

Besides the fact that it doesn't change the fuzzer's behavior much, I'm concerned that the fuzzer will learn that if NUM_SIZE = 5 and the selector is greater than 4, it will reach the round trip. So essentially, a selector with a value of 0, 5, 10, or 15 will reach the same branch (the 0 one). Now imagine if a new message is added and we need to update the `NUM_SIZE` to 6. Now, we would lose all the corpus that have a selector >= 5. Because an input with a selector of 0, 5, 10, or 15 will now point to branches 0, 1, 2, or 3.

bit-aloo: I see the concern, but it feels mostly theoretical for our case. The message set is stable and we’re not adding new types in V2, so corpus invalidation is unlikely to matter.

libFuzzer isn’t reasoning about the modulo anyway; it just follows coverage, so multiple selector values mapping to the same branch doesn’t really hurt. The bound mainly makes the harness clearer and avoids wasting cycles

I guess I should have mentioned an enum based arbitrary approach rather than the explicit modulo, though both are kinda same. Something like this:

```rust
  #[derive(Arbitrary, Debug)]
  enum FuzzInput<'a> {
      SetupConnection(Vec<u8>),
      SetupConnectionError(Vec<u8>),
      ...
  }
```

erickcestari: Upon reviewing this again, I agree that the modulo or enum approach is better if the target doesn't expand. Sorry for the confusion. The new changes you propose are LGTM.

By the way, if it's ever necessary to expand it and then change the `NUM_SIZE`, we can take the corpus and update the first byte to be moduled by the `NUM_SIZE` before increasing the `NUM_SIZE`.

fuzz/fuzz_targets/end_to_end_serialization_for_datatypes.rs
lucasbalieiro: these variants can not be added due to unsatisfied trait bounds.

For example:

```
test_datatype_roundtrip!(Seq0255<B032>, data)
```

produces:

```bash
error[E0599]: the method `to_bytes` exists for struct `Seq0255<'_, ...>`,
but its trait bounds were not satisfied
```

`Seq0255<T>` does not implement the required traits (`Serialize` / `Into<EncodableField<'_>>`) for the instantiated inner type, so the macro expansion fails at compile time.

I also tested other nested types (e.g. `Sv2Option<...>`) and encountered the same issue: some combinations do not satisfy the trait bounds required by the serialization path.

Handling nesting with `bool` would introduce additional macro complexity. `bool` requires a dedicated validation path, which means adding explicit match arms such as:

* `Seq064K<bool>`
* `Sv2Option<Seq064K<bool>>`

This is technically feasible (for `bool`), but it increases branching and maintenance cost in the macro. We should decide whether supporting these extra cases justifies the added complexity.

fuzz/fuzz_targets/end_to_end_serialization_for_common_messages.rs
lucasbalieiro: added the modulo

fuzz/fuzz_targets/end_to_end_serialization_for_job_declaration_messages.rs
lucasbalieiro: added the modulo

fuzz/fuzz_targets/end_to_end_serialization_for_mining_messages.rs
lucasbalieiro: added the modulo

fuzz/fuzz_targets/end_to_end_serialization_for_template_distribution_messages.rs
lucasbalieiro: added the modulo

fuzz/fuzz_targets/end_to_end_serialization_for_datatypes.rs
lucasbalieiro: added the modulo

fuzz/Cargo.toml
lucasbalieiro: it was added automagically using `cargo +nightly fuzz add <target_name>`

removed

fuzz/fuzz_targets/common.rs
lucasbalieiro: addressed

fuzz/fuzz_targets/common.rs
lucasbalieiro: addressed

fuzz/fuzz_targets/end_to_end_serialization_for_datatypes.rs
bit-aloo: We need to implement `Encodable` for `B032`, a bit of a bummer we missed that. We should add it here:

```rust
impl_into_encodable_field_for_seq!(bool);
impl_into_encodable_field_for_seq!(u8);
impl_into_encodable_field_for_seq!(u16);
impl_into_encodable_field_for_seq!(U24);
impl_into_encodable_field_for_seq!(u32);
impl_into_encodable_field_for_seq!(u64);
impl_into_encodable_field_for_seq!(U256<'a>);
impl_into_encodable_field_for_seq!(Signature<'a>);
impl_into_encodable_field_for_seq!(B0255<'a>);
impl_into_encodable_field_for_seq!(B064K<'a>);
impl_into_encodable_field_for_seq!(B016M<'a>);
```

Since `Seq` is a container type, it shouldn’t exclude any SV2 types, it should support all valid SV2 field types.

Feel free to open an issue for this and get this in a separate PR.

lucasbalieiro: @Shourya742 , added an issue to track this: https://github.com/stratum-mining/stratum/issues/2093

PR #2077, opened 2026-02-06
feat(channels): add set_nominal_hashrate to client ExtendedChannel

gimballock:
Add a setter for `nominal_hashrate` on the client-side `ExtendedChannel`. This allows downstream consumers (e.g. tProxy) to update the locally tracked hashrate when the downstream vardiff produces an `UpdateChannel` message, keeping it consistent with actual miner performance.

See: [stratum-mining/sv2-apps#242](https://github.com/stratum-mining/sv2-apps/issues/242)

companion https://github.com/stratum-mining/sv2-apps/pull/245

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -182,6 +182,11 @@ impl<'a> ExtendedChannel<'a> {
         self.nominal_hashrate
     }
 
+    /// Sets the nominal hashrate for the channel, in h/s.
+    pub fn set_nominal_hashrate(&mut self, hashrate: f32) {
+        self.nominal_hashrate = hashrate;
+    }
+
     /// Returns a reference to the currently active job, if any.
     pub fn get_active_job(&self) -> Option<&ExtendedJob<'a>> {
         self.active_job.as_ref()


sv2/channels-sv2/src/client/extended.rs
```diff
@@ -182,6 +182,15 @@ impl<'a> ExtendedChannel<'a> {
         self.nominal_hashrate
     }
 
+    /// Sets the nominal hashrate for the channel, in h/s.
+    ///
+    /// This should be called when the downstream vardiff produces an `UpdateChannel`
+    /// message, so that the locally tracked hashrate stays consistent with the
+    /// actual miner performance.
```
```suggestion
```

GitGab19:
Can you squash last commit?

lucasbalieiro:
> @lucasbalieiro can you double check if we're executing integration tests with the companion here?
> 
> I added "companion [stratum-mining/sv2-apps#245](https://github.com/stratum-mining/sv2-apps/pull/245)" the description but not sure if we're getting the right trigger

From the latest integration test execution seems that every thing is alright:

See: https://github.com/stratum-mining/stratum/actions/runs/21768667046/job/62810936243?pr=2077#step:9:18

image

PR #2081, opened 2026-02-12
Fix ci

Shourya742:
Our current CI is failing because it pulls a specific quickcheck version whose transitive dependency requires Rust 2024 edition, which is not supported by Rust 1.75. This PR pins quickcheck to a compatible version to restore CI stability.

Additionally, Miri recently regressed (see https://github.com/rust-lang/miri/issues/4855). This PR temporarily pins the nightly toolchain to avoid that issue.

### DIFF
--- a/.github/workflows/miri.yaml
+++ b/.github/workflows/miri.yaml
@@ -13,17 +13,20 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install Rust nightly + Miri
-        uses: dtolnay/rust-toolchain@nightly
+        # FIXME: Pin nightly due to a regression in miri on nightly-2026-02-12.
+        # See https://github.com/rust-lang/miri/issues/4855.
+        # Revert to plain `channel: nightly` once this is fixed upstream.
+        uses: dtolnay/rust-toolchain@master
         id: toolchain
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/miri.yaml
```diff
@@ -12,18 +12,20 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Install Rust nightly + Miri
-        uses: dtolnay/rust-toolchain@nightly
-        id: toolchain
-        with:
-          components: miri
```
GitGab19, line 19: Can't we pin the nightly version here?

Cargo.toml
```diff
@@ -8,4 +8,8 @@ members = [
 exclude = [
     "integration-test-framework",
     "fuzz"
-]
\ No newline at end of file
+]
+
```
GitGab19, line 12: Can you add a comment here to explain why we're choosing these versions?

bit-aloo: Done

.github/workflows/miri.yaml
bit-aloo: It does have just not under stable or nightly decorators, changed

PR #2082, opened 2026-02-12
move dependencies to workspace root

Shourya742:
Way better to manage dependecies semantics versioning


GitGab19:
By doing this, it's way more difficult to detect which crate has to be bumped after a major bump in one of its dependencies, during PR review.

PR #2083, opened 2026-02-13
Add Prop test infra to binary_sv2

Shourya742:
This test just fix the prop test dependency, adds a coverage shell file and implement prop test of datatypes/non-copy-datatypes/inner.rs.

Part of #2071

### DIFF
--- a/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
+++ b/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
@@ -445,3 +445,583 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::codec::GetSize;
+    use crate::datatypes::{PubKey, Signature, Str0255, U32AsRef, B016M, B0255, B032, B064K, U256};
+    use alloc::boxed::Box;
+    use quickcheck::{Arbitrary, Gen, TestResult};
+    use quickcheck_macros::quickcheck;
+
+    #[derive(Debug, Clone)]
+    struct BytesUpTo32(Vec<u8>);
+
+    impl Arbitrary for BytesUpTo32 {
+        fn arbitrary(g: &mut Gen) -> Self {
+            let len = (u8::arbitrary(g) as usize) % 33;
+            BytesUpTo32((0..len).map(|_| u8::arbitrary(g)).collect())
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    struct BytesUpTo255(Vec<u8>);
+
+    impl Arbitrary for BytesUpTo255 {
+        fn arbitrary(g: &mut Gen) -> Self {
+            let len = u8::arbitrary(g) as usize;
+            BytesUpTo255((0..len).map(|_| u8::arbitrary(g)).collect())
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    struct BytesUpTo64K(Vec<u8>);
+
+    impl Arbitrary for BytesUpTo64K {
+        fn arbitrary(g: &mut Gen) -> Self {
+            let len = (u16::arbitrary(g) as usize).min(g.size());
+            BytesUpTo64K((0..len).map(|_| u8::arbitrary(g)).collect())
+        }
+    }
+
+    #[derive(Debug, Clone)]
+    struct BytesUpTo16M(Vec<u8>);
+
+    impl Arbitrary for BytesUpTo16M {
+        fn arbitrary(g: &mut Gen) -> Self {
+            let max_len = (1 << 20).min(g.size());
+            let len = (u32::arbitrary(g) as usize % max_len).min(0xFFFFFF);
+            BytesUpTo16M((0..len).map(|_| u8::arbitrary(g)).collect())
+        }
+    }
+
+    #[derive(Debug, Clone, Copy)]
+    struct Bytes32([u8; 32]);
+
+    impl Arbitrary for Bytes32 {
+        fn arbitrary(g: &mut Gen) -> Self {
+            let mut arr = [0u8; 32];
+            for byte in arr.iter_mut() {
+                *byte = u8::arbitrary(g);
+            }
+            Bytes32(arr)
+        }
+    }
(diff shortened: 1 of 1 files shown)


sv2/binary-sv2/coverage.sh
lucasbalieiro, line 1: what if this becomes a more generic script, instead of dealing with just binary_s2, and change its path to `scripts` folder?

bit-aloo: Yup, agree. I had to follow the same setups while writing test for other crates as well.

lucasbalieiro: Is not working on my mac, as soon as understand what is failing I'll post here

bit-aloo: you need to have cargo-llvm-cov installed, along with some other llvm profilers. I will add them in the script.

lucasbalieiro: I have them, I probably messed up with the other one i use for the fuzzing coverage:

https://github.com/stratum-mining/stratum-fuzzing-corpus/blob/main/scripts/fuzz_coverage.sh

This is one is working for me

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -445,3 +445,583 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
...
+    }
+
+    #[quickcheck]
+    fn prop_b0255_rejects_oversized(data: Vec<u8>) -> TestResult {
+        if data.len() > 255 {
+            TestResult::from_bool(B0255::try_from(data).is_err())
+        } else {
+            TestResult::discard()
+        }
+    }
```
lucasbalieiro, line 577: for me this test is always `discard`ing the data. Because, this value is too big for the testing session. Maybe we should have some `Arbitrary` impl for some `BytesOver255` (?)

How i tested it? the `panic!` that I added is never reached

```suggestion
    fn prop_b0255_rejects_oversized(data: Vec<u8>) -> TestResult {
        if data.len() > 255 {
             panic!(
                "Data length {} exceeds 255, which should be rejected by B0255",
                data.len()
            );
            TestResult::from_bool(B0255::try_from(data).is_err())
        } else {
            TestResult::discard()
        }
    }
```

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -445,3 +445,583 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
...
+
+    #[quickcheck]
+    fn prop_pubkey_roundtrip(data: Bytes32) {
+        let original = data.0;
+        let mut input = data.0;
+        let pubkey = PubKey::try_from(&mut input[..]).unwrap();
+        let mut buf = vec![0u8; pubkey.get_size()];
+        pubkey.to_slice_unchecked(&mut buf);
+        let decoded = PubKey::from_bytes_unchecked(&mut buf);
+        assert_eq!(decoded.inner_as_ref(), &original[..]);
```
lucasbalieiro, line 731: why not compare directly to `data`?
I'm ok with both, just wanted to know is there is a reason.

```suggestion
        assert_eq!(decoded.inner_as_ref(), &data.0[..]);
```

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -445,3 +445,583 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
...
+        assert_eq!(str0255.get_size(), 1 + data.0.len());
+    }
+
+    #[quickcheck]
+    fn prop_str0255_rejects_oversized(data: Vec<u8>) -> TestResult {
+        if data.len() <= 255 {
+            return TestResult::discard();
+        }
+        TestResult::from_bool(Str0255::try_from(data).is_err())
+    }
```
lucasbalieiro, line 818: all the data is being discarded, same comment as in `prop_b0255_rejects_oversized`. the test bellow never panics


```suggestion
    #[quickcheck]
    fn prop_str0255_rejects_oversized(data: Vec<u8>) -> TestResult {
        if data.len() <= 255 {
            return TestResult::discard();
        }
        panic!("Data length {} exceeds 255", data.len());
        TestResult::from_bool(Str0255::try_from(data).is_err())
    }
```

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -445,3 +445,583 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
...
+    fn prop_b0255_from_vec_roundtrip(data: BytesUpTo255) {
+        let original = data.0.clone();
+        let b0255 = B0255::try_from(data.0).unwrap();
+        let vec_back = b0255.to_vec();
+        assert_eq!(vec_back, original);
+    }
+
+    #[quickcheck]
+    fn prop_str0255_from_utf8_string(text: alloc::string::String) -> TestResult {
+        if text.len() > 255 {
```
lucasbalieiro, line 901: this is  if is never reached, but i think in that case is OK

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -445,3 +445,583 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
...
+    }
+
+    #[quickcheck]
+    fn prop_u256_from_array(data: Bytes32) {
+        let u256: U256 = data.0.into();
+        assert_eq!(u256.inner_as_ref(), &data.0[..]);
+    }
+
+    #[quickcheck]
+    fn prop_b032_rejects_oversized_(size: u8) -> TestResult {
```
lucasbalieiro, line 931:
```suggestion
    fn prop_b032_rejects_oversized(size: u8) -> TestResult {
```

Shourya742:
Closing this considering we anyway gonna change a lot of stuff in binary_sv2 in future.

PR #2087, opened 2026-02-15
add codec prop test

Shourya742:
Part of: #2069

### DIFF
--- a/sv2/codec-sv2/src/decoder.rs
+++ b/sv2/codec-sv2/src/decoder.rs
@@ -436,3 +436,422 @@ mod tests {
         assert_eq!(actual, expect);
     }
 }
+
+#[cfg(test)]
+mod prop_tests {
+    use crate::{decoder::Buffer, encoder::Encoder, StandardDecoder};
+    #[cfg(feature = "noise_sv2")]
+    use crate::{HandshakeRole, NoiseEncoder, StandardNoiseDecoder, State};
+    use binary_sv2::{self, Deserialize, Serialize};
+    use buffer_sv2::Buffer as IsBuffer;
+    #[cfg(feature = "noise_sv2")]
+    use framing_sv2::framing::Frame;
+    use framing_sv2::framing::Sv2Frame;
+    #[cfg(feature = "noise_sv2")]
+    use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
+    #[cfg(feature = "noise_sv2")]
+    use noise_sv2::{ELLSWIFT_ENCODING_SIZE, INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE};
+    use quickcheck::{Arbitrary, Gen, TestResult};
+    use quickcheck_macros::quickcheck;
+    #[cfg(feature = "noise_sv2")]
+    use std::convert::TryInto;
+    #[cfg(feature = "noise_sv2")]
+    use std::time::Duration;
+
+    #[cfg(feature = "noise_sv2")]
+    const AUTHORITY_PUBLIC_K: &str = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72";
+    #[cfg(feature = "noise_sv2")]
+    const AUTHORITY_PRIVATE_K: &str = "mkDLTBBRxdBv998612qipDYoTK3YUrqLe8uWw7gu3iXbSrn2n";
+
+    type Slice = <Buffer as IsBuffer>::Slice;
+
+    #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+    struct TestMessage {
+        value: u16,
+    }
+
+    impl Arbitrary for TestMessage {
+        fn arbitrary(g: &mut Gen) -> Self {
+            TestMessage {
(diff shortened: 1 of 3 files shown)


sv2/codec-sv2/src/decoder.rs
```diff
@@ -436,3 +436,372 @@ mod tests {
...
+        };
+
+        let mut encoder = Encoder::<TestMessage>::new();
+        let encoded = match encoder.encode(frame) {
+            Ok(e) => e,
+            Err(_) => return TestResult::failed(),
+        };
+
+        let mut decoder = StandardDecoder::<TestMessage>::new();
+        let encoded_bytes: &[u8] = encoded.as_ref();
```
lucasbalieiro: why do we need this? I see the same pattern on the other tests but doesn't seem to affect anything.

sv2/codec-sv2/src/decoder.rs
```diff
@@ -436,3 +436,372 @@ mod tests {
...
+                        };
+                    return TestResult::from_bool(decoded_msg == original_msg);
+                }
+                Err(crate::Error::MissingBytes(_)) => continue,
+                Err(_) => return TestResult::failed(),
+            }
+        }
+
+        TestResult::failed()
+    }
```
lucasbalieiro: most of this tests here follow the same pattern, why don't we put some of this tests together and do the assertions?

We probably could put together the tests that the assert the preservation of `extension_type`, `message_type`, for example

sv2/codec-sv2/src/decoder.rs
```diff
@@ -436,3 +436,372 @@ mod tests {
...
+        let mut encoder = Encoder::<TestMessage>::new();
+        let encoded1 = match encoder.encode(frame1) {
+            Ok(e) => e,
+            Err(_) => return TestResult::failed(),
+        };
+
+        let encoded2 = match encoder.encode(frame2) {
+            Ok(e) => e,
+            Err(_) => return TestResult::failed(),
+        };
```
lucasbalieiro, line 628: would be cool if we concat this frames together to see if it will be able to decode 2 separate different frames?

sv2/codec-sv2/src/encoder.rs
```diff
@@ -290,3 +290,172 @@ impl<T: Serialize + GetSize> WithoutNoise<Buffer, T> {
...
+                None => return TestResult::discard(),
+            };
+
+        let mut encoder = Encoder::<TestMessage>::new();
+        let encoded = encoder.encode(frame.clone());
+
+        match encoded {
+            Ok(data) => {
+                let bytes: &[u8] = data.as_ref();
+                let expected_max_len = frame.encoded_length() + 100;
```
lucasbalieiro: just to check: the 100 is arbitrary right? or is there a reason for this specific number?

sv2/codec-sv2/src/encoder.rs
```diff
@@ -290,3 +290,172 @@ impl<T: Serialize + GetSize> WithoutNoise<Buffer, T> {
...
+                TestResult::from_bool(!bytes.is_empty())
+            }
+            Err(_) => TestResult::failed(),
+        }
+    }
+
+    #[quickcheck]
+    fn prop_encoder_output_size_bounded(msg: TestMessage, msg_type: u8) -> TestResult {
+        let frame =
+            match Sv2Frame::<TestMessage, Slice>::from_message(msg.clone(), msg_type, 0, false) {
```
lucasbalieiro: why `clone` the `msg`?

sv2/codec-sv2/src/encoder.rs
lucasbalieiro, line 1: Like in the decoder, I would recommend putting some of this tests together, because they follow the same structure and just differ in the assert

sv2/codec-sv2/src/encoder.rs
bit-aloo: Removed

sv2/codec-sv2/src/encoder.rs
bit-aloo: Clubbed very similar assertion in a single method.

sv2/codec-sv2/src/decoder.rs
bit-aloo: Clubbed.

sv2/codec-sv2/src/decoder.rs
bit-aloo: Its a very annoying thing with buffer_sv2 drop implementation, where you need to be explicit with drop otherwise it kinda gets stuck in spin loop

sv2/codec-sv2/src/decoder.rs
bit-aloo: That what we are doing here?

lucasbalieiro: > That what we are doing here?

no, here we are decoding two separated frame in different steps.

The idea is to join the two frames:
`let unified_frames = |encoded_1||encoded_2|`

And then pass it to the `decode_frame` to see if it will be able to correctly generate two separate frames while receiving 1 big chunk with two frames in it


But, thats just a crazy idea that popped in my head, feel free to solve this thread if it does not make sense or it will give too much work. non-blocker for me

PR #2090, opened 2026-02-18
defer `channels_sv2::client` share acceptance to upstream acknowledgement

### LINKED ISSUE #2089: ShareAccounting in `channels_sv2` has to be refined wrt shares acceptance management
I was analyzing some monitoring data in tProxy, and I dived deeper into the current way we are managing the `shares_accepted` value.

Currently that value is updated by the `update_share_accounting()` which is called internally by the `validate_share()`.

Now, let's focus on the following scenario:
- a tProxy connected to a Pool
- an Extended Channel opened between the two

When a share arrives to tProxy, it's validated against the `client::ExtendedChannel`, and if it satisfies channel target, the `validate_share` calls the `update_share_accounting()`, which increments the `shares_accepted` value.

Is this the right thing to do in this case? Shouldn't we increase the `shares_accepted` only after we get a `SubmitShares.Success` message from the Pool?

cc @plebhash

plebhash:
close #2089

breaking change on `channels_sv2`, bumps major version

we replace `channels_sv2::client::ShareAccounting::update_share_accounting` with:
- `track_validated_share` (still called from `validate_share`)
- `on_share_acknowledgement` (called from the `on_share_acknowledgement` introduced into the channels layer)

application layer calls `on_share_acknowledgement` on channels every time a `SubmitSharesSuccess` arrives

companion https://github.com/stratum-mining/sv2-apps/pull/283

### DIFF
--- a/stratum-core/stratum-translation/Cargo.toml
+++ b/stratum-core/stratum-translation/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "stratum_translation"
-version = "0.1.2"
+version = "0.1.3"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "Stratum V1 ↔ Stratum V2 translation utilities for reuse across proxies, apps, and firmware"


sv2/channels-sv2/src/client/extended.rs
```diff
@@ -212,6 +212,17 @@ impl<'a> ExtendedChannel<'a> {
         &self.share_accounting
     }
 
+    /// Updates share accounting based on a [`SubmitSharesSuccess`] message from the
+    /// upstream server. Delegates to [`ShareAccounting::on_share_acknowledgement`].
+    pub fn on_share_acknowledgement(
+        &mut self,
+        new_submits_accepted_count: u32,
+        new_shares_sum: f64,
```
bit-aloo, line 220: Yeah, we can make `new_share_sum` a `u64`, based on the comments in the sv2-apps PR: https://github.com/stratum-mining/sv2-apps/pull/283#discussion_r2825699396 . There’s no real need to keep it as an `f64`, it looks like an inaccuracy that slipped through during the initial iteration.

stratum-core/stratum-translation/Cargo.toml
GitGab19: we need to bump PATCH here to publish it on our next release

sv2/channels-sv2/src/client/extended.rs
GitGab19: IIRC, the `new_shares_sum` represent the amount of work acknowledged by the server in a specific `SubmitSharesSuccess`, that's why it's a f64.

But I might be wrong here.

plebhash: I remember seeing work being reported as floating point on Braiins UI, and also contemplating the fact that strictly speaking, work is a unit of difficulty, and therefore it can (and maybe should) be tracked as floating point numbers (like rust bitcoin does)

this is specially true when we're operating at diff ranges `0 < diff < 1`, where we have no relevant data unless we're working with floating point numbers

but as @Shourya742 correctly points out, this was kind of a failed attempt at retaining granular data, because `SubmitSharesSuccess.new_shares_sum` is `U64`, which forces everyone to truncate and makes an internal representation based on `f64` pretty much useless

FWIW there's some extra context on https://github.com/stratum-mining/sv2-spec/pull/126

---

with that said, I'd like to avoid expanding the scope of this PR to also cover for this, because that's not the problem #2089 we're trying to solve

created https://github.com/stratum-mining/stratum/issues/2092 to keep track of it

bit-aloo: Acc to spec, its the sum of share acknowledged in the batch,

bit-aloo: I was not aware of that PR. Thanks for mentioning that.

plebhash: > Acc to spec, its the sum of share acknowledged in the batch,

no, you're confusing it with `new_submits_accepted_count`

please see https://github.com/stratum-mining/sv2-spec/pull/126, I drafted this while interacting with Braiins

stratum-core/stratum-translation/Cargo.toml
plebhash: bebe82fdc5a963bd83131539cdd0dc919597fb26

PR #2097, opened 2026-03-02
Check for duplicate share on block found

### LINKED ISSUE #2096: `validate_share` in `channels_sv2` is not checking for duplicate shares in case the share is a valid block
During `bitcoin++ gets exploited` conference we got to know from a guy who was attending the hackathon that `validate_share` in `channels_sv2` does not check for duplicate shares in case the share is a valid block.

We need to add that check there, as we have for the other cases.

GitGab19:
Closes #2096


Shourya742 (approved):
ACK

PR #2098, opened 2026-03-02
Add `extranonce_manager` module in `channels_sv2`

### LINKED ISSUE #1924: Improved Extranonce Prefix Generation

GitGab19:
This PR introduces a new `extranonce_manager` module in `channels_sv2` centered around `ExtranonceAllocator`

It's a unified, reusable allocator for extranonce prefixes for both standard and extended channels. 

It replaces the legacy `ExtendedExtranonce` factory from `mining_sv2` and provides a clearer, spec‑aligned, and more configurable way to size and manage extranonce space across pools, proxies (JDC/Translator), and other applications.

Internally, the allocator uses a compact bitmap (1 bit per channel) to track local_prefix_id usage, ensuring fast allocations, deterministic reuse when channels close, and no overlap between standard and extended channel prefixes.

This PR is inspired by the @coleFD's bitvec approach (see https://github.com/stratum-mining/stratum/pull/2049), but it doesn't add any new external dependencies to our crates.

Closes #1924 

companion https://github.com/stratum-mining/sv2-apps/pull/310


sv2/channels-sv2/src/extranonce_manager.rs
```diff
@@ -0,0 +1,849 @@
...
+//! |       (fixed)         | server_id  | local_prefix_id   |         (for rolling)          |
+//! |<----------------------- total_extranonce_len (≤ 32) ------------------------------>|
+//! ```
+//!
+//! - **upstream_prefix**: Fixed bytes assigned by the upstream node (empty for a Pool server).
+//! - **local_prefix**: Per-channel bytes controlled by *this* node, composed of:
+//!   - **server_id** (optional): Static identifier for this server instance.
+//!   - **local_prefix_id**: Dynamic bytes tracked by a bit vector — each channel gets a unique value.
+//! - **rollable_extranonce_size**: Bytes left for the downstream miner to roll (extended channels).
+//!   Standard channels zero-pad this portion in the prefix since they don't roll.
```
plebhash:
```suggestion
//! - **rollable_extranonce_size**: Bytes left for the downstream miner to roll (extended channels), or do further nested allocation layers.
//!   Standard channels zero-pad this portion in the prefix since they don't roll.
```

sv2/channels-sv2/src/extranonce_manager.rs
```diff
@@ -0,0 +1,849 @@
...
+//! extended mining channels. It replaces the previous `ExtendedExtranonce` factory
+//! from `mining_sv2`, adding support for prefix reuse when channels close.
+//!
+//! # Extranonce layout
+//!
+//! ```text
+//! |<-- upstream_prefix -->|<-------- local_prefix -------->|<-- rollable_extranonce_size -->|
+//! |       (fixed)         | server_id  | local_prefix_id   |         (for rolling)          |
+//! |<----------------------- total_extranonce_len (≤ 32) ------------------------------>|
+//! ```
```
plebhash: fixing a few things here:
- point out that `rollable_extranonce_size` can also be used for downstream allocation
- improve table rendering

```suggestion
//! | upstream_prefix | server_id | local_prefix_id | rollable_extranonce_size |
//! |:---------------:|:---------:|:---------------:|:-----------------------:|
//! | (fixed) | (optional static) | (dynamic, unique per channel) | (for rolling or downstream allocation) |
```

Image

sv2/subprotocols/mining/src/new_mining_job.rs
```diff
@@ -153,8 +153,23 @@ impl NewExtendedMiningJob<'_> {
...
-    use crate::tests::from_arbitrary_vec_to_array;
-    use core::convert::TryFrom;
+    use alloc::vec::Vec;
+    use core::convert::{TryFrom, TryInto};
+
+    fn from_arbitrary_vec_to_array(vec: Vec<u8>) -> [u8; 32] {
+        if vec.len() >= 32 {
+            vec[0..32].try_into().unwrap()
+        } else {
+            let mut result = Vec::new();
```
coleFD: This would be simpler and faster
```
let mut result = [0_u8; 32];
result[start..end].copy_from_slice(&vec[start..end];
```

GitGab19: This function is only used by a couple of tests, but I changed it with your suggestion anyway.

plebhash:
@pavlenex had some issues while trying to connect rented hashrate from Nicehash to a tProxy (exposed over internet)

image

while reviewing/testing this PR, we should make sure nicehash use-case is not blocked

sv2/channels-sv2/src/extranonce_manager.rs
```diff
@@ -0,0 +1,847 @@
+//! Extranonce prefix allocation for downstream SV2 channels.
+//!
+//! [`ExtranonceAllocator`] manages unique extranonce prefixes for both standard and
+//! extended mining channels. It replaces the previous `ExtendedExtranonce` factory
+//! from `mining_sv2`, adding support for prefix reuse when channels close.
```
plebhash:
```suggestion
//! [`ExtranonceAllocator`] manages unique extranonce prefixes for both standard and
//! extended mining channels.
```

No need to reference old designs in docs.

sv2/channels-sv2/src/extranonce_manager.rs
```diff
@@ -0,0 +1,847 @@
...
+            local_prefix_id,
+            self.local_prefix_id_len,
+        ));
+        prefix
+    }
+
+    /// Build the prefix bytes for a standard channel.
+    ///
+    /// Same as the extended prefix, plus zero-padded rollable bytes at the end
+    /// (standard channels don't roll, so the pool fills the full extranonce).
```
plebhash:
```suggestion
    /// (standard channels don't roll, so we fill the rollable bytes with 0).
```

JDC could also be doing this, not only Pool

plebhash:
the suffix `*_id` is being overloaded with multiple competing meanings, and it's making me pretty confused while trying to build a mental map about this new approach

for example:
- `server_id` represents raw bytes
- `local_prefix_id` represents an index in the `BitVector`
- `local_prefix_id` ALSO represents the encoded bytes written into the extranonce prefix (AFAIU `local_prefix_id_len` follows this interpretation)

we should establish terminology that clearly separates concerns.

I suggest we rename variables to something like:
- `server_id` -> `server_bytes`
- `local_prefix_id`  for `BitVector` indices -> `local_slot` or `local_index`
- `local_prefix_id` for encoded bytes written into the extranonce prefix -> `local_slot_bytes` or `local_index_bytes` 

on the suggestion above, I'm leaving `*_bytes` with the exclusive meaning of raw/encoded bytes, while `*_slot`/`*_index` is used exclusively for `BitVector` indices

what's currently described as `local_prefix` could remain the same, because the suffix `*_prefix` would be used consistently with `upstream_prefix`

IMO this would bring more consistency to how the new concepts map into the code, and make the code easier to reason about and maintain on the long term.

plebhash:
aside from the suggestion above, I think I got a much better understanding of this new approach

I'll make another round of review tomorrow (my clanker found some potential edge cases that could cause problems, but I need to sanity-check them before posting here)

GitGab19:
@plebhash I included all your suggestions and now we have:
- `server_id` -> `server_bytes`
- `local_prefix_id` for `BitVector` indexes -> `local_index`
- `local_prefix_id` for encoded bytes written into the extranonce prefix -> `local_index_bytes`

sv2/channels-sv2/src/extranonce_manager.rs
```diff
@@ -0,0 +1,826 @@
...
+    fn mark_allocated(&mut self, index: usize) {
+        self.allocation_bitmap.set(index, true);
+        self.last_allocated_index = Some(index);
+    }
+
+    /// Find the next available local index.
+    ///
+    /// Scans forward from the last allocation point, wrapping around.
+    /// Prefers any free index over `last_freed_index` to avoid immediately reusing
+    /// a prefix whose previous channel may still have in-flight shares.
```
plebhash: we're trying to avoid reusing the most recently disconnected channel's prefix because that channel may still have in-flight shares

but if more than one channel is disconnected before the next allocation (e.g.: batch disconnection), only the last freed index is skipped (which is pretty much arbitrary/pseudorandom in case of a batch disconnection), while earlier freed indexes can still be reused immediately.

so for scenarios of batch disconnections, this logic is not really doing what we'd expect.

but before adding more complex logic to cover such scenarios, we should clarify what' we're really trying to protect against:

is that something we **really** need to worry about? be it for single or batch disconnections

why is it a problem that one channel now contains the same `extranonce_prefix` of a channel that existed in the recent past?

as long as there's no immediate `channel_id` re-use (which is a separate problem that IMO shouldn't be solved on this module), share validation of the current channel remains completely unaffected, and the channel that was closed... well, it no longer exists, so those shares will simply be dropped

so I don't really see any issue to be avoided here, and I think we can simplify things, because the current protection logic is incomplete (and arguably unnecessary)

more specifically:
```diff
diff --git a/sv2/channels-sv2/src/extranonce_manager.rs b/sv2/channels-sv2/src/extranonce_manager.rs
index eedcb786..973910e8 100644
--- a/sv2/channels-sv2/src/extranonce_manager.rs
+++ b/sv2/channels-sv2/src/extranonce_manager.rs
@@ -125,7 +125,6 @@ pub struct ExtranonceAllocator {
     total_extranonce_len: usize,
     allocation_bitmap: BitVector,
     last_allocated_index: Option<usize>,
-    last_freed_index: Option<usize>,
 }
 
 impl core::fmt::Debug for ExtranonceAllocator {
@@ -229,7 +228,6 @@ impl ExtranonceAllocator {
     pub fn free(&mut self, local_index: usize) {
         if local_index < self.max_channels() {
             self.allocation_bitmap.set(local_index, false);
-            self.last_freed_index = Some(local_index);
         }
     }
 
@@ -302,7 +300,6 @@ impl ExtranonceAllocator {
             total_extranonce_len,
             allocation_bitmap,
             last_allocated_index: None,
-            last_freed_index: None,
         })
     }
 
@@ -314,50 +311,21 @@ impl ExtranonceAllocator {
 
     /// Find the next available local index.
     ///
     /// Scans forward from the last allocation point, wrapping around.
-    /// Prefers any free index over `last_freed_index` to avoid immediately reusing
-    /// a prefix whose previous channel may still have in-flight shares.
-    /// Falls back to `last_freed_index` if it is the only one available.
     fn find_free_index(&self) -> Option<usize> {
         let max = self.max_channels();
         let start = self.last_allocated_index.map_or(0, |idx| (idx + 1) % max);
 
-        if let Some(idx) = self.first_free_index_wrapping(start, self.last_freed_index) {
-            return Some(idx);
-        }
-
-        self.last_freed_index
-            .filter(|&idx| idx < max && !self.allocation_bitmap.get(idx))
-    }
-
-    /// Scan the bitmap starting at `from`, wrapping around to cover all indexes.
-    /// Skips `skip` (if provided) so the caller can avoid a specific index.
-    fn first_free_index_wrapping(&self, from: usize, skip: Option<usize>) -> Option<usize> {
-        let max = self.max_channels();
-        self.first_free_index_in_range(from, max, skip).or_else(|| {
-            if from > 0 {
-                self.first_free_index_in_range(0, from, skip)
-            } else {
-                None
-            }
-        })
-    }
-
-    /// Scan `[from, to)` for the first free index, skipping `skip` if provided.
-    fn first_free_index_in_range(
-        &self,
-        from: usize,
-        to: usize,
-        skip: Option<usize>,
-    ) -> Option<usize> {
-        let mut cursor = from;
-        while cursor < to {
-            match self.allocation_bitmap.find_first_zero_in_range(cursor, to) {
-                Some(idx) if Some(idx) == skip => cursor = idx + 1,
-                result => return result,
-            }
-        }
-        None
+        self.allocation_bitmap
+            .find_first_zero_in_range(start, max)
+            .or_else(|| {
+                if start > 0 {
+                    self.allocation_bitmap.find_first_zero_in_range(0, start)
+                } else {
+                    None
+                }
+            })
     }
 
     /// Build the prefix bytes for an extended channel.
@@ -747,7 +715,7 @@ mod tests {
     }
 
     #[test]
-    fn free_and_skip_last_freed() {
+    fn free_reuses_first_available_index() {
         let mut alloc = ExtranonceAllocator::new(6, None, 256).unwrap();
 
         let mut ids = Vec::new();
@@ -755,18 +723,13 @@ mod tests {
             ids.push(alloc.allocate_extended(5).unwrap().local_index());
         }
 
-        let id_a = ids[10];
-        let id_b = ids[11];
+        let id_a = ids[20];
+        let id_b = ids[10];
         alloc.free(id_a);
         alloc.free(id_b);
 
         let reused = alloc.allocate_extended(5).unwrap();
-        assert_ne!(
-            reused.local_index(),
-            id_b,
-            "should skip the last-freed index"
-        );
-        assert_eq!(reused.local_index(), id_a);
+        assert_eq!(reused.local_index(), id_b);
     }
 
     #[test]
```

GitGab19: I agree on your point here, I'll remove that logic.

GitGab19: Removed it.

plebhash:
semver concern coming from my clanker:

> The dependent crate version bumps should be reconsidered.
> 
> Even if these crates do not expose the specific `mining_sv2` APIs that were removed, they do expose `mining_sv2` types in public signatures. Moving those signatures from `mining_sv2` `8.0.*` types to `mining_sv2` `9.0.*` types is still a breaking public API change by Rust type identity.
> 
> With the current patch bumps, downstream users could have something like this in their `Cargo.toml` files:
> 
> ```toml
> channels_sv2 = "^4.0.0"
> mining_sv2 = "^8.0.0"
> # etc
> ```
> 
> and a normal `cargo update` could select `channels_sv2` `4.0.1`, which now expects `mining_sv2` `9.0.*` message types. Their code would then fail to compile because `mining_sv2` `8::Foo` and `mining_sv2 9::Foo` are different Rust types, even if the actual message struct did not change.
>
> So crates exposing `mining_sv2` message types publicly should get incompatible version bumps too, otherwise we risk semver-compatible updates causing unexpected downstream compile breaks.

I think this is less of a concern if consumers only used `stratum-core`, because then this repo controls the internally consistent set of `mining_sv2`, `channels_sv2`, etc. versions that get re-exported.

but if anyone imports `channels_sv2` + `mining_sv2` + etc as seperate dependencies, it feels like a legitimate concern (and unfortunately we cannot force people to only use `stratum-core` instead of separate deps)

---

speaking more from the human side, I empathize with the psychological urge to be conservative and avoid major bumps as much as possible, especially when due to indirect changes on the dependency chain

however, I think this is more of an undesired consequence of the way SRI core crates were structured (excessive crate fragmentation), and I've somewhat accepted the tradeoff of living with frequent major bumps (in name of avoiding forcing dependency hell on downstream consumers)

GitGab19:
Should we start bumping MAJOR on every dependent crate though?

plebhash:
> Should we start bumping MAJOR on every dependent crate though?

hmm that feels too generic...`channels_sv2` does deserve a MAJOR bump, but I'd say there's exceptions.

## v0.x.y crates

Here's a table my agent to put together, which after careful examination, I agree with:

> | Crate | Current PR bump | More appropriate bump | Why |
> |---|---:|---:|---|
> | `parsers_sv2` | `0.2.2 -> 0.2.3` | `0.2.2 -> 0.3.0` | Public enums wrap `mining_sv2` message types. |
> | `handlers_sv2` | `0.2.2 -> 0.2.3` | `0.2.2 -> 0.3.0` | Public traits expose `mining_sv2` and `parsers_sv2` types. |
> | `stratum_translation` | `0.1.3 -> 0.1.4` | `0.1.3 -> 0.2.0` | Public functions expose `mining_sv2` types; public error also exposes a `channels_sv2` error type. |
> | `stratum-core` | `0.2.1 -> 0.2.2` | `0.2.1 -> 0.3.0` | Re-exports the affected crates, so its public type graph changes. |

Those crates are still on `v0.x.y`, so they are exceptions because a breaking change can be signaled as a MINOR bump.

## dependencies that don't touch public APIs

Let's say crate B depends on crate A, which is undergoing some breaking change.

The main criterion is whether **B public API has any types from A**. Not whether the breaking change in A leaks over B in any significant way.

If B only uses types from A without exposure via public APIs, then a PATCH bump would be fine.

More specifically with the case of this PR:
- it does NOT matter that the breaking changes of `mining_sv2` have no effect on `channels_sv2` public APIs
- it DOES matter that there are types from `mining_sv2` exposed via `channels_sv2` public APIs

if there were no types from `mining_sv2` exposed via `channels_sv2` public APIs, a PATCH bump would have been fine.

---

PR #2133  is formalizing these rules into `CONTRIBUTING.md` so we have more clarity moving forward (both for humans and agents). It's also updating `RELEASE.md`, which got outdated since the repo split.

plebhash:
my final review point on this PR is somewhat subjective, and therefore it's not blocking

---

I feel we could have better conceptual organization by splitting into submodules
- `mod.rs` retains the header Rust Docs module description, `const MAX_EXTRANONCE_LEN` and helper `bytes_needed` 
- `ExtranoncePrefix` goes to public `extranonce_manager/prefix.rs` submodule
- `ExtranonceAllocator` goes to public `extranonce_manager/allocator.rs` submodule
- `BitVector` goes to private `extranonce_manager/bitvector.rs` submodule

commit c89e150c2b61b122aea2cb5c2feee635e54bfd9d implements these suggestions, and could be either cherry-picked or squashed into other commits on this PR if agreed upon

if not (especially given this could also affect https://github.com/stratum-mining/sv2-apps/pull/310), I think this PR is already good to go (aside from the versioning issues detailed above)

plebhash:
about the Nicehash comment above: https://github.com/stratum-mining/stratum/pull/2098#issuecomment-4237715430

I haven't had time to do some actual testing with this PR + https://github.com/stratum-mining/sv2-apps/pull/310, but from what I understood of this PR on a conceptual level, I don't see any major blockers

at `channels_sv2::extranonce_manager`, we just need to make sure that:
- Pool can establish arbitrary sizes for Extranonce Prefix and Full Extranonce
- if running in aggregated mode, tProxy can establish arbitrary sizes for local Extranonce Prefix (aka extranonce1) allocation, while respecting sizes for Extranonce Prefix and Full Extranonce that were imposed by Pool

both points seem sufficiently covered with these new APIs, and any potential edge cases with Sv1 extranonce1 or extranonce2 could only arise from specific combinations of configuration parameters of Pool + tProxy

---

TLDR I don't see any material blockers on this PR with regards to the issues observed on Nicehash, at least on a conceptual level

GitGab19:
@plebhash I cherry-picked your commit and squashed it in https://github.com/stratum-mining/stratum/pull/2098/commits/536480dbd3e3181712f4fe6754b22452abd2d767 (don't know why your contribution in the commit went away, sorry).

I also changed the different crates bumps according to your analysis.

About the companion PR, I still need to test it and do some changes probably, so let's wait a bit there.

Once that's ready, I'll do some tests with Nicehash (I created an account an hour ago), I'll keep you posted.

plebhash (approved):

Shourya742:
I will be reviewing it over the weekends.. adding myself as a reviewer.

GitGab19:
@plebhash I completed all the adaptations on the apps, so https://github.com/stratum-mining/sv2-apps/pull/310 is finally ready for a full review.

In addition to that, I just tried to connect NiceHash to a tmux session on our VPS which is running my PR, and everything looks fine!

image

GitGab19:
I was also able to reproduce the error @pavlenex got, by selecting SHA256 (instead of the correct SHA256AsicBoost) algorithm on NiceHash:

image

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+/// make it available for reuse.
+///
+/// # Memory
+///
+/// The allocator uses an internal bitmap of `max_channels` bits (`max_channels / 8`
+/// bytes). See the module-level documentation for a reference table.
+pub struct ExtranonceAllocator {
+    upstream_prefix: Vec<u8>,
+    server_bytes: Vec<u8>,
+    total_extranonce_len: usize,
```
Shourya742, line 27: u8 should work, considering MAX_EXTRANONCE_SIZE is 32

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// - `total_extranonce_len`: Total extranonce length in bytes (≤ 32).
+    /// - `server_bytes`: Optional static server identifier (placed at the start of local_prefix).
+    /// - `max_channels`: Maximum concurrent channels. Determines the size of the
+    ///   internal allocation bitmap (`max_channels / 8` bytes). See the
+    ///   [Memory](#memory) section on [`ExtranonceAllocator`] for a reference table.
+    ///
+    /// `local_prefix_len` is derived automatically:
+    /// `server_bytes.len() + bytes_needed(max_channels)`.
+    pub fn new(
+        total_extranonce_len: usize,
```
Shourya742, line 60:
```suggestion
        total_extranonce_len: u8,
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        server_bytes: Option<Vec<u8>>,
+        max_channels: usize,
```
Shourya742, line 62: We have a hard limit of 32 bytes, so max channels will fit in u32.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+
+    /// Free a previously allocated local index, making it available for reuse.
+    pub fn free(&mut self, local_index: usize) {
+        if local_index < self.max_channels() {
+            self.allocation_bitmap.set(local_index, false);
+        }
+    }
+
+    /// Bytes available for downstream rolling (extended channels).
+    pub fn rollable_extranonce_size(&self) -> usize {
```
Shourya742, line 137:
```suggestion
    #[inline]
    pub fn rollable_extranonce_size(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.total_extranonce_len - self.upstream_prefix.len() - self.local_prefix_len()
+    }
+
+    /// Length of the upstream prefix portion.
+    pub fn upstream_prefix_len(&self) -> usize {
```
Shourya742, line 142:
```suggestion
    #[inline]
    pub fn upstream_prefix_len(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.upstream_prefix.len()
+    }
+
+    /// The upstream prefix bytes.
+    pub fn upstream_prefix(&self) -> &[u8] {
```
Shourya742, line 147:
```suggestion
    #[inline]
    pub fn upstream_prefix(&self) -> &[u8] {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        &self.upstream_prefix
+    }
+
+    /// Total extranonce length.
+    pub fn total_extranonce_len(&self) -> usize {
```
Shourya742, line 152:
```suggestion
    #[inline]
    pub fn total_extranonce_len(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.total_extranonce_len
+    }
+
+    /// Length of the local prefix portion (`server_bytes.len() + local_index_len`).
+    pub fn local_prefix_len(&self) -> usize {
```
Shourya742, line 157:
```suggestion
   #[inline]
    pub fn local_prefix_len(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.server_bytes.len() + self.local_index_len()
+    }
+
+    /// Number of currently allocated channels.
+    pub fn allocated_count(&self) -> usize {
```
Shourya742, line 162:
```suggestion
   #[inline]
    pub fn allocated_count(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.allocation_bitmap.count_ones()
+    }
+
+    /// Maximum number of concurrent channels.
+    pub fn max_channels(&self) -> usize {
```
Shourya742, line 167:
```suggestion
   #[inline]
    pub fn max_channels(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.allocation_bitmap.capacity()
+    }
+
+    /// Number of bytes used to encode each local index.
+    fn local_index_len(&self) -> usize {
```
Shourya742, line 172:
```suggestion
    #[inline]
    fn local_index_len(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+            upstream_prefix,
+            server_bytes,
+            total_extranonce_len,
+            allocation_bitmap,
+            last_allocated_index: None,
+        })
+    }
+
+    /// Mark a local index as in-use and record it as the latest allocation.
+    fn mark_allocated(&mut self, index: usize) {
```
Shourya742, line 209:
```suggestion
    #[inline]
    fn mark_allocated(&mut self, index: usize) {
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+pub(crate) struct BitVector {
+    /// Packed allocation bits. `words[i]` covers indexes `i*64 .. (i+1)*64`.
+    words: Vec<u64>,
+    /// Number of valid indexes tracked (= `max_channels`). May be less than
+    /// `words.len() * 64` when `max_channels` is not a multiple of 64.
+    capacity: usize,
+}
+
+impl BitVector {
+    pub(crate) fn new(capacity: usize) -> Self {
```
Shourya742, line 40:
```suggestion
    #[inline]
    pub(crate) fn new(capacity: usize) -> Self {
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+        let num_words = capacity.div_ceil(64);
+        Self {
+            words: vec![0u64; num_words],
+            capacity,
+        }
+    }
+
+    pub(crate) fn capacity(&self) -> usize {
```
Shourya742, line 48:
```suggestion
    #[inline]
    pub(crate) fn capacity(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub(crate) fn get(&self, index: usize) -> bool {
+        debug_assert!(index < self.capacity);
+        let word = index / 64;
+        let bit = index % 64;
+        (self.words[word] >> bit) & 1 == 1
+    }
+
+    pub(crate) fn set(&mut self, index: usize, value: bool) {
+        debug_assert!(index < self.capacity);
+        let word = index / 64;
```
Shourya742, line 62:
```suggestion
        let word = index >> 6;
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+        let bit = index % 64;
```
Shourya742, line 63:
```suggestion
        let bit = index & 63;
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Find the first zero bit in `[from, to)`, skipping fully-set `u64` words.
+    /// Returns `None` if every bit in the range is set.
+    pub(crate) fn find_first_zero_in_range(&self, from: usize, to: usize) -> Option<usize> {
+        if from >= to {
+            return None;
+        }
+
+        let mut idx = from;
+        while idx < to {
+            let word_idx = idx / 64;
```
Shourya742, line 84:
```suggestion
            let word_idx = idx & 63;
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+            let word = self.words[word_idx];
+
+            if word == u64::MAX {
+                idx = (word_idx + 1) * 64;
+                continue;
+            }
+
+            let bit_offset = idx % 64;
```
Shourya742, line 92:
```suggestion
            let bit_offset = idx & 63;
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,181 @@
...
+        let word = index / 64;
+        let bit = index % 64;
+        if value {
+            self.words[word] |= 1u64 << bit;
+        } else {
+            self.words[word] &= !(1u64 << bit);
+        }
+    }
+
+    pub(crate) fn count_ones(&self) -> usize {
```
Shourya742, line 71:
```suggestion
    #[inline]
    pub(crate) fn count_ones(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,38 @@
...
+impl ExtranoncePrefix {
+    pub(crate) fn new(local_index: usize, prefix: Vec<u8>) -> Self {
+        Self {
+            local_index,
+            prefix,
+        }
+    }
+
+    /// The raw prefix bytes — pass these to channel constructors.
+    pub fn as_bytes(&self) -> &[u8] {
```
Shourya742, line 24:
```suggestion
    #[inline]
    pub fn as_bytes(&self) -> &[u8] {
```

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,38 @@
...
+        &self.prefix
+    }
+
+    /// Consume and return the prefix bytes.
+    pub fn into_bytes(self) -> Vec<u8> {
```
Shourya742, line 29:
```suggestion
    #[inline]
    pub fn into_bytes(self) -> Vec<u8> {
```

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,38 @@
...
+        self.prefix
+    }
+
+    /// The assigned local index — store this and pass to [`super::ExtranonceAllocator::free`]
+    /// when the channel closes.
+    pub fn local_index(&self) -> usize {
```
Shourya742, line 35:
```suggestion
    #[inline]
    pub fn local_index(&self) -> usize {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+        upstream_prefix: Vec<u8>,
+        server_bytes: Vec<u8>,
+        total_extranonce_len: usize,
+        max_channels: usize,
+    ) -> Result<Self, ExtranonceAllocatorError> {
+        if total_extranonce_len > MAX_EXTRANONCE_LEN {
+            return Err(ExtranonceAllocatorError::ExceedsMaxLength);
+        }
+        if max_channels == 0 {
+            return Err(ExtranonceAllocatorError::InvalidLengths);
```
Shourya742, line 187: This variant name can be improved?

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Stores both the raw prefix bytes (to pass to channel constructors) and the
+/// `local_index` (to pass back to [`super::ExtranonceAllocator::free`] when the channel closes).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ExtranoncePrefix {
+    local_index: usize,
+    prefix: Vec<u8>,
+}
+
+impl ExtranoncePrefix {
+    pub(crate) fn new(local_index: usize, prefix: Vec<u8>) -> Self {
```
Shourya742, line 16:
```suggestion
    #[inline]
    pub(crate) fn new(local_index: usize, prefix: Vec<u8>) -> Self {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,449 @@
...
+    pub fn allocate_standard(&mut self) -> Result<ExtranoncePrefix, ExtranonceAllocatorError> {
+        let idx = self
+            .find_free_index()
+            .ok_or(ExtranonceAllocatorError::CapacityExhausted)?;
+        self.mark_allocated(idx);
+        Ok(ExtranoncePrefix::new(idx, self.build_standard_prefix(idx)))
+    }
+
+    /// Free a previously allocated local index, making it available for reuse.
+    pub fn free(&mut self, local_index: usize) {
```
Shourya742, line 130: Is this is expected to be called by the user?

bit-aloo (commented):
First pass,

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: If I change this to `u8`, I have to add casting to `usize` in every place this value is used.

Not sure it's worth it.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: If I change this to `u8`, I have to add casting to `usize` in every place this value is used.

Not sure it's worth it.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: `max_channels` can theoretically be up to 2^32, which doesn't fit in u32.

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
GitGab19: Here you wanted to suggest `idx >> 6;`, right?

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: Done.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: Yes, this is expected to be called by the user once a channel is closed/freed. 

I improved the docs to clearly state this.

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
Shourya742: Yup

sv2/channels-sv2/src/extranonce_manager/allocator.rs
Shourya742: Channels are not zero-indexed, and `usize` is still architecture-dependent, better to use `u64` instead?

sv2/channels-sv2/src/extranonce_manager/allocator.rs
Shourya742: I was thinking about whether we could introduce semantics for implicitly freeing a registered extranonce, rather than requiring it to be done explicitly. I mentioned this in the companion PR as well: https://github.com/stratum-mining/sv2-apps/pull/310#discussion_r3108994891

sv2/channels-sv2/src/extranonce_manager/allocator.rs
Shourya742: But why does every place need it to be a `usize`? I thought the reason it was a `usize` elsewhere was because of the range-based method we had before this approach.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: Ok, I'll use u32 as you initially suggested.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: I'll change it to u8 soon.

GitGab19: Done

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: Changed to u8.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: Done

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: @Shourya742 I pushed an additional commit with a proposal which should improve this, without making everything too complex: ~~https://github.com/stratum-mining/stratum/pull/2098/changes/2dd596c7570883defb2906d849e6e7d5d4c0c8ec~~ https://github.com/stratum-mining/stratum/pull/2098/commits/462f59584c15bed12c4b5d7debe580361735d450

Let me know what you think, I'll adapt the companion PR in the meantime to experiment with it.

If you like this, I'll squash it in the first commit here.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,569 @@
...
+    #[inline]
+    pub fn max_channels(&self) -> u32 {
+        // The bitmap was constructed from a `u32` passed to `new`/`from_upstream`,
+        // so this cast is lossless.
+        self.allocation_bitmap.capacity() as u32
+    }
+
+    /// Number of bytes used to encode each local index.
+    #[inline]
+    fn local_index_len(&self) -> usize {
```
Shourya742:
```suggestion
    fn local_index_len(&self) -> u8 {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,569 @@
...
+        Ok(ExtranoncePrefix::new(
+            idx,
+            self.build_standard_prefix(idx),
+            Arc::downgrade(&self.allocation_bitmap),
+        ))
+    }
+
+    /// Bytes available for downstream rolling (extended channels).
+    #[inline]
+    pub fn rollable_extranonce_size(&self) -> usize {
```
Shourya742:
```suggestion
    pub fn rollable_extranonce_size(&self) -> u8 {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,569 @@
...
+        self.total_extranonce_len as usize - self.upstream_prefix.len() - self.local_prefix_len()
+    }
+
+    /// Length of the upstream prefix portion.
+    #[inline]
+    pub fn upstream_prefix_len(&self) -> usize {
```
Shourya742:
```suggestion
    pub fn upstream_prefix_len(&self) -> u8 {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,569 @@
...
+
+    /// Total extranonce length.
+    #[inline]
+    pub fn total_extranonce_len(&self) -> u8 {
+        self.total_extranonce_len
+    }
+
+    /// Length of the local prefix portion (`server_bytes.len() + local_index_len`).
+    #[inline]
+    pub fn local_prefix_len(&self) -> usize {
```
Shourya742:
```suggestion
    pub fn local_prefix_len(&self) -> u8 {
```

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -0,0 +1,569 @@
...
+        self.server_bytes.len() + self.local_index_len()
+    }
+
+    /// Number of currently allocated channels.
+    #[inline]
+    pub fn allocated_count(&self) -> usize {
```
Shourya742:
```suggestion
    pub fn allocated_count(&self) -> u32 {
```

sv2/channels-sv2/src/extranonce_manager/bitvector.rs
```diff
@@ -0,0 +1,206 @@
...
+        let mask = 1u64 << bit;
+        if value {
+            self.words[word].fetch_or(mask, Ordering::AcqRel);
+        } else {
+            self.words[word].fetch_and(!mask, Ordering::AcqRel);
+        }
+    }
+
+    #[inline]
+    pub(crate) fn count_ones(&self) -> usize {
```
Shourya742:
```suggestion
    pub(crate) fn count_ones(&self) -> u32 {
```

sv2/channels-sv2/src/extranonce_manager/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+
+pub mod allocator;
+mod bitvector;
+pub mod prefix;
+
+pub use allocator::{ExtranonceAllocator, ExtranonceAllocatorError};
+pub use prefix::ExtranoncePrefix;
+
+/// Maximum extranonce length in bytes (Sv2 spec).
+pub const MAX_EXTRANONCE_LEN: usize = 32;
```
Shourya742:
```suggestion
pub const MAX_EXTRANONCE_LEN: u8 = 32;
```

sv2/channels-sv2/src/extranonce_manager/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Minimum bytes needed to represent `n` distinct values (`0..n`).
+///
+/// Always returns at least 1. Exposed so consumers can derive byte counts
+/// from their `max_channels` value without hardcoding (and risking drift
+/// from what [`ExtranonceAllocator`] uses internally).
+pub const fn bytes_needed(n: u32) -> usize {
```
Shourya742:
```suggestion
pub const fn bytes_needed(n: u32) -> u8 {
```

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,101 @@
...
+        core::mem::take(&mut self.prefix)
+    }
+
+    /// The assigned local index within the allocator's bitmap.
+    ///
+    /// Useful for diagnostics, logging, or tests. Not required for normal
+    /// use: the allocation is released automatically when the prefix is
+    /// dropped.
+    #[inline]
+    pub fn local_index(&self) -> usize {
```
Shourya742:
```suggestion
    pub fn local_index(&self) -> u32 {
```

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl Eq for ExtranoncePrefix {}
+
+impl Drop for ExtranoncePrefix {
+    fn drop(&mut self) {
+        if let Some(bitmap) = self.bitmap.upgrade() {
+            bitmap.set(self.local_index, false);
+        }
+    }
+}
```
Shourya742, line 101: This is pretty smart

GitGab19: Do you like it? If so, I would squash the commit which introduced this safety mechanism into the very first one.

Shourya742: Yes

GitGab19: Alright, just squashed!

bit-aloo: I think we need to make some adjustments to support this working with channel drop: https://github.com/stratum-mining/sv2-apps/pull/310#discussion_r3122241696, without forcing us to maintain an additional hashmap just for cleanup. It makes sense to tie the lifetime of the extranonce prefix to the channel’s lifetime.

GitGab19: Just pushed https://github.com/stratum-mining/stratum/pull/2098/commits/07d7d05c2025c81217ae604f217e9df76d529df7 for using `ExtranoncePrefix` type (instead of bytes) ONLY in the server-side channels (standard and extended), since the client-side channels don't need any allocation/deallocation mechanism.

GitGab19: I'll adapt the companion PR in a moment, but I'd like an ACK from both of you on these changes.

cc @Shourya742 @plebhash

plebhash: ACK

bit-aloo: looks good

GitGab19:
@plebhash @Shourya742 I pushed a new commit (https://github.com/stratum-mining/stratum/commit/5a36d11ca19677649ef3d0430a3c9ccc87344bce) to extend the usage of `ExtranoncePrefix` to client channels, as we discussed during the PR Review Club call.

I'll work on adapting the tProxy on the companion PR soon.

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -75,43 +125,22 @@ impl ExtranoncePrefix {
...
-
-    /// The assigned local index within the allocator's bitmap.
-    ///
-    /// Useful for diagnostics, logging, or tests. Not required for normal
-    /// use: the allocation is released automatically when the prefix is
-    /// dropped.
-    #[inline]
-    pub fn local_index(&self) -> u32 {
-        self.local_index
-    }
```
bit-aloo: We should add these accessors for `PrefixAllocation`

GitGab19: Why would we need them?

GitGab19:
@plebhash @Shourya742 I pushed a new commit (https://github.com/stratum-mining/stratum/pull/2098/commits/ffb63c24e4990536a62cc85b14a2f738ba363bc4) which renames the old `server_bytes` field into a more generic `local_prefix_bytes`, usable in both `new()` and `from_upstream_prefix()` constructors.

In this way tProxy can use it (in the `from_upstream_prefix()`) to 0-pad the difference between `extranonce_prefix` received from its upstream, and the `downstream_extranonce2_size` value we set in the config file.

I pushed all the adaptations on the companion PR, and I did a quick round of testing, and everything looks good. But I want to do some deeper tests tomorrow.

If you ACK this last commit, I'll squash it in b1c669e378006318e8898c97573ef66c3fdc854b.

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,145 @@
...
+impl ExtranoncePrefix {
+    /// Build an [`ExtranoncePrefix`] from bytes received over the wire.
+    ///
+    /// Intended for client-side channels whose extranonce prefix is chosen
+    /// by the upstream server (e.g. via `OpenExtendedMiningChannelSuccess`,
+    /// `OpenStandardMiningChannelSuccess`, or `SetExtranoncePrefix`).
+    /// The resulting prefix carries no allocation record and its [`Drop`]
+    /// impl is a silent no-op.
+    #[inline]
+    pub fn from_wire(prefix: Vec<u8>) -> Self {
```
bit-aloo: We need to make this method falliable as prefix from wire can be more than 32 breaking our extranonce invariant

bit-aloo:
I was thinking about the current design, and at the moment the server side of the channel can accept the `from_wire` variant of `ExtranoncePrefix`, which renders the cleanup useless since it doesn’t have an allocation pointer reference (`Weak` reference). I was thinking it might be necessary to make such combinations impossible using types, something like:

For client_side of channel:
```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct WireExtranoncePrefix {
    prefix: Vec<u8>,
}
```

For server_side of channel
```rust
pub struct ExtranoncePrefix {
    prefix: Vec<u8>,
    allocation: PrefixAllocation,
}

#[derive(Debug)]
struct PrefixAllocation {
    local_index: u32,
    bitmap: Weak<BitVector>,
}

```

sv2/channels-sv2/src/extranonce_manager/prefix.rs
bit-aloo: These are related to freeing, and end user should not worry. Yup, makes sense.

sv2/channels-sv2/src/extranonce_manager/prefix.rs
GitGab19: Done, it now returns a `ExtranoncePrefixError` if the length is more than 32 bytes.

bit-aloo:
> I was thinking about the current design, and at the moment the server side of the channel can accept the `from_wire` variant of `ExtranoncePrefix`, which renders the cleanup useless since it doesn’t have an allocation pointer reference (`Weak` reference). I was thinking it might be necessary to make such combinations impossible using types, something like:
> 
> For client_side of channel:
> 
> ```rust
> #[derive(Debug, Clone, PartialEq, Eq)]
> pub struct WireExtranoncePrefix {
>     prefix: Vec<u8>,
> }
> ```
> 
> For server_side of channel
> 
> ```rust
> pub struct ExtranoncePrefix {
>     prefix: Vec<u8>,
>     allocation: PrefixAllocation,
> }
> 
> #[derive(Debug)]
> struct PrefixAllocation {
>     local_index: u32,
>     bitmap: Weak<BitVector>,
> }
> ```

@GitGab19 what you think of this?

GitGab19:
@Shourya742 I just pushed a new commit (https://github.com/stratum-mining/stratum/pull/2098/commits/b61eaf4ff856fc68c5b15c81beca1324dfb802fa) which introduces a new `AllocatedExtranoncePrefix` (wrapper of `ExtranoncePrefix`) which is enforced at compile-time on the server-side channels.

In the case of client-side channels, we need to keep things as they are, because tProxy is using both pure prefixes coming from the wire and prefixes coming from the allocator (to extend the extranonce for its downstreams).

bit-aloo:
> @Shourya742 I just pushed a new commit ([b61eaf4](https://github.com/stratum-mining/stratum/commit/b61eaf4ff856fc68c5b15c81beca1324dfb802fa)) which introduces a new `AllocatedExtranoncePrefix` (wrapper of `ExtranoncePrefix`) which is enforced at compile-time on the server-side channels.
> 
> In the case of client-side channels, we need to keep things as they are, because tProxy is using both pure prefixes coming from the wire and prefixes coming from the allocator (to extend the extranonce for its downstreams).

This works as well, as long as server channels stay safe. Looks good to me

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -0,0 +1,290 @@
...
+/// There is no public constructor: an `AllocatedExtranoncePrefix` can only
+/// be obtained from
+/// [`ExtranonceAllocator::allocate_extended`](super::allocator::ExtranonceAllocator::allocate_extended)
+/// or
+/// [`allocate_standard`](super::allocator::ExtranonceAllocator::allocate_standard).
+/// It converts into the wider [`ExtranoncePrefix`] via
+/// [`From`]/[`Into`] — the conversion is one-way; allocation provenance
+/// cannot be forged in the other direction.
+#[derive(Debug)]
+pub struct AllocatedExtranoncePrefix(ExtranoncePrefix);
```
bit-aloo, line 90: Won't let anyone touch this except this crate.

```suggestion
pub(crate) struct AllocatedExtranoncePrefix(ExtranoncePrefix);
```

GitGab19: I can't do that, because it's used in the channels constructors, which are public.

plebhash:
a non-blocking comment about the current design (where standard and extended channels share a unified allocation space):

standard channel allocation is evolving in this pattern:
```
0000000000000000000000000000000000000000
0000000100000000000000000000000000000000
0000000200000000000000000000000000000000
0000000300000000000000000000000000000000
...
FFFFFFFF00000000000000000000000000000000
```

one could argue we're wasting search space with the 0-padding

if standard channel allocation space was separate from extended, it could evolve under this pattern:
```
0000000000000000000000000000000000000000
0000000000000000000000000000000000000001
0000000000000000000000000000000000000002
0000000000000000000000000000000000000003
...
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
```

this would lift the ceiling for max concurrent standard channels to an astronomical scale, with the tradeoffs:
- small (and arguably negligible) collisions with allocated extended channels rollable space
- impossible to track over a bitvec (RAM would explode), therefore the old wrapping-counter approach would be required for standard channels, leaving reusable bitvecs for extended channels only 

---

it's hard to evaluate whether this will really have a meaningful impact in production scenarios, as a unified allocation space that caters for e.g.: `max_channels = 16_777_216` is probably be more than sufficient

anyways I don't think we need to do anything about this now (and perhaps ever)

but I feel it's at least worthy of note, since this is something I've always contemplated about the previous design

bit-aloo (approved):
ACK.

PR #2106, opened 2026-03-03
add fuzz target for the `noise_sv2` crate.

### LINKED ISSUE #2105: add fuzz coverage to `noise_sv2`
the main things to cover there are the handshake process and the encryption/decryption roundtrip

lucasbalieiro:
to be merged after #2104 

closes #2105 

the target aims to fuzz the public apis of the handshake and also sanity check the encryption roundtrip.


fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+use noise_sv2::{Initiator, Responder};
+use secp256k1::{Keypair, Secp256k1, XOnlyPublicKey};
+
+use rand::{rngs::StdRng, SeedableRng};
+
+fn generate_key(rand_seed: u64) -> Keypair {
+    //spec says 4.3.1.1: No assumption is made about the parity of Y-coordinate. For the purpose of signing (e.g. certificate) and ECDH (handshake) it is not necessary to "grind" the private key.
```
GitGab19: Can you put a direct link to that part of the specs?

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
```diff
@@ -0,0 +1,127 @@
...
+    let secp = Secp256k1::new();
+    let mut rng = StdRng::seed_from_u64(rand_seed);
+
+    let (secret_key, _public_key) = secp.generate_keypair(&mut rng);
+
+    Keypair::from_secret_key(&secp, &secret_key)
+}
+
+#[derive(PartialEq, Arbitrary, Debug)]
+enum PeerMode {
```
GitGab19, line 28: Can you add a comment on top of this?

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
```diff
@@ -0,0 +1,127 @@
...
+    Unknown,
+    Known,
+    PeerDifferentFromResponder,
+}
+
+#[derive(Debug, Arbitrary)]
+struct FuzzInput {
```
GitGab19, line 40: Can you add a comment on top of some fields like `should_corrupt_message` and `corruption_index`?

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
```diff
@@ -0,0 +1,127 @@
...
+    peer_mode: PeerMode,
+    cert_validity: u32,
+    message: Vec<u8>,
+    should_corrupt_message: bool,
+    corruption_index: usize,
+    rand_seed: u64,
+}
+
+fn maybe_corrupt_message(should_corrupt: bool, corruption_index: usize, message: &mut [u8]) {
```
GitGab19, line 62: Can you add a comment on top of this?

fuzz/Cargo.toml
```diff
@@ -17,13 +17,16 @@ cargo-fuzz = true
...
+noise_sv2 = { path = "../sv2/noise-sv2" }
 parsers_sv2 = { path = "../sv2/parsers-sv2" }
 framing_sv2 = { path = "../sv2/framing-sv2" }
 codec_sv2 = { path = "../sv2/codec-sv2", features = ["noise_sv2"]}
 common_messages_sv2 = { path = "../sv2/subprotocols/common-messages" }
 job_declaration_sv2 = { path = "../sv2/subprotocols/job-declaration" }
 mining_sv2 = {path = "../sv2/subprotocols/mining" }
 template_distribution_sv2 = { path = "../sv2/subprotocols/template-distribution" }
+secp256k1 = { version = "0.28.2", default-features = false, features = ["alloc", "rand"] }
+rand = { version = "0.8.5", default-features = false }
```
GitGab19, line 29: Can't we use the dependencies we already have at the workspace level for these two?

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
lucasbalieiro: addressed

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
lucasbalieiro: addressed

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
lucasbalieiro: addressed

fuzz/Cargo.toml
lucasbalieiro: I think we can't do that because the fuzz is excluded from the workspace.

see:

https://github.com/stratum-mining/stratum/blob/2a52ad5beb3a3eb23e7879604ea71f4719a57617/Cargo.toml#L8-L11

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
```diff
@@ -0,0 +1,186 @@
...
+            return;
+        }
+    }
+
+    // Verify the decrypted message matches the original
+    assert_eq!(
+        input.message, secret_message,
+        "Decrypted message does not match original"
+    );
+});
```
plebhash, line 186: `step_1` and `step_2` each read wall-clock time internally

https://github.com/stratum-mining/stratum/blob/2c70113b7dc7111d90e7cd4227e90474899ff4d7/sv2/noise-sv2/src/responder.rs#L270-L277

https://github.com/stratum-mining/stratum/blob/2c70113b7dc7111d90e7cd4227e90474899ff4d7/sv2/noise-sv2/src/initiator.rs#L295-L303

if `cert_validity` is small (e.g.: 0 or 1) a legitimate handshake can fail purely due to elapsed time between those calls, then panic as if it were a protocol bug.

---

this is not necessarily a comment on fuzz harness, but perhaps on what `noise_sv2` APIs should really allow for

because ultimately, what's the point in a 0s certificate validity time?

lucasbalieiro: I'm still thinking about this but just to add to the conversation, we added in https://github.com/stratum-mining/stratum/pull/2017 a time leeway of 10 seconds of tolerance.

So even if the fuzzer or the user sets to 0, the lib allow for 10 secs of tolerance

plebhash: hmm good point... so maybe that already provides safety guards against these edge cases?

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
```diff
@@ -0,0 +1,186 @@
...
+// Corrupts a message by XORing a byte at the specified index with 0xFF.
+// This simulates bit-flipping noise on the wire, testing error handling.
+//
+// # Arguments
+// * `should_corrupt` - If true, performs the corruption
+// * `corruption_index` - Index of byte to flip (wrapped via modulo)
+// * `message` - Mutable slice to corrupt in-place
+fn maybe_corrupt_message(should_corrupt: bool, corruption_index: usize, message: &mut [u8]) {
+    if should_corrupt {
+        let index = corruption_index % message.len();
```
bit-aloo, line 64: This can panic if the message length is 0

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
lucasbalieiro: > hmm good point... so maybe that already provides safety guards against these edge cases?

I’d say yes.

To expand on what we discussed in person, we could also address this by fuzzing the APIs that receive a `now` parameter, such as `noise_sv2::initiator::step_2_with_now` and `noise_sv2::responder::step_1_with_now_rng`.

If the `now` values between the initiator and responder (i.e., client and server) diverge too much, the code will raise an `InvalidCertificate` error. This effectively simulates what happened in the Docker issue (#2015), which later triggered the time-leeway change in PR #2017. In other words, the fuzzer would be exercising scenarios where the client and server clocks differ significantly.

Additionally, if the fuzzer happens to find an input that causes the API to take more than 10 seconds between `step_1` and `step_2`, this will be logged automatically. That gives us a concrete artifact we can inspect later. For example:

```bash
INFO: seed corpus: files: 158 min: 25b max: 3697b total: 42551b rss: 36Mb
#2	pulse  ft: 984 exec/s: 0 rss: 38Mb
Slowest unit: 11 s:
artifact_prefix='/Users/lucasbalieiro/dev/stratum/fuzz/artifacts/fuzz_noise_handshake_and_roundtrip_encryption/'; Test unit written to /Users/lucasbalieiro/dev/stratum/fuzz/artifacts/fuzz_noise_handshake_and_roundtrip_encryption/slow-unit-eca8b7ae09164aae73d4e2b93ac0c547f70d2a57
Base64: kpKSkpJubW1tbW1tjnOSkpKSkpKSkgoliw==
```

That said, I don't think we will gain much by using the `with_now` APIs

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
lucasbalieiro: 0_0, thanks, addressed

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
plebhash: cool, resolving then

sorry for the noise (🥁)

fuzz/fuzz_targets/fuzz_noise_handshake_and_roundtrip_encryption.rs
bit-aloo: What is the change? I still see we are using "%" with message length?

lucasbalieiro: sorry, the change got stucked locally:

image


now we go!

PR #2109, opened 2026-03-04
Fix version mask unvalid len

### LINKED ISSUE #2108: subtraction underflow occurs when `mining.set_version_mask` is greater then 8 bytes
### summary

While fuzzing SV1 RPC, an arithmetic overflow was found in `HexU32Be::try_from(&str)` when parsing a `mining.set_version_mask` message with a mask value longer than 8 hex characters.

The overflow occurs here:
https://github.com/stratum-mining/stratum/blob/2a52ad5beb3a3eb23e7879604ea71f4719a57617/sv1/src/utils.rs#L115

`expected_len` is a hardcoded `usize` with value `8`. If the input string is longer than 8 characters, the subtraction underflows and panics in debug builds.

---

### reproduction

Apply this diff to `sv1/src/lib.rs` to introduce a test with a `set_version_mask` longer then 8 bytes:
```diff
diff --git a/sv1/src/lib.rs b/sv1/src/lib.rs
index 2b608490..72d497b9 100644
--- a/sv1/src/lib.rs
+++ b/sv1/src/lib.rs
@@ -720,4 +720,16 @@ mod tests {
             other => panic!("Expected Error::Method, got {:?}", other),
         }
     }
+
+    #[test]
+    fn version_mask_invalid_len() {
+        let raw = serde_json::json!([
+            "mining.set_version_mask",
+            ["123456789"] // len > 8 bytes
+        ]);
+        let msg: Result<Message, _> = serde_json::from_value(raw);
+        if let Ok(msg) = msg {
+            let _ = Method::try_from(msg);
+        }
+    }
 }

```

Run the test:
```
cargo test version_mask_invalid_len
```

Expected result:
```
test tests::version_mask_invalid_len ... FAILED

failures:

---- tests::version_mask_invalid_len stdout ----

thread 'tests::version_mask_invalid_len' (26438) panicked at sv1/src/utils.rs:115:25:
attempt to subtract with overflow


failures:
    tests::version_mask_invalid_len
```

---

### fix

Add an explicit length check before the subtraction:
```diff
impl TryFrom<&str> for HexU32Be {
    fn try_from(value: &str) -> Result<Self, Error<'static>> {
        let expected_len = 8;
+        if value.len() > expected_len {
+            return Err(...);
+        }
        let delta_len = expected_len - value.len();
        ...
    }
}
```

This converts a panic into an explicit, recoverable error.

exd02:
Fixes #2108 

This PR introduces a new error variant, `Error::InvalidHexLen`, to handle cases where `mining.set_version_mask` exceeds 8 bytes, which previously resulted in an underflow.

The validation is added in `HexU32Be::try_from`, returning `InvalidHexLen` when the input length is greater than the expected 8 hex characters.

A test is included to ensure that a `mining.set_version_mask` message with a length greater than 8 bytes fails with `ParsingMethodError::InvalidHexLen`.

### DIFF
--- a/sv1/src/lib.rs
+++ b/sv1/src/lib.rs
@@ -720,4 +720,26 @@ mod tests {
             other => panic!("Expected Error::Method, got {:?}", other),
         }
     }
+
+    #[test]
+    fn version_mask_invalid_len() {
+        let raw = serde_json::json!([
+            "mining.set_version_mask",
+            ["123456789"] // len > 8 bytes
+        ]);
+
+        let msg: Result<Message, _> = serde_json::from_value(raw);
+
+        if let Ok(msg) = msg {
+            let result = Method::try_from(msg);
+            assert!(result.is_err(), "Expected error for invalid hex length");
+
+            match result.unwrap_err() {
+                MethodError::ParsingMethodError((ParsingMethodError::InvalidHexLen(_), _)) => {}
+                other => panic!("Expected InvalidHexLen, got {:?}", other),
+            }
+        } else {
+            panic!("Message parsing failed unexpectedly");
+        }
+    }
 }


sv1/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sv1_api"
-version = "2.1.3"
+version = "3.0.0"
```
GitGab19, line 3: Why did you bump MAJOR here? Was semver CI check complaining?

exd02: Yes, it was asking for major: `Summary semver requires new major version: 1 major and 0 minor checks failed`. Should I change it?

GitGab19: It might be because you're adding a new error variant.

It's fine for me 👍

GitGab19 (approved):

PR #2118, opened 2026-03-16
channels_sv2: clamp target to max_target instead of erroring

vnprc:
## Problem

When a pool's vardiff computes a target easier than the client's declared
`max_target`, `update_channel()` returns `Err(RequestedMaxTargetOutOfRange)`.
This silently breaks vardiff for the affected channel: the error is swallowed by
the caller, no difficulty update is applied, and the miner is permanently stuck
at its last successfully-set difficulty. The pool logs a `WARN` roughly every
60 seconds per affected channel while mining continues at the wrong difficulty.

This is not a hypothetical edge case. It triggers reliably for low-hashrate
miners whose natural vardiff target overshoots their declared `max_target`.

## Root Cause

The `max_target` field in `OpenExtendedMiningChannel` means *"the easiest
difficulty I will accept."* When vardiff computes a target easier than
`max_target`, the correct response is to assign exactly `max_target` — the
client has already declared that difficulty acceptable. The current code errors
instead, which is both incorrect per the spec and operationally harmful.

## Fix

Replace the `RequestedMaxTargetOutOfRange` error path with a clamp in four
places:

- `ExtendedChannel::new()` — channel creation
- `ExtendedChannel::update_channel()` — vardiff updates
- `StandardChannel::new()` — channel creation
- `StandardChannel::update_channel()` — vardiff updates

Also removes two stray `println!` debug statements from `ExtendedChannel::new()`.

## Tests

The existing `test_update_channel` tests in both `extended.rs` and `standard.rs`
already exercise this path (`very_small_hashrate = 0.1`). They previously
asserted `is_err()` / `RequestedMaxTargetOutOfRange`. They now assert `is_ok()`
and verify the channel target was clamped to `not_so_permissive_max_target`.
All 50 unit tests pass.

## Notes

- `RequestedMaxTargetOutOfRange` is preserved in both error enums — it remains
reachable from the `UpdateChannel` message handler path where the client itself
sends a bad `max_target`.
- No API surface changes (no new public types, no signature changes).

## Pre-existing CI failure

`cargo clippy` currently fails on upstream `main` due to `needless_lifetimes` in
`binary_sv2` — unrelated to this PR.

---

companion https://github.com/stratum-mining/sv2-apps/pull/447

### DIFF
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -346,8 +346,13 @@ where
 
     /// Updates channel configuration with a new nominal hashrate.
     ///
-    /// Adjusts target difficulty and internal state. Returns an error if
-    /// any input parameters are invalid or constraints are violated.
+    /// Recomputes target difficulty and updates channel state.
+    ///
+    /// If the recomputed target is easier than the effective `requested_max_target`,
+    /// the target is clamped to `requested_max_target`.
+    ///
+    /// Returns [`ExtendedChannelError::InvalidNominalHashrate`] when
+    /// `new_nominal_hashrate` cannot be converted into a valid target.
     ///
     /// This can be used in two scenarios:
     /// - Client sent `UpdateChannel` message, which contains a `requested_max_target` parameter
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -208,11 +208,10 @@ where
                 }
             };
 
-        if target > max_target {
-            println!("target: {:?}", target.to_be_bytes());
-            println!("max_target: {:?}", max_target.to_be_bytes());
-            return Err(ExtendedChannelError::RequestedMaxTargetOutOfRange);
-        }
+        // Clamp to max_target rather than error. The client declared max_target as
+        // an acceptable difficulty floor, so using it when the initial target would
+        // otherwise exceed it is always valid.
+        let target = target.min(max_target);
 
         if extranonce_prefix.len() > MAX_EXTRANONCE_PREFIX_LEN {
             return Err(ExtendedChannelError::ExtranoncePrefixTooLarge);
@@ -396,11 +395,10 @@ where
             bytes_to_hex(&max_target_bytes)
         );
 
-        let new_target: Target = target;
-
-        if new_target > *requested_max_target {
-            return Err(ExtendedChannelError::RequestedMaxTargetOutOfRange);
-        }
+        // Clamp to max_target rather than error. The client declared max_target as
+        // an acceptable difficulty floor, so using it when vardiff would otherwise
+        // exceed it is always valid.
+        let new_target = target.min(*requested_max_target);
 
         self.nominal_hashrate = new_nominal_hashrate;
         self.target = new_target;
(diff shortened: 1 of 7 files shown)


Alkamal01:
Tested this locally and the math checks out. Moving from an error to a clamp in `update_channel` is the right move for vardiff—it stops the lock-up for smaller miners without touching the high-end.

I did a quick check on the f64 limits too. Since the clamp is unidirectional (target>max_target), it won't trigger for high-hashrate ASICs, so no new precision loss there. Also, good call on dropping those `println!` calls in the constructor; we don't need the I/O lag in high-concurrency.

All 50 `channels_sv2` tests passed. Concept is solid. LGTM.

vnprc:
Thanks for looking at this @Alkamal01!

plebhash:
thanks for spotting this @vnprc 

apologies for taking so long to review it, this is a great improvement!

overall the PR is pretty solid, but there's a few things left... I'm taking the liberty of implementing them to make the review cycles a bit more agile, can you rebase against `main` and cherry pick these commits from my fork here?
- 4894c9859bd4950ec669351c987ada720fe6c239
- bde2a2e09bb6f276bb457250aa9a36982c4fc4d1
- 19b6c86d9b361477621e3fd9d3c4d18d7a3672ac

plebhash:
> can you rebase against main and cherry pick these commits from my fork here?

actually nevermind @vnprc , I just realized I'm able to push here

plebhash (approved):

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -208,11 +208,14 @@ where
...
-            return Err(ExtendedChannelError::RequestedMaxTargetOutOfRange);
-        }
+        // Clamp to max_target rather than error. The client declared max_target as
+        // an acceptable difficulty floor, so using it when the initial target would
+        // otherwise exceed it is always valid.
+        let target = if target > max_target {
+            max_target
+        } else {
+            target
+        };
```
bit-aloo:
```suggestion
        let target = target.min(max_target);
```

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -396,11 +404,14 @@ where
...
-            return Err(ExtendedChannelError::RequestedMaxTargetOutOfRange);
-        }
+        // Clamp to max_target rather than error. The client declared max_target as
+        // an acceptable difficulty floor, so using it when vardiff would otherwise
+        // exceed it is always valid.
+        let new_target: Target = if target > *requested_max_target {
+            *requested_max_target
+        } else {
+            target
+        };
```
bit-aloo:
```suggestion
        let new_target = target.min(*requested_max_target);
```

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -195,11 +195,14 @@ where
...
-            return Err(StandardChannelError::RequestedMaxTargetOutOfRange);
-        }
+        // Clamp to max_target rather than error. The client declared max_target as
+        // an acceptable difficulty floor, so using it when the initial target would
+        // otherwise exceed it is always valid.
+        let target: Target = if calculated_target > requested_max_target {
+            requested_max_target
+        } else {
+            calculated_target
+        };
```
bit-aloo:
```suggestion
        let new_target = calculated_target.min(requested_max_target);
```

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -344,11 +352,14 @@ where
...
-            return Err(StandardChannelError::RequestedMaxTargetOutOfRange);
-        }
+        // Clamp to max_target rather than error. The client declared max_target as
+        // an acceptable difficulty floor, so using it when vardiff would otherwise
+        // exceed it is always valid.
+        let new_target: Target = if target > requested_max_target {
+            requested_max_target
+        } else {
+            target
+        };
```
bit-aloo:
```suggestion
        let new_target = target.min(requested_max_target);
```

vnprc:
awesome! thanks for looking at it! lmk if i can help in any way.

PR #2134, opened 2026-04-17
Make sv1 server trait methods return results

bit-aloo:
Part of: https://github.com/stratum-mining/sv2-apps/issues/436

Luckily sv1 trait error type already return UnknownID when the downstream is not present, so we can use that to return the error for scenario mentioned in the issue above.

companion https://github.com/stratum-mining/sv2-apps/pull/437

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -552,6 +552,8 @@ dependencies = [
  "buffer_sv2",
  "criterion",
  "noise_sv2",
+ "quickcheck",
+ "quickcheck_macros",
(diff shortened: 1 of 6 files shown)


plebhash (approved):

GitGab19 (commented):
Don't we need a MAJOR bump of `sv1_api` here?

Shourya742:
> 41142f7

We should, done here: 41142f7

GitGab19 (approved):

PR #2138, opened 2026-04-22
represent share work as `u64` (where appropriate)

### LINKED ISSUE #2092: need to switch `ShareAccounting` work representation from `f64` to `u64` on `channels_sv2`
when `channels_sv2` was originally written, I took some inspiration from Braiins UI, which represents work as floating point numbers

I also remember contemplating the fact that strictly speaking, work is a unit of difficulty, and therefore it can (and should) be tracked as floating point numbers (like rust bitcoin does)

this is specially true when we're operating at diff ranges `0 < diff < 1`, where we have no relevant data unless we're working with floating point numbers

but as we noted on https://github.com/stratum-mining/stratum/pull/2090#discussion_r2825911382, this was kind of a failed attempt at retaining granular data, because `SubmitSharesSuccess.new_shares_sum` is `U64`, which forces everyone to truncate and makes an internal representation based on `f64` pretty much useless

plebhash:
close #2092

companion https://github.com/stratum-mining/sv2-apps/pull/449

---

## client side

work acknowledged by upstream arrives over the wire as `u64` (`SubmitSharesSuccess.new_shares_sum`), but for auditability we now track two values:

- `acknowledged_work_sum` (`u64`): what upstream acknowledged over SV2 messages
- `validated_work_sum` (`f64`): what the client validated locally (tracked by job difficulty)

this keeps wire accounting aligned with protocol types, while still allowing client to detect mismatches between what it validated vs what server acknowledged

## server side

if we truncated every share's work into `u64` before storing it in memory, repeated truncation would cause accumulated accounting drift over time

so on the server side, truncation stays limited to `get_last_batch_work_sum` (used for crafting `SubmitSharesSuccess`), while in-memory work accounting remains `f64`

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -236,7 +236,7 @@ impl<'a> ExtendedChannel<'a> {
     pub fn on_share_acknowledgement(
         &mut self,
         new_submits_accepted_count: u32,
-        new_shares_sum: f64,
+        new_shares_sum: u64,
     ) {
         self.share_accounting
             .on_share_acknowledgement(new_submits_accepted_count, new_shares_sum);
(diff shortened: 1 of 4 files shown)


GitGab19 (approved):

plebhash:
reverting back to draft, see: https://github.com/stratum-mining/stratum/issues/2092#issuecomment-4317129404

plebhash:
> reverting back to draft, see: [#2092 (comment)](https://github.com/stratum-mining/stratum/issues/2092#issuecomment-4317129404)

adapted to the new direction

now client side keeps track of `acknowledged_work_sum: u64` vs `validated_work_sum: f64`

server side only truncates on `get_last_batch_work_sum` for the creation of `SubmitSharesSuccess` message

GitGab19 (approved):
LGTM

PR #2139, opened 2026-04-22
channels_sv2::client::share_accounting keeps track of `error_code` for rejected shares

### LINKED ISSUE #2132: `channels_sv2::client::share_accounting` should keep track of `SubmitShare.Error.error_code`
```diff
pub struct ShareAccounting {
    last_share_sequence_number: u32,
    acknowledged_shares: u32,
    validated_shares: u32,
-   rejected_shares: u32,
+   rejected_shares: HashMap<String, u32>
    share_work_sum: f64,
    seen_shares: HashSet<Hash>,
    best_diff: f64,
    blocks_found: u32,
}
```

every time a `SubmitShares.Error` arrives to an Sv2 client, the specific string in `error_code` is incremented in `rejected_shares_error_codes`

that way, UI can display a more detailed view on the reasons that are leading to share rejection

plebhash:
close #2132

companion https://github.com/stratum-mining/sv2-apps/pull/450

### DIFF
--- a/sv2/channels-sv2/src/client/share_accounting.rs
+++ b/sv2/channels-sv2/src/client/share_accounting.rs
@@ -4,7 +4,9 @@
 //! statistics, and reporting share validation results and errors. These abstractions
 //! are intended for use in Mining Clients.
 
-use super::HashSet;
+extern crate alloc;
+use super::{HashMap, HashSet};
+use alloc::string::String;
 use bitcoin::hashes::sha256d::Hash;
 
 /// The outcome of share validation, as seen by a Mining Client.
@@ -66,7 +68,7 @@ pub struct ShareAccounting {
     last_share_sequence_number: u32,
     acknowledged_shares: u32,
     validated_shares: u32,
-    rejected_shares: u32,
+    rejected_shares: HashMap<String, u32>, // <error_code, count>
     share_work_sum: f64,
     seen_shares: HashSet<Hash>,
     best_diff: f64,
@@ -86,7 +88,7 @@ impl ShareAccounting {
             last_share_sequence_number: 0,
             acknowledged_shares: 0,
             validated_shares: 0,
-            rejected_shares: 0,
+            rejected_shares: HashMap::new(),
             share_work_sum: 0.0,
             seen_shares: HashSet::new(),
             best_diff: 0.0,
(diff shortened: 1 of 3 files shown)


plebhash:
cc @gimballock

sv2/channels-sv2/src/client/share_accounting.rs
```diff
@@ -153,9 +158,9 @@ impl ShareAccounting {
         self.validated_shares
     }
 
-    /// Returns the total number of shares rejected by upstream.
-    pub fn get_rejected_shares(&self) -> u32 {
-        self.rejected_shares
+    /// Returns a reference to the map of rejected shares by error code.
+    pub fn get_rejected_shares(&self) -> &HashMap<String, u32> {
+        &self.rejected_shares
     }
```
bit-aloo, line 164: We should not directly allow access to HashMap. We can have three alternative API.

```rust
pub fn rejected_count(&self, error: &str) -> u32 {
    self.rejected_shares.get(error).copied().unwrap_or(0)
}

pub fn total_rejected(&self) -> u32 {
    self.rejected_shares.values().sum()
}

pub fn rejected_shares(&self) -> impl Iterator<Item = (&str, u32)> {
    self.rejected_shares
        .iter()
        .map(|(k, v)| (k.as_str(), *v))
}
```

plebhash: I agree with your point in the sense of API hygiene... the main downside of this is the fact that we're exposing internal representation over public APIs, which makes the crate more prone to avoidable breaking changes on the long term

but I would prefer not doing this change on this PR for a few reasons:

- there's nothing objectively wrong with this approach, it's just ergonomics and maintainability
- this pattern is already established in many other parts of the crate, for which similar adaptations would require a much broader companion PR on `sv2-apps`

therefore I'm creating issue https://github.com/stratum-mining/stratum/issues/2140 so we can implement this improvement later

bit-aloo: Cool, we can do it separately

bit-aloo (approved):

sv2/channels-sv2/src/client/share_accounting.rs
```diff
@@ -66,7 +68,7 @@ pub struct ShareAccounting {
     last_share_sequence_number: u32,
     acknowledged_shares: u32,
     validated_shares: u32,
-    rejected_shares: u32,
+    rejected_shares: HashMap<String, u32>,
```
GitGab19: Can you add a comment to clarify what the String and u32 represent?

bit-aloo: Hmm… I think having wrapper types here (in fact, across all of `channels_sv2`) would make sense. Since these are plain types, it can sometimes be difficult to understand them what they represent.

plebhash: added a comment on this line

about wrapper types, let's do that later: https://github.com/stratum-mining/stratum/issues/2140#issuecomment-4317173281

bit-aloo (approved):
ACK

GitGab19 (approved):

PR #2148, opened 2026-05-07
replace vardiff hardcoded threshold ladder with parametric noise floor

gimballock:
Compute the should-update threshold live from `delta_time` and `shares_per_minute` as a 99% predictive Poisson bound on the share count, plus a 5pp safety margin. Replaces the hardcoded 100/60/50/45/30/15 rungs which were calibrated to a different statistical regime than the default share rate produces, so most rungs sat *inside* the Poisson noise band and fired on sampling noise rather than real hashrate changes.

Drops the unconditional >=100% rung specifically: at small `delta_time` (16-30s window) the noise band can span beyond +-100%, so this rung also fired on noise. The `>1000%` Phase 1 ramp clamp is unaffected and still handles miners starting far from truth.

Tests:
- Replace the multi-step `with_less_spm_than_expected` check (calibrated to old rungs) with a single sustained-underperformance scenario.
- Drop `no_shares_less_than_30s_decrease` — case is intentionally unreachable at typical share rates.
- Bump `simulation_duration` in `no_shares_30_to_60s_decrease` (31 -> 55) and `hashrate_clamps_to_minimum` (16 -> 60) so they sit outside the new noise floor.
- Add `no_fire_within_noise` and `fires_outside_noise` covering the core property of the change.

### DIFF
--- a/sv2/channels-sv2/src/vardiff/classic.rs
+++ b/sv2/channels-sv2/src/vardiff/classic.rs
@@ -5,7 +5,37 @@ use tracing::debug;
 /// Default minimum hashrate (H/s) if not specified.
 const DEFAULT_MIN_HASHRATE: f32 = 1.0;
 
+/// Two-sided 99% normal quantile, used for the predictive Poisson noise floor.
+const Z_99_TWO_SIDED: f64 = 2.576;
+
+/// Safety margin (in percentage points) added on top of the noise floor.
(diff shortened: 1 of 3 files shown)


gimballock:
We can close this, I don't expect it to be integrated as-is, it serves more as an example demonstration that optimizations are out there. This may in fact be a good one but a more thorough investigation should drive that decision.

Towards that end I have created PR #2154, which as a follow up can be used to properly evaluate this algorithm.

Issue #2142, opened 2026-04-30
define types for commonly-used `error_code` strings


plebhash:
this was discussed on Sv2 Working Group call of April 30th 2026 and proposed by @coleFD 

this is tied to https://github.com/stratum-mining/sv2-spec/pull/194 and https://github.com/stratum-mining/sv2-spec/discussions/165

the core concepts here are:
- the spec allows for implementations to take automated actions on `error_code` strings
- the spec does not force implementations to use specific `error_code` strings, only provides suggestions for common scenarios

so we end up in a situation where different implementations could be aiming for the same `error_code` string, but still be subject to small typo or encoding issues

while that is hard to avoid without major spec changes, SRI can greatly reduce the room for such scenarios by providing types for commonly-used `error_code` strings

these should be provided on their respective message-crates:
- `common_messages_sv2`
- `mining_sv2`
- `job_declaration_sv2`
- `template_distribution_sv2`

plebhash:
we should add constants for all `error_code` strings that are currently used on `sv2-apps`

plebhash:
tbh, I don't think we need to wait for https://github.com/stratum-mining/sv2-spec/pull/194 to be merged

as is, SRI is already diverging from the spec sentence discussed in https://github.com/stratum-mining/sv2-spec/discussions/165, so the improvement proposed here will not change anything with regards to that

---

moreover, after considering @GitGab19 's suggestion https://github.com/stratum-mining/stratum/pull/2149#discussion_r3224814100 I think addressing this first would be a good pre-requisite for a more concise approach in #2149

I'll start tackling this now

PR #2149, opened 2026-05-08
`channels_sv2::server::share_accounting` keeps track of rejected shares

plebhash:
close #2119 

companion https://github.com/stratum-mining/sv2-apps/pull/491

### DIFF
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -671,6 +671,8 @@ where
         let is_stale_job = self.job_store.get_stale_job(job_id).is_some();
 
         if is_stale_job {
+            self.share_accounting
+                .increment_rejected_shares(ERROR_CODE_SUBMIT_SHARES_STALE_SHARE);
             return Err(ShareValidationError::Stale(
                 ERROR_CODE_SUBMIT_SHARES_STALE_SHARE,
             ));
@@ -678,6 +680,8 @@ where
 
         // if job_id is not active, past or stale, return error
         if !is_active_job && !is_past_job && !is_stale_job {
+            self.share_accounting
+                .increment_rejected_shares(ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID);
             return Err(ShareValidationError::InvalidJobId(
                 ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID,
             ));
@@ -696,7 +700,6 @@ where
                 .get_stale_job(job_id)
                 .expect("stale job must exist")
         };
-
         let job_target = self
             .job_id_to_target
             .get(&job_id)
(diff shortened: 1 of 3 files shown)


sv2/channels-sv2/src/server/extended.rs
```diff
@@ -435,6 +435,11 @@ where
         &self.share_accounting
     }
 
+    /// Updates rejected-share accounting for an emitted `SubmitShares.Error`.
+    pub fn increment_rejected_shares(&mut self, error_code: String) {
```
gimballock: miner nit: this could be &str instead of String to avoid the memory alloc

plebhash: thanks! addressed

bit-aloo (approved):
ACK

sv2/channels-sv2/src/server/extended.rs
```diff
@@ -435,6 +435,11 @@ where
         &self.share_accounting
     }
 
+    /// Updates rejected-share accounting for an emitted `SubmitShares.Error`.
+    pub fn increment_rejected_shares(&mut self, error_code: &str) {
```
GitGab19: Shouldn't this be called internally by `validate_share()`, in case of an error?

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -410,6 +410,11 @@ where
         &self.share_accounting
     }
 
+    /// Updates rejected-share accounting for an emitted `SubmitShares.Error`.
+    pub fn increment_rejected_shares(&mut self, error_code: &str) {
```
GitGab19: Shouldn't this be called internally by `validate_share()`, in case of an error?

sv2/channels-sv2/src/server/extended.rs
plebhash: yeah I think this would be a more concise approach, since we would avoid the need for the app layer to explicitly call this

(or more importantly, we would eliminate the possibility of the app layer forgetting to call this while it should)

plebhash: but while trying to implement this approach, I actually started to think that it might make sense to tackle #2142 first

that way, `validate_share` could return `ShareValidationError` variants that already carry the needed `error_code` embedded into itself

the app layer would then be able to use those `error_code` without having to think too much about it

plebhash: btw I don't think we need to wait for https://github.com/stratum-mining/sv2-spec/pull/194 before we tackle #2142

as is, SRI is already diverging from the spec sentence discussed in https://github.com/stratum-mining/sv2-spec/discussions/165, so the doing #2142 will not change anything with regards to that

plebhash: reverting this to draft while I work on #2142

plebhash: done, now those are called from inside `validate_share`

GitGab19: Awesome!

GitGab19 (approved):

PR #2151, opened 2026-05-12
add `error_code` constants

### LINKED ISSUE #2142: define types for commonly-used `error_code` strings

plebhash:
close #2142

companion https://github.com/stratum-mining/sv2-apps/pull/495

### DIFF
--- a/sv2/channels-sv2/src/server/error.rs
+++ b/sv2/channels-sv2/src/server/error.rs
@@ -2,14 +2,24 @@
 
 use crate::server::jobs::error::JobFactoryError;
 
+/// Errors that can occur while operating an extended channel on the server side.
+///
+/// Variants carrying `&'static str` are intended to be used as `error_code` values in protocol
+/// error messages (for example
+/// [`OpenMiningChannelError`](mining_sv2::OpenMiningChannelError) and
+/// [`UpdateChannelError`](mining_sv2::UpdateChannelError)).
+///
+/// Variants without `&'static str` SHOULD lead to a client disconnection or application
+/// shutdown.
 #[derive(Debug)]
 pub enum ExtendedChannelError {
+    OpenChannelInvalidNominalHashrate(&'static str),
+    UpdateChannelInvalidNominalHashrate(&'static str),
+    RequestedMinExtranonceSizeTooLarge(&'static str),
     JobFactoryError(JobFactoryError),
-    InvalidNominalHashrate,
     ChainTipNotSet,
     TemplateIdNotFound,
     JobIdNotFound,
-    RequestedMinExtranonceSizeTooLarge,
     ExtranoncePrefixTooLarge,
     ScriptSigSizeTooLarge,
     InvalidJobOrigin,
@@ -24,10 +34,20 @@ pub enum GroupChannelError {
     ScriptSigSizeTooLarge,
 }
 
+/// Errors that can occur while operating a standard channel on the server side.
+///
+/// Variants carrying `&'static str` are intended to be used as `error_code` values in protocol
+/// error messages (for example
+/// [`OpenMiningChannelError`](mining_sv2::OpenMiningChannelError) and
+/// [`UpdateChannelError`](mining_sv2::UpdateChannelError)).
+///
+/// Variants without `&'static str` SHOULD lead to a client disconnection or application
+/// shutdown.
 #[derive(Debug)]
 pub enum StandardChannelError {
+    OpenChannelInvalidNominalHashrate(&'static str),
+    UpdateChannelInvalidNominalHashrate(&'static str),
     TemplateIdNotFound,
-    InvalidNominalHashrate,
     ExtranoncePrefixTooLarge,
     JobFactoryError(JobFactoryError),
     ChainTipNotSet,
--- a/sv2/subprotocols/job-declaration/src/lib.rs
+++ b/sv2/subprotocols/job-declaration/src/lib.rs
@@ -46,3 +46,13 @@ pub const CHANNEL_BIT_DECLARE_MINING_JOB_ERROR: bool = false;
 pub const CHANNEL_BIT_PROVIDE_MISSING_TRANSACTIONS: bool = false;
 pub const CHANNEL_BIT_PROVIDE_MISSING_TRANSACTIONS_SUCCESS: bool = false;
 pub const CHANNEL_BIT_PUSH_SOLUTION: bool = true;
+
+// Commonly used DeclareMiningJobError error_code values.
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_MINING_JOB_TOKEN: &str = "invalid-mining-job-token";
+pub const ERROR_CODE_DECLARE_MINING_JOB_MISSING_TXS: &str = "missing-txs";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_COINBASE_TX: &str = "invalid-coinbase-tx";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_COINBASE_TX_INPUT: &str =
+    "invalid-coinbase-tx-input";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR: &str = "internal-error";
+pub const ERROR_CODE_DECLARE_MINING_JOB_STALE_CHAIN_TIP: &str = "stale-chain-tip";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB: &str = "invalid-job";
--- a/sv2/subprotocols/mining/src/lib.rs
+++ b/sv2/subprotocols/mining/src/lib.rs
@@ -98,5 +98,55 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
 pub const CHANNEL_BIT_SUBMIT_SHARES_EXTENDED: bool = true;
 pub const CHANNEL_BIT_SUBMIT_SHARES_STANDARD: bool = true;
 pub const CHANNEL_BIT_SUBMIT_SHARES_SUCCESS: bool = true;
+
+// Commonly used OpenMiningChannelError error_code values.
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_STANDARD_CHANNELS_NOT_SUPPORTED_FOR_CUSTOM_WORK: &str =
+    "standard-channels-not-supported-for-custom-work";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_INVALID_USER_IDENTITY: &str = "invalid-user-identity";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_INVALID_NOMINAL_HASHRATE: &str =
+    "invalid-nominal-hashrate";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_MIN_EXTRANONCE_SIZE_TOO_LARGE: &str =
+    "min-extranonce-size-too-large";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_MAX_TARGET_OUT_OF_RANGE: &str = "max-target-out-of-range";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_UNSUPPORTED_MIN_EXTRANONCE_SIZE: &str =
+    "unsupported-min-extranonce-size";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_UNKNOWN_USER: &str = "unknown-user";
+
+// Commonly used UpdateChannelError error_code values.
+pub const ERROR_CODE_UPDATE_CHANNEL_INVALID_NOMINAL_HASHRATE: &str = "invalid-nominal-hashrate";
+pub const ERROR_CODE_UPDATE_CHANNEL_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+
+// Commonly used SubmitSharesError error_code values.
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE: &str = "invalid-share";
+pub const ERROR_CODE_SUBMIT_SHARES_STALE_SHARE: &str = "stale-share";
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID: &str = "invalid-job-id";
+pub const ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW: &str = "difficulty-too-low";
+pub const ERROR_CODE_SUBMIT_SHARES_DUPLICATE_SHARE: &str = "duplicate-share";
+pub const ERROR_CODE_SUBMIT_SHARES_BAD_EXTRANONCE_SIZE: &str = "bad-extranonce-size";
+pub const ERROR_CODE_VERSION_ROLLING_NOT_ALLOWED: &str = "version-rolling-not-allowed";
+
+// Commonly used SetCustomMiningJobError error_code values.
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_JD_NOT_SUPPORTED: &str = "jd-not-supported";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN: &str =
+    "invalid-mining-job-token";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_JOB_NOT_YET_VALIDATED: &str = "job-not-yet-validated";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_STALE_CHAIN_TIP: &str = "stale-chain-tip";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MIN_NTIME: &str = "invalid-min-ntime";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_NBITS: &str = "invalid-nbits";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_VERSION: &str = "invalid-version";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_COINBASE_TX: &str = "invalid-coinbase-tx";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_COINBASE_TX_VERSION: &str =
+    "invalid-coinbase-tx-version";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_COINBASE_PREFIX: &str =
+    "invalid-coinbase-prefix";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_COINBASE_TX_INPUT_N_SEQUENCE: &str =
+    "invalid-coinbase-tx-input-n-sequence";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_COINBASE_TX_OUTPUTS: &str =
+    "invalid-coinbase-tx-outputs";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_COINBASE_TX_LOCKTIME: &str =
+    "invalid-coinbase-tx-locktime";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MERKLE_PATH: &str = "invalid-merkle-path";
 pub const CHANNEL_BIT_UPDATE_CHANNEL: bool = true;
 pub const CHANNEL_BIT_UPDATE_CHANNEL_ERROR: bool = true;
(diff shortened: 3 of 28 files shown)


sv2/subprotocols/job-declaration/src/lib.rs
```diff
@@ -46,3 +46,14 @@ pub const CHANNEL_BIT_DECLARE_MINING_JOB_ERROR: bool = false;
...
+// Commonly used DeclareMiningJobError error_code values.
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_MINING_JOB_TOKEN: &str = "invalid-mining-job-token";
+pub const ERROR_CODE_DECLARE_MINING_JOB_MISSING_TXS: &str = "missing-txs";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX: &str =
+    "invalid-job-param-value-coinbase-tx";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX_INPUT: &str =
+    "invalid-job-param-value-coinbase-tx-input";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR: &str = "internal-error";
+pub const ERROR_CODE_DECLARE_MINING_JOB_STALE_CHAIN_TIP: &str = "stale-chain-tip";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB: &str = "invalid-job";
```
GitGab19, line 58: Where are these in the specs? I'm looking at the JD chapter, and I only see this sentence:

```
Possible error codes:
   - invalid-mining-job-token
   - invalid-job-param-value-{} - {} is replaced by a particular field name from DeclareMiningJob message
```

sv2/subprotocols/mining/src/lib.rs
```diff
@@ -98,5 +98,59 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
...
+    "standard-channels-not-supported-for-custom-work";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_INVALID_USER_IDENTITY: &str = "invalid-user-identity";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_INVALID_NOMINAL_HASHRATE: &str =
+    "invalid-nominal-hashrate";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_MIN_EXTRANONCE_SIZE_TOO_LARGE: &str =
+    "min-extranonce-size-too-large";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_MAX_TARGET_OUT_OF_RANGE: &str = "max-target-out-of-range";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_UNSUPPORTED_MIN_EXTRANONCE_SIZE: &str =
+    "unsupported-min-extranonce-size";
+pub const ERROR_CODE_OPEN_MINING_CHANNEL_UNKNOWN_USER: &str = "unknown-user";
```
GitGab19, line 113: In the `OpenMiningChannel.Error` paragraph on the specs I only see this:
```
Possible error codes:
    - unknown-user
    - max-target-out-of-range

```

sv2/subprotocols/mining/src/lib.rs
```diff
@@ -98,5 +98,59 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
...
+
+// Commonly used UpdateChannelError error_code values.
+pub const ERROR_CODE_UPDATE_CHANNEL_INVALID_NOMINAL_HASHRATE: &str = "invalid-nominal-hashrate";
```
GitGab19, line 116: In the paragraph on the specs I see we have `max-target-out-of-range` but no `invalid-nominal-hashrate`

sv2/subprotocols/mining/src/lib.rs
```diff
@@ -98,5 +98,59 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
...
+pub const ERROR_CODE_UPDATE_CHANNEL_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+
+// Commonly used SubmitSharesError error_code values.
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE: &str = "invalid-share";
+pub const ERROR_CODE_SUBMIT_SHARES_STALE_SHARE: &str = "stale-share";
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID: &str = "invalid-job-id";
+pub const ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW: &str = "difficulty-too-low";
+pub const ERROR_CODE_SUBMIT_SHARES_DUPLICATE_SHARE: &str = "duplicate-share";
+pub const ERROR_CODE_SUBMIT_SHARES_BAD_EXTRANONCE_SIZE: &str = "bad-extranonce-size";
```
GitGab19, line 126: On the specs we have:
```
Possible error codes:
    - invalid-channel-id
    - stale-share
    - difficulty-too-low
    - invalid-job-id
```

sv2/subprotocols/mining/src/lib.rs
```diff
@@ -98,5 +98,59 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
...
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_PREFIX: &str =
+    "invalid-job-param-value-coinbase-prefix";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX_INPUT_N_SEQUENCE:
+    &str = "invalid-job-param-value-coinbase-tx-input-n-sequence";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX_OUTPUTS: &str =
+    "invalid-job-param-value-coinbase-tx-outputs";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX_LOCKTIME: &str =
+    "invalid-job-param-value-coinbase-tx-locktime";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_JOB_PARAM_VALUE_MERKLE_PATH: &str =
+    "invalid-job-param-value-merkle-path";
```
GitGab19: On the specs we have:
```
Possible errors:
    - invalid-channel-id
    - invalid-mining-job-token
    - invalid-job-param-value-{} - {} is replaced by a particular field name from SetCustomMiningJob message

```

GitGab19 (commented):
I looked on the specs and reported almost on every "message" the error_code(s) that we have there, to simplify the review and being sure we're covering all we have on the specs.

That said, I think my initial understanding was to define the error codes (as constants) only for the ones which are explicitly defined in the current specs.

But reviewing this PR made me realize that probably your goal was to cover all the codes we have on the specs + everything we are already using at the application level (which very likely would be useful also for some adopters).

Is my understanding correct?

plebhash:
> I looked on the specs and reported almost on every "message" the error_code(s) that we have there, to simplify the review and being sure we're covering all we have on the specs.
> 
> That said, I think my initial understanding was to define the error codes (as constants) only for the ones which are explicitly defined in the current specs.
> 
> But reviewing this PR made me realize that probably your goal was to cover all the codes we have on the specs + everything we are already using at the application level (which very likely would be useful also for some adopters).
> 
> Is my understanding correct?

as of today, SRI already implements `error_code`s that are not explicitly listed on the spec

this PR covers all `error_code`s from current SRI

---

I cannot say I introduced **all** `error_code`s that are not listed on spec, but I do remember introducing many

I've always treated the `error_code`s listed on the spec as merely suggestions, not a comprehensive list

so whenever I was faced with some edge case that the spec didn't explicitly list, I just established the `error_code` on SRI without going through the bureaucracy of proposing spec change (which would severely slow down whatever I was working on SRI)

GitGab19:
I see, and I guess it makes sense.

My question now is: should we have constants for all the error-codes which are currently listed on the specs? Even if they are not currently used by our implementation?

plebhash:
> should we have constants for all the error-codes which are currently listed on the specs? Even if they are not currently used by our implementation?

hmm makes sense, currently this PR is not covering 100% of the `error_code`s listed on the spec, and it should

changing

plebhash:
ok now this PR covers 100% of the `error_code`s defined on spec, plus the ones defined on SRI

sv2/subprotocols/mining/src/lib.rs
```diff
@@ -98,5 +98,60 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
...
+pub const ERROR_CODE_UPDATE_CHANNEL_INVALID_NOMINAL_HASHRATE: &str = "invalid-nominal-hashrate";
+pub const ERROR_CODE_UPDATE_CHANNEL_MAX_TARGET_OUT_OF_RANGE: &str = "max-target-out-of-range";
+pub const ERROR_CODE_UPDATE_CHANNEL_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+
+// Commonly used SubmitSharesError error_code values.
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE: &str = "invalid-share";
+pub const ERROR_CODE_SUBMIT_SHARES_STALE_SHARE: &str = "stale-share";
+pub const ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID: &str = "invalid-job-id";
+pub const ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW: &str = "too-low-difficulty";
```
GitGab19: this should be `difficulty-too-low` (?)

sv2/subprotocols/mining/src/submit_shares.rs
```diff
@@ -129,7 +129,7 @@ pub struct SubmitSharesError<'decoder> {
     ///
     /// - invalid-channel-id
     /// - stale-share
-    /// - difficulty-too-low
+    /// - too-low-difficulty
```
GitGab19: this should be `difficulty-too-low` (?)

sv2/subprotocols/mining/src/lib.rs
GitGab19: Aren't we missing `invalid-job-param-value-min-ntime`?

sv2/subprotocols/job-declaration/src/lib.rs
```diff
@@ -46,3 +46,14 @@ pub const CHANNEL_BIT_DECLARE_MINING_JOB_ERROR: bool = false;
...
 pub const CHANNEL_BIT_PROVIDE_MISSING_TRANSACTIONS_SUCCESS: bool = false;
 pub const CHANNEL_BIT_PUSH_SOLUTION: bool = true;
+
+// Commonly used DeclareMiningJobError error_code values.
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_MINING_JOB_TOKEN: &str = "invalid-mining-job-token";
+pub const ERROR_CODE_DECLARE_MINING_JOB_MISSING_TXS: &str = "missing-txs";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX: &str =
+    "invalid-job-param-value-coinbase-tx";
+pub const ERROR_CODE_DECLARE_MINING_JOB_INVALID_JOB_PARAM_VALUE_COINBASE_TX_INPUT: &str =
+    "invalid-job-param-value-coinbase-tx-input";
```
GitGab19: These should be `coinbase-tx-prefix` and `coinbase-tx-suffix`?

sv2/subprotocols/job-declaration/src/lib.rs
```diff
@@ -46,3 +46,14 @@ pub const CHANNEL_BIT_DECLARE_MINING_JOB_ERROR: bool = false;
...
+pub const ERROR_CODE_DECLARE_MINING_JOB_INTERNAL_ERROR: &str = "internal-error";
+pub const ERROR_CODE_DECLARE_MINING_JOB_STALE_CHAIN_TIP: &str = "stale-chain-tip";
```
GitGab19, line 57: How can we have this error code if there's not chain-tip in the message?

sv2/subprotocols/job-declaration/src/lib.rs
GitGab19: Aren't we missing `invalid-job-param-value-version`, `invalid-job-param-value-wtxid_list`, and `invalid-job-param-value-excess-data` here?

sv2/subprotocols/mining/src/lib.rs
plebhash: yeah that's a good question

my agent pointed out this one:

https://github.com/stratum-mining/sv2-spec/blob/4cdeecd5f97a0c969952e831c17ada50fab6ff36/03-Protocol-Overview.md?plain=1#L227

and I told it to use it... but we also have:

https://github.com/stratum-mining/sv2-spec/blob/4cdeecd5f97a0c969952e831c17ada50fab6ff36/05-Mining-Protocol.md?plain=1#L392

I guess I would just fix the spec so both are `difficulty-too-low`, shouldn't be controversial/bureaucratic?

sv2/subprotocols/job-declaration/src/lib.rs
plebhash: this is an internal implementation detail of `jd_server_sv2::job_declarator::job_validation::bitcoin_core_ipc` + `bitcoin_core_sv2::job_declaration_protocol::handlers`

for example: if there's a `ProvideMissingTransactions` round-trip, and the `MempoolMirror` context chain-tip drifts in the mean time, `DeclareMiningJob` failes with this error code

sv2/subprotocols/mining/src/lib.rs
```diff
@@ -98,5 +98,60 @@ pub const CHANNEL_BIT_SUBMIT_SHARES_ERROR: bool = true;
...
+pub const ERROR_CODE_SUBMIT_SHARES_BAD_EXTRANONCE_SIZE: &str = "bad-extranonce-size";
+
+// Commonly used SetCustomMiningJobError error_code values.
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_JD_NOT_SUPPORTED: &str = "jd-not-supported";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_CHANNEL_ID: &str = "invalid-channel-id";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_MINING_JOB_TOKEN: &str =
+    "invalid-mining-job-token";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_JOB_NOT_YET_VALIDATED: &str = "job-not-yet-validated";
+pub const ERROR_CODE_SET_CUSTOM_MINING_JOB_INVALID_PREV_HASH_LENGTH: &str =
+    "invalid-prev-hash-length";
```
plebhash: this shouldn't be needed

sv2/subprotocols/job-declaration/src/lib.rs
plebhash: all `invalid-job-param-value-*` were changed into `invalid-*`
 
if a specific parameter is invalid, we write that in a straightforward way, no templates or patterns induced by spec

sv2/subprotocols/mining/src/lib.rs
plebhash: yes we are

I'm adding it to this PR (as `invalid-min-ntime`), and taking note of JDS validation gap https://github.com/stratum-mining/sv2-apps/issues/500

sv2/subprotocols/job-declaration/src/lib.rs
plebhash: note: I removed the `invalid-job-param-value-` pattern, so the error messages on this comment now are:

- `invalid-coinbase-tx`

as discussed on the call, this error comes up when `BitcoinCoreIPCEngine::handle_declare_mining_job` fails to deserialize the declared coinbase tx via `rust-bitcoin` APIs

it's kind of a generic error because `rust-bitcoin` APIs don't give us much information about it

- `invalid-coinbase-tx-input`

after the declared coinbase tx has been deserialized on `BitcoinCoreIPCEngine::handle_declare_mining_job`, we check whether it only has 1 single input

otherwise, it's not an actual coinbase tx

---

overall, with regards to explicit references to `coinbase-tx-prefix` and `coinbase-tx-suffix`, since we're no longer following `invalid-job-param-value-` pattern, I think we don't need to cover those?

tbh, I wonder if it would even be possible to cover them

plebhash:
@GitGab19 FYI since your last review I'm doing some additional changes here, which are one extra step aligned with the vision described in https://github.com/stratum-mining/stratum/pull/2149#discussion_r3229125136 

more specifically, with 4d80cc58bdaf70a05ccd7e000654065e84354644 I'm making sure that error `enum`s of `channels_sv2` (e.g.: `ExtendedChannelError`, `StandardChannelError`, and most importantly `ShareValidationError`, all carry a `(&'static str)`

but not all error variants carry a `(&'static str)`, so hopefully Rust Docs comments should make the pattern clear:
```rust
/// Variants carrying `&'static str` are intended to be used as `error_code` values in the response message.
/// 
/// Variants without `&'static str` SHOULD lead to a client disconnection or application
/// shutdown.
/// ...
```

sv2/subprotocols/mining/src/lib.rs
plebhash: yes, changed

sv2/subprotocols/mining/src/submit_shares.rs
plebhash: yes, changed

sv2/subprotocols/job-declaration/src/lib.rs
plebhash: another example is BIP34 validation:

`MempoolMirror` keeps track of the current block height (via BIP34 of the coinbase tx on the template it receives from Bitcoin Core)

every `DeclareMiningJob` message carries `BIP34` as part of `coinbase_tx_prefix`

so if that's behind the one on `MempoolMirror`, `DeclareMiningJob` is rejected with `stale-chain-tip` error

plebhash:
reproducing https://github.com/stratum-mining/stratum/pull/2151#discussion_r3234767483:

> Aren't we missing `invalid-job-param-value-version`, `invalid-job-param-value-wtxid_list`, and `invalid-job-param-value-excess-data` here?

perhaps @GitGab19 's intention was more of a rethorical question to point out the issues we discussed around `invalid-job-param-value-{}` pattern

nevertheless, it might be worth addressing the question for the sake of clarity on this PR

## `DeclareMiningJob.{version,wtxid_list}` 

those parameters are both validated during `checkBlock` IPC call, and on failure result in a generic `error_code = "invalid-job"`

maybe one could argue that those deserve more granular `error_code`s, which is a fair point

however, I don't really have a clear idea if/how that could be easily achieved inside `bitcoin_core_sv2`, and I don't think that should be addressed on the scope of this PR

## `DeclareMiningJob.excess_data`

so far this parameter has never been used on SRI, and I don't know what kind of edge cases could arise from it

sv2/channels-sv2/src/server/error.rs
```diff
@@ -2,10 +2,20 @@
 
 use crate::server::jobs::error::JobFactoryError;
 
+/// Errors that can occur while operating an extended channel on the server side.
+///
+/// Variants carrying `&'static str` are intended to be used as `error_code` values in protocol
+/// error messages (for example
+/// [`OpenExtendedMiningChannelError`](mining_sv2::OpenExtendedMiningChannelError) and
```
GitGab19:
```suggestion
/// [`OpenMiningChannelError`](mining_sv2::OpenMiningChannelError) and
```

sv2/channels-sv2/src/server/error.rs
```diff
@@ -24,10 +34,20 @@ pub enum GroupChannelError {
     ScriptSigSizeTooLarge,
 }
 
+/// Errors that can occur while operating a standard channel on the server side.
+///
+/// Variants carrying `&'static str` are intended to be used as `error_code` values in protocol
+/// error messages (for example
+/// [`OpenStandardMiningChannelError`](mining_sv2::OpenStandardMiningChannelError) and
```
GitGab19:
```suggestion
/// [`OpenMiningChannelError`](mining_sv2::OpenMiningChannelError) and
```

GitGab19 (commented):
Now it looks way better, very good job!

I only left a couple of nits.

sv2/channels-sv2/src/server/error.rs
plebhash: done

GitGab19: I still see the old docs (`OpenStandardMiningChannelError`) here.

plebhash: sorry done

GitGab19 (approved):

PR #2152, opened 2026-05-12
add `stable_hashrate` flag to `channels_sv2::server`

### LINKED ISSUE #2143: `channels_sv2::server` should expose whether `vardiff` has converged (via `stable_hashrate: bool` statevar)
[`channels_sv2::vardiff::classic::VardiffState::try_vardiff`](https://docs.rs/channels_sv2/4.0.0/channels_sv2/vardiff/trait.Vardiff.html#tymethod.try_vardiff) returns:
- `Ok(Some(new_hashrate))` when an update occurs
- `Ok(None)` when conditions don’t warrant an update
- `Err` for actual processing errors

we need to add the following methods to `channels_sv2::server::{standard,extended}`:
- `set_stable_hashrate(bool: stable_hashrate)`
- `get_stable_hashrate() -> bool`

the application layer would be responsible for calling these on its `try_vardiff` tasks, by following this logic:
```rust
match try_vardiff(...) {
  Ok(Some(new_hashrate)) => channel.set_stable_hashrate(false),
  Ok(None) => channel.set_stable_hashrate(true),
  Err(_) => {},
}
```

that way, `stratum-apps` can poll `get_stable_hashrate` to expose HTTP APIs that allow `sv2-ui` to display informations as suggested by @average-gary :

Image

note: I don't think https://github.com/stratum-mining/sv2-apps/issues/396 has much to do with this

at best, making vardiff cycles shorter (which is not a low-hanging fruit on `sv2-apps`) only would make the proposed vardiff progress indicator converge to ✅ a bit faster, which is good but not the point to be unblocked here

plebhash:
close #2143 

companion https://github.com/stratum-mining/sv2-apps/pull/496

### DIFF
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -73,6 +73,7 @@ use tracing::debug;
 /// - the channel's current target
 /// - the channel's mapping between `job_id` and target
 /// - the channel's nominal hashrate
+/// - whether the channel's nominal hashrate is treated as stable
 /// - the channel's [`JobStore`]
 /// - the channel's [`JobFactory`]
 /// - the channel's [`ShareAccounting`]
@@ -92,6 +93,7 @@ where
     target: Target,
     job_id_to_target: HashMap<u32, Target>,
     nominal_hashrate: f32,
+    stable_hashrate: bool,
     job_store: J,
     job_factory: JobFactory,
     share_accounting: ShareAccounting,
@@ -240,6 +242,7 @@ where
             target,
             job_id_to_target: HashMap::new(),
             nominal_hashrate,
+            stable_hashrate: false,
             job_store,
             job_factory: JobFactory::new(version_rolling_allowed, pool_tag, miner_tag),
             share_accounting: ShareAccounting::new(share_batch_size),
@@ -338,6 +341,16 @@ where
     /// Returns the nominal hashrate for this channel.
     pub fn get_nominal_hashrate(&self) -> f32 {
         self.nominal_hashrate
+    }
+
+    /// Sets whether this channel's nominal hashrate should be treated as stable.
+    pub fn set_stable_hashrate(&mut self, stable_hashrate: bool) {
+        self.stable_hashrate = stable_hashrate;
+    }
+
+    /// Returns whether this channel's nominal hashrate is treated as stable.
+    pub fn get_stable_hashrate(&self) -> bool {
+        self.stable_hashrate
     }
 
     /// Updates the nominal hashrate for this channel.
--- a/sv2/channels-sv2/src/server/standard.rs
+++ b/sv2/channels-sv2/src/server/standard.rs
@@ -72,6 +72,7 @@ use tracing::debug;
 /// - the channel's current target
 /// - the channel's mapping between `job_id` and target
 /// - the channel's nominal hashrate
+/// - whether the channel's nominal hashrate is treated as stable
 /// - the channel's [`JobStore`]
 /// - the channel's share accounting state
 /// - the channel's expected share per minute
@@ -89,6 +90,7 @@ where
     target: Target,
     job_id_to_target: HashMap<u32, Target>,
     nominal_hashrate: f32,
+    stable_hashrate: bool,
     share_accounting: ShareAccounting,
     expected_share_per_minute: f32,
     job_store: J,
@@ -225,6 +227,7 @@ where
             target,
             job_id_to_target: HashMap::new(),
             nominal_hashrate,
+            stable_hashrate: false,
             share_accounting: ShareAccounting::new(share_batch_size),
             expected_share_per_minute,
             job_factory: JobFactory::new(true, pool_tag_string, miner_tag_string),
@@ -275,6 +278,16 @@ where
     /// Updates the nominal hashrate for this channel.
     pub fn set_nominal_hashrate(&mut self, nominal_hashrate: f32) {
         self.nominal_hashrate = nominal_hashrate;
+    }
+
+    /// Sets whether this channel's nominal hashrate should be treated as stable.
+    pub fn set_stable_hashrate(&mut self, stable_hashrate: bool) {
+        self.stable_hashrate = stable_hashrate;
+    }
+
+    /// Returns whether this channel's nominal hashrate is treated as stable.
+    pub fn get_stable_hashrate(&self) -> bool {
+        self.stable_hashrate
     }
 
     /// Returns the requested maximum target for this channel.


GitGab19 (approved):

sv2/channels-sv2/src/server/extended.rs
GitGab19, line 75: We need to add the new `stable_hashrate` here.

sv2/channels-sv2/src/server/standard.rs
GitGab19, line 74: We need to add the new `stable_hashrate` here.

GitGab19 (changes requested):
I just noticed we need to add the new `stable_hashrate` field in the channels docs.

sv2/channels-sv2/src/server/extended.rs
plebhash: good catch, thanks

fixed

sv2/channels-sv2/src/server/standard.rs
plebhash: good catch, thanks

fixed

sv2/channels-sv2/src/server/extended.rs
GitGab19: I still don't see it on the `ExtendedChannel`.

plebhash: shit forgot to stage before I amended the commit, sorry

should be good now

GitGab19 (approved):

bit-aloo (commented):
I have some other thoughts on the implementation here. I don't think `stable_hashrate` is the right abstraction to represent vardiff stability. I think associating an `Instant` with the nominal hashrate would be a better way to convey the idea. For example, it could show how long the current hashrate has remained stable, something like “stable since xx minutes.”

It should also be very easy to track through the `update_channel` method, where we already update the nominal hashrate, without requiring any additional changes related to stability tracking.

plebhash:
thanks for the suggestion @Shourya742

how do you see this approach shaping up impact the final UX at `sv2-ui` level?

recently @pavlenex wondered whether we could display some notion of "progress", indicating whether hashrate estimation is close to becoming stable or not... I think this would be the ideal UX, but I cannot contemplate a straightforward way to implement it inside `channels_sv2`

your proposal seems to be going in a different direction however... IIUC the user would be able to see how much time has elapsed since the hashrate estimation changed, correct?

bit-aloo:
> thanks for the suggestion @Shourya742
> 
> how do you see this approach shaping up impact the final UX at `sv2-ui` level?
> 
> recently @pavlenex wondered whether we could display some notion of "progress", indicating whether hashrate estimation is close to becoming stable or not... I think this would be the ideal UX, but I cannot contemplate a straightforward way to implement it inside `channels_sv2`
> 
> your proposal seems to be going in a different direction however... IIUC the user would be able to see how much time has elapsed since the hashrate estimation changed, correct?

I don’t know how we can show progress here, especially with our ad hoc classic vardiff algorithm. We can probably provide some estimations, though I’m not sure how accurate they would be. If we move to PID vardiff, we could approximate the actual time more reliably, since statistically we can measure it and the real time would only differ by some ± delta. That would allow us to achieve the UX Pav proposed above.

My suggestions were mostly about the abstraction we are introducing. A stable_hashrate boolean feels misleading, because the hashrate may appear stable for some duration, fluctuate again, and then become stable once more. At least for the time being, a better approach might be to simply show the duration of stability (e.g., “stable for X minutes”).

I also don’t understand how stable_hashrate helps with the final UX design for vardiff progress.

GitGab19:
> A stable_hashrate boolean feels misleading, because the hashrate may appear stable for some duration, fluctuate again, and then become stable once more.

If hashrate is stable, its estimation should be stable as well. So IMO this is the best thing we can do now with what we have.

> At least for the time being, a better approach might be to simply show the duration of stability (e.g., “stable for X minutes”).

Knowing for how much time the hashrate estimated through vardiff is stable does not help in any way from a UX point of view. Users want to have clarity about what's going on when looking at the hashrate displayed on the UI, especially when it deviates a lot from what they see on the firmware side.

That's why I think the `stable-hashrate` boolean is the simplest thing to achieve that. If the initial hashrate displayed is not similar to what the firmware measures (with way more accurate precision of course), the vardiff needs to adjust diff, until it reaches some kind of stability.

Then if we're arguing here about the fact that with the current vardiff algorithm we don't have too much stability, I think it's a separate discussion, and we should probably run some more precise numbers and do a proper analysis to improve that.

If we see some issues with the current algorithm (no stability even when reaching the real hashrate range), we're gonna improve it, and this approach based on the `stable-hashrate` will automatically benefit by that.

plebhash:
> I also don’t understand how stable_hashrate helps with the final UX design for vardiff progress.

I tried to elaborate the idea on https://github.com/stratum-mining/sv2-ui/issues/151

bit-aloo:
If you look at the companion PR, we are toggling stable_hashrate between vardiff cycles. So, let’s say that during a vardiff epoch, when we call try_vardiff, it returns Ok(None), and we set stable_hashrate to true. However, this can happen in two scenarios: either there was genuinely no update, or the cycle was triggered too early (in our case, before 15 seconds).

```rust
        if delta_time <= 15 {
            return Ok(None);
        }

.... 

        if !should_update {
            return Ok(None);
        }


```

Because of this, we may observe fluctuating stability, where the value keeps switching between true and false. Doesn’t that make the behavior somewhat confusing? Considering the graphs we have been seeing for vardiff cycles already show noticeable jitter, having a boolean like this feels difficult to justify, since at one moment it may report “stable,” and shortly after report not stable.

Let me know if you think I might be missing stuff here?

GitGab19:
Given that vardiff cycles are hardcoded to 60s, the `if delta_time <= 15` condition can never happen, so if we get `Ok(None)`, it means there's no need for adjustments "for real".

bit-aloo:
> Given that vardiff cycles are hardcoded to 60s, the `if delta_time <= 15` condition can never happen, so if we get `Ok(None)`, it means there's no need for adjustments "for real".

It can happen because the 60-second timer is global and not relative to each channel. So, let’s say a channel is added at the 50th second and the vardiff cycle runs at the 60th second, this condition will be hit. However, it will only happen the first time, since after that all channels will follow the global timer cycle.

plebhash:
I just want to emphasize that I'm not trying to push for the upmost elegant or perfect UX here

yeah, there's many ways we can frame this as very KISS

the reason I'm pushing for this is because I see a lot of confusion in the community about how to visually interpret hashrate on `sv2-ui`, and the efforts here are meant to act on a low-hanging fruit with significant impact on the short term, even if not perfect and overly simplistic

if the changes proposed here are remain perceived as a net negative, happy to close the PR and move on to something else (as I have plenty other stuff to do on this milestone)

(and I don't think we should pivot towards PID vardiff in the short term)

bit-aloo:
Yeah, that makes sense regarding perfect UX. I think we can iterate on it again later. I just wanted to share my thoughts about the instability of stable_hashrate, so we’re all aware of it. I’m okay with merging this for now, and if we see confusion later, we can improve it..

bit-aloo (approved):
ACK

plebhash:
temporarily reverting to draft for some adjustments to companion (avoid accidental merging)

PR #2156, opened 2026-05-14
fix validate_share panic after on_set_new_prev_hash in custom-work mode

### LINKED ISSUE #2155: validate_share panics after on_set_new_prev_hash in custom-work mode
## Bug

`ExtendedChannel::on_set_new_prev_hash` clears `job_id_to_target` up front (`server/extended.rs:549`), then in the `has_future_jobs() == false` branch only calls `mark_past_jobs_as_stale()`. That helper doesn't touch `active_job`, so the old active job stays alive but its `job_id_to_target` entry is gone.

Next `validate_share` for that `job_id` hits:

```rust
// server/extended.rs:672-675
self.job_id_to_target
    .get(&job_id)
    .expect("job target must exist");
```

`is_active_job` is true, so the stale and `InvalidJobId` early returns are skipped and the `expect` panics.

### Repro

1. Open an extended channel with `REQUIRES_CUSTOM_WORK`.
2. JDC sends `SetCustomMiningJob` -> active job `J`, no future jobs queued.
3. `SetNewPrevHash` arrives -> false branch fires, map cleared, `J` still active.
4. Miner submits an in-flight share for `J` -> panic.

Fires every Bitcoin block on any JD-mode channel since shares are always in flight when the tip flips.

**`StandardChannel::on_set_new_prev_hash` clears the map at `standard.rs:515` then returns `TemplateIdNotFound` in the false branch. If the caller keeps the channel alive, next share panics at `standard.rs:594-597`.**

### Fix

Probably both:

- In the false arm, move `active_job` into `past_jobs` before `mark_past_jobs_as_stale`, so old shares come back `Stale` instead of panicking.
- Replace the two `expect("job target must exist")` sites with `ok_or(InvalidJobId)`.

rx18-eng:
## Body

Closes #2155.

- new `JobStore::deactivate_job` method that moves `active_job` into `past_jobs`
- `ExtendedChannel::on_set_new_prev_hash` false arm now calls `deactivate_job()` before `mark_past_jobs_as_stale()`, so the old active goes through past then stale and late shares for it return `Stale` instead of panicking
- `StandardChannel::on_set_new_prev_hash` reordered so `job_id_to_target.clear()` only runs after `activate_future_job(...)` returns true, so the error path can't corrupt state
- both `expect("job target must exist")` sites in `validate_share` (`extended.rs:675`, `standard.rs:597`) replaced with `ok_or(ShareValidationError::InvalidJobId)?`

### Two new tests

- `test_set_new_prev_hash_without_future_jobs_marks_active_as_stale` in `extended.rs` reproduces the panic from the issue and asserts `Err(Stale)` after the fix
- `test_set_new_prev_hash_without_future_jobs_preserves_state` in `standard.rs` confirms state survives the error and the next share validates without panic

Both tests fail with the original panic on unpatched code; full crate suite passes (74/74) after the fix.

### DIFF
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -569,15 +569,21 @@ where
         &mut self,
         set_new_prev_hash: SetNewPrevHashTdp<'a>,
     ) -> Result<(), ExtendedChannelError> {
-        // clear the job id to target mapping
-        self.job_id_to_target.clear();
-
         // extended channels dedicated to custom work don't need to keep track of future jobs
         match self.job_store.has_future_jobs() {
             false => {
+                // demote the previously-active job to past so that the subsequent
+                // mark_past_jobs_as_stale call moves it into the stale set. without this,
+                // a late share for the still-active old job would skip the stale check in
+                // validate_share and panic on the missing job_id_to_target entry that we
+                // just cleared.
+                self.job_store.deactivate_job();
                 // explicitly mark past jobs as stale, because we're not going to
                 // do it implicitly via activate_future_job in case this extended channel is doing custom work
                 self.job_store.mark_past_jobs_as_stale();
+                // there is no active job in this branch, so any previous job target
+                // mappings are obsolete after the chain tip update.
+                self.job_id_to_target.clear();
             }
             true => {
                 // try to activate the future job, and also mark past jobs as stale
@@ -587,7 +593,9 @@ where
                 ) {
                     return Err(ExtendedChannelError::TemplateIdNotFound);
                 }
-
+                // clear the job id to target mapping only after a successful activation,
+                // so that an early-return error path does not corrupt channel state.
+                self.job_id_to_target.clear();
                 // associate the new active job with the current target
                 let job_id = self
                     .job_store
@@ -700,10 +708,13 @@ where
                 .get_stale_job(job_id)
                 .expect("stale job must exist")
         };
-        let job_target = self
-            .job_id_to_target
-            .get(&job_id)
-            .expect("job target must exist");
+        let Some(job_target) = self.job_id_to_target.get(&job_id) else {
+            self.share_accounting
+                .increment_rejected_shares(ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID);
+            return Err(ShareValidationError::InvalidJobId(
+                ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID,
+            ));
+        };
 
         let extranonce_size = share.extranonce.inner_as_ref().len();
         if extranonce_size != self.rollable_extranonce_size as usize {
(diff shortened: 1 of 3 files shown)


rx18-eng:
@plebhash would appreciate a review from you !

rx18-eng:
I fixed the failing semver check !

rx18-eng:
@plebhash shall i modify anything on this pr ?

sv2/channels-sv2/src/server/extended.rs
plebhash:
```suggestion
```

plebhash: not sure why the suggestion is rendering like this

what I mean is:
```diff
-        // clear the job id to target mapping
-        self.job_id_to_target.clear();
```

in other words: remove L561-L562

plebhash, line 587:
```suggestion
                // there is no active job in this branch, so any previous job target
                // mappings are obsolete after the chain tip update.
                self.job_id_to_target.clear();
            }
```

plebhash:
```suggestion
                // clear the job id to target mapping only after a successful activation,
                // so that an early-return error path does not corrupt channel state.
                self.job_id_to_target.clear();
```

plebhash: why was this resolved without being properly addressed?

plebhash (changes requested):
@rx18-eng you re-requested my review while not only ignoring the main suggestion, but also resolving it.

all that in less than a minute after I left my comments.

honestly I would prefer that you took the time to read and understand what the suggestions mean, instead of blindly  clicking buttons on Github UI

with this current workflow, you're just closing your eyes and doing whatever you can to rush your PR getting merged, despite correctness or sanity of the contribution.

that's not a good way to make FOSS contributions.

sv2/channels-sv2/src/server/extended.rs
plebhash: by the way, the other two suggestions are meaningless without this one

I hoped that would have been obvious (assuming @rx18-eng actually read them and tried to understand what they meant)

rx18-eng:
@plebhash you're right, I rushed. The first suggestion rendered as an empty code block in my view, so I read it as a no-op and clicked through the other two. I didn't go back to  your "in other words: remove L561-L562" reply until after I'd already pushed and re-requested review. That's on me, I should have read every thread end to end first. I sincerely apologize for this !

rx18-eng:
Going back through it now, the point is clear: with the top-level clear at L561-L562 still in place, the arm-internal   clears from suggestions 2 and 3 don't actually defend  against the early-return on `TemplateIdNotFound`, because    
 `job_id_to_target` is already empty before `activate_future_job` even runs.

plebhash:
it seems you started your branch before #2151 had been merged, so please rebase against `main` and fix CI

also please squash all commits into one

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -39,6 +39,23 @@ pub trait JobStore<T: Job>: Send + Sync + Debug {
...
+    ///
+    /// Returns `true` if a job was deactivated, `false` if there was no active job.
+    ///
+    /// The default implementation is a no-op that returns `false`. It exists so that adding
+    /// this method does not break existing downstream `JobStore` implementations.
+    /// Implementations that own an active job slot should override it.
+    fn deactivate_job(&mut self) -> bool {
+        false
+    }
+
```
plebhash:
```suggestion
/// Moves the active job (if any) into past jobs.
///
/// Default impl is a no-op.
fn deactivate_job(&mut self) { }
```

there's no point in returning `bool`

also rust docs were too verbose

please note that I'm not making this suggestion for it to be merged via github UI... it's not properly formatted and it would lead to an unnecessary extra commit

this is meant to be manually applied into the code and squashed into the original commit

plebhash: sorry my suggested rust doc comments were still assuming it would return `bool`

edited

plebhash (approved):

PR #2183, opened 2026-06-03
test: cover transport-mode handshake state

### LINKED ISSUE #2153: Repeated tests on the codec library
I was exploring the library and when I got to the tests it caught my attention as the tests have different name but the exact same code.

https://github.com/stratum-mining/stratum/blob/6ab03af290c94fb080419cb2a744bdd10e51c279/sv2/codec-sv2/src/lib.rs#L287-L301

Also the second test's code doesn't seem to match its name, so perhaps someone duplicated the first while creating the second and then forgot to actually implement the new test. Just a hypothesis.

0xjc65eth:
## Summary
- Replace the duplicated not-initialized handshake test with a transport-mode state check.
- Build a minimal initiator/responder Noise handshake in the test helper.
- Assert `step_0` returns `NotInHandShakeState` after the codec has reached transport mode.

Closes #2153.

## Why
Issue #2153 points out that the two tests had different names but identical setup. The second test claimed to cover transport mode, but it was still using `State::NotInitialized(32)`. This changes only that test path so it exercises the state named by the test.

## Tests
- `cargo fmt --check`
- `cargo test -p codec_sv2 --features noise_sv2 handshake_step_fails_if_state --lib`
- `cargo test -p codec_sv2 --features noise_sv2 --lib`

### DIFF
--- a/sv2/codec-sv2/src/lib.rs
+++ b/sv2/codec-sv2/src/lib.rs
@@ -283,6 +283,44 @@ impl State {
 #[cfg(feature = "noise_sv2")]
 mod tests {
     use super::*;
+    use key_utils::{Secp256k1PublicKey, Secp256k1SecretKey};
+    use noise_sv2::{ELLSWIFT_ENCODING_SIZE, INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE};
+    use std::time::Duration;
+
+    const AUTHORITY_PUBLIC_K: &str = "9auqWEzQDVyd2oe1JVGFLMLHZtCo2FFqZwtKA5gd9xbuEu7PH72";
+    const AUTHORITY_PRIVATE_K: &str = "mkDLTBBRxdBv998612qipDYoTK3YUrqLe8uWw7gu3iXbSrn2n";
+
+    fn make_transport_state() -> State {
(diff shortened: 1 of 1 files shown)


edilmedeiros:
AI slop?

0xjc65eth:
The PR body formatting was bad; I fixed that.

The intended change is narrow: #2153 noted that the transport-mode test had the same `State::NotInitialized(32)` setup as the not-initialized test. This helper performs the normal initiator/responder Noise handshake to obtain a real `State::Transport`, then verifies `step_0` returns `Error::NotInHandShakeState` from that state.

If there is a concrete issue with the handshake setup or if maintainers prefer deleting these old tests during the codec refactor, happy to adjust.

bit-aloo (commented):
We’re going to get rid of these tests, and the codec API surface will change soon. I don’t see a point in having this right now. We also mentioned in the issue that we would eventually remove it, but kept the issue open just for tracking purposes.

Thanks for the contribution, but we’re going to close this as not required.

plebhash:
also fuck off with your slop spamming @0xjc65eth 

I hear you're pissing off A LOT of people on the Bitcoin community 

I'm not reviewing nor merging your PRs anymore

PR #2184, opened 2026-06-03
buffer_sv2: expose POOL_CAPACITY

### LINKED ISSUE #2130: Expose `POOL_CAPACITY` from `buffer_sv2`
`Buffer_sv2` does not expose `POOL_CAPACITY`, even though it is defined internally, and we need access to it for benchmarking.

0xjc65eth:
## Summary
- Re-export `buffer_pool::POOL_CAPACITY` from the `buffer_sv2` crate root.
- Allows benchmarks and external users to access `buffer_sv2::POOL_CAPACITY` without reaching into the private `buffer_pool` module.

Closes #2130.

## Tests
- `cargo fmt --check`
- `cargo check -p buffer_sv2`
- `git diff --check`

### DIFF
--- a/sv2/buffer-sv2/src/lib.rs
+++ b/sv2/buffer-sv2/src/lib.rs
@@ -53,7 +53,7 @@ use alloc::vec::Vec;
 
 pub use crate::buffer::BufferFromSystemMemory;
 pub use aes_gcm::aead::Buffer as AeadBuffer;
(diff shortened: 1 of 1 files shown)


bit-aloo (commented):
We can't get this in now, considering this would require us to bump the version. We gonna batch planned changes to buffer_sv2 later. Thanks for the PR, but I have to close this as not planned.

PR #2185, opened 2026-06-08
Add absolute path in binary sv2 proc macro

### LINKED ISSUE #2167: Binary_sv2(Proc_macros): Use absolute path instead of relative path to import binary_sv2
The generated code assumes the caller imported binary_sv2 into scope. Use ::binary_sv2::... instead so the derive works from outside the repo too.

bit-aloo:
closes: #2167


bit-aloo:
Nice it didn't break anything on sv2_apps.

bit-aloo:
Though we need to do cleanup on sv2-apps once this gets in.

GitGab19 (approved):

plebhash:
> Though we need to do cleanup on sv2-apps once this gets in.

@Shourya742 can you elaborate on this?

plebhash (approved):

PR #2186, opened 2026-06-09
Use compile error in proc macro

### LINKED ISSUE #2166: Binary_sv2(Proc macro): Replace panics with compile_errors
Right now unsupported syntax often crashes the binary_sv2 proc macro, with not good diagnostic error. Ideally we should have them known during compile time itself with spans.

bit-aloo:
closes: #2166

### DIFF
--- a/sv2/binary-sv2/derive_codec/src/lib.rs
+++ b/sv2/binary-sv2/derive_codec/src/lib.rs
@@ -245,16 +245,24 @@ struct ParsedStruct {
 struct ParsedField {
     // Name of the field.
     name: String,
+    // Span of the field name.
+    span: Span,
     // Type of the field.
     type_: String,
     // Generics associated with the field, if any.
     generics: String,
 }
 
+struct MacroError {
+    span: Span,
+    message: String,
+}
+
 impl ParsedField {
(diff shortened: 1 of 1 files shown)


plebhash (approved):
non-blocking clanker suggestion:

> add compile-fail tests (e.g., in binary-sv2 tests) for unsupported syntax + reserved field names so this behavior stays locked in and future refactors don’t reintroduce panic paths

bit-aloo:
> non-blocking clanker suggestion:
> 
> > add compile-fail tests (e.g., in binary-sv2 tests) for unsupported syntax + reserved field names so this behavior stays locked in and future refactors don’t reintroduce panic paths

We have another issue on proc macro to restrict binary_sv2 only for sv2 types, maybe this can be added then.

plebhash:
> We have another issue on proc macro to restrict binary_sv2 only for sv2 types, maybe this can be added then.

can you please take note there?

bit-aloo:
> > We have another issue on proc macro to restrict binary_sv2 only for sv2 types, maybe this can be added then.
> 
> can you please take note there?

Done

PR #2187, opened 2026-06-09
Fix binary_sv2 std build

bit-aloo:
closes: #2170

PR #2189, opened 2026-06-10
Remove unused entities in `binary_sv2` + `common_messages_sv2` + `template_distribution_sv2`

### LINKED ISSUE #2174: Binary_sv2: Remove prop test injectors
In binary_sv2 crate, we have a number of prop test injects which we don't use. We can safely remove them.

bit-aloo:
closes: #2174 

Also removes variable and two owned trait.

Also removes prop_test feature from common_message and template_distribution_message_Sv2.


GitGab19:
I see you're touching `template_distribution_sv2` and `common_messages_sv2` as well: why aren't they mentioned in the PR title and/or description?

bit-aloo:
> I see you're touching `template_distribution_sv2` and `common_messages_sv2` as well: why aren't they mentioned in the PR title and/or description?

Ah!! thanks for mentioning. I completely forgot to mention them. Though I thought of adding them when I was removing prop_test.

plebhash:
this PR has a similar versioning blindspot as described in https://github.com/stratum-mining/stratum/pull/2187#issuecomment-4685604192

I assume there's significant intersection between these two PRs and after the first one is merged, we can probably remove many bumps from the other one

---

unfortunately keeping sanitify of versioning across inter-dependent crates on a monorepo is non-trivial and hard to enforce via `cargo semver-checks` alone, so we have to be hypervigilant (always resorting to `CONTRIBUTING.md` when in doubt, ideally with regular clanker assistance to mitigate blindspots)

bit-aloo:
Updated the bumps here as well.

bit-aloo:
Rebased, and removed bumping commit.

sv2/subprotocols/template-distribution/src/lib.rs
plebhash: this should be cleaned up

plebhash: also here:

https://github.com/stratum-mining/stratum/blob/a15c224eed3f5933dd70e955b9c98626d2206b37/sv2/subprotocols/template-distribution/README.md#L15-L19

sv2/subprotocols/common-messages/src/lib.rs
plebhash: this should be cleaned up

plebhash: also here:

https://github.com/stratum-mining/stratum/blob/a15c224eed3f5933dd70e955b9c98626d2206b37/sv2/subprotocols/common-messages/README.md#L12-L16

bit-aloo: Done thanks.

sv2/subprotocols/template-distribution/src/lib.rs
bit-aloo: Done thanks.

plebhash (approved):

PR #2190, opened 2026-06-12
deprecate `channels_sv2::server::jobs::job_store::JobStore`

### LINKED ISSUE #2181: need to deprecate `channels_sv2::server::jobs::job_store::JobStore`
`channels_sv2::server::jobs::job_store::JobStore` was introduced in 2025 as an attempt to abide by some restrictions that some specific Sv2 adopter had with regards to a custom/opinionated implementation of a Template Provider

since then we learned that `JobStore` wasn't enough, and this adopter decided not to use `channels_sv2` crate

personally speakling, I'm not a fan of the `JobStore` abstraction, as I feel it's overengineered for the goals that `channels_sv2` aims to achieve

moreover, it's starting to accumulate technical debt

therefore, I think we should simply remove this abstraction from the crate

plebhash:
close #2181

companion https://github.com/stratum-mining/sv2-apps/pull/554

### DIFF
--- a/sv2/channels-sv2/src/server/jobs/job_store.rs
+++ b/sv2/channels-sv2/src/server/jobs/job_store.rs
@@ -16,7 +16,7 @@ use super::Job;
 /// Maintains collections for future, active, past, and stale jobs, and tracks template-to-job ID
 /// mappings for future job activation.
 #[derive(Debug)]
-pub(crate) struct JobStore<T: Job + Clone> {
+pub(crate) struct JobStore<T: Job> {
     future_template_to_job_id: HashMap<u64, u32>,
     // Future jobs are indexed with job_id (u32)
     future_jobs: HashMap<u32, T>,
@@ -27,7 +27,7 @@ pub(crate) struct JobStore<T: Job + Clone> {
     stale_jobs: HashMap<u32, T>,
 }
 
-impl<T: Job + Clone> JobStore<T> {
+impl<T: Job> JobStore<T> {
     /// Creates a new empty job store.
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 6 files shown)


sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -89,7 +27,7 @@ pub struct DefaultJobStore<T: Job + Clone> {
     stale_jobs: HashMap<u32, T>,
 }
 
-impl<T: Job + Clone> DefaultJobStore<T> {
+impl<T: Job + Clone> JobStore<T> {
```
bit-aloo: Clone can be removed here.

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -102,22 +40,25 @@ impl<T: Job + Clone> DefaultJobStore<T> {
     }
 }
 
-impl<T: Job + Clone> Default for DefaultJobStore<T> {
+impl<T: Job + Clone> Default for JobStore<T> {
```
bit-aloo: Here as well, we can remove Clone

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -102,22 +40,25 @@ impl<T: Job + Clone> DefaultJobStore<T> {
...
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<T: Job + Clone + Debug> JobStore<T> for DefaultJobStore<T> {
-    fn add_future_job(&mut self, template_id: u64, new_job: T) -> u32 {
+impl<T: Job + Clone> JobStore<T> {
```
bit-aloo: Here as well.

bit-aloo (commented):
Pretty smooth changes. Few suggestions rest looks good.

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: it was already done on the commit afterwards

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: it was already done on the commit afterwards

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: it was already done on the commit afterwards

bit-aloo (approved):
ACK. Nice cleanup.

GitGab19 (approved):

PR #2200, opened 2026-06-16
Binary sv2 cleanup

### LINKED ISSUE #2171: Binary_sv2: Remove from_vec_ and from_vec_unchecked
Both `from_vec_` and `from_vec_unchecked` implementation doesn't respect the length field of variable size sv2 types.

### LINKED ISSUE #2173: Binary_sv2: Length of fixed type is always 1
```rust
    pub fn len(&self) -> usize {
        match (self, ISFIXED) {
            (Inner::Ref(data), false) => data.len(),
            (Inner::Owned(data), false) => data.len(),
            (_, true) => 1,
        }
    }
```

### LINKED ISSUE #2175: Binary_sv2: What to do with types we not use, but in specs? and vice versa
Missing: MAC, SHORT_TX_ID, PUBKEY

Edit:

What I meant is that we have types defined in both the specs and `binary_sv2` that we do not actually use. What should we do with those? Also, there are types that exist in the specs but are not implemented in `binary_sv2` or anywhere else in our implementation. What should we do about them?

### LINKED ISSUE #2176: Binary_sv2: From<Vec<T>> for Seq0255 and From<Vec<T>> for Seq064K bypass the length invariants
```rust

impl<T> From<Vec<T>> for Seq0255<'_, T> {
    fn from(v: Vec<T>) -> Self {
        Seq0255(v, PhantomData)
    }
}

impl<T> From<Vec<T>> for Seq064K<'_, T> {
    fn from(v: Vec<T>) -> Self {
        Seq064K(v, PhantomData)
    }
}

```

### LINKED ISSUE #2177: Binary_sv2: to_writer doesn't respect variable sized primitive encoding
Currently, the write just dumps all the bytes instead of appending length to it.

### LINKED ISSUE #2178: Binary_sv2: Reader reads unbounded bytes
The reader uses `read_to_end` leading to unbounded bytes read

### LINKED ISSUE #2179: Binary_sv2: Primitive wrappers shouldn't expose internal types
Sv2 Primitives shouldn't expose internal types which might allow bypassing type invariants.

### LINKED ISSUE #2213: Expand conversion APIs in binary_sv2
Conversions for copy types should be lossless and ergonomic, while conversions for variable-sized types should be explicit and fallible through TryFrom. These changes are largely motivated by experience working with the sv2-apps codebase and aim to provide clearer semantics.

### LINKED ISSUE #2214: Replace handwritten From implementations with mbe in binary_sv2
We have a lot of from implementation in impl.rs in binary_sv2 which can be abstracted via macros.

### LINKED ISSUE #2215: Remove U32Ref and U8Owned
We don't really need them.

bit-aloo:
closes: #2171 #2173 #2175 #2176 #2177 #2178 #2179 #2213 #2214 #2215 #2216

Apart from the issues mentioned above, this PR introduces several additional improvements:

1. Removes a significant amount of redundant code.
2. Expands the API surface. In particular, conversions for copy types are now lossless, while conversions for variable-sized types are performed through `TryFrom`. (More details can be found in the companion `sv2-apps` PR. I'm open to suggestions for alternative APIs these additions are primarily based on my experience working with the `sv2-apps` codebase.)
3. Introduces a dedicated `MAC` type, bringing us into full compliance with the specification.
4. Replaces a large number of handwritten `From` implementations with MBE macros, significantly reducing duplication.
5. Fixes issues in the streaming APIs. Previously, certain code paths could lead to unbounded reads, creating potential attack vectors. Reads are now properly bounded, and recursion-related bounds have been removed from the writer implementation.
6. Removes `U32Ref` and `U8Owned`, as they do not provide any meaningful structural advantages.
7. Adds iterator support for `Seq` types, allowing users to iterate over elements and perform in-place modifications more ergonomically.

companion https://github.com/stratum-mining/sv2-apps/pull/576

### DIFF
--- a/sv2/binary-sv2/src/codec/decodable.rs
+++ b/sv2/binary-sv2/src/codec/decodable.rs
@@ -1,6 +1,6 @@
 use crate::{
     codec::{GetSize, SizeHint},
-    datatypes::{Signature, Sv2DataType, B016M, B0255, B032, B064K, U24, U256},
+    datatypes::{Mac, Signature, Sv2DataType, B016M, B0255, B032, B064K, U24, U256},
     Error,
 };
 use alloc::vec::Vec;
@@ -79,6 +79,7 @@ pub enum PrimitiveMarker {
     Bool,
     U24,
     U256,
+    Mac,
     Signature,
     U32,
     F32,
@@ -121,6 +122,7 @@ pub enum DecodablePrimitive<'a> {
     Bool(bool),
     U24(U24),
     U256(U256<'a>),
+    Mac(Mac<'a>),
     Signature(Signature<'a>),
     U32(u32),
     F32(f32),
@@ -160,6 +162,7 @@ impl SizeHint for PrimitiveMarker {
             Self::Bool => bool::size_hint(data, offset),
             Self::U24 => U24::size_hint(data, offset),
             Self::U256 => U256::size_hint(data, offset),
+            Self::Mac => Mac::size_hint(data, offset),
             Self::Signature => Signature::size_hint(data, offset),
             Self::U32 => u32::size_hint(data, offset),
             Self::F32 => f32::size_hint(data, offset),
@@ -306,6 +309,7 @@ impl PrimitiveMarker {
             Self::Bool => Ok(DecodablePrimitive::Bool(bool::from_reader_(reader)?)),
             Self::U24 => Ok(DecodablePrimitive::U24(U24::from_reader_(reader)?)),
             Self::U256 => Ok(DecodablePrimitive::U256(U256::from_reader_(reader)?)),
+            Self::Mac => Ok(DecodablePrimitive::Mac(Mac::from_reader_(reader)?)),
             Self::Signature => Ok(DecodablePrimitive::Signature(Signature::from_reader_(
                 reader,
             )?)),
@@ -328,6 +332,7 @@ impl GetSize for DecodablePrimitive<'_> {
             DecodablePrimitive::Bool(v) => v.get_size(),
             DecodablePrimitive::U24(v) => v.get_size(),
             DecodablePrimitive::U256(v) => v.get_size(),
+            DecodablePrimitive::Mac(v) => v.get_size(),
             DecodablePrimitive::Signature(v) => v.get_size(),
             DecodablePrimitive::U32(v) => v.get_size(),
             DecodablePrimitive::F32(v) => v.get_size(),
--- a/sv2/binary-sv2/src/codec/decodable.rs
+++ b/sv2/binary-sv2/src/codec/decodable.rs
@@ -1,6 +1,6 @@
 use crate::{
     codec::{GetSize, SizeHint},
-    datatypes::{Signature, Sv2DataType, U32AsRef, B016M, B0255, B032, B064K, U24, U256},
+    datatypes::{Signature, Sv2DataType, B016M, B0255, B032, B064K, U24, U256},
     Error,
 };
 use alloc::vec::Vec;
@@ -81,7 +81,6 @@ pub enum PrimitiveMarker {
     U256,
     Signature,
     U32,
-    U32AsRef,
     F32,
     U64,
     B032,
@@ -124,7 +123,6 @@ pub enum DecodablePrimitive<'a> {
     U256(U256<'a>),
     Signature(Signature<'a>),
     U32(u32),
-    U32AsRef(U32AsRef<'a>),
     F32(f32),
     U64(u64),
     B032(B032<'a>),
@@ -164,7 +162,6 @@ impl SizeHint for PrimitiveMarker {
             Self::U256 => U256::size_hint(data, offset),
             Self::Signature => Signature::size_hint(data, offset),
             Self::U32 => u32::size_hint(data, offset),
-            Self::U32AsRef => U32AsRef::size_hint(data, offset),
             Self::F32 => f32::size_hint(data, offset),
             Self::U64 => u64::size_hint(data, offset),
             Self::B032 => B032::size_hint(data, offset),
@@ -313,9 +310,6 @@ impl PrimitiveMarker {
                 reader,
             )?)),
             Self::U32 => Ok(DecodablePrimitive::U32(u32::from_reader_(reader)?)),
-            Self::U32AsRef => Ok(DecodablePrimitive::U32AsRef(U32AsRef::from_reader_(
-                reader,
-            )?)),
             Self::F32 => Ok(DecodablePrimitive::F32(f32::from_reader_(reader)?)),
             Self::U64 => Ok(DecodablePrimitive::U64(u64::from_reader_(reader)?)),
             Self::B032 => Ok(DecodablePrimitive::B032(B032::from_reader_(reader)?)),
@@ -336,7 +330,6 @@ impl GetSize for DecodablePrimitive<'_> {
             DecodablePrimitive::U256(v) => v.get_size(),
             DecodablePrimitive::Signature(v) => v.get_size(),
             DecodablePrimitive::U32(v) => v.get_size(),
-            DecodablePrimitive::U32AsRef(v) => v.get_size(),
             DecodablePrimitive::F32(v) => v.get_size(),
             DecodablePrimitive::U64(v) => v.get_size(),
             DecodablePrimitive::B032(v) => v.get_size(),
--- a/sv2/binary-sv2/src/codec/decodable.rs
+++ b/sv2/binary-sv2/src/codec/decodable.rs
@@ -150,9 +150,9 @@ pub enum DecodableField<'a> {
 }
 
 impl SizeHint for PrimitiveMarker {
-    // PrimitiveMarker needs introspection to return a size hint. This method is not implementable.
+    // PrimitiveMarker requires a concrete marker instance to determine the size.
     fn size_hint(_data: &[u8], _offset: usize) -> Result<usize, Error> {
-        unimplemented!()
+        Err(Error::UnInitializedDecoder)
     }
 
     fn size_hint_(&self, data: &[u8], offset: usize) -> Result<usize, Error> {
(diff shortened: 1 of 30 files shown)


bit-aloo:
Awesome, the test are passing, atleast I have not broken any invariants. Need to improve API ergonomics and should be good to go.

bit-aloo:
I will open issues for all the extra points mentioned in the description.

GitGab19:
My clanker found this:
```
High: the new bounded from_reader can still block/read until EOF on an invalid B032 length. [decodable.rs (line 70)](/tmp/stratum-pr2200/sv2/binary-sv2/src/codec/decodable.rs:70) treats every ReadError as “need one more byte”, but [inner.rs (line 109)](/tmp/stratum-pr2200/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs:109) also returns ReadError when a variable-sized type declares more than its max, e.g. B032 header 33. For messages with B032 fields, a peer can send an oversized length and keep the socket open; the loop keeps reading instead of rejecting the message. I’d split “insufficient bytes” from “declared size exceeds max”, or make from_reader retry only on the former.
```

GitGab19:
@Shourya742 just a suggestion about the point raised by the clanker:

Yes. The core fix is to stop using `ReadError` for two different meanings:

1. “I need more bytes to know the size”
2. “The size is known and invalid”

Right now both flow through `ReadError`, and [decodable.rs](/tmp/stratum-pr2200/sv2/binary-sv2/src/codec/decodable.rs:70) treats every `ReadError` as recoverable by reading one more byte.

The best fix is in `Inner::expected_length` / `expected_length_for_reader`:

```rust
// pseudo-shape
if payload_len <= MAXSIZE {
    Ok(payload_len + HEADERSIZE)
} else {
    Err(Error::ValueExceedsMaxSize(
        ISFIXED,
        SIZE,
        HEADERSIZE,
        MAXSIZE,
        data.to_vec(),
        payload_len,
    ))
}
```

For the reader path:

```rust
if expected_length <= MAXSIZE {
    Ok(expected_length)
} else {
    Err(Error::ValueExceedsMaxSize(
        ISFIXED,
        SIZE,
        HEADERSIZE,
        MAXSIZE,
        header.to_vec(),
        expected_length,
    ))
}
```

Then `Decodable::from_reader` can keep retrying only on true partial-read errors:

```rust
Err(Error::OutOfBound | Error::ReadError(_, _)) => read_one_more_byte()
Err(error) => return Err(error)
```

With that change, `B032` input `[33]` would return immediately because the header is complete and declares a payload larger than 32. It would no longer read/block waiting for 33 bytes.

A slightly cleaner design would add a dedicated error like `Error::InvalidDeclaredLength { declared, max }`, because `ValueExceedsMaxSize` carries a value vector and is a bit awkward for “length header is invalid”. But if the PR wants a small surgical fix, returning `ValueExceedsMaxSize` from the known-invalid length paths is enough.

sv2/binary-sv2/src/codec/impls.rs
```diff
@@ -9,789 +9,209 @@ use crate::{
...
-
-    fn try_from(value: DecodablePrimitive<'a>) -> Result<Self, Self::Error> {
-        match value {
-            DecodablePrimitive::U32AsRef(val) => Ok(val),
-            _ => Err(Error::PrimitiveConversionError),
-        }
-    }
-}
-
-// IMPL TRY_FROM DECODEC FIELD FOR PRIMITIVES
```
GitGab19, line 360: Why did you remove the `TryFrom` implementations for `EncodableField`, without replacing them with a macro?

sv2/binary-sv2/src/lib.rs
```diff
@@ -342,9 +342,9 @@ impl From<Vec<u8>> for EncodableField<'_> {
 }
 
 #[cfg(feature = "with_buffer_pool")]
-impl From<buffer_sv2::Slice> for EncodableField<'_> {
-    fn from(_v: buffer_sv2::Slice) -> Self {
-        unreachable!()
```
GitGab19, line 347: Do you know why this was labeled as `unreachable` earlier?

GitGab19:
Given that on 012536aa31d907d9643b1af8f082f2fe05a4e194 you're fixing two issues found by Loupe, can't you add tests reported by it as well?

sv2/binary-sv2/src/datatypes/copy_data_types.rs
GitGab19: Remove the mention to `from_vec_`.

GitGab19:
I would add tests mentioned by Loupe on commit 173e162e919b73cbc059c3737faa7800d98b3613 as well.

GitGab19:
Add Loupe's tests on 5a0d4bb27780e98413d4c2ef8d630f6d68539777 as well.

sv2/binary-sv2/src/codec/decodable.rs
GitGab19: Why are you still keeping this one?

GitGab19:
Why isn't b4aebeb7c28c2c9289ac852c485cbd4552e6d09f right after cf697b249c2f804d4c4b90fd476c95b8840b5725?

They are both tackling https://github.com/stratum-mining/stratum/issues/2215, one removing `U32Ref` and the other removing `OwnedU8`.

sv2/binary-sv2/src/codec/decodable.rs
GitGab19: In e1720b337c612da7648ac7e8190b6cdd78e34015 you added a new `fn decode_owned()` which then you removed in 09d67d2751fc4a272d0184326411f9d5b53c5959?

What's going on?

sv2/binary-sv2/src/codec/impls.rs
bit-aloo: Because we were not using them.

Conceptually, converting from a type `T` to `EncodableField` makes sense, because `EncodableField` is the wire-format representation used for encoding. The reverse conversion, from `EncodableField` back to `T`, fits better with `DecodableField` and the decoding side of the codec.

I’ll mention this, in the commit message so the removal is clear.

sv2/binary-sv2/src/lib.rs
bit-aloo: My best guess is that this comes from how BYTES is treated in the spec versus in binary_sv2.

The SV2 spec defines BYTES as a primitive, but if you look closely, binary_sv2 does not really have a direct mapping for it. BYTES is just an arbitrary-length byte sequence, and in practice it usually appears in the frame payload.
The main reason is that BYTES has no length prefix of its own. Its size must come from external context. In our case, that usually means framing, where the payload length is given by msg_length. So BYTES is not really missing from binary_sv2; it is just handled indirectly as raw bytes whose length is already known from the surrounding protocol context. That is why it does not appear as a dedicated primitive enum variant, and why its usage in the codebase is mostly tied to framing or other context-driven decoding.

We do have Vec<u8> as a representation of raw bytes, which is probably why buffer_sv2::Slice was also made encodable. But now I am wondering whether having a dedicated representation for this in binary_sv2 even makes sense. BYTES seems to be mostly a framing-level concept, used to describe how many bytes belong to a complete message, rather than a standalone datatype that binary_sv2 needs to model directly.

Similarly, from_bytes in binary_sv2 does not decode a generic BYTES value by itself. It expects the caller to already know what concrete type those bytes should be decoded into. That also suggests that BYTES is more contextual than a real first-class datatype in this crate

I think removing both this and Vec one above would make more sense here.

bit-aloo:
> Then `Decodable::from_reader` can keep retrying only on true partial-read errors:
> 
> ```rust
> Err(Error::OutOfBound | Error::ReadError(_, _)) => read_one_more_byte()
> Err(error) => return Err(error)
> ```
> 
> With that change, `B032` input `[33]` would return immediately because the header is complete and declares a payload larger than 32. It would no longer read/block waiting for 33 bytes.
> 

This makes sense, taken.

sv2/binary-sv2/src/codec/decodable.rs
bit-aloo: re-writing history, hope this gets removed.

sv2/binary-sv2/src/lib.rs
bit-aloo: I reverted this, I will open a separate issue and remove it altogether.

bit-aloo:
Loupe tests are in, and all are passing, I will link the commits to corresponding loupe issue for reference.

bit-aloo:
@GitGab19 the commits might be a bit different now, I rewrote the history with test and your suggestions.

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -405,6 +415,32 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
...
+            Inner::Ref(data) => data
+                .try_into()
+                .expect("fixed-size SV2 byte wrapper must always match SIZE"),
+            Inner::Owned(data) => data
+                .try_into()
+                .expect("fixed-size SV2 byte wrapper must always match SIZE"),
+        }
+    }
+}
+
```
plebhash, line 443: I asked `gpt-5.3-codex` whether these `.expect`s were infallible:

> Short answer: **conditionally**, not absolutely.
> 
> - They are infallible **if** the invariant holds: for `Inner<'_, true, SIZE, 0, 0>`, payload length is always exactly `SIZE`.
> - With current API, that invariant is usually enforced by checked constructors (`TryFrom`, `from_bytes_`, `from_reader_`), so in normal flow these `expect`s should not fire.
> - But `Inner` is a `pub enum`, so callers can still construct invalid fixed values directly (e.g. `Inner::Owned(vec![0; SIZE - 1])`), and then `as_array()` / `into_array()` will panic.
> 
> So: **infallible by convention**, not by type-level encapsulation.

not a blocker for me, but I thought I'd bring it up anyways

plebhash (approved):

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
bit-aloo: We gonna be removing Inner, thats the end goal.

GitGab19 (approved):

Issue #896, opened 2024-05-04
tproxy: `HexError(OddLength)` in `mining.subscribe`


plebhash:
recently we had PR https://github.com/stratum-mining/stratum/pull/851 by @nikicat

I noticed we didn't have an issue here on github to track this (potential) bug

plebhash:
also, it took me some time to figure out how to reproduce this bug deterministically.

this issue is a good place to register my findings

here are instructions to reproduce the bug:

1. start tProxy
2. start netcat: `nc localhost 34255`
3. send a `mining.subscribe` with an extranonce of even length (this should work fine): `{"id": 1, "method": "mining.subscribe", "params": ["", "00"]}`
4. start netcat again: `nc localhost 34255`
5. send a `mining.subscribe` with an extranonce of off length (this should trigger the bug on tProxy): `{"id": 1, "method": "mining.subscribe", "params": ["", "0"]}`

plebhash:
I'm also backtracing the bug to find the root cause:

inside `pub async fn new_downstream` we have a call to `handle_incoming_sv1 `:

https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/roles/translator/src/lib/downstream_sv1/downstream.rs#L180

which calls `handle_message`:

https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/roles/translator/src/lib/downstream_sv1/downstream.rs#L375-L381

which calls `handle_request`:

https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/protocols/v1/src/lib.rs#L77

the issue seems to be in the `try_into()?`, which is implemented at:

https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/protocols/v1/src/methods/mod.rs#L142-L146
 
the nested `try_into()?` (line 146 above) is implemented at:
https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/protocols/v1/src/methods/mod.rs#L214-L229

the nested `try_into()?` (line 226 above) is implemented at:
https://github.com/stratum-mining/stratum/blob/24b8b1d8666d6055c4a964d2aac8ef2445734a10/protocols/v1/src/methods/client_to_server.rs#L313-L342

plebhash:
[...]

PR #2205, opened 2026-06-17
adjust hex conversion on sv1 to use `bitcoin::hashes::hex`

### LINKED ISSUE #2201: consider moving away from `bitcoin_hashes`
while working on a fuzz target for `sv1_api`, I found an input that causes a panic in the `bitcoin_hashes` crate due to the presence of non-ASCII characters at specific positions within a hex string. The panic occurs on the second `unwrap()` in the implementation below:

[https://github.com/rust-bitcoin/bitcoin_hashes/blob/16c0cb9375d3e12887ec9a5be2e758e3d907359e/src/hex.rs#L64-L85](https://github.com/rust-bitcoin/bitcoin_hashes/blob/16c0cb9375d3e12887ec9a5be2e758e3d907359e/src/hex.rs#L64-L85)

The `sv1_api` crate currently depends on `bitcoin_hashes@0.3.2`. While debugging the issue, I incrementally upgraded the dependency and observed that the panic no longer occurs by version `0.5.0`.

While investigating further, I noticed that the `bitcoin_hashes` repository has been archived and has not received updates in several years. The README states:

> Development of this library has now been moved over to the [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin) repository (see rust-bitcoin/src/hashes).

see: https://github.com/rust-bitcoin/bitcoin_hashes/blob/ce3ca68bd29a4e597340963a2af256838582e54c/README.md?plain=1#L3-L4

A quick search through the codebase also suggests that `bitcoin_hashes` is only used within the `sv1_api` crate.

During the investigation, I also noticed that one of the conditions that makes this crash possible is the padding logic implemented in `hex_decode()`:

[https://github.com/stratum-mining/stratum/blob/e4cbd6175845eeefb54392f1fb866890240d00c6/sv1/src/utils.rs#L50-L58](https://github.com/stratum-mining/stratum/blob/e4cbd6175845eeefb54392f1fb866890240d00c6/sv1/src/utils.rs#L50-L58)

Given all of the above, I can think of a few possible approaches:

* Upgrade `bitcoin_hashes` to a version that no longer panics on this input.
* Add validation before calling the decoder to reject non-ASCII characters. This would likely be the least invasive change.
* Replace `bitcoin_hashes` with the hash implementation from `rust-bitcoin`, which we already depend on elsewhere in the codebase. This seems like the cleanest long-term solution.

there is also the existing `FIXME` comment above `hex_decode()` regarding the padding behavior. I think it is worth discussing whether silently padding malformed input is the right approach.

my preference would be to stop "fixing" invalid inputs through padding and instead return an explicit error to the caller. That would make malformed data easier to detect and avoid surprising behavior.

lucasbalieiro:
closes #2201.

This PR replaces the use of `bitcoin_hashes@0.3.2`, which can panic when processing certain multibyte characters, with `bitcoin::hashes::hex`.

It also centralizes hex decoding behind a single utility function, `crate::utils::hex_decode`. Prior to this change, there were multiple hex conversion implementations scattered throughout the codebase, making behavior harder to reason about and maintain consistently.

While working on this issue, I also noticed inconsistent handling of odd-length hex strings. Some call sites would pad the input before decoding, while others would not.

After investigating the historical context (see #896), I introduced a dedicated wrapper named `hex_decode_with_padding`. The goal is to make padding an explicit and deliberate choice rather than an implicit side effect hidden within the decoding logic.

At the moment, `hex_decode_with_padding` is only used in the `TryFrom` implementation for `Extranonce`, which is the use case discussed in #896. The padding behavior for `Extranonce` also appears to be required to satisfy the unit test introduced in #1791.

### DIFF
--- a/sv1/src/methods/server_to_client.rs
+++ b/sv1/src/methods/server_to_client.rs
@@ -1,4 +1,3 @@
-use bitcoin_hashes::hex::FromHex;
 use serde_json::{
     Value,
     Value::{Array as JArrary, Bool as JBool, Number as JNumber, String as JString},
@@ -10,7 +9,7 @@ use crate::{
     error::Error,
     json_rpc::{Message, Notification, Response},
     methods::ParsingMethodError,
-    utils::{Extranonce, HexBytes, HexU32Be, MerkleNode, PrevHash},
+    utils::{hex_decode, Extranonce, HexBytes, HexU32Be, MerkleNode, PrevHash},
 };
 
 // client.get_version()
@@ -268,7 +267,7 @@ impl TryFrom<Notification> for SetExtranonce<'_> {
             .ok_or_else(|| ParsingMethodError::not_array_from_value(msg.params.clone()))?;
         let (extra_nonce1, extra_nonce2_size) = match &params[..] {
             [JString(a), JNumber(b)] => (
-                Extranonce::try_from(Vec::<u8>::from_hex(a)?)?,
+                Extranonce::try_from(hex_decode(a)?)?,
                 b.as_u64()
                     .ok_or_else(|| ParsingMethodError::not_unsigned_from_value(b.clone()))?
                     as usize,
--- a/sv1/src/methods/server_to_client.rs
+++ b/sv1/src/methods/server_to_client.rs
@@ -728,6 +728,23 @@ fn subscribe_response_crash_on_object_in_subscriptions() {
     assert!(result.is_err());
 }
 
+#[test]
+fn subscribe_response_crash_on_non_ascii_hex_extranonce() {
+    let json = r#"{
+      "id": 0,
+      "error": null,
+      "result": [
+        ["mining.notify", "1"],
+        "55555ʛ",
+        4
+      ]
+    }"#;
+
+    let response: crate::json_rpc::Response = serde_json::from_str(json).unwrap();
+    let result = crate::methods::Server2ClientResponse::try_from(response);
+    assert!(result.is_err(), "{:?}", result);
+}
+
 impl VersionRollingParams {
     pub fn new(
         version_rolling_mask: HexU32Be,
--- a/sv1/src/utils.rs
+++ b/sv1/src/utils.rs
@@ -68,7 +68,7 @@ pub fn hex_decode_with_padding(s: &str) -> Result<Vec<u8>, Error<'static>> {
         s
     };
 
-    Ok(hex_decode(s)?)
+    hex_decode(s)
 }
 
 impl<'a> TryFrom<&str> for Extranonce<'a> {
--- a/sv1/src/utils.rs
+++ b/sv1/src/utils.rs
@@ -1,6 +1,6 @@
 use crate::error::{self, Error};
 use binary_sv2::{B032, U256};
-use bitcoin_hashes::hex::{FromHex, ToHex};
+use bitcoin::hashes::hex::{DisplayHex, FromHex};
 use byteorder::{BigEndian, ByteOrder, LittleEndian, WriteBytesExt};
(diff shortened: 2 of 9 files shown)


sv1/src/methods/server_to_client.rs
```diff
@@ -729,6 +728,22 @@ fn subscribe_response_crash_on_object_in_subscriptions() {
...
+fn subscribe_response_crash_on_non_ascii_hex_extranonce() {
+    let json = r#"{
+        "id": 34560326215225189,
+        "error": null,
+        "result": [
+            [["", [-32088499539804252, {"": null}]], null, false],
+            "55555ʛ",
+            -1.0900715865557387e-175
+        ]
+    }"#;
```
plebhash, line 741: non-blocking clanker review:

> This regression test includes multiple invalid fields (subscriptions shape and extra_nonce2_size as a negative float), so it may pass without exercising the non-ASCII extranonce decode path.
>
> Suggestion: keep non-target fields valid (e.g. valid subscriptions + extra_nonce2_size: 4) so the test isolates the non-ASCII hex behavior.

plebhash (approved):

sv1/src/utils.rs
```diff
@@ -47,27 +47,41 @@ impl<'a> From<Extranonce<'a>> for Value {
...
+
+    hex_decode(s)
 }
 
 impl<'a> TryFrom<&str> for Extranonce<'a> {
     type Error = error::Error<'a>;
 
     fn try_from(value: &str) -> Result<Self, Error<'a>> {
-        Ok(Extranonce(B032::try_from(hex_decode(value)?)?))
+        Ok(Extranonce(B032::try_from(hex_decode_with_padding(value)?)?))
```
bit-aloo, line 78: does any such scenario even happen, odd length hex string? discussion is too long, some tldr would be helpful.

And considering decode_with_padding is used on once, maybe inlining would be ok.

sv1/src/methods/server_to_client.rs
lucasbalieiro: addresed, force-pushed [d501a1a](https://github.com/stratum-mining/stratum/commit/d501a1a15bc91aba9d3ccb04cd184791ba91bc71)

sv1/src/utils.rs
lucasbalieiro: TL;DR: At the time of the original issue, a faulty firmware was sending an odd-length hex `extranonce`, which caused errors in `tproxy`.

I agree that this scenario should not occur, and ideally we should reject invalid `extranonce` values instead of trying to "fix" them via left-padding (IMO).

however, following the history of the codebase, this behavior was intentionally carried over into the New TProxy work:
https://github.com/stratum-mining/stratum/pull/1791/changes/f0a69aa7f1410f26fa3c3802f9af067c900c1eea

If I remove the left-padding behavior now, it will break an existing unit test (which tells me that this is intentional). Since changing the semantics of `Extranonce` parsing is not the goal of this PR, I opted to preserve the current behavior. I also don't know whether any downstream users are relying on this behavior today, so keeping it seemed like the safer option for a targeted fix.

Regarding inlining `hex_decode_with_padding`, I wouldn't oppose that. My main concern is that, if we decide to keep left-padding in specific scenarios, we should avoid ending up in the same situation we had before, where hex decoding logic was scattered across the codebase with slightly different implementations.

One of the goals of this PR was to centralize hex decoding behind a small set of helpers so that any future changes to validation or decoding behavior can be made in a single place.

bit-aloo (approved):

GitGab19 (approved):

PR #2209, opened 2026-06-19
accelerate CI via `cargo binstall`

plebhash:
accelerate CI via [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall)

instead of compiling `cargo-nextest`, `cargo-semver-checks`, and `cargo-fuzz` on every run, we install pre-built binaries via `cargo-binstall`.

additional hardening:
- disable `quick-install` option to only use upstream release artifacts (instead of third-party quick-install artifacts)
- pass `GITHUB_TOKEN` to `cargo-binstall` to avoid unauthenticated GitHub API rate-limit fallback
- use `--locked` for deterministic fallback behavior if source install is ever needed
- run fuzz with an explicit rustc host target to avoid sanitizer/libc mismatches

estimated impact (latest successful PR run vs median of 5 recent comparable successful PR runs on the same workflows/jobs):

- `ci (ubuntu-latest)`:
  - tool install: `~2m33s -> ~2s` (**~99% reduction**)
  - total job: `~31m27s -> ~29m05s` (**~2m22s faster**, ~8%)
- `ci (macos-latest)`:
  - tool install: `~2m18s -> ~4s` (**~97% reduction**)
  - total job: `~30m16s -> ~29m01s` (**~1m15s faster**, ~4%)
- `semver-check`:
  - tool install: `~3m53s -> ~2s` (**~99% reduction**)
  - total job: `~7m12s -> ~3m17s` (**~3m55s faster**, ~54%)
- `fuzz`:
  - tool install: `~20s -> ~2s` (**~90% reduction**)
  - total job: `~1m14s -> ~1m01s` (**~13s faster**, ~18%)

critical-path wall clock improves (`~31m27s -> ~29m05s`, ~2m22s), while matrix jobs still run in parallel.

### DIFF
--- a/.github/workflows/fuzz_tests.yaml
+++ b/.github/workflows/fuzz_tests.yaml
@@ -25,8 +25,18 @@ jobs:
         run: |
           rustup override set ${{ steps.toolchain.outputs.name }}
 
+      # Install cargo-binstall so tool installs below use prebuilt binaries.
+      - name: Install cargo-binstall
+        uses: cargo-bins/cargo-binstall@v1.20.0
+        with:
+          version: "1.20.0"
+
+      # Install cargo-fuzz via cargo-binstall (no source build).
+      # Keep quick-install disabled to only use upstream release artifacts.
       - name: Install cargo-fuzz
-        run: cargo install cargo-fuzz
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: cargo binstall --no-confirm --disable-telemetry --disable-strategies quick-install --locked cargo-fuzz
 
       - name: Checkout fuzzing corpus
         uses: actions/checkout@v4
@@ -36,6 +46,16 @@ jobs:
 
       - name: Run fuzz regressions
         run: |
+          # cargo-fuzz binaries installed via cargo-binstall can default to a
+          # musl target triple; derive the runner's rustc host triple so
+          # sanitizer builds run against the native target instead.
+          HOST_TARGET="$(rustc -vV | sed -n 's/^host: //p')"
+
+          if [ -z "$HOST_TARGET" ]; then
+            echo "Failed to detect rustc host target" >&2
+            exit 1
+          fi
+
           for TARGET in $(cargo fuzz list); do
             CORPUS_DIR="fuzz/repo-corpus/corpus/$TARGET"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/fuzz_tests.yaml
```diff
@@ -25,8 +25,16 @@ jobs:
         run: |
           rustup override set ${{ steps.toolchain.outputs.name }}
 
+      # Install cargo-binstall so tool installs below use prebuilt binaries.
+      - name: Install cargo-binstall
+        uses: cargo-bins/cargo-binstall@v1.20.0
```
GitGab19, line 30:
```suggestion
      - name: Install cargo-binstall
        uses: cargo-bins/cargo-binstall@v1.20.0
        with:
            version: "1.20.0"
```

GitGab19 (commented):
One nit found by my clanker, which pins the version for real.

Without it, the action still downloads the latest it seems: https://github.com/stratum-mining/stratum/actions/runs/27847218527/job/82418833360?pr=2209#step:5:35

.github/workflows/fuzz_tests.yaml
plebhash: done everywhere here and on https://github.com/stratum-mining/sv2-apps/pull/578

GitGab19 (approved):

.github/workflows/fuzz_tests.yaml
```diff
@@ -25,8 +25,18 @@ jobs:
...
```
bit-aloo, line 30: Is this version beside binstall required anymore, considering now we have with version attribute? I think we can remove it

plebhash: my knowledge on`.yaml` syntax for Github Actions is admittedly shallow

clanker answer:

> It looks redundant, but it is **not** redundant in behavior.
> 
> - `uses: cargo-bins/cargo-binstall@v1.20.0` pins the **GitHub Action implementation**.
> - `with: version: "1.20.0"` pins the **cargo-binstall binary that the action installs**.
> 
> If you remove either one, one layer becomes floating:
> - remove `with.version` => installs latest binary
> - remove exact `@v1.20.0` (e.g. use `@v1`) => action code can change
> 
> So keeping both is the correct fully-pinned setup.

bit-aloo: I didn't knew it pointed to action code, the first version. Cool, makes sense.

bit-aloo:
@plebhash Can you squash the commits, as kinda related?

bit-aloo (approved):

PR #2211, opened 2026-06-23
Return Header from Sv2Frame::get_header

### LINKED ISSUE #2112: Sv2Frame `get_header()` returns `Option<Header>` but is always Some
Every Sv2Frame always has a header. The return type should be Header directly. Currently every caller has to .unwrap() or pattern-match on a None that can never happen.

```rust
  pub fn get_header(&self) -> Option<crate::header::Header> {
      Some(self.header) 
  }
```

DeepCogNeural:
## Summary
- Change `Sv2Frame::get_header()` to return `Header` directly instead of `Option<Header>`.
- Update related tests and examples to remove impossible `None` handling.

Fixes #2112.
companion https://github.com/stratum-mining/sv2-apps/pull/584

## Verification
- `cargo fmt --all --check`
- `cargo test -p framing_sv2`
- `cargo test -p codec_sv2`
- `cargo check --workspace --all-targets`
- `git diff --check`

Companion verification:
- `cargo fmt --all --check` in `integration-tests`, `miner-apps`, `pool-apps`, and `stratum-apps`
- `cargo check --manifest-path integration-tests/Cargo.toml` with `stratum-core` patched to this PR checkout

### DIFF
--- a/sv2/codec-sv2/examples/encrypted.rs
+++ b/sv2/codec-sv2/examples/encrypted.rs
@@ -228,9 +228,7 @@ fn main() {
     }
 
     // Parse the decoded frame header and payload
-    let decoded_frame_header = decoded_frame
-        .get_header()
-        .expect("Failed to get the frame header");
+    let decoded_frame_header = decoded_frame.get_header();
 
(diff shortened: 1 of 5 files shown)


bit-aloo:
This API is now changed.

PR #2212, opened 2026-06-23
remove custom impls of `serde::Deserialize` and `serde::Serialize` for HexBytes and HexU32Be

### LINKED ISSUE #2208: HexBytes::Serialize/Deserialize impls are broken leftovers from Message Generator
https://github.com/stratum-mining/stratum/blob/f7687927c97472ab6bd44edfb05da43f35bb95a6/sv1/src/utils.rs#L397-L408

the `Serialize` and `Deserialize` implementations for `HexBytes` appear to be buggy and are likely leftovers from the Message Generator testing tool introduced in PR #602.

While building a fuzz target for `sv1_api`, I found some issues with these implementations:

* `Serialize` uses `String::from_utf8_lossy(bytes)` instead of hex encoding. As a result, any non-UTF-8 bytes  are silently corrupted. Invalid bytes are replaced with `U+FFFD`. 

* `Deserialize` stores `hex_string.as_bytes().to_vec()` directly without performing any hex decoding. An input of `"dead"` becomes the four ASCII bytes `[0x64, 0x65, 0x61, 0x64]` instead of the two bytes `[0xde, 0xad]`. Arbitrary non-hex strings are also accepted without validation.

Both implementations contain the comment `// Example of serialization for testing purpose`, which made sense when they were added for the Message Generator. However, that tool has since been deprecated, while these implementations remain exposed through the public API.

Although the internal production code paths use manual `From`/`TryFrom` implementations and avoid these serde implementations, `HexBytes` is still a public type with public `Serialize` and `Deserialize` implementations. Any downstream user who:

* Derives `Serialize` on a struct containing `HexBytes`, or
* Uses `serde_json::to_value()` / `serde_json::from_value::<HexBytes>()`

can end up silently corrupting their data during serialization or accepting invalid input during deserialization.

possible resolutions:

1. Remove the `serde` implementations entirely. They were originally added for a deprecated testing tool, and their continued existence exposes downstream users to incorrect behavior. And just use the proc_macros provided by serde

2. Fix the implementations to match the documented contract of `HexBytes` ("represented as hex strings in JSON"):

   * Serialize using `self.0.to_lower_hex_string()`.
   * Deserialize using `hex_decode(&s)` and return an error for invalid hex input.
   * Remove the outdated testing-related comments.

lucasbalieiro:
closes #2208

### DIFF
--- a/sv1/src/utils.rs
+++ b/sv1/src/utils.rs
@@ -2,7 +2,6 @@ use crate::error::{self, Error};
 use binary_sv2::{B032, U256};
 use bitcoin::hashes::hex::{DisplayHex, FromHex};
 use byteorder::{BigEndian, ByteOrder, LittleEndian, WriteBytesExt};
-use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::fmt;
 use std::{convert::TryFrom, mem::size_of, ops::BitAnd};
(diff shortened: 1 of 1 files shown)


sv1/src/utils.rs
```diff
@@ -100,7 +100,7 @@ impl<'a> From<B032<'a>> for Extranonce<'a> {
 }
 
 /// Big-endian alternative of the HexU32
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
```
lucasbalieiro: another point is that none of the other types in this file implement (or derive) Serde's `Serialize`/`Deserialize`. The only ones that do are the examples introduced.

for example, types like `Extranonce` rely exclusively on their `From`/`TryFrom` implementations for serialization and parsing.

Given that the rest of the codebase already follows this pattern, it might actually be safer to remove the Serde implementations from these example types altogether. That would reduce the public API surface and avoid exposing serialization behavior that was originally intended only for testing purposes.

GitGab19 (approved):

bit-aloo (approved):
This looks good. We should have bumped the sv1_api version, but we already did that in one of our earlier PRs.

During the call, I was a bit confused about how we convert from Value to our SV1 types. After looking at the code, I realized we have a custom parser that converts Value into specific SV1 types. This is mainly because the same SV1 messages can have a variety of different shapes. We can look into improving this in the future, but it's not a priority at the moment.

PR #2219, opened 2026-06-30
remove `unwrap` calls when using `as_u64`

### LINKED ISSUE #2218: unwraps triggered inside client_to_server.rs
while fuzzing the sv1 crate, some `unwrap`s were triggered while doing some roundtrip checks.

e.g: 
https://github.com/stratum-mining/stratum/blob/6be1267d261fccddb326d0a78d8a12cdc2606533/sv1/src/methods/client_to_server.rs#L196-L208

While looking to see if server_to_client.rs has the same behavior, noticed that this path was already correctly handled there

lucasbalieiro:
closes #2218

### DIFF
--- a/sv1/src/methods/client_to_server.rs
+++ b/sv1/src/methods/client_to_server.rs
@@ -202,8 +202,12 @@ impl TryFrom<StandardRequest> for Submit<'_> {
                         a.into(),
                         b.into(),
                         Extranonce::try_from(hex_decode(c)?)?,
-                        HexU32Be(d.as_u64().unwrap() as u32),
-                        HexU32Be(e.as_u64().unwrap() as u32),
+                        HexU32Be(d.as_u64().ok_or_else(|| {
+                            ParsingMethodError::ImpossibleToParseAsU64(Box::new(d.clone()))
+                        })? as u32),
+                        HexU32Be(e.as_u64().ok_or_else(|| {
+                            ParsingMethodError::ImpossibleToParseAsU64(Box::new(e.clone()))
+                        })? as u32),
                         Some((f.as_str()).try_into()?),
                     ),
                     [JString(a), JString(b), JString(c), JString(d), JString(e), JString(f)] => (
@@ -218,8 +222,12 @@ impl TryFrom<StandardRequest> for Submit<'_> {
                         a.into(),
                         b.into(),
                         Extranonce::try_from(hex_decode(c)?)?,
-                        HexU32Be(d.as_u64().unwrap() as u32),
-                        HexU32Be(e.as_u64().unwrap() as u32),
+                        HexU32Be(d.as_u64().ok_or_else(|| {
+                            ParsingMethodError::ImpossibleToParseAsU64(Box::new(d.clone()))
+                        })? as u32),
+                        HexU32Be(e.as_u64().ok_or_else(|| {
+                            ParsingMethodError::ImpossibleToParseAsU64(Box::new(e.clone()))
+                        })? as u32),
                         None,
                     ),
                     [JString(a), JString(b), JString(c), JString(d), JString(e)] => (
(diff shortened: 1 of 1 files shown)


sv1/src/methods/client_to_server.rs
```diff
@@ -198,14 +198,20 @@ impl TryFrom<StandardRequest> for Submit<'_> {
...
-                        Some((f.as_str()).try_into()?),
-                    ),
+                    [JString(a), JString(b), JString(c), JNumber(d), JNumber(e), JString(f)] => {
+                        (
+                            a.into(),
+                            b.into(),
+                            Extranonce::try_from(hex_decode(c)?)?,
+                            HexU32Be(d.as_u64().ok_or_else(|| {
+                                ParsingMethodError::not_unsigned_from_value(d.clone())
+                            })? as u32),
```
plebhash: `as_u64()` followed by `as u32` gives room for silent truncation

```suggestion
                            HexU32Be(
                                u32::try_from(
                                    d.as_u64()
                                        .ok_or_else(|| ParsingMethodError::not_unsigned_from_value(d.clone()))?,
                                )
                                .map_err(|_| ParsingMethodError::not_unsigned_from_value(d.clone()))?,
                            ),
```

---

leaving this suggestion here, but also applies to other places on this PR

plebhash: and since we're adding unit test coverage, it would be good to include truncation as well

lucasbalieiro: but removing the `as u32` will not preserve the behavior that the crate already has which is:

`HexU32Be(d.as_u64().unwrap() as u32)`

plebhash: oops, maybe `ParsingMethodError::ValueNotAnUnsigned` is not the right variant for the `u32::try_from` failure above

maybe `ParsingMethodError::UnexpectedValue`, or add a a new `ParsingMethodError::ValueOutOfRange`?

plebhash: > but removing the `as u32` will not preserve the behavior that the crate already has which is:
> 
> `HexU32Be(d.as_u64().unwrap() as u32)`

yeah I guess one could argue that fixing this silent truncation is out of scope from the original issue that was found via fuzzying

but if I had to guess, I'd say that the original `unwrap()` followed by `as u32` feels like a quick-and-dirty approach rather than having silent truncation as an intentional decision

so the suggestion above is just trying to take advantage of the fact that we're touching this part of the code, so we can solve two problems while we're at it

---

with that said, no hard opinions, fine to leave it as-is and fix this later if it causes problems

lucasbalieiro: > maybe ParsingMethodError::UnexpectedValue, or add a a new ParsingMethodError::ValueOutOfRange?

re-used the existing `ImpossibleToParseAsU64` error and force-pushed(https://github.com/stratum-mining/stratum/compare/d5b6e8608b30d5d1a53e244e8cadaa58d6875701..27dd92b1bb6b07c1375107ef220411b0f14389de) 

Regarding the removal of the silent truncation, I think that may have a broader impact and would be better handled in a separate PR if we decide it's necessary.

a few notes to link in a follow-up issue, if necessary:

* `HexU32Be` is a `u32` wrapper, and `serde_json::Number` does not provide an `as_u32()` API. Avoiding the cast would require introducing a `TryFrom<u64>` and updating all call sites. (touching only the files of this PR will leave the codebase inconsistent).

* From a quick grep, all `HexU32Be` fields appear to represent inherently 32-bit Bitcoin header values (`nTime`, `nonce`, `nBits`, `version`).

* The canonical SV1 path seems to already prevents overflow: `HexU32Be::try_from(&str)` rejects hex strings longer than 8 characters. see:
  https://github.com/stratum-mining/stratum/blob/6be1267d261fccddb326d0a78d8a12cdc2606533/sv1/src/utils.rs#L126-L130

plebhash (approved):

bit-aloo (approved):

Issue #2206, opened 2026-06-18
sv1_api crate panics when extranonce is too long


lucasbalieiro:
test case proving it:
```rust
#[test]
fn subscribe_response_crash_on_too_long_extranonce() {
    let json = r#"{
      "id": 0,
      "error": null,
      "result": [
        ["mining.notify", "1"],
        "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
        4
      ]
    }"#;
    let response: crate::json_rpc::Response = serde_json::from_str(json).unwrap();
    let result = crate::methods::Server2ClientResponse::try_from(response);
    assert!(result.is_err(), "{:?}", result);
}
```
failure observed:

```bash
---- methods::server_to_client::subscribe_response_crash_on_too_long_extranonce stdout ----
thread 'methods::server_to_client::subscribe_response_crash_on_too_long_extranonce' panicked at sv1/src/methods/mod.rs:65:18:
v1 Error does not implement this ParsingMethodError, but probably should
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #2220, opened 2026-06-30
add fuzz targets for sv1 crate

### LINKED ISSUE #2124: add fuzz coverage for `sv1` crate

lucasbalieiro:
closes #2124  (to be merged after #2219)

two fuzz targets are added:
- `fuzz_sv1_wire`: throws random bytes at the SV1 JSON parser (mimicking an attacker sending garbage over TCP). This is the most untrusted input path for this crate (`sv1_connection` in `network_helpers`). 
    - caught: 
        - #2197
        - #2201
        - #2206
        - #2208 

- `fuzz_sv1_method_parsers`:  random bytes alone (like we are doing on the first target) rarely produce a well-formed `mining.submit` or similar structured message, so this target uses biased generators to craft semi-valid params that reach deeper parser paths. caught: #2218.


fuzz/rust-toolchain.toml
GitGab19, line 1: Why this change?

lucasbalieiro: I added this so I don't have to type `+nightly` every time I have to run a target.

fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+fn gen_valid_hex(
+    u: &mut Unstructured<'_>,
+    len_range: RangeInclusive<usize>,
+) -> arbitrary::Result<String> {
+    let len = u.int_in_range(len_range)?;
+    let nibbles: String = (0..len)
+        .map(|_| match u.int_in_range(0..=15).unwrap() {
```
plebhash: clanker review:

> `gen_valid_hex` can panic in the harness due to `unwrap()`
> 
> This can create false-positive fuzz crashes that are not parser bugs.

fuzz/Cargo.lock
GitGab19, line 1: Can you generate this with Cargo 1.75?

My clanker found that `cargo +1.75.0 check --manifest-path fuzz/Cargo.toml` fails.

lucasbalieiro: that is expected. the fuzzer is crafted to run on nightly. The `edition2024` is not stabilized in +1.75

```bash
error: failed to parse manifest at `/Users/lucasbalieiro/dev/stratum/fuzz/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.75.0 (1d8b05cdd 2023-11-20)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
  ```

File: fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs (hunk 1 above)
Comment (lucasbalieiro) [reply fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs]: force-pushed the removal these `unwrap`s.

**Review by plebhash — APPROVED** [review]

File: fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs [hunk 2]
```diff
@@ -0,0 +1,280 @@
...
+}
+
+fn gen_notify_params(u: &mut Unstructured<'_>) -> arbitrary::Result<Value> {
+    let branch_len = u.int_in_range::<usize>(0..=8)?;
+    let branch: Vec<Value> = (0..branch_len)
+        .map(|_| Value::String(gen_valid_hex(u, 1..=8).unwrap_or_default()))
+        .collect();
+    Ok(Value::Array(vec![
+        Value::String(u.arbitrary()?),
+        Value::String(gen_valid_hex(u, 32..=32)?),
```
GitGab19: Another point found by the clanker:
```
For the biased `mining.notify` path, should these hashes be generated as 64 hex chars? `PrevHash::try_from` expects 32 decoded bytes, so `gen_valid_hex(u, 32..=32)` only gives 16 bytes and likely makes this path fail before it reaches the rest of notify parsing. Same for merkle branch entries, since `MerkleNode` is `U256`-backed too.
```

plebhash: I tried to fix it locally and the diff looks like this:
```diff
diff --git a/fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs b/fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs
index 3154a6d82..e7e768ff7 100644
--- a/fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs
+++ b/fuzz/fuzz_targets/fuzz_sv1_method_parsers.rs
@@ -102,11 +102,11 @@ fn gen_configure_params(u: &mut Unstructured<'_>) -> arbitrary::Result<Value> {
 fn gen_notify_params(u: &mut Unstructured<'_>) -> arbitrary::Result<Value> {
     let branch_len = u.int_in_range::<usize>(0..=8)?;
     let branch: Vec<Value> = (0..branch_len)
-        .map(|_| Value::String(gen_valid_hex(u, 1..=8).unwrap_or_default()))
+        .map(|_| Value::String(gen_valid_hex(u, 64..=64).unwrap_or_default()))
         .collect();
     Ok(Value::Array(vec![
         Value::String(u.arbitrary()?),
-        Value::String(gen_valid_hex(u, 32..=32)?),
+        Value::String(gen_valid_hex(u, 64..=64)?),
         Value::String(gen_valid_hex(u, 1..=8)?),
         Value::String(gen_valid_hex(u, 1..=8)?),
         Value::Array(branch),
~
```

lucasbalieiro: makes sense

lucasbalieiro: force-pushed the solution

GitGab19 (approved):

PR #2224, opened 2026-07-09
allow access to all sv1 message props

coleFD:
make fields in sv1 messages public

### DIFF
--- a/sv1/src/methods/client_to_server.rs
+++ b/sv1/src/methods/client_to_server.rs
@@ -385,8 +385,8 @@ impl TryFrom<StandardRequest> for Subscribe<'_> {
 
 #[derive(Debug, Clone)]
 pub struct Configure {
-    extensions: Vec<ConfigureExtension>,
-    id: u64,
+    pub extensions: Vec<ConfigureExtension>,
+    pub id: u64,
 }
 
 impl fmt::Display for Configure {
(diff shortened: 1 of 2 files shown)


sv1/src/methods/server_to_client.rs
plebhash: missing this one

is that intentional, or not needed for your use-case @coleFD ?

sv1/src/methods/client_to_server.rs
```diff
@@ -679,8 +679,8 @@ impl From<ConfigureExtension> for serde_json::Map<String, Value> {
 
 #[derive(Debug, Clone)]
 pub struct VersionRollingParams {
-    mask: Option<HexU32Be>,
-    min_bit_count: Option<HexU32Be>,
+    pub mask: Option<HexU32Be>,
+    pub min_bit_count: Option<HexU32Be>,
 }
```
plebhash, line 694: making these fields public means `VersionRollingParams` can now be instantiated with combinations we currently treat as invalid when parsing.

for example:
```rust
VersionRollingParams { mask: None, min_bit_count: Some(HexU32Be(0x10)) }
```

`ConfigureExtension::from_value` explicitly rejects this shape (`(None, Some(_)) => Err(...)`).

so I wonder whether it wouldn't be safer to change the PR approach, while keeping internal values private and exposing getters instead?

sv1/src/methods/server_to_client.rs
coleFD: nope just missed it

sv1/src/methods/client_to_server.rs
plebhash: pushing 134ba2be3b9cb775792b15fee4eb16405beb1197 here to address this

plebhash (approved):

PR #2225, opened 2026-07-13
split inner to owned and ref variants

### LINKED ISSUE #2180: Binary_sv2: Add owned primitive sv2 types
Based on our findings in #2136, the incremental path forward, without breaking all lower-level dependencies, is to introduce owned SV2 primitives instead of deprecating the `Inner` representation of SV2 primitives immediately.

Ideally, in the future, we would like to fully deprecate the `Inner` representation and instead have distinct `Owned` and `Ref` types. This would allow users of `binary_sv2` to decide how they want to structure and scope their ADTs, very similar to how `serde` approaches ownership and borrowing.

With the introduction of owned SV2 primitive types, changes will need to be made across `parser_sv2` and the subprotocol crates by introducing owned variants of SV2 messages. Alongside this, the `into_static` method should be updated to return owned message variants instead of owned variants of `Inner` binary SV2 constructs.

After that, we can begin migrating `sv2-apps` to this owned-type framework.

Once the migration has successfully propagated up the stack to `sv2-apps`, we can start removing the `Inner` types entirely and replace them with concrete `Ref` types that can be converted into `Owned` variants when required.

At a high level, this is essentially a transition from a single `Inner` type abstraction into separate `Ref` and `Owned` representations with explicit ownership semantics.

bit-aloo:
closes: #2180

companion https://github.com/stratum-mining/sv2-apps/pull/616

### DIFF
--- a/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
+++ b/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
@@ -1,208 +1,432 @@
-// Provides a flexible container for managing either owned or mutable references to byte arrays.
-//
-// # Overview
-// Defines the `Inner` enum to manage both mutable references to byte slices and owned vectors
-// (`Vec<u8>`). Accommodates both fixed-size and variable-size data using const generics, offering
-// control over size and header length constraints.
-//
-// # `Inner` Enum
-// The `Inner` enum has two variants for data management:
-// - `Ref(&'a mut [u8])`: A mutable reference to a byte slice, allowing in-place data modification.
-// - `Owned(Vec<u8>)`: An owned byte vector, providing full control over data and supporting move
-//   semantics.
-//
-// ## Const Parameters
-// Configured using const generics for the following constraints:
-// - `ISFIXED`: Indicates whether the data has a fixed size.
-// - `SIZE`: Specifies the size when `ISFIXED` is true.
-// - `HEADERSIZE`: Defines the size of the header, useful for variable-size data with a prefix
-//   length.
-// - `MAXSIZE`: Limits the maximum allowable size of the data.
-//
-// # Usage
-// `Inner` offers several methods for data manipulation, including:
-// - `as_bytes()` and `as_mut_bytes()`: Provide immutable or mutable access to the data.
-// - `to_owned_bytes()` and `into_bytes()`: Return owned payload bytes.
-// - `len()` and `is_empty()`: Inspect payload size directly.
-// - `expected_length(data: &[u8])`: Computes the expected length, validating it against
-//   constraints.
-// - `get_header()`: Returns the data's header based on `HEADERSIZE`.
-//
-// # Implementations
-// The `Inner` enum implements `PartialEq`, `Eq`, `GetSize`, `SizeHint`, and `Sv2DataType` traits,
-// enabling buffer size calculations, reading, and writing to byte slices.
-//
-// # Error Handling
-// Methods return `Error` types when data exceeds size limits or deviates from the configuration,
-// ensuring compliance with defined constraints.
+// Provides separate borrowed and owned byte wrappers for SV2 byte-array primitives.
 use crate::{
     codec::{GetSize, SizeHint},
     datatypes::Sv2DataType,
     Error,
 };
 
-use alloc::vec::Vec;
-use core::convert::{TryFrom, TryInto};
-#[cfg(not(feature = "no_std"))]
-use std::io::{Error as E, Read, Write};
+use alloc::{string::String, vec::Vec};
+use core::{
+    convert::{TryFrom, TryInto},
+    fmt,
+};
 
-// The `Inner` enum represents a flexible container for managing both reference to mutable
-// slices and owned bytes arrays (`Vec<u8>`). This design allows the container to either own
-// its data or simply reference existing mutable data. It uses const generics to differentiate
-// between fixed-size and variable-size data, as well as to specify key size-related parameters.
-//
-// It has two variants:
-// - `Ref(&'a mut [u8])`: A mutable reference to an external byte slice.
-// - `Owned (Vec<u8>)`: A vector that owns its data, enabling dynamic ownership.
-//
-// The const parameters that govern the behavior of this enum are:
-//  - `ISFIXED`: A boolean indicating whether the data has a fixed size.
-//  - `SIZE`: The size of the data if `ISFIXED` is true.
-//  - `HEADERSIZE`: The size of the header, which is used for types that require a prefix to
-//    describe the content's length.
-//  - `MAXSIZE`: The maximum allowable size for the data.
-
-#[derive(Debug)]
-pub enum Inner<
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct Inner<
     'a,
     const ISFIXED: bool,
     const SIZE: usize,
     const HEADERSIZE: usize,
     const MAXSIZE: usize,
 > {
-    Ref(&'a mut [u8]),
-    Owned(Vec<u8>),
+    data: &'a [u8],
 }
 
-impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXSIZE: usize>
-    PartialEq for Inner<'_, ISFIXED, SIZE, HEADERSIZE, MAXSIZE>
-{
-    // Provides equality comparison between two `Inner` instances by checking the equality
-    // of their data, regardless of whether they are references or owned vectors.
-    fn eq(&self, other: &Self) -> bool {
-        match (self, other) {
-            (Inner::Ref(b), Inner::Owned(a)) => *b == &a[..],
-            (Inner::Owned(b), Inner::Ref(a)) => *a == &b[..],
-            (Inner::Owned(b), Inner::Owned(a)) => b == a,
-            (Inner::Ref(b), Inner::Ref(a)) => b == a,
-        }
-    }
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct InnerOwned<
+    const ISFIXED: bool,
+    const SIZE: usize,
+    const HEADERSIZE: usize,
+    const MAXSIZE: usize,
+> {
+    data: Vec<u8>,
 }
 
-impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXSIZE: usize> Eq
-    for Inner<'_, ISFIXED, SIZE, HEADERSIZE, MAXSIZE>
-{
-}
+trait InnerBytes {
+    const ISFIXED: bool;
+    const SIZE: usize;
+    const HEADERSIZE: usize;
+    const MAXSIZE: usize;
 
-impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXSIZE: usize>
-    Inner<'_, ISFIXED, SIZE, HEADERSIZE, MAXSIZE>
-{
-    // Calculates the expected length of the data based on the type's parameters (fixed-size
-    // or variable-size). It checks if the length conforms to the specified constraints like
-    // `SIZE`, `MAXSIZE`, and `HEADERSIZE`, returning the length or an error if the data
-    // exceeds the limits.
-    fn expected_length(data: &[u8]) -> Result<usize, Error> {
-        match ISFIXED {
-            true => Ok(Self::expected_length_fixed()),
-            false => Self::expected_length_variable(data),
-        }
-    }
+    fn as_bytes(&self) -> &[u8];
 
-    // For fixed-size data, the expected length is always `SIZE`.
-    fn expected_length_fixed() -> usize {
-        SIZE
-    }
-
-    // For variable-size data, this method calculates the size based on the header.
-    // The header describes the length of the data, and this method ensures the data
-    // is correctly sized relative to the header information.
-    fn expected_length_variable(data: &[u8]) -> Result<usize, Error> {
-        if data.len() >= HEADERSIZE {
-            let payload_len = match HEADERSIZE {
-                1 => Ok(data[0] as usize),
-                2 => Ok(u16::from_le_bytes([data[0], data[1]]) as usize),
-                3 => Ok(u32::from_le_bytes([data[0], data[1], data[2], 0]) as usize),
-                // HEADERSIZE for Sv2 datatypes is at maximum 3 bytes
-                // When HEADERSIZE is 0 datatypes ISFIXED only exception is Bytes datatypes but is
-                // not used
-                _ => unreachable!(),
-            };
-            payload_len.and_then(|payload_len| {
-                if payload_len <= MAXSIZE {
-                    Ok(payload_len + HEADERSIZE)
-                } else {
-                    Err(Error::ValueExceedsMaxSize(
-                        ISFIXED,
-                        SIZE,
-                        HEADERSIZE,
-                        MAXSIZE,
-                        data.to_vec(),
-                        payload_len,
-                    ))
-                }
-            })
+    fn len(&self) -> usize {
+        let len = self.as_bytes().len();
+        if Self::ISFIXED {
+            debug_assert_eq!(len, Self::SIZE);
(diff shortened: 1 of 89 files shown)


sv2/subprotocols/common-messages/src/setup_connection.rs
```diff
@@ -221,6 +254,34 @@ pub struct SetupConnectionSuccess {
...
+        }
+    }
+
+    pub fn as_owned(&self) -> SetupConnectionSuccessOwned {
+        (*self).into_owned()
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Copy)]
+pub struct SetupConnectionSuccessOwned {
```
plebhash: in the case of `SetupConnectionSuccess`, there's no lifetime on the message fields, therefore the `SetupConnectionSuccessOwned` variant is not implemented via macro derivation

so we ended up copy-pasting the entire struct just to enforce the pattern where **every** message has its `*Owned` counterpart

we should definitely enforce this pattern for every message, but I wonder if a simple alias wouldn't have done the trick here?

imagine there's a small spec change... having to apply it 2x always introduces the risk for human error 

(even though LLMs vastly reduce this kind of risk nowadays, and we're also choosing to live with this kind of risk in `bitcoin_core_sv2` as well)

the alias also gives us the opportunity to have some rust docs above it explaining the rationale behind it (e.g.: why this is an alias and not a macro derivation like most other messages)

(I'm writing this on `SetupConnection`, but I'm sure there's other similar instances where the suggestion also applies)

---

anyways, just a small nit... not really something I'm highly opinionated about, feel free to ignore if you feel it's too much labor for very little gain

sv2/binary-sv2/tests/test.rs
plebhash, line 692: clanker review:

> `b032_from_reader_should_reject_oversize_payload` calls `B032::from_reader_(&mut cursor)` and asserts `is_err()`. But `Inner::from_reader_ (inner.rs:642)` now unconditionally returns `Err(Error::PrimitiveConversionError)` — a borrowed `Inner` can never be built from a reader anymore. The test passes vacuously and no longer exercises the oversize-rejection logic it documents. It should use `B032Owned::from_reader_` to actually test the 33-byte-payload rejection.

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
```diff
@@ -1,508 +1,854 @@
...
 }
 
 impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXSIZE: usize>
-    TryFrom<Vec<u8>> for Inner<'_, ISFIXED, SIZE, HEADERSIZE, MAXSIZE>
+    fmt::Display for InnerOwned<ISFIXED, SIZE, HEADERSIZE, MAXSIZE>
 {
-    type Error = Error;
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt_hex(self.as_bytes(), f)
+    }
```
plebhash: clanker review:

> this umbrella `impl Display for InnerOwned` doesn't work — `Display` semantics live at the alias level, not the const generics (`U256Owned` and `PubKeyOwned` are the same type: `InnerOwned<true, 32, 0, 0>`).
> 
> and it's already a bug: `U256<'_>` prints reversed hex, but `U256Owned` prints plain here — so owned messages (`prev_hash`, `merkle_root`, `target`) render in the opposite byte order from their borrowed counterparts. `B032Owned`/`B064KOwned` also lose the `B032(...)`/`B064K(...)` wrappers.
> 
> please drop this impl and add per-alias impls in `mod.rs` mirroring the borrowed ones: `U256Owned` (reversed), `B032Owned` and `B064KOwned` (plain, with wrapper). the other owned aliases don't need `Display`.

plebhash:
recently @Shourya742 asked this over Discord:

> 
> At a high level, we're moving away from the old Inner enum, which previously had both Ref and Owned variants. We still have Inner, but now it only represents the Ref variant, and we've introduced InnerOwned to represent the Owned variant. This is purely an internal implementation detail of binary_sv2. Inner can be converted into the owned variant, similar to how the old Static variant worked. That's the main idea behind all the changes in binary_sv2.
> 
> In parsers_sv2 and the protocol-specific message crates, we now support both the owned and the previous reference-based variants with lifetimes, along with conversions from the reference type to the owned type. The handlers also have owned variants now. So far, all of that feels good.
> 
> The part I'm unsure about is channels_sv2, since we need to be careful about the API we expose to users.
> 
> My initial thought was that if channels_sv2 needs to store a value, it should require the owned variant; otherwise, it can just work with the reference-based type. The catch is that, in our sv2-apps implementation, we almost always convert reference types to 'static so they can be used in async Rust. So while the public API could be generic over reference-based types, internally we immediately turn them into 'static anyway.
> 
> That feels like an implementation detail driven by our heavy use of async Rust, but as a library we probably shouldn't be opinionated about how users structure their code.
> 
> Right now, I've made everything in channels_sv2 use the owned types, essentially saying, "these types own their memory, so the library is free to store them or do whatever it needs." I'm not sure if that's the right direction, though.
> 
> The other option would be a hybrid approach where channels_sv2 supports both owned and borrowed variants, but that seems like it would introduce a lot of duplication (unless we abstract it behind a trait).
> 
> Curious to hear your thoughts on which direction makes more sense. 

it took me a while to find bandwidth to look at this carefully, but I was finally able to do it today

first, thanks @Shourya742  for the detailed explanation... the reasoning feels very clear and I can get a very good understanding of the problem at hand

overall, PR looks solid too, and I'm excited for the much leaner codebase we're achieving as a result

---

now, looking at `channels_sv2` crate... I always like to think of it as something that's designed to work as a state-machine that represents how the Sv2 channels evolve over the app lifetime

conversely, main purpose of `buffer_sv2` is the ability to "recycle" the same (relatively small) amount of Heap memory in a sustainable way (i.e. without fragmentation or over-consumption as time goes on) for millions and millions of messages across the app's lifetime

so the only part of the message above that I would push back is:
> but as a library we probably shouldn't be opinionated about how users structure their code.

while this principle definitely holds true for crates such as `parsers_sv2` and `handlers_sv2` (and it's good to be careful to avoid overlooking it), if we were to provide ref-based APIs on `channels_sv2`, the channel state lifetime would be essentially holding buffer space hostage, which defeats the purpose of `buffer_sv2` in the first place

so for the case of `channels_sv2` crate, I wouldn't necessarily frame this design decision as being opinionated... it's simply a constraint of the broader design space

perhaps a good way to summarize this into a rule-of-thumb:

as frames arrive over the wire:
- if the message can be dropped immediately after being processed (e.g.: `Sniffer` use-case), then reference-based message type is the right approach, because it allows the underlying `BufferPool` slots to be recycled (as intended), while keeping the global app execution fast and efficient (no cpu/memory "wasted" copying/cloning stuff around)
- if the message needs to persist in memory for a long time (e.g.: Job Factory where `NewTemplate` message acts as `JobOrigin` of `ExtendedJob`, for which we might want to reconstruct a block/solution later), then we should convert it into the `Owned` variant so we can free-up the underlying `BufferPool` as soon as possible (while accepting the price for this extra copying/cloning, which is the right trade-off for this kind of scenario)

I'm sure none of this will not come as a surprise to you, as I vaguely recall that we've had similar discussions in dev calls in the past (and your mental model on these kinds of concepts is usually way sharper than mine 😅)

nevertheless, it's a good exercise to re-visit this point now so we can confidently move on in this direction without looking back later

---

TLDR: with regards to `channels_sv2`, this PR is already in the right direction

plebhash:
I'm proof-reading my message above and I realized that it's maybe too centered around `buffer_sv2`

and given the main sentence that I was trying to push back against is:

> but as a library we probably shouldn't be opinionated about how users structure their code.

perhaps one could argue that using `channels_sv2` without `buffer_sv2` is a valid use-case, and therefore the sentence above is true.

but I think the core of my rationale remains, while `buffer_sv2` is just one possible instance of the real constraint:

channel state will always outlive whatever buffer a frame was decoded from, so borrowing a message that can be stored somewhere else is either impossible (if that memory is reused, the compiler won't allow it) or just indirect ownership with bad ergonomics (a lifetime parameter leaked into every type that holds a channel) and terrible consequences with regards to memory fragmentation

so TLDR: taking `*Owned` at the `channels_sv2` boundary is basically a domain constraint, not really opinionated API design

sv2/subprotocols/common-messages/src/setup_connection.rs
bit-aloo: Initially, I used type alias, but then I thought someone would might point this out and ask to have separate struct itself. I am ok with both.

sv2/binary-sv2/tests/test.rs
bit-aloo: Fixed thanks.

sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
bit-aloo: Remove thanks.

sv2/subprotocols/common-messages/src/setup_connection.rs
bit-aloo: Added alias back.

GitGab19:
Initial clanker review.

---

## Review summary

Overall this is a well-executed refactor — the downstream propagation is remarkably clean: all Owned parser dispatch tables pair the correct message-type/channel-bit constants (including the two easily-confused `SetNewPrevHash` messages), the five `*_owned.rs` handler modules are semantically identical to their ref counterparts with full trait coverage parity, and sv1/stratum-translation wire behavior is byte-for-byte unchanged. Tests and CI are green.

That said, I found **two real issues in the core `binary-sv2` changes** that I think should be addressed before merge, plus a few smaller items.

### 🔴 `Decodable::from_reader` no longer works for any borrowed message type

`PrimitiveMarker::from_reader` (`sv2/binary-sv2/src/codec/decodable.rs`) now always returns the `*Owned` variants of `DecodablePrimitive`, but the `TryFrom<DecodablePrimitive>` impls for the borrowed types (`impls.rs`) still only accept the borrowed variant — so decoding any message with a byte field via `from_reader` now fails with `Error::PrimitiveConversionError`. There's also no replacement path: the macro-generated `*Owned` structs get `GetSize`/`Encodable` but **no `Decodable` impl**, so owned messages can't be decoded from a reader either. CI stays green only because nothing tests this path.

Repro (passes on `main`, fails on this branch with `PrimitiveConversionError`):

```rust
#[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
struct TestBorrowed<'decoder> {
    a: u8,
    b: U256<'decoder>,
}

let expected = TestBorrowed { a: 7, b: U256::from(&[9u8; 32]) };
let encoded = to_bytes(expected.clone()).unwrap();
let mut cursor = Cursor::new(encoded);
let decoded = <TestBorrowed as Decodable>::from_reader(&mut cursor); // Err(PrimitiveConversionError)
```

Suggested direction: implement `Decodable` for the generated Owned structs and make `from_reader` an owned-only API (documented as such), or explicitly remove/deprecate `from_reader` for borrowed types — rather than leaving it silently broken.

### 🔴 `Inner::from_reader_` deliberately leaks memory

The borrowed `Sv2DataType::from_reader_` (`inner.rs`, commit "Fix from_reader for Ref variant") is implemented as `Self::new(dst.leak())` — one permanently leaked heap allocation per call, and per *element* when reached through a borrowed sequence's `from_reader`. The new `b032_from_reader_should_build_borrowed_payload` test locks this in as intended behavior. SRI's internal codec path avoids it, but it's a public-API landmine for any long-lived consumer decoding from a `Read` source. A borrowed type can't be built from a reader without a backing buffer — this should be an error (or not exist on the Ref type), not a leak.

### 🟡 Medium

- **`SetupConnectionOwned` is missing inherent methods** (`common-messages/src/setup_connection.rs`): it gets `set_requires_standard_job`/`allow_full_template_mode` but not `get_version()`/`requires_standard_job()`, unlike other messages (e.g. `NewMiningJobOwned` replicates all its methods). Anyone storing an owned `SetupConnection` must re-derive version negotiation by hand.
- **Fragile type-name matching in the derive macro** (`derive_codec/src/lib.rs`): `owned_primitive_type` only recognizes bare identifiers (`U256`, `B032`, …). A field typed via a qualified path (`binary_sv2::U256<'decoder>`) or an alias silently falls through to the `'decoder → 'static` substitution and would generate a broken Owned struct. Worth a guard or at least a doc note now that this macro generates public API types.

### 🟢 Minor

- `parse_message_frame_with_tlvs` (`parsers-sv2/src/lib.rs`) changed from returning `AnyMessage<'static>` to a borrowed `AnyMessage<'a>` — a breaking contract change, but the doc comment wasn't updated to mention the borrow or that `.into_owned()` is the migration path.
- Too-short input to fixed-size owned types now yields `Error::ReadError(len, SIZE)` where it previously yielded `ValueExceedsMaxSize` — nothing in-repo matches on the variant, but external consumers might.
- Broken intra-doc links in channels-sv2 (`client/extended.rs`, `client/standard.rs`, `client/group.rs`, `server/jobs/extended.rs`) still reference the ref type names (`NewExtendedMiningJob`, `NewMiningJob`, `SetCustomMiningJob`) no longer in scope → `cargo doc` warnings / dead links.
- `sv1/src/lib.rs` (comment touched by this PR) claims a "catch-all panic arm" in `From<Error> for ParsingMethodError` — no such panic exists; the impl maps to `UnhandledSv1Error`.
- Test helper `scan_to_u256_sequence` (`mining/src/new_mining_job.rs`) now does `Box::leak` per 32-byte chunk per quickcheck iteration — `U256Owned` would be the honest fixture type.

### ⚙️ Performance notes (non-blocking)

- **Job factory** (`channels-sv2/src/server/jobs/factory.rs`): pre-existing `.clone()` sites on templates/custom jobs now deep-copy merkle paths and coinbase outputs 2–4× per job build (previously cheap ref clones). Runs per channel per new template on a pool server — worth an `Arc` or restructuring in a follow-up.
- **Share validation** now takes `SubmitShares*Owned` by value → one small allocation per share vs. the old zero-copy frame validation. Intended design and the validation bodies are unchanged; just noting the trade-off on the hottest path.

bit-aloo:
@GitGab19 I addressed all the suggestions above.

For now, I removed the `from_reader` and `writer` APIs that we would use for streaming. Since supporting that would require changes across framing, the codec, and the network helpers, I think we can tackle it later when we actually want streaming support. This should take care of the high-priority issues.

I also added the missing methods for `SetupConnectionOwned` (I somehow missed them in the initial iteration 😅).

All the doc breakages should be fixed now.

Regarding the last paragraph about performance, I think the model didn't see how we use the APIs with all the statics, so it ended up flagging it 😅.

plebhash:
@bit-aloo I merged https://github.com/stratum-mining/stratum/pull/2234 since the scope was very small, felt like a really low hanging fruit

the PR was bumping `channels_sv2` to `7.0.1`, which now introduced conflicts here

this PR should still bump it to `8.0.0`, but just a heads up on this small conflict here

---

edit: actually, https://github.com/stratum-mining/stratum/pull/2232 is bumping `channels_sv2` to 8.0.0`, so the conflict here is that you no longer need to bump that

plebhash (approved):
code changes LGTM

first 14 commits are ok, but then commit history started to stack up review pivots, which I generally prefer to avoid

but since PR scope is wide, I'm not going to frame it as a blocking point, and leave it up to @bit-aloo to decide whether to clean it up or not

bit-aloo:
> code changes LGTM
> 
> first 14 commits are ok, but then commit history started to stack up review pivots, which I generally prefer to avoid
> 
> but since PR scope is wide, I'm not going to frame it as a blocking point, and leave it up to @bit-aloo to decide whether to clean it up or not

I will scope them, now and wait for review by @GitGab19

GitGab19 (commented):
ACK

If you rebase this and the companion PR, we can proceed with merging both.

GitGab19 (approved):

bit-aloo:
> ACK
> 
> If you rebase this and the companion PR, we can proceed with merging both.

Awesome!!! sure!

PR #2227, opened 2026-07-19
Do not round up SV1 difficulties

ckolivas:
Rounding up difficulty to a power of 2 compared to the upstream pool guarantees hashrate loss if the pool provides non power of 2 difficulties.  If power of 2 is desired for some compatibility reason, you should round down instead of up to not lose hashrate. This will however lead to shares being rejected that are above the power of 2 difficulty but below the pool's difficulty.  Whether a filter should be applied for shares higher than the power of 2 difficulty but below the pool's difficulty, or just pass the pool's actual difficulty instead is for debate. I'd be inclined to simply pass the pool's difficulty as is instead. I have in addition added the necessary code to do the filtering in mining-apps but as rust is not my speciality it would require co-authorship from claude and I'm unsure of the current policy wrt AI code here. The AI second patch and the following pull request https://github.com/stratum-mining/sv2-apps/pull/630 are dependent on this one.

### DIFF
--- a/stratum-core/stratum-translation/src/sv2_to_sv1.rs
+++ b/stratum-core/stratum-translation/src/sv2_to_sv1.rs
@@ -133,8 +133,14 @@ pub fn build_sv1_set_difficulty_from_sv2_target(target: Target) -> Result<json_r
 /// Difficulties below `minimum_difficulty_for_integer_power_of_two_rounding` are emitted unchanged
 /// as JSON numbers built from `f64`, preserving fractional values.
 ///
-/// Difficulties at or above the threshold are rounded up to the next integer power of two and
-/// emitted as JSON numbers built from `u64`.
+/// Difficulties at or above the threshold are rounded down to the largest integer power of two
+/// not exceeding the difficulty and emitted as JSON numbers built from `u64` (difficulties between
+/// the threshold and 1 round up to 1).
+///
+/// Rounding must never round up: the upstream credits shares against its own (unrounded) target,
+/// so a miner told a higher difficulty than upstream silently discards every share between the
+/// upstream target and the next power of two - up to half its hashrate. Rounding down merely
+/// makes the miner submit some shares below the upstream target, which the proxy filters.
 pub fn build_sv1_set_difficulty_from_sv2_target_with_integer_power_of_two_rounding(
     target: Target,
     minimum_difficulty_for_integer_power_of_two_rounding: f64,
@@ -170,14 +176,17 @@ fn integer_power_of_two_sv1_difficulty_value_from_difficulty(
         return Ok(Value::Number(value));
     }
 
-    let integer_difficulty = difficulty.ceil();
-    if integer_difficulty > u64::MAX as f64 {
+    let integer_difficulty = difficulty.floor();
+    if integer_difficulty >= u64::MAX as f64 {
         return Err(StratumTranslationError::Sv1DifficultyOverflow(difficulty));
     }
 
-    let power_of_two = (integer_difficulty as u64)
-        .checked_next_power_of_two()
-        .ok_or(StratumTranslationError::Sv1DifficultyOverflow(difficulty))?;
+    let integer_difficulty = integer_difficulty as u64;
+    let power_of_two = if integer_difficulty == 0 {
+        1
+    } else {
+        1u64 << (63 - integer_difficulty.leading_zeros())
+    };
 
     Ok(Value::from(power_of_two))
 }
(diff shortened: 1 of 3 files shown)


stratum-core/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -133,8 +133,14 @@ pub fn build_sv1_set_difficulty_from_sv2_target(target: Target) -> Result<json_r
 /// Difficulties below `minimum_difficulty_for_integer_power_of_two_rounding` are emitted unchanged
 /// as JSON numbers built from `f64`, preserving fractional values.
 ///
-/// Difficulties at or above the threshold are rounded up to the next integer power of two and
-/// emitted as JSON numbers built from `u64`.
+/// Difficulties at or above the threshold are rounded down to the largest integer power of two
+/// not exceeding the difficulty and emitted as JSON numbers built from `u64` (difficulties between
+/// the threshold and 1 round up to 1).
```
GitGab19: This still allows rounding up when the difficulty is below 1 and the threshold is below 1. We should require threshold to always be >= 1 IMO, and remove any rounding up there.

stratum-core/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -165,21 +177,53 @@ fn integer_power_of_two_sv1_difficulty_value_from_difficulty(
...
+    Ok(power_of_two as f64)
+}
+
+/// Returns the SV1 difficulty value that
+/// [`build_sv1_set_difficulty_from_sv2_target_with_integer_power_of_two_rounding`] advertises
+/// for `target`, so callers can reconstruct the exact downstream difficulty in force.
+pub fn sv1_advertised_difficulty_from_sv2_target(
+    target: Target,
+    minimum_difficulty_for_integer_power_of_two_rounding: f64,
+) -> Result<f64> {
```
GitGab19: Could this expose the advertised `Target` instead, or in addition? Callers that validate Sv1 shares need the target corresponding to the rounded advertised difficulty, not just the f64 difficulty number (at least our tProxy works in that way).

By checking your companion PR (https://github.com/stratum-mining/sv2-apps/pull/630), I saw you had to create the helper which gives you the `Target` there, but IMO it should live here.

Or you think there's a strong need for returning the `Result<f64>` here instead?

stratum-core/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -146,10 +152,16 @@ pub fn build_sv1_set_difficulty_from_sv2_target_with_integer_power_of_two_roundi
...
+/// Proxies use this to validate downstream shares against exactly the difficulty the miner was
+/// given, and separately filter which of those shares also meet the upstream target.
+fn advertised_sv1_difficulty_from_difficulty(
     difficulty: f64,
     minimum_difficulty_for_integer_power_of_two_rounding: f64,
-) -> Result<Value> {
+) -> Result<f64> {
     if !difficulty.is_finite() || difficulty <= 0.0 {
         return Err(StratumTranslationError::InvalidSv1Difficulty(difficulty));
     }
```
GitGab19, line 163: Should this reject thresholds below 1.0 instead of only <= 0.0? If the threshold is below 1.0, a difficulty like 0.25 can enter the rounding branch and become 1, which is a round-up and conflicts with the correct “must never round up” invariant you're suggesting.

GitGab19 (commented):
ACK on changing the rounding behavior, but I see a couple of improvements to be made before proceeding with this PR.

stratum-core/stratum-translation/src/sv2_to_sv1.rs
ckolivas: Stratum v1 doesn't usually support diffs below 1.

stratum-core/stratum-translation/src/sv2_to_sv1.rs
ckolivas: As per my previous comment, stratum v1 normally doesn't support diffs below 1.

stratum-core/stratum-translation/src/sv2_to_sv1.rs
GitGab19: Having diffs below 1 is useful for integration tests, and manual tests with a cpu-miner, that's why the current code allows for that.

stratum-core/stratum-translation/src/sv2_to_sv1.rs
ckolivas: No strong need at all.  I wasn't even sure of why it was being rounded to a power of 2 in the first place.

ckolivas:
If there is no particular reason for rounding to a power of 2, then the target should just be exposed as is. Should I close this and the companion PR and just offer that as a simple PR?

GitGab19:
The reason behind the rounding is described under this past issue: https://github.com/stratum-mining/sv2-apps/issues/561

From my side your PR here is good enough, as it rounds down (and not up) diffs, which is the major problem of the current implementation.

It also provides a way for tProxy to validate shares with the advertised diffs, which is also the correct thing to do.

I would simply modify the two things I pointed out above, but if you prefer I can push a commit on top of yours directly on this PR, or give you a commit to cherry-pick.

ckolivas:
Easiest and fastest would be for you to push a commit on top I suspect. Thanks for review.

GitGab19:
@ckolivas I just pushed two commits on top of yours.

Do you want to update the PR on `sv2-apps` (https://github.com/stratum-mining/sv2-apps/pull/630) to these changes, or you prefer to have me doing that?

ckolivas:
Please just wrap this up :)

GitGab19 (approved):

PR #2239, opened 2026-08-01
`channels_sv2`: harden arithmetic against overflow, underflow and division by zero

### LINKED ISSUE #2238: `channels_sv2`: need to use saturating arithmetic for all `ShareAccounting` counters (`client` + `server`)
All `u32` counters in the two `ShareAccounting` types use non-saturating arithmetic. 

Past `u32::MAX`, the next increment panics under overflow checks (debug, `cargo test`, `overflow-checks=true` releases) or silently wraps to 0 in default release builds.

## server

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/share_accounting.rs#L119-L125

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/share_accounting.rs#L133-L152

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/share_accounting.rs#L201-L203

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/share_accounting.rs#L241-L243

## client

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/client/share_accounting.rs#L116-L123

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/client/share_accounting.rs#L129-L134

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/client/share_accounting.rs#L142-L152

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/client/share_accounting.rs#L183-L185

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/client/share_accounting.rs#L225-L227

---

## impact

- overflow-checked builds: panic aborts the channel's validation/accounting path (DoS)
- default release: counters wrap to 0, silently corrupting statistics any consumer relies on for throttling/banning/reconciliation

Per-share counters need ~4.3B submissions, so practical severity is low; the upstream-driven paths (`on_share_acknowledgement`, `on_share_rejection`) are cheap for a malicious/compromised pool to exercise repeatedly.

## suggested fix

Mechanical: `saturating_add` for all increments, `fold(0, u32::saturating_add)` for the two `sum()` aggregations. Regression tests can seed private counters at `u32::MAX` from the in-file test modules.

plebhash:
close #2238 
close https://github.com/project-loupe/audit-stratum/issues/54
close https://github.com/project-loupe/audit-stratum/issues/12
close https://github.com/project-loupe/audit-stratum/issues/75
close https://github.com/project-loupe/audit-stratum/issues/38
close https://github.com/project-loupe/audit-stratum/issues/74
close https://github.com/project-loupe/audit-stratum/issues/36
close https://github.com/project-loupe/audit-stratum/issues/16
close https://github.com/project-loupe/audit-stratum/issues/40

### DIFF
--- a/sv2/channels-sv2/src/vardiff/classic.rs
+++ b/sv2/channels-sv2/src/vardiff/classic.rs
@@ -32,11 +32,20 @@ impl VardiffState {
     /// Creates a new `VardiffState` with a specific minimum hashrate.
     ///
     /// # Arguments
-    /// * `min_allowed_hashrate` - The minimum hashrate to enforce.
+    /// * `min_allowed_hashrate` - The minimum hashrate to enforce. A non-positive or non-finite
+    ///   value is meaningless as a floor (and would reintroduce the division-by-zero that
+    ///   [`Vardiff::try_vardiff`] guards against), so it falls back to the default.
     pub fn new_with_min(min_allowed_hashrate: f32) -> Result<Self, VardiffError> {
         let timestamp_secs = std::time::SystemTime::now()
             .duration_since(std::time::UNIX_EPOCH)?
             .as_secs();
+
+        let min_allowed_hashrate = if min_allowed_hashrate.is_finite() && min_allowed_hashrate > 0.0
+        {
+            min_allowed_hashrate
+        } else {
+            DEFAULT_MIN_HASHRATE
+        };
 
         Ok(VardiffState {
             shares_since_last_update: 0,
@@ -128,6 +137,35 @@ impl Vardiff for VardiffState {
             return Ok(None);
         }
 
+        // `min_allowed_hashrate` is validated in `new_with_min`, but the field is `pub`,
+        // so direct assignment can still plant a non-finite or non-positive floor;
+        // sanitize it here as well before relying on it.
+        let min_hashrate =
+            if self.min_allowed_hashrate.is_finite() && self.min_allowed_hashrate > 0.0 {
+                self.min_allowed_hashrate
+            } else {
+                DEFAULT_MIN_HASHRATE
+            };
+
+        // `hashrate` is the channel's nominal hashrate, which originates from the
+        // miner-supplied `nominal_hash_rate` and is only checked for negativity upstream.
+        // It is the divisor for the delta percentage below and the base every special-case
+        // cap scales (`hashrate * 10.0`, `hashrate / 1.5`), so a value at or below the floor
+        // just pins the result back to that floor instead of converging, and `0.0`/`NaN`
+        // produce `inf`/`NaN` percentages that disable every `should_update` arm outright.
+        // The output is already clamped to the floor below, so clamp the baseline to the
+        // same floor. `is_finite` is still needed: `inf` survives a plain `max()` and makes
+        // the percentage `inf / inf == NaN`.
+        let hashrate = if hashrate.is_finite() {
+            hashrate.max(min_hashrate)
+        } else {
+            debug!(
+                target: "vardiff",
+                "Prior hashrate {hashrate} unusable; using minimum {min_hashrate}",
+            );
+            min_hashrate
+        };
+
         let realized_share_per_min =
             self.shares_since_last_update as f64 / (delta_time as f64 / 60.0);
 
@@ -198,14 +236,14 @@ impl Vardiff for VardiffState {
                 _ => hashrate * 3.0,
             };
         }
-        if new_hashrate < self.min_allowed_hashrate {
+        if new_hashrate < min_hashrate {
             debug!(
                 target: "vardiff",
                 "New hashrate {:.2} H/s below minimum threshold {:.2} H/s — clamping",
                 new_hashrate,
-                self.min_allowed_hashrate
+                min_hashrate
             );
-            new_hashrate = self.min_allowed_hashrate;
+            new_hashrate = min_hashrate;
         }
         self.reset_counter()?;
 
--- a/sv2/channels-sv2/src/vardiff/classic.rs
+++ b/sv2/channels-sv2/src/vardiff/classic.rs
@@ -104,7 +104,25 @@ impl Vardiff for VardiffState {
             .map_err(VardiffError::TimeError)?
             .as_secs();
 
-        let delta_time = now - self.timestamp_of_last_update;
+        let delta_time = match now.checked_sub(self.timestamp_of_last_update) {
+            Some(delta_time) => delta_time,
+            None => {
+                // The clock stepped backwards (e.g. NTP correction), leaving the recorded
+                // timestamp in the future, so elapsed time is unmeasurable. Re-anchor the
+                // window to `now` and skip just this round: the `delta_time <= 15` guard
+                // below returns before `reset_counter()` runs, so without re-anchoring here
+                // every later round would measure against the same stale future timestamp
+                // and vardiff would stall for the whole length of the backwards step.
+                debug!(
+                    target: "vardiff",
+                    "Clock stepped backwards (recorded {}, now {}); re-anchoring vardiff window",
+                    self.timestamp_of_last_update,
+                    now
+                );
+                self.reset_counter()?;
+                return Ok(None);
+            }
+        };
 
         if delta_time <= 15 {
             return Ok(None);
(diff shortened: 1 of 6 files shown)


bit-aloo (commented):
Can you extend saturating arithmetic to places where we use vanilla arithmetic. I see there are few.

plebhash:
> Can you extend saturating arithmetic to places where we use vanilla arithmetic. I see there are few.

added 8a22ded1249e529e3305da01d5482f4c51989fb3

focused on three sites that can actually panic:

- `hash_rate_from_target`: `target - 1` underflows when the target is zero. Reachable from vardiff with an upstream-supplied target, so it now returns `InputError::ArithmeticOverflow`. Saturating would be wrong here: the true numerator for `t = 0` is `2^256`, which isn't representable in a `U256`.
- `hash_rate_to_target`: `h_times_s + 1` overflows at `u128::MAX` (the `f64 as u128` cast saturates there). The existing `max(x, x + 1)` was already reaching for saturation; now `saturating_add(1)`.
- `try_vardiff`:  `now - timestamp_of_last_update` underflows on a backwards clock step (NTP). Now `saturating_sub`, falling through the existing `delta_time <= 15` guard.


also added 4133371d3061d65b6710405e3d4442cc392346cb

guarding `try_vardiff` against a zero or NaN prior hashrate — it's the divisor for the delta percentage *and* the base every special-case cap scales, so a miner declaring `nominal_hash_rate = 0.0` could drive difficulty to the floor under a share flood.

---

the rest I left alone:

- `share_work_sum`, `validated_work_sum`, `last_batch_work_sum` are `f64` — floats don't panic on overflow, they go to inf. Nothing to guard.
- `JobIdFactory::next`: `saturating_add` would be a bug here, it'd hand out duplicate job IDs forever once saturated. Wrapping is arguably right but that's a protocol question about job-ID reuse, not an overflow fix; I'd rather not decide it in this PR.
- `shares_since_last_update += 1` : `u32`, but reset on every vardiff adjustment, so not reachable.
- `local_index_to_bytes`: both subtractions already guarded by `copy_len = be_bytes.len().min(len)`.
- `bytes_needed`: the `n - 1` is guarded by the `if n <= 1` early return.
- The `a.len() + b.len()` usize adds — not reachable on 64-bit.

sv2/channels-sv2/src/vardiff/classic.rs
```diff
@@ -104,12 +113,32 @@ impl Vardiff for VardiffState {
             .map_err(VardiffError::TimeError)?
             .as_secs();
 
-        let delta_time = now - self.timestamp_of_last_update;
+        // A backwards clock step (e.g. NTP correction) can put `now` behind the
+        // recorded timestamp; saturate to 0 so we skip this round instead of panicking.
+        let delta_time = now.saturating_sub(self.timestamp_of_last_update);
```
GitGab19: Aren't we gonna skip the updates always, in case of a backwards clock?

sv2/channels-sv2/src/vardiff/classic.rs
```diff
@@ -104,12 +113,32 @@ impl Vardiff for VardiffState {
...
         }
 
+        // `hashrate` is the channel's nominal hashrate, which originates from the
+        // miner-supplied `nominal_hash_rate` and is only checked for negativity upstream.
+        // A zero or non-finite value is division-by-zero fuel: it makes the delta
+        // percentage below `inf`/`NaN`, and it is also the base every special-case cap
+        // multiplies (`hashrate * 10.0`) or divides (`hashrate / 1.5`), so a share flood
+        // would drive difficulty to the floor (or, for `NaN`, silently never adjust).
+        // Fall back to the configured floor so there is a usable baseline.
+        let hashrate = if hashrate.is_finite() && hashrate > 0.0 {
```
GitGab19: Even if hashrate is 0.01, we end up in the same problem.

So we should probably take the max between the `hashrate` and the `min_allowed_hashrate` here?

GitGab19: Something like this:
```
let min_hashrate =
    if self.min_allowed_hashrate.is_finite() && self.min_allowed_hashrate > 0.0 {
        self.min_allowed_hashrate
    } else {
        DEFAULT_MIN_HASHRATE
    };

let hashrate = if hashrate.is_finite() {
    hashrate.max(min_hashrate)
} else {
    min_hashrate
};
```

sv2/channels-sv2/src/vardiff/classic.rs
plebhash: Good catch — yes. As written it skips for the whole length of the backwards step, not just one round.

The `delta_time <= 15` early return fires before `reset_counter()`, which is the only thing inside `try_vardiff` that refreshes `timestamp_of_last_update`. So once the recorded timestamp is in the future, every later call recomputes `delta_time = 0` and returns `Ok(None)` — vardiff is off until wall-clock time passes `timestamp_of_last_update + 15`. A normal NTP correction is milliseconds and harmless, but a big step (bad RTC, VM snapshot restore) disables difficulty adjustment for however long the step was, which is the same bypass window as the zero-hashrate finding.

Now handled with `checked_sub`: on an inverted clock it re-anchors via `reset_counter()` and skips a single round, so the next round measures against a sane baseline. The regression test now asserts the re-anchor too — the old one only checked that we didn't panic, which is how this slipped past me.

sv2/channels-sv2/src/vardiff/classic.rs
plebhash: Right, `0.01` is broken the same way — I traced it: `new_hashrate` comes out ~375000, the delta percentage is ~3.75e9, so the `> 1000` branch caps it to `0.01 * 10 = 0.1` and the output clamp pins that straight back to the floor. Same first round as `0.0`; the only difference is that `0.0` is absorbing under `* 10.0` while `0.01` crawls back at 10x per vardiff window.

Taking the max against `min_allowed_hashrate` now, as you suggested — and it's the symmetric thing to do, since the output is already clamped to that same floor a few lines down. Kept the `is_finite()` check on top of it though: `inf.max(floor)` is still `inf`, and an infinite baseline makes the percentage `inf / inf == NaN`, which makes every `should_update` arm false and silently disables vardiff.

Folded the zero/NaN tests into one table-driven case over `[0.0, 0.01, -5.0, NaN, inf]`. Only the `0.01` row failed before this change, which is exactly the gap you spotted.

sv2/channels-sv2/src/vardiff/classic.rs
GitGab19: Yeah that should properly align the timestamp again, nice!

sv2/channels-sv2/src/vardiff/classic.rs
GitGab19: That's better, but we're still not checking if `self.min_allowed_hashrate` is finite and greater than 0.0 there.

So that check is still missing.

plebhash: 🫩 fucking retarded Opus, sorry about that

throwing Fable at it, hopefully last round:

> Good point — `new_with_min` validates it, but `min_allowed_hashrate` is a `pub` field, so that validation can be bypassed by direct assignment.
> 
> Added the check in `try_vardiff` following your suggested shape: the floor is sanitized to `DEFAULT_MIN_HASHRATE` when non-finite or non-positive, and the prior hashrate falls back via `hashrate.max(min_hashrate)`. The final output clamp now uses the sanitized floor too, since an `inf` floor would otherwise clamp the result to `inf`.
> 
> Also added a regression test (`test_try_vardiff_sanitizes_unusable_floor`) that plants `0.0`/`-5.0`/`NaN`/`inf` directly into the field and asserts `try_vardiff` still produces a finite, positive hashrate.

GitGab19 (approved):
ACK

PR #2243, opened 2026-08-02
`channels_sv2`: fix coinbase `scriptSig` size/serialization defects that yield consensus-invalid coinbases

### LINKED ISSUE #2242: `channels_sv2`: need to fix coinbase tx `scriptSig` size/serialization defects that yield consensus-invalid coinbase txs
## summary

The following defects in server-side coinbase tx construction can produce a coinbase that fails Bitcoin consensus (100-byte `scriptSig` cap, or serialization that doesn't round-trip).

General impact: a block found on an affected channel is rejected by Bitcoin Nodes — the reward is silently lost — and in the group-channel case the pool can distribute unmineable work.

## 1. `GroupChannel::on_new_template` never re-checks the `scriptSig` budget

`GroupChannel::new` enforces the 100-byte `scriptSig` limit with a hardcoded `5 // BIP34` allowance, but `on_new_template` accepts the actual `NewTemplate::coinbase_prefix` (up to 8 bytes per the in-tree docs) without re-validation.

A config at the edge of the budget (e.g. 58-byte pool tag + 32-byte extranonce = exactly 100) silently becomes 103 bytes with an 8-byte prefix.

Fix: compute the real assembled size in `on_new_template` and return `ScriptSigSizeTooLarge`.

## 2. `ExtendedChannel::set_extranonce_prefix` drops the constructor's size invariant

The constructor checks `5 + 1 + 3 + pool_tag + miner_tag + 1 + prefix + rollable ≤ 100`

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/extended.rs#L230-L232

the setter only re-checks `MAX_EXTRANONCE_LEN`:

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/extended.rs#L291-L302

Swapping in a larger (individually valid) prefix pushes the total past 100. 

Note: the same issue also happens on `StandardChannel::set_extranonce_prefix` and `GroupChannel::set_full_extranonce_size`

Fix: re-run the full invariant in the setter.

## 3. `BlockFound` coinbase uses the channel's current prefix length, not the job's

`script_sig.push(self.extranonce_prefix.len() as u8)` followed by `script_sig.extend(job.get_extranonce_prefix())` — after a `set_extranonce_prefix` rotation to a different length, the `OP_PUSHBYTES` opcode disagrees with the byte count, so the reconstructed coinbase's txid diverges from the job's `merkle_root`. 

https://github.com/stratum-mining/stratum/blob/408a6fd6aea3893fa4851d4632bcd756ff8acac9/sv2/channels-sv2/src/server/standard.rs#L681-L684

Fix: `script_sig.push(job.get_extranonce_prefix().len() as u8)`.

## regression tests

Each item has a boundary-case test: 

assemble a coinbase at the relevant size boundary, assert it round-trips through `bitcoin::consensus::deserialize` and (where applicable) that its `txid` equals the job's `merkle_root`.

plebhash:
close #2242
close https://github.com/project-loupe/audit-stratum/issues/20
close https://github.com/project-loupe/audit-stratum/issues/30
close https://github.com/project-loupe/audit-stratum/issues/79
close https://github.com/project-loupe/audit-stratum/issues/88

companion https://github.com/stratum-mining/sv2-apps/pull/658

### DIFF
--- a/sv2/channels-sv2/src/server/standard.rs
+++ b/sv2/channels-sv2/src/server/standard.rs
@@ -715,7 +715,10 @@ impl StandardChannel {
 
             let mut script_sig = job.get_template().coinbase_prefix.to_owned_bytes();
             script_sig.extend(op_pushbytes_pool_miner_tag);
-            script_sig.push(self.extranonce_prefix.len() as u8); // OP_PUSHBYTES_X (for the extranonce)
+            // the opcode must describe the job's extranonce prefix, not the channel's current one:
+            // `set_extranonce_prefix` may have rotated to a different length since this job was
+            // created, and the job's `merkle_root` is committed to the job's prefix
+            script_sig.push(job.get_extranonce_prefix().len() as u8); // OP_PUSHBYTES_X (for the extranonce)
             script_sig.extend(job.get_extranonce_prefix());
 
             let tx_in = TxIn {
@@ -791,7 +794,9 @@ mod tests {
         },
     };
     use binary_sv2::Sv2OptionOwned as Sv2Option;
-    use bitcoin::{transaction::TxOut, Amount, ScriptBuf, Target};
+    use bitcoin::{
+        consensus::deserialize, transaction::TxOut, Amount, ScriptBuf, Target, Transaction,
+    };
     use mining_sv2::{
         NewMiningJobOwned as NewMiningJob, SubmitSharesStandardOwned,
         ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW,
@@ -1242,6 +1247,143 @@ mod tests {
             standard_channel.get_share_accounting().get_blocks_found(),
             0
         );
+    }
+
+    #[test]
+    fn test_share_validation_block_found_after_extranonce_prefix_rotation() {
+        // note:
+        // same test vectors as `test_share_validation_block_found`, plus an extranonce prefix
+        // rotation (to a prefix of a different length) in between job creation and share
+        // submission.
+        //
+        // the job (and therefore its merkle root, and therefore the winning nonce) is committed to
+        // the prefix that was current at `on_new_template` time, so the coinbase reconstructed on
+        // BlockFound must also be committed to that same prefix.
+
+        let standard_channel_id = 1;
+        let user_identity = "user_identity".to_string();
+
+        let extranonce_prefix = [
+            83, 116, 114, 97, 116, 117, 109, 32, 86, 50, 32, 83, 82, 73, 32, 80, 111, 111, 108, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+        ]
+        .to_vec();
+        let max_target = Target::from_le_bytes([0xff; 32]);
+        let nominal_hashrate = 1.0;
+        let share_batch_size = 100;
+        let expected_share_per_minute = 1.0;
+        let mut standard_channel = StandardChannel::new(
+            standard_channel_id,
+            user_identity,
+            ExtranoncePrefix::from_wire(extranonce_prefix.clone()).unwrap(),
+            max_target,
+            nominal_hashrate,
+            share_batch_size,
+            expected_share_per_minute,
+            None,
+            None,
+        )
+        .unwrap();
+
+        // channel target: 04325c53ef368eb04325c53ef368eb04325c53ef368eb04325c53ef368eb0431
+        let template = NewTemplate {
+            template_id: 1,
+            future_template: false,
+            version: 536870912,
+            coinbase_tx_version: 2,
+            coinbase_prefix: vec![2, 159, 0, 0].try_into().unwrap(),
+            coinbase_tx_input_sequence: 4294967294,
+            coinbase_tx_value_remaining: SATS_AVAILABLE_IN_TEMPLATE,
+            coinbase_tx_outputs_count: 1,
+            coinbase_tx_outputs: vec![
+                0, 0, 0, 0, 0, 0, 0, 0, 38, 106, 36, 170, 33, 169, 237, 226, 246, 28, 63, 113, 209,
+                222, 253, 63, 169, 153, 223, 163, 105, 83, 117, 92, 105, 6, 137, 121, 153, 98, 180,
+                139, 235, 216, 54, 151, 78, 140, 249,
+            ]
+            .try_into()
+            .unwrap(),
+            coinbase_tx_locktime: 158,
+            merkle_path: vec![].try_into().unwrap(),
+        };
+
+        // match the original script format used to generate the coinbase_reward_outputs for the
+        // expected job
+        let pubkey_hash = [
+            235, 225, 183, 220, 194, 147, 204, 170, 14, 231, 67, 168, 111, 137, 223, 130, 88, 194,
+            8, 252,
+        ];
+        let mut script_bytes = vec![0]; // SegWit version 0
+        script_bytes.push(20); // Push 20 bytes (length of pubkey hash)
+        script_bytes.extend_from_slice(&pubkey_hash);
+        let script = ScriptBuf::from(script_bytes);
+        let coinbase_reward_outputs = vec![TxOut {
+            value: Amount::from_sat(SATS_AVAILABLE_IN_TEMPLATE),
+            script_pubkey: script,
+        }];
+
+        // network target: 7fffff0000000000000000000000000000000000000000000000000000000000
+        let ntime = 1745596910;
+        let prev_hash = [
+            251, 175, 106, 40, 35, 87, 122, 90, 58, 51, 78, 32, 202, 236, 228, 36, 154, 174, 206,
+            144, 147, 195, 21, 224, 195, 103, 214, 189, 51, 190, 24, 98,
+        ]
+        .into();
+        let n_bits = 545259519;
+        let chain_tip = ChainTip::new(prev_hash, n_bits, ntime);
+
+        // prepare standard channel with non-future job
+        standard_channel.set_chain_tip(chain_tip);
+        standard_channel
+            .on_new_template(template.clone(), coinbase_reward_outputs)
+            .unwrap();
+
+        // capture what the job is committed to, before rotating the channel's prefix
+        let active_standard_job = standard_channel.get_active_job().unwrap();
+        let job_id = active_standard_job.get_job_id();
+        let job_merkle_root = active_standard_job.get_merkle_root().to_array();
+        let job_extranonce_prefix = active_standard_job.get_extranonce_prefix().to_vec();
+        assert_eq!(job_extranonce_prefix, extranonce_prefix);
+
+        // rotate the channel to a prefix of a different length
+        let rotated_extranonce_prefix = vec![0xab; 16];
+        standard_channel
+            .set_extranonce_prefix(
+                AllocatedExtranoncePrefix::for_test(rotated_extranonce_prefix).unwrap(),
+            )
+            .unwrap();
+
+        // this share has hash 3c34f63de61283c907b68e3127146d7d11f1fb14e50020a8317a292d11e2dab6
+        // which satisfied the network target
+        // 7fffff0000000000000000000000000000000000000000000000000000000000
+        let share_valid_block = SubmitSharesStandardOwned {
+            channel_id: standard_channel_id,
+            sequence_number: 0,
+            job_id,
+            nonce: 0,
+            ntime: 1745596932,
+            version: 536870912,
+        };
+
+        let Ok(ShareValidationResult::BlockFound(_, _, serialized_coinbase)) =
+            standard_channel.validate_share(share_valid_block)
+        else {
+            panic!("expected BlockFound");
+        };
+
+        let coinbase: Transaction = deserialize(&serialized_coinbase).unwrap();
+
+        // merkle_path is empty in this template, so the merkle root IS the coinbase txid
+        let coinbase_txid: [u8; 32] = *coinbase.compute_txid().as_ref();
+        assert_eq!(coinbase_txid, job_merkle_root);
+
+        // the scriptSig must push the job's extranonce prefix, with a matching OP_PUSHBYTES opcode
+        let script_sig = coinbase.input[0].script_sig.as_bytes();
+        let extranonce_start = script_sig.len() - job_extranonce_prefix.len();
+        assert_eq!(
+            script_sig[extranonce_start - 1],
+            job_extranonce_prefix.len() as u8
+        );
+        assert_eq!(&script_sig[extranonce_start..], &job_extranonce_prefix[..]);
     }
 
     #[test]
--- a/sv2/channels-sv2/src/server/standard.rs
+++ b/sv2/channels-sv2/src/server/standard.rs
@@ -113,6 +113,10 @@ impl StandardChannel {
     ///
     /// For non-JD jobs, `pool_tag_string` is added to the coinbase scriptSig as
     /// `Sv2/pool_tag_string//`.
+    ///
+    /// Returns [`StandardChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
+    /// extranonce prefix and a worst-case coinbase prefix do not fit within the coinbase
+    /// `scriptSig` budget, see [`JobFactory::fits_script_sig_budget`].
     #[allow(clippy::too_many_arguments)]
     pub fn new_for_pool(
         channel_id: u32,
@@ -147,6 +151,10 @@ impl StandardChannel {
     ///
     /// The `pool_tag_string` and `miner_tag_string` are added to the coinbase scriptSig as
     /// `Sv2/pool_tag_string/miner_tag_string/`.
+    ///
+    /// Returns [`StandardChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
+    /// extranonce prefix and a worst-case coinbase prefix do not fit within the coinbase
+    /// `scriptSig` budget, see [`JobFactory::fits_script_sig_budget`].
     #[allow(clippy::too_many_arguments)]
     pub fn new_for_job_declaration_client(
         channel_id: u32,
@@ -204,16 +212,12 @@ impl StandardChannel {
             return Err(StandardChannelError::ExtranoncePrefixTooLarge);
         }
 
-        let script_sig_size = 5 + // BIP34
-            1 + // OP_PUSHBYTES
-            3 + // "Sv2"
-            3 + // `/` delimiters
-            pool_tag_string.as_ref().map_or(0, |s| s.len()) +
-            miner_tag_string.as_ref().map_or(0, |s| s.len()) +
-            1 + // OP_PUSHBYTES
-            extranonce_prefix.len();
+        let job_factory = JobFactory::new(true, pool_tag_string, miner_tag_string);
 
-        if script_sig_size > 100 {
+        // conservative check against the spec's worst-case `NewTemplate::coinbase_prefix`.
+        // the exact size is re-checked against each actual template in `JobFactory::coinbase`.
+        // standard channels have no rollable extranonce, so the prefix is the full extranonce
+        if !job_factory.fits_script_sig_budget(extranonce_prefix.len()) {
             return Err(StandardChannelError::ScriptSigSizeTooLarge);
         }
 
@@ -229,7 +233,7 @@ impl StandardChannel {
             share_accounting: ShareAccounting::new(share_batch_size),
             expected_share_per_minute,
             job_store: JobStore::new(),
-            job_factory: JobFactory::new(true, pool_tag_string, miner_tag_string),
+            job_factory,
             chain_tip: None,
         })
     }
@@ -441,6 +445,13 @@ impl StandardChannel {
     /// Only meant to be used in case we want to broadcast standard jobs.
     /// In case we want to broadcast extended jobs via group channel, use `on_group_channel_job`
     /// instead.
+    ///
+    /// Returns [`StandardChannelError::JobFactoryError`] wrapping
+    /// [`JobFactoryError::ScriptSigSizeTooLarge`](crate::server::jobs::error::JobFactoryError::ScriptSigSizeTooLarge)
+    /// if the template's `coinbase_prefix` pushes the assembled coinbase `scriptSig` past
+    /// its budget. The constructor can only check against the spec's worst-case prefix (see
+    /// [`JobFactory::fits_script_sig_budget`]), so this is where an out-of-spec Template Provider
+    /// is caught.
     pub fn on_new_template(
         &mut self,
         template: NewTemplateOwned,
@@ -789,6 +800,7 @@ mod tests {
         },
         server::{
             error::StandardChannelError,
+            jobs::factory::{MAX_COINBASE_PREFIX_SIZE, MAX_SCRIPT_SIG_SIZE},
             share_accounting::{ShareValidationError, ShareValidationResult},
             standard::StandardChannel,
         },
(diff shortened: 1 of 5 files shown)


sv2/channels-sv2/src/server/standard.rs
```diff
@@ -203,15 +214,14 @@ impl<'a> StandardChannel<'a> {
...
-            1 + // OP_PUSHBYTES
-            extranonce_prefix.len();
+        let job_factory = JobFactory::new(true, pool_tag_string, miner_tag_string);
 
-        if script_sig_size > 100 {
+        // conservative check against the spec's worst-case `NewTemplate::coinbase_prefix`.
+        // the exact size is re-checked against each actual template in `JobFactory::coinbase`.
+        // standard channels have no rollable extranonce, so the prefix is the full extranonce
+        if job_factory.script_sig_size(MAX_COINBASE_PREFIX_SIZE, extranonce_prefix.len())
+            > MAX_SCRIPT_SIG_SIZE
```
bit-aloo: This check we can move inside the job_factory and call with extranonce_prefix size? with methods like these:

```rust
impl JobFactory {
    pub fn check_script_sig_budget(&self, coinbase_prefix_size: usize, full_extranonce_size: usize)
        -> Result<(), JobFactoryError>;

    pub fn check_worst_case_script_sig_budget(&self, full_extranonce_size: usize)
        -> Result<(), JobFactoryError>;
}
```

bit-aloo (commented):
One suggestion, and rest looks ok

sv2/channels-sv2/src/server/standard.rs
plebhash: thanks for the suggestion, incorporating in a slightly different (more concise) way

adding `pub fn fits_script_sig_budget(&self, full_extranonce_size: usize) -> bool` to `JobFactory`

the following diff is being rebased into commit history:

```diff
diff --git a/sv2/channels-sv2/src/server/extended.rs b/sv2/channels-sv2/src/server/extended.rs
index 75ed7f19..972fbe10 100644
--- a/sv2/channels-sv2/src/server/extended.rs
+++ b/sv2/channels-sv2/src/server/extended.rs
@@ -45,12 +45,7 @@ use crate::{
     merkle_root::merkle_root_from_path,
     server::{
         error::ExtendedChannelError,
-        jobs::{
-            extended::ExtendedJob,
-            factory::{JobFactory, MAX_COINBASE_PREFIX_SIZE, MAX_SCRIPT_SIG_SIZE},
-            job_store::JobStore,
-            JobOrigin,
-        },
+        jobs::{extended::ExtendedJob, factory::JobFactory, job_store::JobStore, JobOrigin},
         share_accounting::{ShareAccounting, ShareValidationError, ShareValidationResult},
     },
     target::{bytes_to_hex, hash_rate_to_target, u256_to_block_hash},
@@ -123,8 +118,8 @@ impl ExtendedChannel {
     /// and `//` delimiters: `/pool_tag_string//`
     ///
     /// Returns [`ExtendedChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
-    /// full extranonce and a worst-case [`MAX_COINBASE_PREFIX_SIZE`] coinbase prefix do not fit
-    /// within [`MAX_SCRIPT_SIG_SIZE`].
+    /// full extranonce and a worst-case coinbase prefix do not fit within the coinbase `scriptSig`
+    /// budget, see [`JobFactory::fits_script_sig_budget`].
     #[allow(clippy::too_many_arguments)]
     pub fn new_for_pool(
         channel_id: u32,
@@ -165,8 +160,8 @@ impl ExtendedChannel {
     /// `/` delimiters: `/pool_tag_string/miner_tag_string/`
     ///
     /// Returns [`ExtendedChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
-    /// full extranonce and a worst-case [`MAX_COINBASE_PREFIX_SIZE`] coinbase prefix do not fit
-    /// within [`MAX_SCRIPT_SIG_SIZE`].
+    /// full extranonce and a worst-case coinbase prefix do not fit within the coinbase `scriptSig`
+    /// budget, see [`JobFactory::fits_script_sig_budget`].
     #[allow(clippy::too_many_arguments)]
     pub fn new_for_job_declaration_client(
         channel_id: u32,
@@ -234,10 +229,8 @@ impl ExtendedChannel {

         // conservative check against the spec's worst-case `NewTemplate::coinbase_prefix`.
         // the exact size is re-checked against each actual template in `JobFactory::coinbase`
-        if job_factory.script_sig_size(
-            MAX_COINBASE_PREFIX_SIZE,
-            extranonce_prefix.len() + rollable_extranonce_size as usize,
-        ) > MAX_SCRIPT_SIG_SIZE
+        if !job_factory
+            .fits_script_sig_budget(extranonce_prefix.len() + rollable_extranonce_size as usize)
         {
             return Err(ExtendedChannelError::ScriptSigSizeTooLarge);
         }
@@ -298,7 +291,8 @@ impl ExtendedChannel {
     /// and share validation will be performed accordingly.
     ///
     /// Returns an error if the new extranonce prefix is too large, or if it would push the
-    /// assembled coinbase `scriptSig` past [`MAX_SCRIPT_SIG_SIZE`]. The channel is left unchanged
+    /// assembled coinbase `scriptSig` past its budget (see
+    /// [`JobFactory::fits_script_sig_budget`]). The channel is left unchanged
     /// in both error cases.
     pub fn set_extranonce_prefix(
         &mut self,
@@ -310,11 +304,9 @@ impl ExtendedChannel {

         // re-run the constructor's invariant: a prefix that is individually valid can still push
         // the assembled scriptSig past the consensus cap
-        if self.job_factory.script_sig_size(
-            MAX_COINBASE_PREFIX_SIZE,
+        if !self.job_factory.fits_script_sig_budget(
             extranonce_prefix.len() + self.rollable_extranonce_size as usize,
-        ) > MAX_SCRIPT_SIG_SIZE
-        {
+        ) {
             return Err(ExtendedChannelError::ScriptSigSizeTooLarge);
         }

@@ -481,8 +473,9 @@ impl ExtendedChannel {
     /// Returns [`ExtendedChannelError::JobFactoryError`] wrapping
     /// [`JobFactoryError::ScriptSigSizeTooLarge`](crate::server::jobs::error::JobFactoryError::ScriptSigSizeTooLarge)
     /// if the template's `coinbase_prefix` pushes the assembled coinbase `scriptSig` past
-    /// [`MAX_SCRIPT_SIG_SIZE`]. The constructor can only check against the spec's worst-case
-    /// [`MAX_COINBASE_PREFIX_SIZE`], so this is where an out-of-spec Template Provider is caught.
+    /// its budget. The constructor can only check against the spec's worst-case prefix (see
+    /// [`JobFactory::fits_script_sig_budget`]), so this is where an out-of-spec Template Provider
+    /// is caught.
     pub fn on_new_template(
         &mut self,
         template: NewTemplateOwned,
diff --git a/sv2/channels-sv2/src/server/group.rs b/sv2/channels-sv2/src/server/group.rs
index 27217d84..d6ed3423 100644
--- a/sv2/channels-sv2/src/server/group.rs
+++ b/sv2/channels-sv2/src/server/group.rs
@@ -34,11 +34,7 @@ use crate::{
     chain_tip::ChainTip,
     server::{
         error::GroupChannelError,
-        jobs::{
-            extended::ExtendedJob,
-            factory::{JobFactory, MAX_COINBASE_PREFIX_SIZE, MAX_SCRIPT_SIG_SIZE},
-            job_store::JobStore,
-        },
+        jobs::{extended::ExtendedJob, factory::JobFactory, job_store::JobStore},
     },
 };
 use bitcoin::transaction::TxOut;
@@ -82,8 +78,8 @@ impl GroupChannel {
     /// and `//` delimiters: `/pool_tag_string//`
     ///
     /// Returns [`GroupChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
-    /// extranonce and a worst-case [`MAX_COINBASE_PREFIX_SIZE`] coinbase prefix do not fit within
-    /// [`MAX_SCRIPT_SIG_SIZE`].
+    /// extranonce and a worst-case coinbase prefix do not fit within the coinbase `scriptSig`
+    /// budget, see [`JobFactory::fits_script_sig_budget`].
     pub fn new_for_pool(
         group_channel_id: u32,
         full_extranonce_size: usize,
@@ -110,8 +106,8 @@ impl GroupChannel {
     /// `/` delimiters: `/pool_tag_string/miner_tag_string/`
     ///
     /// Returns [`GroupChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
-    /// extranonce and a worst-case [`MAX_COINBASE_PREFIX_SIZE`] coinbase prefix do not fit within
-    /// [`MAX_SCRIPT_SIG_SIZE`].
+    /// extranonce and a worst-case coinbase prefix do not fit within the coinbase `scriptSig`
+    /// budget, see [`JobFactory::fits_script_sig_budget`].
     pub fn new_for_job_declaration_client(
         group_channel_id: u32,
         full_extranonce_size: usize,
@@ -138,9 +134,7 @@ impl GroupChannel {

         // conservative check against the spec's worst-case `NewTemplate::coinbase_prefix`.
         // the exact size is re-checked against each actual template in `JobFactory::coinbase`
-        if job_factory.script_sig_size(MAX_COINBASE_PREFIX_SIZE, full_extranonce_size)
-            > MAX_SCRIPT_SIG_SIZE
-        {
+        if !job_factory.fits_script_sig_budget(full_extranonce_size) {
             return Err(GroupChannelError::ScriptSigSizeTooLarge);
         }

@@ -184,17 +178,16 @@ impl GroupChannel {
     /// Also clears all channel IDs, as no channels can belong to the same group while having different `full_extranonce_size`s.
     ///
     /// Returns [`GroupChannelError::ScriptSigSizeTooLarge`] if the new size would push the
-    /// assembled coinbase `scriptSig` past [`MAX_SCRIPT_SIG_SIZE`], leaving the group channel
-    /// unchanged.
+    /// assembled coinbase `scriptSig` past its budget (see
+    /// [`JobFactory::fits_script_sig_budget`]), leaving the group channel unchanged.
     pub fn set_full_extranonce_size(
         &mut self,
         full_extranonce_size: usize,
     ) -> Result<(), GroupChannelError> {
         // re-run the constructor's invariant, before touching any state
-        if self
+        if !self
             .job_factory
-            .script_sig_size(MAX_COINBASE_PREFIX_SIZE, full_extranonce_size)
-            > MAX_SCRIPT_SIG_SIZE
+            .fits_script_sig_budget(full_extranonce_size)
         {
             return Err(GroupChannelError::ScriptSigSizeTooLarge);
         }
@@ -267,9 +260,10 @@ impl GroupChannel {
     ///
     /// Returns [`GroupChannelError::JobFactoryError`] wrapping
     /// [`JobFactoryError::ScriptSigSizeTooLarge`](crate::server::jobs::error::JobFactoryError::ScriptSigSizeTooLarge)
-    /// if the template's `coinbase_prefix` pushes the assembled coinbase `scriptSig` past
-    /// [`MAX_SCRIPT_SIG_SIZE`]. The constructor can only check against the spec's worst-case
-    /// [`MAX_COINBASE_PREFIX_SIZE`], so this is where an out-of-spec Template Provider is caught.
+    /// if the template's `coinbase_prefix` pushes the assembled coinbase `scriptSig` past its
+    /// budget. The constructor can only check against the spec's worst-case prefix (see
+    /// [`JobFactory::fits_script_sig_budget`]), so this is where an out-of-spec Template Provider
+    /// is caught.
     pub fn on_new_template(
         &mut self,
         template: NewTemplateOwned,
diff --git a/sv2/channels-sv2/src/server/jobs/factory.rs b/sv2/channels-sv2/src/server/jobs/factory.rs
index 4d7eeb57..6215a681 100644
--- a/sv2/channels-sv2/src/server/jobs/factory.rs
+++ b/sv2/channels-sv2/src/server/jobs/factory.rs
@@ -46,8 +46,8 @@ use template_distribution_sv2::NewTemplateOwned;
 /// consensus rules.
 pub const MAX_SCRIPT_SIG_SIZE: usize = 100;

-/// Maximum number of bytes that [`NewTemplate::coinbase_prefix`] is allowed to contribute to the
-/// coinbase `scriptSig`, as mandated by the Sv2 Template Distribution Protocol spec.
+/// Maximum number of bytes that [`NewTemplateOwned::coinbase_prefix`] is allowed to contribute to
+/// the coinbase `scriptSig`, as mandated by the Sv2 Template Distribution Protocol spec.
 ///
 /// The spec phrase "up to 8 bytes (not including the length byte)" refers to the `B0255` wire
 /// length prefix. The BIP34 script push opcode is already part of these 8 bytes.
@@ -149,8 +149,8 @@ impl JobFactory {
     }

     /// Returns the number of bytes of the coinbase `scriptSig` assembled by this factory, for a
-    /// [`NewTemplate::coinbase_prefix`] of `coinbase_prefix_size` bytes and a full extranonce of
-    /// `full_extranonce_size` bytes.
+    /// [`NewTemplateOwned::coinbase_prefix`] of `coinbase_prefix_size` bytes and a full extranonce
+    /// of `full_extranonce_size` bytes.
     ///
     /// The assembled layout is:
     /// `coinbase_prefix || OP_PUSHBYTES_N || /pool_tag/miner_tag/ || OP_PUSHBYTES_M || extranonce`
@@ -168,6 +168,19 @@ impl JobFactory {
             + full_extranonce_size
     }

+    /// Returns whether the coinbase `scriptSig` assembled by this factory fits within
+    /// [`MAX_SCRIPT_SIG_SIZE`], for a full extranonce of `full_extranonce_size` bytes and the
+    /// worst-case [`NewTemplateOwned::coinbase_prefix`] allowed by the spec
+    /// ([`MAX_COINBASE_PREFIX_SIZE`]).
+    ///
+    /// Meant for callers that do not have a template at hand, such as channel constructors and
+    /// extranonce setters. Since it assumes the largest in-spec prefix, a `true` here guarantees
+    /// that no in-spec template can overflow the budget; the exact size is still re-checked
+    /// against each actual template in `JobFactory::coinbase`.
+    pub fn fits_script_sig_budget(&self, full_extranonce_size: usize) -> bool {
+        self.script_sig_size(MAX_COINBASE_PREFIX_SIZE, full_extranonce_size) <= MAX_SCRIPT_SIG_SIZE
+    }
+
     /// Creates a new job from a template.
     ///
     /// This job (and related shares) is fully committed to:
diff --git a/sv2/channels-sv2/src/server/standard.rs b/sv2/channels-sv2/src/server/standard.rs
index 90af70c3..043a4efa 100644
--- a/sv2/channels-sv2/src/server/standard.rs
+++ b/sv2/channels-sv2/src/server/standard.rs
@@ -39,10 +39,7 @@ use crate::{
     server::{
         error::StandardChannelError,
         jobs::{
-            extended::ExtendedJob,
-            factory::{JobFactory, MAX_COINBASE_PREFIX_SIZE, MAX_SCRIPT_SIG_SIZE},
-            job_store::JobStore,
-            standard::StandardJob,
+            extended::ExtendedJob, factory::JobFactory, job_store::JobStore, standard::StandardJob,
         },
         share_accounting::{ShareAccounting, ShareValidationError, ShareValidationResult},
     },
@@ -116,8 +113,8 @@ impl StandardChannel {
     /// and `//` delimiters: `/pool_tag_string//`
     ///
     /// Returns [`StandardChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
-    /// extranonce prefix and a worst-case [`MAX_COINBASE_PREFIX_SIZE`] coinbase prefix do not fit
-    /// within [`MAX_SCRIPT_SIG_SIZE`].
+    /// extranonce prefix and a worst-case coinbase prefix do not fit within the coinbase
+    /// `scriptSig` budget, see [`JobFactory::fits_script_sig_budget`].
     #[allow(clippy::too_many_arguments)]
     pub fn new_for_pool(
         channel_id: u32,
@@ -154,8 +151,8 @@ impl StandardChannel {
     /// `/` delimiters: `/pool_tag_string/miner_tag_string/`
     ///
     /// Returns [`StandardChannelError::ScriptSigSizeTooLarge`] if the tags, the delimiters, the
-    /// extranonce prefix and a worst-case [`MAX_COINBASE_PREFIX_SIZE`] coinbase prefix do not fit
-    /// within [`MAX_SCRIPT_SIG_SIZE`].
+    /// extranonce prefix and a worst-case coinbase prefix do not fit within the coinbase
+    /// `scriptSig` budget, see [`JobFactory::fits_script_sig_budget`].
     #[allow(clippy::too_many_arguments)]
     pub fn new_for_job_declaration_client(
         channel_id: u32,
@@ -218,9 +215,7 @@ impl StandardChannel {
         // conservative check against the spec's worst-case `NewTemplate::coinbase_prefix`.
         // the exact size is re-checked against each actual template in `JobFactory::coinbase`.
         // standard channels have no rollable extranonce, so the prefix is the full extranonce
-        if job_factory.script_sig_size(MAX_COINBASE_PREFIX_SIZE, extranonce_prefix.len())
-            > MAX_SCRIPT_SIG_SIZE
-        {
+        if !job_factory.fits_script_sig_budget(extranonce_prefix.len()) {
             return Err(StandardChannelError::ScriptSigSizeTooLarge);
         }

@@ -259,7 +254,8 @@ impl StandardChannel {
     /// Sets a new extranonce prefix for this channel.
     ///
     /// Returns an error if the new prefix is too large, or if it would push the assembled coinbase
-    /// `scriptSig` past [`MAX_SCRIPT_SIG_SIZE`]. The channel is left unchanged in both error cases.
+    /// `scriptSig` past its budget (see [`JobFactory::fits_script_sig_budget`]). The channel is
+    /// left unchanged in both error cases.
     pub fn set_extranonce_prefix(
         &mut self,
         extranonce_prefix: AllocatedExtranoncePrefix,
@@ -270,10 +266,9 @@ impl StandardChannel {

         // re-run the constructor's invariant: a prefix that is individually valid can still push
         // the assembled scriptSig past the consensus cap
-        if self
+        if !self
             .job_factory
-            .script_sig_size(MAX_COINBASE_PREFIX_SIZE, extranonce_prefix.len())
-            > MAX_SCRIPT_SIG_SIZE
+            .fits_script_sig_budget(extranonce_prefix.len())
         {
             return Err(StandardChannelError::ScriptSigSizeTooLarge);
         }
@@ -451,8 +446,9 @@ impl StandardChannel {
     /// Returns [`StandardChannelError::JobFactoryError`] wrapping
     /// [`JobFactoryError::ScriptSigSizeTooLarge`](crate::server::jobs::error::JobFactoryError::ScriptSigSizeTooLarge)
     /// if the template's `coinbase_prefix` pushes the assembled coinbase `scriptSig` past
-    /// [`MAX_SCRIPT_SIG_SIZE`]. The constructor can only check against the spec's worst-case
-    /// [`MAX_COINBASE_PREFIX_SIZE`], so this is where an out-of-spec Template Provider is caught.
+    /// its budget. The constructor can only check against the spec's worst-case prefix (see
+    /// [`JobFactory::fits_script_sig_budget`]), so this is where an out-of-spec Template Provider
+    /// is caught.
     pub fn on_new_template(
         &mut self,
         template: NewTemplateOwned,
```

bit-aloo (approved):
ACK

GitGab19:
Clanker review:

> Verdict: request changes on [[PR #2243](https://github.com/stratum-mining/stratum/pull/2243)](https://github.com/stratum-mining/stratum/pull/2243).
> 
> ### Finding
> 
> - **[High] The `scriptSig` guard is bypassed for custom jobs.** The new “authoritative” check exists only in [`[JobFactory::coinbase](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L626-L635)`](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L626-L635). [`[new_custom_job](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L416)`](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L416) and the receiving [`[custom_coinbase](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L542)`](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L542) path never call it.
> 
>   I reproduced this with the PR’s boundary configuration: a 52-byte pool tag, 32-byte extranonce, and 9-byte template prefix produce a 101-byte `scriptSig`. Both `new_custom_job` and `new_extended_job_from_custom_job` return `Ok`, so the JDC can still create—and the pool can still accept—consensus-invalid work. Larger custom prefixes also invalidate the custom prefix/suffix code’s assumption that the CompactSize script length occupies exactly one byte.
> 
>   I’d factor an exact-size validator out of `coinbase()` and call it from `new_custom_job`; the receiving path should independently enforce:
> 
>   ```rust
>   set_custom_mining_job.coinbase_prefix.len() + full_extranonce_size <= MAX_SCRIPT_SIG_SIZE
> ```

plebhash:
> Clanker review:
> 
> > Verdict: request changes on [[PR #2243](https://github.com/stratum-mining/stratum/pull/2243)](#2243).
> > ### Finding
> > 
> > * **[High] The `scriptSig` guard is bypassed for custom jobs.** The new “authoritative” check exists only in [`[JobFactory::coinbase](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L626-L635)`](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L626-L635). [`[new_custom_job](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L416)`](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L416) and the receiving [`[custom_coinbase](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L542)`](https://github.com/stratum-mining/stratum/blob/18029b1e86b19a2d3d76ac5cb0046f5414cce625/sv2/channels-sv2/src/server/jobs/factory.rs#L542) path never call it.
> >   I reproduced this with the PR’s boundary configuration: a 52-byte pool tag, 32-byte extranonce, and 9-byte template prefix produce a 101-byte `scriptSig`. Both `new_custom_job` and `new_extended_job_from_custom_job` return `Ok`, so the JDC can still create—and the pool can still accept—consensus-invalid work. Larger custom prefixes also invalidate the custom prefix/suffix code’s assumption that the CompactSize script length occupies exactly one byte.
> >   I’d factor an exact-size validator out of `coinbase()` and call it from `new_custom_job`; the receiving path should independently enforce:
> >   ```rust
> >   set_custom_mining_job.coinbase_prefix.len() + full_extranonce_size <= MAX_SCRIPT_SIG_SIZE
> >   ```

addressed, clanker answer:

> Good catch on the custom job path — confirmed, and fixed in `fa1cd824`.
> 
> I reproduced all three cases before touching anything:
> 
> ```
> JDC new_custom_job (52-byte tag, 9-byte prefix): Ok -> 101-byte scriptSig
> pool new_extended_job_from_custom_job:           Ok
> hostile 250-byte prefix from downstream:         Ok -> 282-byte scriptSig
> ```
> 
> The delegation argument doesn't hold for the *receiving* path: `new_extended_job_from_custom_job` takes `coinbase_prefix` from a downstream JDC over the wire, so the "upstream caller" is a remote peer and there is nothing in-tree left to delegate to. That one is worth fixing regardless of scope, so I did — the 250-byte case above is a hostile downstream, not a misconfigured local one.
> 
> **The two paths can't share one validator.** Factoring the exact-size check out of `coinbase()` and calling it from both doesn't work, because the arithmetic differs:
> 
> - `new_custom_job` (JDC *builds* the prefix): the factory adds the pool/miner tag itself, so `script_sig_size(template.coinbase_prefix.len(), full_extranonce_size)` applies.
> - `new_extended_job_from_custom_job` (pool *receives* it): the tag is already embedded in the incoming `coinbase_prefix`, so it's `coinbase_prefix.len() + full_extranonce_size` — exactly your second snippet. Reusing `script_sig_size` here would double-count the tag.
> 
> Both are bounded now. The receiving check sits in `custom_coinbase` so the prefix and suffix builders are covered by a single choke point, mirroring how `coinbase()` owns the template path. Bounding the custom path to 100 bytes also closes the CompactSize concern you raised, since the scriptSig can no longer reach 253.
> 
> Covered by `test_new_custom_job_rejects_oversized_script_sig` and `test_custom_job_rejects_oversized_script_sig`: exact-100 → `Ok`, 101 → `Err`, plus the 250-byte hostile case.

GitGab19 (approved):

PR #2255, opened 2026-08-03
`channels_sv2`: fix chain-tip transition correctness gaps in the job state machine

### LINKED ISSUE #2254: `channels_sv2`: need to fix chain-tip transition correctness gaps in the job state machine
On chain-tip transitions, the job lifecycle contract (activate future job, retire old active job to stale, propagate `min_ntime`) is enforced by some channel implementations and silently dropped by others. Most outliers sit next to a sibling that does it correctly; the last one is missed by every client-side implementation.

### 1. `GroupChannel::on_set_new_prev_hash` ignores `activate_future_job`'s result

`JobStore::activate_future_job` returns `false` and no-ops when `template_id` matches no queued future job, but `group.rs` discards the `bool`, unconditionally advances `chain_tip`, and returns `Ok(())` — despite its own doc comment promising an error.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/server/group.rs#L302-L305

`server/standard.rs` and `server/extended.rs` propagate the `bool` and return `Err(TemplateIdNotFound)`.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/server/standard.rs#L525-L530

Result: silent state corruption; subsequent group jobs are built against a tip whose job was never activated.

Fix: return `Err(GroupChannelError::TemplateIdNotFound)` when activation fails, before touching `chain_tip`.

### 2. `ExtendedChannel::on_chain_tip_update` leaves the stale `active_job` in place

The JDC entry point for tip updates clears future jobs and rotates past→stale, but never transitions `active_job`. A share arriving in the window before the next `SetCustomMiningJobSuccess` passes `is_active_job`, bypasses the stale early-return, and is re-hashed against the NEW tip's `prev_hash` with the OLD job's coinbase/merkle path — phantom `Valid`/`BlockFound` credited for work unrelated to the current tip.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/extended.rs#L456-L472

Fix: demote the prior `active_job` into `past_jobs` before the past→stale rotation the sibling `on_set_new_prev_hash` path already performs (that path has its own gap — see 4).

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/extended.rs#L483-L512

### 3. `GroupChannel::on_set_new_prev_hash` doesn't propagate `min_ntime` into the activated job

The future job is promoted to `active_job` as-is, keeping `min_ntime == None`, so `is_future()` still returns `true` on the active job.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/group.rs#L163-L178

`client/standard.rs` and `client/extended.rs` both write `activated_job.min_ntime = Some(set_new_prev_hash.min_ntime)` on activation.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/extended.rs#L487-L491

Consumers dispatching on `min_ntime` misclassify the job.

Fix: `activated_job.set_no_future(set_new_prev_hash.min_ntime)` before promoting.

### 4. Client `on_set_new_prev_hash` drops the prior `active_job` instead of retiring it

Both client channels overwrite `active_job` with the activated future job. The job that was active under the *previous* prev hash is never inserted into `past_jobs`, so the past→stale rotation two lines below cannot see it — it lands in neither `past_jobs` nor `stale_jobs` and is simply gone.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/extended.rs#L487-L501

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/standard.rs#L299-L311

The server side gets this right: `JobStore::activate_future_job` takes the current `active_job` into `past_jobs` before activating, then calls `mark_past_jobs_as_stale`.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/server/jobs/job_store.rs#L88-L99

Result: a late share for the just-retired job falls through to the `else` branch of `validate_share` and is rejected as `InvalidJobId` instead of `Stale`.

Fix: capture the previous job via `self.active_job.replace(activated_job)` and insert it into `past_jobs` — after the `JobIdNotFound` early-return, so the error path doesn't corrupt state (same rule as the "clear only after a successful activation" ordering in `server/standard.rs`).

## impact

Wrong error codes for late shares (`InvalidJobId`/`DoesNotMeetTarget` instead of `Stale`), silent group-channel tip corruption, phantom share validation/accounting, active-vs-future misclassification.

plebhash:
close #2254 
close https://github.com/project-loupe/audit-stratum/issues/13
close https://github.com/project-loupe/audit-stratum/issues/15
close https://github.com/project-loupe/audit-stratum/issues/29

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -518,23 +518,34 @@ impl ExtendedChannel {
     ///
     /// The message could be either directed at this channel, or at a group channel it belongs to.
     ///
-    /// - If the referenced `job_id` is not a future job, returns an error.
+    /// - If the referenced `job_id` is not a future job, returns an error and leaves channel state
+    ///   untouched.
     /// - If it is a future job, activates it as the current job.
-    /// - Marks all past jobs as stale and clears them.
+    /// - Marks the previously active job and all past jobs as stale, and clears past jobs.
     /// - Clears all seen shares as shares for the previous chain tip will be rejected as stale.
     /// - Updates the chain tip for the channel.
     pub fn on_set_new_prev_hash(
         &mut self,
         set_new_prev_hash: SetNewPrevHashMp,
     ) -> Result<(), ExtendedChannelError> {
-        match self.future_jobs.remove(&set_new_prev_hash.job_id) {
+        // the previously active job is only displaced once activation is known to succeed, so
+        // that the JobIdNotFound path below does not corrupt channel state
+        let previously_active_job = match self.future_jobs.remove(&set_new_prev_hash.job_id) {
             Some(mut activated_job) => {
                 activated_job.0.min_ntime = Sv2OptionOwned::new(Some(set_new_prev_hash.min_ntime));
-                self.active_job = Some(activated_job);
+                self.active_job.replace(activated_job)
             }
             None => {
                 return Err(ExtendedChannelError::JobIdNotFound);
             }
+        };
+
+        // the job that was active under the previous chain tip must be retired to stale rather
+        // than silently dropped, otherwise a late share for it would be rejected as
+        // InvalidJobId instead of Stale
+        if let Some(previously_active_job) = previously_active_job {
+            self.past_jobs
+                .insert(previously_active_job.0.job_id, previously_active_job);
         }
 
         // all other future jobs are now useless
@@ -743,6 +754,7 @@ mod tests {
     use crate::{
         chain_tip::ChainTip,
         client::{
+            error::ExtendedChannelError,
             extended::ExtendedChannel,
             share_accounting::{ShareValidationError, ShareValidationResult},
             MAX_FUTURE_JOBS,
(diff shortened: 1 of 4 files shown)


GitGab19 (approved):
ACk

PR #2257, opened 2026-08-03
`channels_sv2`: fix client `set_target` does not refresh targets of already-queued future jobs

### LINKED ISSUE #2256: `channels_sv2`: client `set_target` does not refresh targets of already-queued future jobs
Per the spec (previously quoted in #1981):

> Maximum target is valid until the next SetTarget message is sent and is applicable for all jobs received on the channel in the future or already received with an empty min_ntime.

#1981 fixed target association on the server channels, but the client channels still miss the second half of that rule. 

Both `StandardChannel::set_target` and `ExtendedChannel::set_target` only replace `self.target`. 

Queued future jobs keep the `Target` captured inside their `future_jobs` tuples at arrival time; when `on_set_new_prev_hash` later activates such a job, `validate_share` compares share hashes against the stale target (`job.1` / `job.2`).

## impact

After a `SetTarget` that tightens difficulty, a share that meets the old (easier) target but not the new one is accepted by the local validator even though the upstream server will reject it — distorting validated-work accounting and any downstream gating that relies on this validator.

## suggested fix

In both client channels:

```rust
pub fn set_target(&mut self, target: Target) {
    self.target = target;
    for future_job in self.future_jobs.values_mut() {
        future_job.1 = target; // .2 in extended.rs (tuple is (job, prefix, target) there)
    }
}
```

Note the spec limits this to jobs with empty `min_ntime` — exactly the `future_jobs` set, so no filtering is needed.

## regression test

Queue a future job under a permissive target, call `set_target` with a strict target, activate via `on_set_new_prev_hash`, then submit a share between the two targets: must return `DoesNotMeetTarget`.

plebhash:
close https://github.com/stratum-mining/stratum/issues/2256
close https://github.com/project-loupe/audit-stratum/issues/56

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -193,8 +193,16 @@ impl ExtendedChannel {
     }
 
     /// Sets a new [`Target`] for the channel.
+    ///
+    /// Per the Sv2 spec, the new target also applies to jobs that were already received with an
+    /// empty `min_ntime` (i.e. queued future jobs), so their associated target is refreshed here.
+    /// Jobs that were already received with a set `min_ntime` (active, past and stale jobs) keep
+    /// their target.
     pub fn set_target(&mut self, new_target: Target) {
         self.target = new_target;
+        for future_job in self.future_jobs.values_mut() {
+            future_job.2 = new_target;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


GitGab19 (approved):
Merge conflicts aside, this LGTM.

PR #2259, opened 2026-08-03
`channels_sv2`: fix client `StandardChannel::on_new_group_channel_job` panics on malformed upstream coinbase

### LINKED ISSUE #2258: `channels_sv2`: client `StandardChannel::on_new_group_channel_job` panics on malformed upstream coinbase
`on_new_group_channel_job` feeds the upstream-controlled `coinbase_tx_prefix` / `coinbase_tx_suffix` (`B064K` wire fields, no minimum length) from a `NewExtendedMiningJob` into `merkle_root_from_path`, then calls `.expect("merkle root must be valid")`.

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/standard.rs#L240-L260

The helper returns `None` whenever `prefix || extranonce_prefix || suffix` fails to deserialize as a Bitcoin transaction — e.g. both fields empty. A malicious or compromised upstream pool can therefore panic any connected mining client / JDC.

The extended-channel sibling of this bug class was already fixed: `on_new_extended_mining_job` propagates `try_strip_bip141` failures as `ExtendedChannelError::FailedToTryToStripBip141`. 

The standard-channel path is the remaining panic site.

## fix

change the signature to `Result<(), StandardChannelError>` with a new `InvalidCoinbase`/`MerkleRootInvalid` variant, mirroring the extended channel's error-propagation style.

## regression test

feed a `NewExtendedMiningJob` with empty prefix/suffix, and assert `Err`

plebhash:
close #2258
close https://github.com/project-loupe/audit-stratum/issues/14

### DIFF
--- a/sv2/channels-sv2/src/client/error.rs
+++ b/sv2/channels-sv2/src/client/error.rs
@@ -37,6 +37,10 @@ pub enum StandardChannelError {
 
     /// The provided extranonce prefix exceeds the maximum allowed size.
     NewExtranoncePrefixTooLarge,
+
+    /// The coinbase transaction of a group channel job is malformed, so no merkle root
+    /// could be derived from it.
+    InvalidCoinbase,
 }
 
 /// Errors that can occur within a **group channel** context.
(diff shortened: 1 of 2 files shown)


GitGab19 (approved):
Merge conflicts aside, it LGTM.

bit-aloo (approved):
ACK

PR #2261, opened 2026-08-03
`channels_sv2`: fix `ExtranoncePrefix` allocation released while live jobs still reference its bytes

### LINKED ISSUE #2260: `channels_sv2`: `ExtranoncePrefix` allocation released while live jobs still reference its bytes — extranonce space can overlap across channels
Allocation lifetime is tied to the channel's current `extranonce_prefix` object

`Drop` clears the slot in the allocator's bitmap

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/extranonce_manager/prefix.rs#L265-L273

But jobs created under that prefix embed only the raw prefix bytes — not an allocation handle — and remain valid after `set_extranonce_prefix` rotates the channel's prefix (by design: a job keeps its creation-time prefix, and shares for it are still validated against its merkle root).

Consequence: after a rotation, the old slot returns to the free pool while old jobs are still accepted under those bytes. A subsequent `allocate_extended`/`allocate_standard` can hand the identical prefix to a second live channel. Two live channels then share an `extranonce` search space — the same work is replayable across both, and per-channel duplicate detection no longer prevents double credit.

Reachability: requires an integration path that rotates a channel prefix while prior jobs are still live (e.g. pool acting on upstream `SetExtranoncePrefix` or re-allocating extranonce space). The in-tree channel APIs explicitly support jobs outliving a rotation, so this state is reachable by design, not misuse.

## fix

Channel defers release. On `set_extranonce_prefix`, the channel stashes the old prefix object and drops it only when the last job created under it leaves the job store (past→stale retirement). Localized to channel code; ties slot lifetime to the existing job-lifecycle machinery.

## regression test

Allocate two slots (exhausting a small allocator), create a job under one, rotate the channel prefix onto the second, then attempt a third allocation: must fail with `CapacityExhausted` while the pre-rotation job is still valid.

plebhash:
close #2260
close https://github.com/project-loupe/audit-stratum/issues/58

### DIFF
--- a/sv2/channels-sv2/src/server/jobs/job_store.rs
+++ b/sv2/channels-sv2/src/server/jobs/job_store.rs
@@ -7,9 +7,13 @@
 //! - **Template Mapping**: Tracks mappings from template IDs to job IDs for future jobs.
 //! - **Lifecycle Management**: Ensures correct state transitions when activating jobs or updating
 //!   chain tips.
+//! - **Retired Extranonce Prefixes**: Holds on to extranonce prefixes that were rotated out of the
+//!   channel while jobs created under them can still accept shares, so that their allocator slots
+//!   are not handed to another channel too early.
 use std::collections::HashMap;
 
 use super::Job;
+use crate::extranonce_manager::ExtranoncePrefix;
 
 /// Internal implementation for tracking mining job states in SV2 server channels.
 ///
@@ -25,6 +29,10 @@ pub(crate) struct JobStore<T: Job> {
     past_jobs: HashMap<u32, T>,
     // Stale jobs are indexed with job_id (u32)
     stale_jobs: HashMap<u32, T>,
+    // Extranonce prefixes rotated out of the channel that are still referenced by at least one job
+    // that can accept shares. Holding the object here keeps its allocator slot reserved; dropping
+    // it releases the slot.
+    retired_extranonce_prefixes: Vec<ExtranoncePrefix>,
 }
 
 impl<T: Job> JobStore<T> {
@@ -36,6 +44,7 @@ impl<T: Job> JobStore<T> {
             active_job: None,
             past_jobs: HashMap::new(),
             stale_jobs: HashMap::new(),
+            retired_extranonce_prefixes: Vec::new(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -112,6 +121,48 @@ impl<T: Job> JobStore<T> {
...
+    /// Takes ownership of an extranonce prefix that is no longer the channel's current one,
+    /// releasing it only once no job created under it can accept shares anymore.
+    ///
+    /// Jobs only carry a copy of the extranonce prefix bytes they were created under, and stay
+    /// valid across a prefix rotation. Dropping the prefix object right away would return its
+    /// slot to the allocator while those jobs still validate shares under those bytes, allowing
+    /// the same extranonce space to be handed to a second live channel.
+    pub fn retire_extranonce_prefix(&mut self, extranonce_prefix: ExtranoncePrefix) {
+        self.retired_extranonce_prefixes.push(extranonce_prefix);
+        self.prune_retired_extranonce_prefixes();
```
GitGab19:
```suggestion
        if self.is_extranonce_prefix_in_use(extranonce_prefix.as_bytes()) {
             self.retired_extranonce_prefixes.push(extranonce_prefix);
        }
        // Otherwise `prefix` drops here and releases its slot.
```

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -112,6 +121,48 @@ impl<T: Job> JobStore<T> {
...
+        if self.retired_extranonce_prefixes.is_empty() {
+            return;
+        }
+        // take first, so that the filter closure can read the job collections through `&self`
+        let retired = std::mem::take(&mut self.retired_extranonce_prefixes);
+        let kept: Vec<ExtranoncePrefix> = retired
+            .into_iter()
+            .filter(|prefix| self.is_extranonce_prefix_in_use(prefix.as_bytes()))
+            .collect();
+        self.retired_extranonce_prefixes = kept;
```
GitGab19:
```suggestion
        let future_jobs = &self.future_jobs;
        let active_job = &self.active_job;
        let past_jobs = &self.past_jobs;

        self.retired_extranonce_prefixes.retain(|prefix| {
            future_jobs
                .values()
                .chain(active_job.iter())
                .chain(past_jobs.values())
                .any(|job| job.get_extranonce_prefix() == prefix.as_bytes())
        });
```

GitGab19 (approved):
LGTM

Just a couple of non-blocking suggestions, which are just alternatives way to achieve what you're already achieving in both `retire_extranonce_prefix` and `prune_retired_extranonce_prefixes`.

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: done

sv2/channels-sv2/src/server/jobs/job_store.rs
plebhash: done

GitGab19 (approved):

PR #2236, opened 2026-07-31
`channels_sv2`: BIP323 adaptations + fix version rolling validation

plebhash:
close #2196 
close https://github.com/project-loupe/audit-stratum/issues/57
close https://github.com/project-loupe/audit-stratum/issues/70
close https://github.com/project-loupe/audit-stratum/issues/72
close https://github.com/project-loupe/audit-stratum/issues/59

companion https://github.com/stratum-mining/sv2-apps/pull/655

PR #2263, opened 2026-08-03
`channels_sv2`: bound client `rejected_shares` map

### LINKED ISSUE #2262: `channels_sv2`: client `rejected_shares` map grows unboundedly
`ShareAccounting::on_share_rejection` is fed `error_code` values straight from upstream `SubmitSharesError` messages:

https://github.com/stratum-mining/stratum/blob/30ca82ac20d7a543cc777be39dc2e19c621eff4b/sv2/channels-sv2/src/client/share_accounting.rs#L129-L134

The wire type is `Str0255`, so a malicious or compromised pool controls up to 255 bytes of arbitrary content per message. 

Every distinct value becomes a permanent key in `rejected_shares: HashMap<String, u32>` — no length cap, no eviction, no allowlist, and (unlike `seen_shares`) no flush on chain-tip updates.

Over a sustained connection this is a slow but real memory-exhaustion DoS against the mining client (CWE-400), mirroring the upstream-trust threat model already accepted for the same struct's counter-overflow fixes.

In-repo usage treats this set as a small fixed enumeration (the `ERROR_CODE_SUBMIT_SHARES_*` constants in `mining_sv2`), so bounding it matches intent.

## fix

count only the known `ERROR_CODE_SUBMIT_SHARES_*` codes

fold anything else into an "unknown" bucket.

## regression test

Feed 10,000 unique error codes into `on_share_rejection`, assert the stored map stays bounded.

plebhash:
close #2262
close https://github.com/project-loupe/audit-stratum/issues/28

companion https://github.com/stratum-mining/sv2-apps/pull/693

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -289,7 +289,7 @@ impl ExtendedChannel {
 
     /// Updates share accounting based on a [`SubmitSharesError`](mining_sv2::SubmitSharesError) message from the upstream
     /// server. Delegates to [`ShareAccounting::on_share_rejection`].
-    pub fn on_share_rejection(&mut self, error_code: String) {
+    pub fn on_share_rejection(&mut self, error_code: &str) {
         self.share_accounting.on_share_rejection(error_code);
     }
 
--- a/sv2/channels-sv2/src/client/share_accounting.rs
+++ b/sv2/channels-sv2/src/client/share_accounting.rs
@@ -8,6 +8,37 @@ extern crate alloc;
 use super::{HashMap, HashSet};
 use alloc::string::String;
 use bitcoin::hashes::sha256d::Hash;
+use mining_sv2::{
+    ERROR_CODE_SUBMIT_SHARES_BAD_EXTRANONCE_SIZE, ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW,
+    ERROR_CODE_SUBMIT_SHARES_DUPLICATE_SHARE, ERROR_CODE_SUBMIT_SHARES_INVALID_CHANNEL_ID,
+    ERROR_CODE_SUBMIT_SHARES_INVALID_JOB_ID,
+    ERROR_CODE_SUBMIT_SHARES_INVALID_NON_ROLLABLE_VERSION_BIT,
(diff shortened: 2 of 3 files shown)


plebhash:
todo: adapt `KNOWN_ERROR_CODES` list once https://github.com/stratum-mining/stratum/pull/2236 is merged

sv2/channels-sv2/src/client/share_accounting.rs
```diff
@@ -126,9 +157,19 @@ impl ShareAccounting {
...
+    /// `error_code` is only tracked under its own key if it is one of the known
+    /// `SubmitShares.Error` codes; any other value is counted under [`UNKNOWN_ERROR_CODE`].
+    /// This keeps the internal map bounded, since `error_code` is fully controlled by the
+    /// upstream server.
     pub fn on_share_rejection(&mut self, error_code: String) {
+        let key = if KNOWN_ERROR_CODES.contains(&error_code.as_str()) {
+            error_code
+        } else {
+            String::from(UNKNOWN_ERROR_CODE)
+        };
```
bit-aloo: I think we can remove this string allocation here, considering we only need code for lookup

plebhash: sure, done

but if we're really trying to be strict against unnecessary allocations here, `on_share_rejection` should take `&str` instead of `String`

which means a companion is needed for this PR: https://github.com/stratum-mining/sv2-apps/pull/693

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -283,7 +283,7 @@ impl ExtendedChannel {
 
     /// Updates share accounting based on a [`SubmitSharesError`](mining_sv2::SubmitSharesError) message from the upstream
     /// server. Delegates to [`ShareAccounting::on_share_rejection`].
-    pub fn on_share_rejection(&mut self, error_code: String) {
+    pub fn on_share_rejection(&mut self, error_code: &str) {
```
bit-aloo, line 292: We can make these as str as ref
```suggestion
    pub fn on_share_rejection(&mut self, error_code: AsRef<str>) {
```

plebhash: I'd rather not.

IMHO feels like overengineering with no clear benefit.

can we KISS and move on?

bit-aloo: yup

bit-aloo (approved):

GitGab19 (approved):

PR #2266, opened 2026-08-03
`channels_sv2`: bound future-job storage on client channels

### LINKED ISSUE #2264: `channels_sv2`: need to bound future-job storage on client channels
All three client channel types store every future job (`min_ntime == None`) received from the upstream server in a long-lived `future_jobs` map keyed by the upstream-controlled `job_id`, with no cap, eviction, or backpressure:

- `client/group.rs` — `GroupChannel::on_new_extended_mining_job`
- `client/standard.rs` — `StandardChannel::store_new_mining_job`
- `client/extended.rs` — `ExtendedChannel::on_new_extended_mining_job`

Entries are only cleared when a valid `SetNewPrevHash` (or chain-tip update) arrives. 

A malicious or compromised upstream pool can withhold that message and stream future jobs with distinct `u32` job IDs. Because each retained job carries two `B064K` buffers (up to 65,535 bytes each) plus a `Seq0255<U256>` merkle path, a few thousand messages retain ~GB-scale heap — memory-exhaustion DoS of the mining client.

Legitimate cardinality is tiny: pools normally send one future job per chain tip, and `future_jobs` is fully cleared on every `SetNewPrevHash`. A small bound therefore costs nothing functionally.

## fix

Cap + evict-oldest: define `MAX_FUTURE_JOBS` small constant

on overflow, evict the oldest pending job. 

## regression tests

Flood each client channel with 10,000 unique future jobs (empty payloads to isolate cardinality): stored count must stay bounded.

plebhash:
close #2264
close https://github.com/project-loupe/audit-stratum/issues/84

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -4,7 +4,7 @@
 //! **Extended Channel** within a mining client.
 
 extern crate alloc;
-use super::HashMap;
+use super::{HashMap, MAX_FUTURE_JOBS};
 use crate::{
     bip141::try_strip_bip141,
     chain_tip::ChainTip,
@@ -17,7 +17,7 @@ use crate::{
     target::{bytes_to_hex, u256_to_block_hash},
     MAX_EXTRANONCE_LEN, VERSION_ROLLING_MASK,
 };
-use alloc::{format, string::String, vec, vec::Vec};
+use alloc::{collections::VecDeque, format, string::String, vec, vec::Vec};
 use binary_sv2::Sv2OptionOwned;
 use bitcoin::{
     absolute::LockTime,
@@ -57,8 +57,8 @@ pub type ExtendedJob = (NewExtendedMiningJobOwned, Vec<u8>, Target);
 /// - The channel's current target.
 /// - The channel's nominal hashrate.
 /// - Whether version rolling is supported (see [BIP 323](https://github.com/bitcoin/bips/blob/master/bip-0323.mediawiki)).
-/// - Future jobs (indexed by `job_id`) to be activated by a [`SetNewPrevHash`](SetNewPrevHashMp)
-///   message.
+/// - Future jobs (indexed by `job_id`, capped at [`MAX_FUTURE_JOBS`]) to be activated by a
+///   [`SetNewPrevHash`](SetNewPrevHashMp) message.
 /// - The currently active job.
 /// - Past jobs (previously active under the current chain tip, indexed by `job_id`).
 /// - Stale jobs (previously active and past jobs under the previous chain tip, indexed by
@@ -76,6 +76,9 @@ pub struct ExtendedChannel {
     version_rolling: bool,
     // future jobs are indexed with job_id (u32)
     future_jobs: HashMap<u32, ExtendedJob>,
+    // Future job IDs ordered by receipt, oldest at the front and newest at the back.
+    // Replaced IDs move to the back; overflow evicts from the front.
+    future_job_order: VecDeque<u32>,
     active_job: Option<ExtendedJob>,
     // past jobs are indexed with job_id (u32)
     past_jobs: HashMap<u32, ExtendedJob>,
@@ -105,6 +108,7 @@ impl ExtendedChannel {
             nominal_hashrate,
             version_rolling,
             future_jobs: HashMap::new(),
+            future_job_order: VecDeque::new(),
             active_job: None,
             past_jobs: HashMap::new(),
             stale_jobs: HashMap::new(),
@@ -221,6 +225,8 @@ impl ExtendedChannel {
     }
 
     /// Returns an iterator over all future jobs for this channel.
+    ///
+    /// At most [`MAX_FUTURE_JOBS`] jobs are kept (oldest evicted first).
     pub fn get_future_jobs(&self) -> impl Iterator<Item = (&u32, &ExtendedJob)> + '_ {
         self.future_jobs.iter()
     }
@@ -293,6 +299,8 @@ impl ExtendedChannel {
     ///
     /// - If [`NewExtendedMiningJob::min_ntime`](mining_sv2::NewExtendedMiningJob::min_ntime) is empty, the job is considered a future job and
     ///   added to the future jobs list (see [`get_future_jobs`](ExtendedChannel::get_future_jobs)).
+    ///   At most [`MAX_FUTURE_JOBS`] future jobs are kept: storing a new one beyond that limit
+    ///   evicts the oldest.
     /// - Otherwise, the job is activated and previous active job moves to the past jobs list.
     pub fn on_new_extended_mining_job(
         &mut self,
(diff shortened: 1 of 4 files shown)


sv2/channels-sv2/src/client/standard.rs
```diff
@@ -278,8 +284,19 @@ impl StandardChannel {
...
+                    .insert(job_id, (new_mining_job, self.target))
+                    .is_none()
+                {
+                    self.future_job_order.push_back(job_id);
+                }
+                while self.future_jobs.len() > MAX_FUTURE_JOBS {
+                    if let Some(evicted_job_id) = self.future_job_order.pop_front() {
+                        self.future_jobs.remove(&evicted_job_id);
+                    }
+                }
```
bit-aloo, line 307: Simpler

```suggestion
           let job_id = new_mining_job.job_id;
          
          if self
              .future_jobs
              .insert(job_id, (new_mining_job, self.target))
              .is_none()
          {
              self.future_job_order.push_back(job_id);
          
              if self.future_jobs.len() > MAX_FUTURE_JOBS {
                  if let Some(job_id) = self.future_job_order.pop_front() {
                      self.future_jobs.remove(&job_id);
                  }
              }
          }
```

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -75,6 +75,9 @@ pub struct ExtendedChannel {
     version_rolling: bool,
     // future jobs are indexed with job_id (u32)
     future_jobs: HashMap<u32, ExtendedJob>,
+    // insertion order of future_jobs, used to evict the oldest entry once MAX_FUTURE_JOBS is
+    // reached
+    future_job_order: VecDeque<u32>,
```
bit-aloo, line 81: I was thinking maybe we can make `future_job` BtreeMap (because of its property of keeping keys sorted), but i guess thats not possible because we can't guarantee the job_id to be monotonically increasing?

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -322,14 +330,26 @@ impl ExtendedChannel {
...
+                        ),
+                    )
+                    .is_none()
+                {
+                    self.future_job_order.push_back(job_id);
+                }
+                while self.future_jobs.len() > MAX_FUTURE_JOBS {
+                    if let Some(evicted_job_id) = self.future_job_order.pop_front() {
+                        self.future_jobs.remove(&evicted_job_id);
+                    }
```
bit-aloo, line 359: Same simplification as mentioned before.

bit-aloo (commented):
How do we act on app, in case a template is not found corresponding to a job_id?

plebhash:
> How do we act on app, in case a template is not found corresponding to a job_id?

This PR is touching client-side logic only. There's simply no template to match against `job_id`s here.

A more coherent question would be: what happens if a server sends a `SetNewPrevHash` for a `job_id` that has already been evicted?

In that case, a `JobIdNotFound` error is thrown, and application should act accordingly (probably fallback?)

This is an explicit trade-off, and it is inherent to the fundamental problem we're trying to solve here:

as a client, we cannot expect a server to be able to create an infinite number of future jobs and then activate an extremely old one. As much as the spec arguably allows for that, in a practical sense we have to be defensive with regards to bounding the client's memory allocation.

This PR is arbitrarily setting `MAX_FUTURE_JOBS` as 16. Could be bigger, or smaller. But the main point is that we need a limit.

sv2/channels-sv2/src/client/extended.rs
plebhash: the spirit of this PR is to defend against bad server implementations

> but i guess thats not possible because we can't guarantee the job_id to be monotonically increasing

so yeah, we shouldn't count on this

sv2/channels-sv2/src/client/extended.rs
plebhash: done

sv2/channels-sv2/src/client/standard.rs
plebhash: done

bit-aloo (approved):
ACK

GitGab19 (commented):
Clanker review:

```
### Replaced job IDs retain their old eviction position

When an existing `job_id` is replaced, `HashMap::insert` returns `Some`, so the ID is not moved to the back of `future_job_order`.

For example:

1. Fill the store with IDs `0..15`.
2. Receive an updated job for ID `0`.
3. Receive a new job with ID `16`.
4. ID `0` is evicted, even though it was the second-most-recently received job.
5. A subsequent `SetNewPrevHash` referencing ID `0` returns `JobIdNotFound`.

The protocol does not require job IDs to be monotonically increasing or unique across received messages. Consider removing an existing ID from the deque and pushing it to the back whenever its value is replaced. Since the queue is capped at 16 entries, the additional linear search is negligible.

This applies to the group, standard, and extended channel implementations. A regression test covering “fill → replace oldest ID → insert another → activate replaced ID” would catch the issue.
```

plebhash:
> Clanker review:
> 
> ```
> ### Replaced job IDs retain their old eviction position
> 
> When an existing `job_id` is replaced, `HashMap::insert` returns `Some`, so the ID is not moved to the back of `future_job_order`.
> 
> For example:
> 
> 1. Fill the store with IDs `0..15`.
> 2. Receive an updated job for ID `0`.
> 3. Receive a new job with ID `16`.
> 4. ID `0` is evicted, even though it was the second-most-recently received job.
> 5. A subsequent `SetNewPrevHash` referencing ID `0` returns `JobIdNotFound`.
> 
> The protocol does not require job IDs to be monotonically increasing or unique across received messages. Consider removing an existing ID from the deque and pushing it to the back whenever its value is replaced. Since the queue is capped at 16 entries, the additional linear search is negligible.
> 
> This applies to the group, standard, and extended channel implementations. A regression test covering “fill → replace oldest ID → insert another → activate replaced ID” would catch the issue.
> ```

addressed

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -76,6 +76,9 @@ pub struct ExtendedChannel {
     version_rolling: bool,
     // future jobs are indexed with job_id (u32)
     future_jobs: HashMap<u32, ExtendedJob>,
+    // insertion order of future_jobs, used to evict the oldest entry once MAX_FUTURE_JOBS is
+    // reached
```
```suggestion
    // Future job IDs ordered by receipt, oldest at the front and newest at the back.
    // Replaced IDs move to the back; overflow evicts from the front.
```

sv2/channels-sv2/src/client/group.rs
```diff
@@ -29,6 +31,9 @@ pub struct GroupChannel {
     channel_ids: HashSet<u32>,
     /// Future jobs, indexed by job_id, waiting to be activated
     future_jobs: HashMap<u32, NewExtendedMiningJobOwned>,
+    /// Insertion order of `future_jobs`, used to evict the oldest entry once
+    /// [`MAX_FUTURE_JOBS`] is reached
```
GitGab19:
```suggestion
    // Future job IDs ordered by receipt, oldest at the front and newest at the back.
    // Replaced IDs move to the back; overflow evicts from the front.
```

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -58,6 +58,9 @@ pub struct StandardChannel {
     target: Target,
     nominal_hashrate: f32,
     future_jobs: HashMap<u32, StandardJob>,
+    // insertion order of future_jobs, used to evict the oldest entry once MAX_FUTURE_JOBS is
+    // reached
```
GitGab19:
```suggestion
    // Future job IDs ordered by receipt, oldest at the front and newest at the back.
    // Replaced IDs move to the back; overflow evicts from the front.
```

GitGab19 (approved):
ACK

A non-blocking comment improvement.

sv2/channels-sv2/src/client/extended.rs
plebhash: done

sv2/channels-sv2/src/client/group.rs
plebhash: done

sv2/channels-sv2/src/client/standard.rs
plebhash: done

GitGab19 (approved):

PR #2270, opened 2026-08-04
Remove aes256gcm from noise_sv2

### LINKED ISSUE #2247: Remove GenericCipher and Aes256GCM
We don't use Aes256GCM and once we remove that, we would not require the GenericCipher as well.

### LINKED ISSUE #2269: Remove unused c1 and c2 fields from Noise handshake roles
Both Initiator and Responder store c1 and c2 as optional cipher fields. These fields appear to be unused:

* They are initialized to None.
* They are never assigned Some(...).
* They are reset to None after the transport ciphers are derived.
* Their key-erasure branches are therefore permanent no-ops.

The actual c1 and c2 transport ciphers are created as local variables at the end of the handshake and moved directly into NoiseCodec/NoiseEngine. Removing the struct fields should reduce unnecessary state without changing cipher derivation or directionality.

bit-aloo:
closes: #2247
closes: #2269

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -786,7 +786,7 @@ dependencies = [
 
 [[package]]
 name = "noise_sv2"
-version = "1.4.2"
+version = "2.0.0"
 dependencies = [
  "chacha20poly1305",
  "criterion",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -788,7 +788,6 @@ dependencies = [
 name = "noise_sv2"
 version = "1.4.2"
 dependencies = [
- "aes-gcm",
  "chacha20poly1305",
  "criterion",
  "quickcheck",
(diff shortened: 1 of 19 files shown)


plebhash (approved):

GitGab19:
Clanker finding:
```
### Regenerate the fuzz lockfile

`fuzz/Cargo.lock` still records `noise_sv2` 1.4.2 with `aes-gcm`. Please regenerate it so locked fuzz builds use `noise_sv2` 2.0.0 and the updated dependency graph.
```

bit-aloo:
> Clanker finding:
> 
> ```
> ### Regenerate the fuzz lockfile
> 
> `fuzz/Cargo.lock` still records `noise_sv2` 1.4.2 with `aes-gcm`. Please regenerate it so locked fuzz builds use `noise_sv2` 2.0.0 and the updated dependency graph.
> ```

Nice catch. Pushed.

GitGab19 (approved):
ACK

PR #2287, opened 2026-08-06
`common_messages_sv2`: add protocol-version-mismatch `SetupConnectionError` `error_code`

plebhash:
companion https://github.com/stratum-mining/sv2-apps/pull/695


bit-aloo (approved):
Ah!! we never really checked for it. Makes sense.

GitGab19 (approved):

PR #2290, opened 2026-08-09
`channels_sv2`: bound job-storage retention (future templates, past jobs, replaced group jobs)

### LINKED ISSUE #2289: `channels_sv2`: need to bound job-storage retention (future templates, past jobs, replaced group jobs)
Follow-up to #2266 (client `future_jobs` bound) and #2267 (`template_id`-reuse orphan drop). 

The same unbounded retention under attacker-controlled message streams pattern persists in the remaining job-storage collections. In every case the eviction path is gated on a `SetNewPrevHash` / activation that a malicious upstream can simply withhold, and each retained entry can carry large wire-controlled buffers (`B064K` coinbase outputs, up to 255 merkle hashes), so thousands of messages retain ~GB-scale heap.

### 1. Server future-template storage has no cardinality bound
`JobStore::add_future_job` (server/jobs/job_store.rs:65-75) retains every future template's full job keyed by the upstream-controlled template_id.

#2267 fixed reuse of an ID, but distinct IDs are unbounded. All three server channel types route through it: 

- `server/group.rs`: `on_new_template`
- `server/standard.rs: `on_new_template`
- `server/extended.rs`: `on_new_template`+ custom-job path.

A malicious/compromised Template Distribution peer streams future templates, withholds `SetNewPrevHash`, and grows both `future_jobs` and `future_template_to_job_id` without limit. 

Fix: cap + evict-oldest in `JobStore::add_future_job` (insertion-order tracking, same pattern as #2266) — one change covers all three channels.

### 2. Client past_jobs has no bound

`client/standard.rs:297-300` (`store_new_mining_job`) moves the replaced active job into past_jobs — unbounded, cleared only on `SetNewPrevHash`.

`client/extended.rs` has the same defect at three sites (`on_new_extended_mining_job`, `on_set_custom_mining_job_success`, `on_set_new_prev_hash`).

A malicious upstream pool streams immediately-active jobs (`min_ntime` present) and forces one retained job per message, no PoW required.

(`client/group.rs` is unaffected — it tracks no past jobs.) 

Fix: mirror #2266's machinery — `past_job_order: VecDeque<u32>` + evict-oldest past the cap.

Note past jobs exist for late-share validation, so the cap must stay nonzero; evicted-job shares degrade to InvalidJobId, acceptable within a single tip window.

### 3. Group channels retain replaced active jobs they never need

`server/group.rs:261-279` (non-future `on_new_template` branch) calls `JobStore::add_active_job` (`job_store.rs:78-85`), which moves the previous active job into `past_jobs`.

Group channels never validate shares — the module docs explicitly say they don't need past/stale tracking — so this history is pure retention.

Fix: add `JobStore::replace_active_job` (drops the previous active instead of retaining it) and use it in the group channel.

### 4. Server standard/extended `past_jobs` (unfiled sibling)

Standard and extended server channels also grow `past_jobs` via `add_active_job` (non-future templates / `SetCustomMiningJob`), cleared only on activation. Unlike groups they do validate shares against past jobs, so they need the bounded-eviction treatment of item 2, not the drop of item 3.

### notes
- Bounding `past_jobs` transitively bounds `stale_jobs` everywhere: stale sets are snapshots of `past_jobs` at tip transitions (stale = take(past) / past.clone()).
- Legitimate cardinality is tiny (~1 future job/template and a handful of past jobs per tip), so a small cap is functionally free.
- Cap value: reuse `MAX_FUTURE_JOBS` (16, defined in client/mod.rs) as a shared bound, or introduce a separate `MAX_PAST_JOBS` — maintainer's call. Server-side constant can live next to `JobStore` (crate-private).
- Trust boundaries differ per item (mining-protocol upstream for 2; template-distribution peer for 1, 3, 4) — worth stating so reviewers don't conflate them.

### regression tests

PoCs exist for items 1–3 (10k-flood tests asserting bounded retention in group, standard-client; mirror for the remaining channels). Item 3 additionally asserts `replace_active_job` leaves `past_jobs` empty in group channels.

PoCs available on:
- https://github.com/project-loupe/audit-stratum/issues/101
- https://github.com/project-loupe/audit-stratum/issues/100
- https://github.com/project-loupe/audit-stratum/issues/97

plebhash:
close #2289
close https://github.com/project-loupe/audit-stratum/issues/101
close https://github.com/project-loupe/audit-stratum/issues/100
close https://github.com/project-loupe/audit-stratum/issues/97
close https://github.com/project-loupe/audit-stratum/issues/127
close https://github.com/project-loupe/audit-stratum/issues/134
close https://github.com/project-loupe/audit-stratum/issues/100
close https://github.com/project-loupe/audit-stratum/issues/117
close https://github.com/project-loupe/audit-stratum/issues/101
close https://github.com/project-loupe/audit-stratum/issues/126
close https://github.com/project-loupe/audit-stratum/issues/136
close https://github.com/project-loupe/audit-stratum/issues/129

### DIFF
--- a/sv2/channels-sv2/src/server/jobs/job_store.rs
+++ b/sv2/channels-sv2/src/server/jobs/job_store.rs
@@ -118,6 +118,11 @@ impl<T: Job> JobStore<T> {
             }
         }
 
+        // a dropped job (replaced template ID or evicted-oldest) may have been the last one
+        // holding a retired extranonce prefix alive; release such slots now rather than at the
+        // next chain transition, which a peer can withhold
+        self.prune_retired_extranonce_prefixes();
+
         new_job_id
     }
 
@@ -138,6 +143,10 @@ impl<T: Job> JobStore<T> {
         if self.past_jobs.len() > MAX_PAST_JOBS {
             if let Some(evicted_job_id) = self.past_job_order.pop_front() {
                 self.past_jobs.remove(&evicted_job_id);
+                // the evicted job may have been the last one holding a retired extranonce
+                // prefix alive; release such slots now rather than at the next chain
+                // transition, which a peer can withhold
+                self.prune_retired_extranonce_prefixes();
                 return Some(evicted_job_id);
             }
         }
@@ -394,6 +403,52 @@ mod tests {
             store.get_active_job().map(|job| job.get_job_id()),
             Some(flood_size - 1)
         );
+    }
+
+    struct PrefixedJob {
+        job_id: u32,
+        prefix: Vec<u8>,
+    }
+
+    impl Job for PrefixedJob {
+        fn get_job_id(&self) -> u32 {
+            self.job_id
+        }
+
+        fn get_extranonce_prefix(&self) -> &[u8] {
+            &self.prefix
+        }
+
+        fn activate(&mut self, _prev_hash_header_timestamp: u32) {}
+    }
+
+    #[test]
+    fn dropped_jobs_release_retired_extranonce_prefixes() {
+        let mut store = JobStore::new();
+        let old_prefix = vec![1u8];
+
+        // a future job created under the old prefix keeps the retired prefix alive
+        store.add_future_job(
+            1,
+            PrefixedJob {
+                job_id: 1,
+                prefix: old_prefix.clone(),
+            },
+        );
+        store.retire_extranonce_prefix(ExtranoncePrefix::from_wire(old_prefix).unwrap());
+        assert_eq!(store.retired_extranonce_prefixes.len(), 1);
+
+        // replacing the future job under the same template ID drops the last job referencing
+        // the retired prefix, so its slot must be released without waiting for a chain
+        // transition
+        store.add_future_job(
+            1,
+            PrefixedJob {
+                job_id: 2,
+                prefix: vec![2u8],
+            },
+        );
+        assert!(store.retired_extranonce_prefixes.is_empty());
     }
 
     #[test]
--- a/sv2/channels-sv2/src/server/jobs/job_store.rs
+++ b/sv2/channels-sv2/src/server/jobs/job_store.rs
@@ -26,10 +26,11 @@ pub(crate) const MAX_FUTURE_JOBS: usize = 16;
 
 /// Maximum number of past jobs a server channel retains under the current chain tip.
 ///
-/// Past jobs exist for late-share validation, so the cap must stay nonzero; shares against an
-/// evicted job degrade to `InvalidJobId`, acceptable within a single tip window. Bounding the map
-/// prevents a malicious template-distribution peer from exhausting server memory by streaming
-/// non-future templates while withholding `SetNewPrevHash`.
+/// Past jobs exist for late-share validation, so the cap must stay nonzero. A share against an
+/// evicted job is rejected as `InvalidJobId` even though it would otherwise have been accepted
+/// and credited — a bounded loss of creditable work, the price of bounding memory against a
+/// malicious template-distribution peer streaming non-future templates while withholding
+/// `SetNewPrevHash`.
 ///
 /// 50 buys ample headroom over the reachable submit depth (a miner abandons a job once the next
 /// one arrives, so late shares realistically target the last 1-2 jobs) at a small measured
@@ -127,8 +128,9 @@ impl<T: Job> JobStore<T> {
     }
 
     /// Moves the active job (if any) into past jobs, evicting the oldest past job beyond
-    /// `MAX_PAST_JOBS`. A share against an evicted job degrades to `InvalidJobId` instead of
-    /// `Stale`, acceptable within a single tip window.
+    /// `MAX_PAST_JOBS`. A share against an evicted job is rejected as `InvalidJobId` even though
+    /// it would otherwise have been accepted and credited: a bounded loss of creditable work,
+    /// the price of bounding memory under a hostile upstream.
     ///
     /// Returns the evicted job's ID, if any, so callers can drop metadata they key by job ID.
     fn retire_active_to_past(&mut self) -> Option<u32> {
@@ -153,11 +155,31 @@ impl<T: Job> JobStore<T> {
         None
     }
 
+    /// Moves the active job (if any) into past jobs without the `MAX_PAST_JOBS` cap and without
+    /// pruning retired extranonce prefixes.
+    ///
+    /// Only for tip transitions, where past jobs immediately drain into stale jobs: `stale_jobs`
+    /// stays bounded at `MAX_PAST_JOBS + 1`, no job is dropped from the stale set (which would
+    /// misclassify its late shares as `InvalidJobId` instead of `Stale`), and no prune runs
+    /// while an in-flight future job is outside every collection (which would release its
+    /// retired extranonce prefix while the job goes on to accept shares under it).
+    fn retire_active_to_past_uncapped(&mut self) {
+        if let Some(active_job) = self.active_job.take() {
+            let job_id = active_job.get_job_id();
+            self.past_jobs.insert(job_id, active_job);
+
+            // a replaced job_id moves to the back of the eviction order
+            self.past_job_order.retain(|id| *id != job_id);
+            self.past_job_order.push_back(job_id);
+        }
+    }
+
     /// Adds an active job, moving the previous active job (if any) to past jobs.
     ///
     /// At most `MAX_PAST_JOBS` past jobs are kept: retiring one beyond that limit evicts the
-    /// oldest, and its ID is returned so callers can drop metadata they key by job ID (e.g. the
-    /// per-job target mapping of standard and extended channels).
+    /// oldest (giving up shares that would still have been creditable), and its ID is returned
+    /// so callers can drop metadata they key by job ID (e.g. the per-job target mapping of
+    /// standard and extended channels).
     pub fn add_active_job(&mut self, job: T) -> Option<u32> {
         // Move currently active job to past jobs (so it can be marked as stale)
         let evicted_job_id = self.retire_active_to_past();
@@ -219,8 +241,11 @@ impl<T: Job> JobStore<T> {
                 return false;
             };
 
-        // Move currently active job to past jobs (so it can be marked as stale)
-        self.retire_active_to_past();
+        // Move currently active job to past jobs (so it can be marked as stale). The
+        // retirement is uncapped: past jobs drain into stale jobs below, so the displaced job
+        // must not push another one out of the stale set, and no prune may run while the
+        // in-flight future job is outside every collection.
+        self.retire_active_to_past_uncapped();
 
         // Activate the future job
         future_job.activate(prev_hash_header_timestamp);
@@ -234,12 +259,13 @@ impl<T: Job> JobStore<T> {
         true
     }
 
-    /// Moves the active job (if any) into past jobs.
+    /// Moves the active job (if any) into past jobs, without the `MAX_PAST_JOBS` cap.
     ///
-    /// At most `MAX_PAST_JOBS` past jobs are kept: retiring one beyond that limit evicts the
-    /// oldest, and its ID is returned so callers can drop metadata they key by job ID.
-    pub fn deactivate_job(&mut self) -> Option<u32> {
-        self.retire_active_to_past()
+    /// Only for tip transitions, right before [`Self::mark_past_jobs_as_stale`] drains past
+    /// jobs into stale jobs: the displaced job must go stale with the rest, not push another
+    /// job out of the stale set.
+    pub fn deactivate_job(&mut self) {
+        self.retire_active_to_past_uncapped();
     }
 
     /// Marks all past jobs as stale so shares can be rejected with the proper error code.
@@ -420,6 +446,69 @@ mod tests {
         }
 
         fn activate(&mut self, _prev_hash_header_timestamp: u32) {}
+    }
+
+    #[test]
+    fn tip_transition_moves_displaced_and_all_past_jobs_to_stale() {
+        let mut store = JobStore::new();
+
+        // fill past jobs to the cap, plus an active job
+        for job_id in 0..=MAX_PAST_JOBS as u32 {
+            store.add_active_job(DummyJob { job_id });
+        }
+
+        let future_job_id = 100;
+        store.add_future_job(
+            1,
+            DummyJob {
+                job_id: future_job_id,
+            },
+        );
+        assert!(store.activate_future_job(1, 0));
+
+        // the displaced active job and every past job must land in the stale set: dropping
+        // any of them would misclassify its late shares as InvalidJobId instead of Stale
+        for job_id in 0..=MAX_PAST_JOBS as u32 {
+            assert!(store.get_stale_job(job_id).is_some());
+        }
+        assert_eq!(store.stale_jobs.len(), MAX_PAST_JOBS + 1);
+        assert!(store.past_jobs.is_empty());
+        assert_eq!(
+            store.get_active_job().map(|job| job.get_job_id()),
+            Some(future_job_id)
+        );
+    }
+
+    #[test]
+    fn activation_keeps_retired_prefix_of_activated_job() {
+        let mut store = JobStore::new();
+        let old_prefix = vec![1u8];
+
+        // a future job created under the old prefix, which is then rotated out
+        store.add_future_job(
+            1,
+            PrefixedJob {
+                job_id: 100,
+                prefix: old_prefix.clone(),
+            },
+        );
+        store.retire_extranonce_prefix(ExtranoncePrefix::from_wire(old_prefix).unwrap());
+        assert_eq!(store.retired_extranonce_prefixes.len(), 1);
+
+        // fill past jobs to the cap with jobs under the new prefix, so that a capped
+        // retirement during activation would evict (and prune) mid-flight
+        for job_id in 0..=MAX_PAST_JOBS as u32 {
+            store.add_active_job(PrefixedJob {
+                job_id,
+                prefix: vec![2u8],
+            });
+        }
+
+        assert!(store.activate_future_job(1, 0));
+
+        // the activated job is the only remaining reference to the retired prefix; its slot
+        // must stay reserved while the job keeps accepting shares under those bytes
+        assert_eq!(store.retired_extranonce_prefixes.len(), 1);
     }
 
     #[test]
--- a/sv2/channels-sv2/src/server/jobs/job_store.rs
+++ b/sv2/channels-sv2/src/server/jobs/job_store.rs
@@ -24,6 +24,18 @@ use crate::extranonce_manager::ExtranoncePrefix;
 /// the oldest future job is evicted.
 pub(crate) const MAX_FUTURE_JOBS: usize = 16;
 
+/// Maximum number of past jobs a server channel retains under the current chain tip.
+///
+/// Past jobs exist for late-share validation, so the cap must stay nonzero; shares against an
+/// evicted job degrade to `InvalidJobId`, acceptable within a single tip window. Bounding the map
+/// prevents a malicious template-distribution peer from exhausting server memory by streaming
+/// non-future templates while withholding `SetNewPrevHash`.
+///
+/// 50 buys ample headroom over the reachable submit depth (a miner abandons a job once the next
+/// one arrives, so late shares realistically target the last 1-2 jobs) at a small measured
+/// memory cost — see the load-test data in PR #2290.
+pub(crate) const MAX_PAST_JOBS: usize = 50;
+
 /// Internal implementation for tracking mining job states in SV2 server channels.
 ///
 /// Maintains collections for future, active, past, and stale jobs, and tracks template-to-job ID
@@ -39,6 +51,9 @@ pub(crate) struct JobStore<T: Job> {
     active_job: Option<T>,
     // Past jobs are indexed with job_id (u32)
     past_jobs: HashMap<u32, T>,
+    // Past job IDs ordered by retirement, oldest at the front and newest at the back.
+    // Replaced IDs move to the back; overflow evicts from the front.
+    past_job_order: VecDeque<u32>,
     // Stale jobs are indexed with job_id (u32)
     stale_jobs: HashMap<u32, T>,
     // Extranonce prefixes rotated out of the channel that are still referenced by at least one job
@@ -56,6 +71,7 @@ impl<T: Job> JobStore<T> {
             future_jobs: HashMap::new(),
             active_job: None,
             past_jobs: HashMap::new(),
+            past_job_order: VecDeque::new(),
             stale_jobs: HashMap::new(),
             retired_extranonce_prefixes: Vec::new(),
         }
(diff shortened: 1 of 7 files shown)


GitGab19 (commented):
Clanker review:

## Review findings

I think this needs changes before merging. The main eviction logic is sound, but two resource-retention paths remain.

### 1. High: `job_id_to_target` remains unbounded

`JobStore::retire_active_to_past` evicts the oldest past job, but the corresponding entries in the server channels’ `job_id_to_target` maps are never removed:

- [Standard channel insertion](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/standard.rs#L486-L491)
- [Extended channel insertion](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/extended.rs#L512-L517)
- [Past-job eviction](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L120-L138)

These maps are cleared only during a chain-tip transition. Extending the existing 10,000-job regression test showed that `past_jobs` was correctly limited to 16 entries, while `job_id_to_target` still contained 10,000 entries. Therefore, an attacker-controlled non-future-template stream can still cause unbounded memory growth.

`JobStore` should report the IDs of evicted jobs so the caller can remove their target metadata, or the target should be stored alongside the job. The regression tests should also assert that target metadata is bounded to the active job plus `MAX_PAST_JOBS`.

### 2. Medium: eviction can strand retired extranonce allocator slots

The new future- and past-job eviction paths remove jobs without calling `prune_retired_extranonce_prefixes`:

- [Future-job eviction](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L93-L115)
- [Past-job eviction](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L120-L138)
- [Current pruning logic](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L251-L270)

If an evicted job was the last remaining job using a rotated-out extranonce prefix, the owning `ExtranoncePrefix` remains in `retired_extranonce_prefixes` until a chain transition. A peer withholding that transition can therefore prevent allocator slots from being released.

A reproduction using a two-slot allocator, one prefix rotation, and enough active-job replacements to evict the old job left both slots allocated when one should have been released.

Please prune retired prefixes after future- and past-job removals, including replacement of a future job with the same template ID, and add a regression test covering allocator release after eviction.

plebhash:
> Clanker review:
> ## Review findings
> 
> I think this needs changes before merging. The main eviction logic is sound, but two resource-retention paths remain.
> ### 1. High: `job_id_to_target` remains unbounded
> 
> `JobStore::retire_active_to_past` evicts the oldest past job, but the corresponding entries in the server channels’ `job_id_to_target` maps are never removed:
> 
>     * [Standard channel insertion](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/standard.rs#L486-L491)
> 
>     * [Extended channel insertion](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/extended.rs#L512-L517)
> 
>     * [Past-job eviction](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L120-L138)
> 
> 
> These maps are cleared only during a chain-tip transition. Extending the existing 10,000-job regression test showed that `past_jobs` was correctly limited to 16 entries, while `job_id_to_target` still contained 10,000 entries. Therefore, an attacker-controlled non-future-template stream can still cause unbounded memory growth.
> 
> `JobStore` should report the IDs of evicted jobs so the caller can remove their target metadata, or the target should be stored alongside the job. The regression tests should also assert that target metadata is bounded to the active job plus `MAX_PAST_JOBS`.
> ### 2. Medium: eviction can strand retired extranonce allocator slots
> 
> The new future- and past-job eviction paths remove jobs without calling `prune_retired_extranonce_prefixes`:
> 
>     * [Future-job eviction](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L93-L115)
> 
>     * [Past-job eviction](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L120-L138)
> 
>     * [Current pruning logic](https://github.com/stratum-mining/stratum/blob/456fb817e369d558507d42b63d18f341e82cd96a/sv2/channels-sv2/src/server/jobs/job_store.rs#L251-L270)
> 
> 
> If an evicted job was the last remaining job using a rotated-out extranonce prefix, the owning `ExtranoncePrefix` remains in `retired_extranonce_prefixes` until a chain transition. A peer withholding that transition can therefore prevent allocator slots from being released.
> 
> A reproduction using a two-slot allocator, one prefix rotation, and enough active-job replacements to evict the old job left both slots allocated when one should have been released.
> 
> Please prune retired prefixes after future- and past-job removals, including replacement of a future job with the same template ID, and add a regression test covering allocator release after eviction.

should be addressed, 2 new commits

GitGab19:
Second pass of clanker review:

Solid PR overall — I audited the map/deque bookkeeping on every mutation path and it holds up, evicted-ID handling is correct at all call sites, and `no_std` builds clean. Two real issues and some nits:

### 🔴 Retired extranonce prefix released while still referenced
`server/jobs/job_store.rs`: `activate_future_job` pops the job from `future_jobs` **before** `retire_active_to_past()`, so the eviction-triggered `prune_retired_extranonce_prefixes()` runs while the in-flight job is in neither `future_jobs`, `active_job`, nor `past_jobs`. A retired prefix whose only remaining reference is that job gets dropped and its allocator slot freed — while the job becomes active and keeps validating shares under those prefix bytes (possible extranonce collision with another channel). Repro: future job under P1 → `set_extranonce_prefix` to P2 (P1 retired) → 17+ non-future templates → `SetNewPrevHash`. Pre-PR the only prune on this path ran *after* activation. **Fix:** prune only after `active_job` is assigned.

### 🟠 Tip-transition eviction drops one job from the stale set
At every tip transition the displaced active job is retired through the *capped* past path right before past drains into stale — so with past at cap, the oldest past job lands in neither set and its late share is rejected `invalid-job-id` instead of `stale-share` (the exact misclassification the existing regression tests guard against). The eviction buys nothing: past is emptied on the next statement. Sites: `job_store.rs` `activate_future_job` (retire at 219 → stale at 228), `server/extended.rs` 612→615 and 690→695, `client/extended.rs` 580→588 and 534→538, `client/standard.rs` 376→384. **Fix:** rotate past→stale before retiring the displaced job (stale stays bounded at `MAX_PAST_JOBS + 1`).

### 🟡 Doc wording
`job_store.rs:126` (+ client copies) say eviction degrades `Stale` → `InvalidJobId` "within a single tip window" — but within a tip window past-job shares validate like active ones and are *accepted/credited* (`job_id_to_target` retains past entries). The real degradation is **Accepted → InvalidJobId**, i.e. lost creditable work. Worth stating honestly when sizing `MAX_PAST_JOBS`.

### Nits
- The map+`VecDeque` bounded-insert idiom is now hand-copied 6× (client ext/std past+future, `add_future_job`, `retire_active_to_past`) plus 9 paired `clear()` sites — a crate-private `BoundedJobMap<K,V>` would make the sync invariant structural.
- `retire_active_to_past`'s `past_job_order.retain(...)` is dead server-side (job IDs are strictly monotonic per channel); the client copies are the load-bearing ones.
- `activate_future_job_replacing_active` duplicates the delegate's lookup as a check-then-act; delegate-first + `stale_jobs.clear()` on success is equivalent and simpler.
- `deactivate_job`'s new `Option<u32>` return has no consumers (only caller discards it and clears `job_id_to_target` wholesale).
- `add_future_job` prunes retired prefixes even on the no-drop path, where pruning can't free anything — condition it on the two drop branches.
- The six server-side 10k-iteration flood tests run the full job factory per iteration; ~`2*MAX+2` iterations prove the same invariant (the client/DummyJob floods are cheap, fine as-is).
- Eight near-identical ~25-line `NewTemplate` test literals — a small `test_template(id, future)` helper would collapse them.

plebhash:
> Second pass of clanker review:
> 
> Solid PR overall — I audited the map/deque bookkeeping on every mutation path and it holds up, evicted-ID handling is correct at all call sites, and `no_std` builds clean. Two real issues and some nits:
> ### 🔴 Retired extranonce prefix released while still referenced
> 
> `server/jobs/job_store.rs`: `activate_future_job` pops the job from `future_jobs` **before** `retire_active_to_past()`, so the eviction-triggered `prune_retired_extranonce_prefixes()` runs while the in-flight job is in neither `future_jobs`, `active_job`, nor `past_jobs`. A retired prefix whose only remaining reference is that job gets dropped and its allocator slot freed — while the job becomes active and keeps validating shares under those prefix bytes (possible extranonce collision with another channel). Repro: future job under P1 → `set_extranonce_prefix` to P2 (P1 retired) → 17+ non-future templates → `SetNewPrevHash`. Pre-PR the only prune on this path ran _after_ activation. **Fix:** prune only after `active_job` is assigned.
> ### 🟠 Tip-transition eviction drops one job from the stale set
> 
> At every tip transition the displaced active job is retired through the _capped_ past path right before past drains into stale — so with past at cap, the oldest past job lands in neither set and its late share is rejected `invalid-job-id` instead of `stale-share` (the exact misclassification the existing regression tests guard against). The eviction buys nothing: past is emptied on the next statement. Sites: `job_store.rs` `activate_future_job` (retire at 219 → stale at 228), `server/extended.rs` 612→615 and 690→695, `client/extended.rs` 580→588 and 534→538, `client/standard.rs` 376→384. **Fix:** rotate past→stale before retiring the displaced job (stale stays bounded at `MAX_PAST_JOBS + 1`).
> ### 🟡 Doc wording
> 
> `job_store.rs:126` (+ client copies) say eviction degrades `Stale` → `InvalidJobId` "within a single tip window" — but within a tip window past-job shares validate like active ones and are _accepted/credited_ (`job_id_to_target` retains past entries). The real degradation is **Accepted → InvalidJobId**, i.e. lost creditable work. Worth stating honestly when sizing `MAX_PAST_JOBS`.
> ### Nits
> 
>     * The map+`VecDeque` bounded-insert idiom is now hand-copied 6× (client ext/std past+future, `add_future_job`, `retire_active_to_past`) plus 9 paired `clear()` sites — a crate-private `BoundedJobMap<K,V>` would make the sync invariant structural.
> 
>     * `retire_active_to_past`'s `past_job_order.retain(...)` is dead server-side (job IDs are strictly monotonic per channel); the client copies are the load-bearing ones.
> 
>     * `activate_future_job_replacing_active` duplicates the delegate's lookup as a check-then-act; delegate-first + `stale_jobs.clear()` on success is equivalent and simpler.
> 
>     * `deactivate_job`'s new `Option<u32>` return has no consumers (only caller discards it and clears `job_id_to_target` wholesale).
> 
>     * `add_future_job` prunes retired prefixes even on the no-drop path, where pruning can't free anything — condition it on the two drop branches.
> 
>     * The six server-side 10k-iteration flood tests run the full job factory per iteration; ~`2*MAX+2` iterations prove the same invariant (the client/DummyJob floods are cheap, fine as-is).
> 
>     * Eight near-identical ~25-line `NewTemplate` test literals — a small `test_template(id, future)` helper would collapse them.

addressed

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -10,23 +10,47 @@
...
+pub(crate) const MAX_FUTURE_JOBS: usize = 16;
+
+/// Maximum number of past jobs a server channel retains under the current chain tip.
+///
+/// Past jobs exist for late-share validation, so the cap must stay nonzero. A share against an
+/// evicted job is rejected as `InvalidJobId` even though it would otherwise have been accepted
+/// and credited — a bounded loss of creditable work, the price of bounding memory against a
+/// malicious template-distribution peer streaming non-future templates while withholding
+/// `SetNewPrevHash`. Size the cap with that trade-off in mind.
+pub(crate) const MAX_PAST_JOBS: usize = 16;
```
gimballock: **We support the bound — any cap beats unbounded retention, and that question looks settled.** Our data speaks to a narrower one: where to sit on the capped spectrum. Two measured points give a direction rather than an optimum.

Same hardware (`r7i.8xlarge`), ~110k connections, held to RSS convergence, cap the only variable:

| | cap=300 | cap=50 |
|---|---|---|
| per-conn RSS | 1.429 MiB | 0.907 MiB (**−36.5%**) |
| share-acceptance deficit | 3,453 (**3.16%**) | 6,514 (**5.86%**) |
| `stale-share` | 3,365 | 6,308 |
| `invalid-job-id` | 4 | 0 |
| peak active channels | 105,701 | 104,697 (noise) |

Lowering the cap trades memory against credited work: 300 → 50 saved 36.5% of per-conn RSS and cost +2.7 points of share-acceptance deficit. Both directions carry a cost, so an optimum exists somewhere — but two points locate a gradient, not a minimum. The cap moved density rather than capacity: 105,701 against 104,697 active channels is noise.

**A way to size it that does not rely on round numbers.** The cap is really a retention window — `MAX_PAST_JOBS` ÷ template rate is the late-share tolerance it buys, so its meaning scales with *your* template cadence rather than ours. Two bounds apply, and they differ sharply. The latency bound is generous: at our ~1 template/s, a cap of 16 tolerates roughly 16 s of in-flight share latency. The structural bound is far tighter than either cap under discussion — a miner stops hashing a job as soon as the next arrives, and the share hop is a zero-delay channel, so reachable submit depth is **~1–2 past jobs**. Against that, 16 already carries about 8× margin, and our own analysis put the smallest defensible cap on this rig at **20**; we picked 50 for headroom, not necessity. **On the eviction path we would therefore expect 16 to be fine.**

**What we cannot yet answer is whether the eviction path is the one that matters.** The yield delta we measured probably is not it: `invalid-job-id` stayed at 0–4 in both arms while the deficit tracked `stale-share`, which points at the stale-set coupling rather than at evicted jobs — and no margin argument covers that mechanism. Our runs also could not test eviction at all, because each simulated miner stopped hashing once its first share was accepted, so every channel submitted exactly one share within seconds of connecting and none was ever validated against an evicted job. That makes our `invalid-job-id` zero vacuous rather than reassuring.

Closing that gap would take continuous hashing at fixed N across a sweep bracketing the structural floor — something like 8 / 16 / 25 / 50, chosen to straddle the ~20 estimate rather than for roundness. That is something we could look into if it would be useful; nothing we have contradicts 16 in the meantime.

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -10,23 +10,47 @@
...
+/// Template Distribution peers control `template_id`, so future jobs are stored under a
+/// peer-controlled key. Bounding the map prevents a malicious or buggy peer from exhausting
+/// server memory by streaming future templates while withholding `SetNewPrevHash`. On overflow,
+/// the oldest future job is evicted.
+pub(crate) const MAX_FUTURE_JOBS: usize = 16;
+
+/// Maximum number of past jobs a server channel retains under the current chain tip.
+///
+/// Past jobs exist for late-share validation, so the cap must stay nonzero. A share against an
+/// evicted job is rejected as `InvalidJobId` even though it would otherwise have been accepted
```
gimballock, line 30: Our data supports the conclusion in the second pass's 🟡 note — this is lost creditable work, not a benign reclassification — but complicates the label. In our runs the loss did not surface as `InvalidJobId`. The `invalid-job-id` rejection counter, labelled by error code, stayed between 0 and 4 in **both** cap arms, while the share-acceptance deficit doubled and tracked `stale-share`.

So the observable symptom of a too-small cap was shares rejected as stale and channels never credited. That matters for testing: a regression test asserting on `InvalidJobId` counts passes whether or not this effect exists. Reproducing it takes continuous share submission across a prev-hash rotation under load.

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -102,33 +236,38 @@ impl<T: Job> JobStore<T> {
...
-            self.past_jobs.insert(active_job.get_job_id(), active_job);
-        }
+        self.retire_active_to_past_uncapped();
     }
 
     /// Marks all past jobs as stale so shares can be rejected with the proper error code.
     pub fn mark_past_jobs_as_stale(&mut self) {
         // Transfer past jobs to stale jobs collection and reset past jobs to empty
         self.stale_jobs = std::mem::take(&mut self.past_jobs);
+        self.past_job_order.clear();
```
gimballock, line 274: This couples the two bounds, and it is our leading candidate for the yield loss we measured on `MAX_PAST_JOBS`: because `stale_jobs = std::mem::take(&mut self.past_jobs)`, the cap transitively bounds the **stale** set as well as the past set.

A smaller cap therefore shrinks the stale window too, so shares in flight across a tip transition can land outside both sets. We could not isolate this from our data — the rejections came back as `stale-share` rather than `invalid-job-id` — so treat it as a hypothesis rather than a finding.

sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -61,27 +87,135 @@ impl<T: Job> JobStore<T> {
...
+
+    /// Moves the active job (if any) into past jobs without the `MAX_PAST_JOBS` cap and without
+    /// pruning retired extranonce prefixes.
+    ///
+    /// Only for tip transitions, where past jobs immediately drain into stale jobs: `stale_jobs`
+    /// stays bounded at `MAX_PAST_JOBS + 1`, no job is dropped from the stale set (which would
+    /// misclassify its late shares as `InvalidJobId` instead of `Stale`), and no prune runs
+    /// while an in-flight future job is outside every collection (which would release its
+    /// retired extranonce prefix while the job goes on to accept shares under it).
+    fn retire_active_to_past_uncapped(&mut self) {
```
gimballock, line 172: Independent corroboration: we hit this same overshoot in our own cap patch before this PR existed. The bypass paths let `past_jobs` transiently reach `MAX_PAST_JOBS + 1`, and we recorded the bound as exactly `MAX + 1`.

This method and its test land on the same number deliberately. Nothing to change — just a note that the invariant holds under sustained load, not only in the unit test.

gimballock (commented):
**This began as scale testing, not as a review of this PR.** We have been running distributed load tests against a vendored pin of this stack, and several defects we hit independently turn out to be the ones this PR and its two review passes already address. For a security fix that convergence is worth reporting: two very different methods — an LLM-assisted audit reading the source, and a fleet of miners hammering a pool on EC2 — landed on the same eviction bugs and the same bounds.

Where we ended up matching you:

| What we hit under load | Finding here |
|---|---|
| `past_jobs` overshoot bounded at exactly `MAX_PAST_JOBS + 1` | `retire_active_to_past_uncapped` and its test assert that bound |
| Evicting by smallest `job_id` rather than insertion order | insertion-ordered `past_job_order` replaces it |
| Share-acceptance loss when the cap shrinks | pass 2 🟡, Accepted → InvalidJobId |
| `job_id_to_target` growing unbounded alongside past jobs | pass 1 🔴 |

**What we can add that the thread does not have yet: a measurement of the cap trade-off.** At fixed N (~110k connections) moving the cap from 300 to 50 doubled the share-acceptance deficit, while the memory saved below 50 is small (~0.07 MiB/conn). **We have no data at 16**, and two points establish a direction rather than an optimum. Framing the cap as a retention window rather than a round number, our earlier analysis puts the reachable submit depth at ~1–2 past jobs, which leaves 16 with roughly 8× margin on the eviction path — so nothing we have contradicts it. The open question is whether eviction is the mechanism that matters. Details are inline on the constant.

One small confirmation while we are here: we re-checked the assumption behind dropping the `past_job_order` dedup, and it holds at this revision — every `job_store` insertion in `group.rs`, `standard.rs`, and `extended.rs` is minted by that channel's own `job_factory`, including the custom-job path, so IDs are monotonic per store.

Context so the numbers are interpretable: pool isolated on its own EC2 box, a fleet of client load-generator boxes, every figure scraped pool-side from local `/metrics`. Our ladder ran 7.5k → **300,012** connections, the last held to RSS (resident set size) convergence across a 3,596 s plateau. We vendor this tree at pin `7fa5aa1` with one local patch making `MAX_PAST_JOBS` configurable, so these are not stock-upstream figures — but the cap was the only variable between arms.

We are downstream users rather than maintainers, so tell us which of this is useful and which is noise.

**Contributions.** We measured the memory saving from bounding past jobs at 36.5%, quantified the yield cost of a smaller cap at +2.7 points of share-acceptance deficit at fixed N, and confirmed the per-store job-ID monotonicity this revision relies on. If a continuously-hashing A/B across 16 / 50 / 300 would help pick the constant, that is something we could look into.

GitGab19 (approved):

plebhash:
thanks @gimballock but I'm having difficulty on how to digest the comments above

on a superficial level I interpreted these as more informational comments that corroborate the PR, rather than a actual requests for change, but the following bits got me slightly confused:

> We are downstream users rather than maintainers, so tell us which of this is useful and which is noise.

> If a continuously-hashing A/B across 16 / 50 / 300 would help pick the constant, that is something we could look into

can you please clarify what you were aiming to communicate with the comments above?

(ideally human generated response please)

gimballock:
I was offering to rerun the scale test with a different sized job buffers to sweep for a best value. 

This buffer holds the outstanding jobs (MAX_PAST_JOBS): The size should be bigger than the rate of jobs submitted divided by the rate jobs are completed to not lose inflight work, but we want it as small as possible to minimize memory footprint. 

The current value you have (16) was picked arbitrarily, we tested some other arbitrary values, 300 and 50, and 50 was better.  More data points are needed to say what the best setting might be.

(not a blocking issue)

plebhash:
thanks @gimballock I have a clear understanding now

also thank you for this analysis, it's very valuable!

I'm switching `MAX_PAST_JOBS` from 16 to 50 as a safety headroom against lost work (while keeping memory footprint small)

a valuable follow-up could consist of:
- gathering empirical data for the optimal value as you suggested
- making `MAX_PAST_JOBS` into a configurable value rather than a constant

although I must say that would be placed as relatively low priority on SRI review pipeline, since we have a lot of work stacking up from audits in other areas of the codebase

gimballock:
I can create the followup PR (unless you would prefer to), feel free to prioritize as needed

plebhash:
> I can create the followup PR (unless you would prefer to), feel free to prioritize as needed

sure go ahead

ideally there should be a new `max_past_jobs: Option<usize>` parameter added to the server and client variants of `ExtendedChannel` and `StandardChannel` objects, to be set via their constructors

that means a companion PR against `sv2-apps` will be needed due to API breaking change

moreover, I wouldn't necessarily completely throw away `MAX_PAST_JOBS` constant... just fine-tune it to the empirically-validated optimal value

then we always do `max_past_jobs.unwrap_or(MAX_PAST_JOBS)` so that the user can opt into the optimal constant by choosing `None` in the constructor

---

for now, merging this as is

PR #2291, opened 2026-08-10
fix rounded Sv1 difficulty target

GitGab19:
Follow-up to https://github.com/stratum-mining/stratum/pull/2227

Closes https://github.com/project-loupe/audit-stratum/issues/96

### DIFF
--- a/stratum-core/stratum-translation/src/sv2_to_sv1.rs
+++ b/stratum-core/stratum-translation/src/sv2_to_sv1.rs
@@ -122,6 +122,18 @@ pub fn build_sv1_set_difficulty_from_sv2_target(target: Target) -> Result<json_r
     Ok(set_target.into())
 }
 
+/// The SV1 difficulty selected by the integer power-of-two rounding policy.
+///
+/// Keeping the rounding decision in the result prevents callers from trying to infer it from the
+/// rounded value, which may itself be below a non-power-of-two rounding threshold.
+enum AdvertisedSv1Difficulty {
+    Unrounded,
+    /// A rounded difficulty equal to `2^exponent`.
+    Rounded {
+        exponent: u32,
+    },
+}
+
 /// Builds an SV1 `mining.set_difficulty` message using integer power-of-two rounding above a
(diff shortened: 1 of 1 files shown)


stratum-core/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -122,6 +122,15 @@ pub fn build_sv1_set_difficulty_from_sv2_target(target: Target) -> Result<json_r
     Ok(set_target.into())
 }
 
+/// The SV1 difficulty selected by the integer power-of-two rounding policy.
+///
+/// Keeping the rounding decision in the result prevents callers from trying to infer it from the
+/// rounded value, which may itself be below a non-power-of-two rounding threshold.
+enum AdvertisedSv1Difficulty {
+    Unrounded(f64),
```
bit-aloo: Considering the places its been used, we don't really need to store f64, in the unrounded variant, as in all those methods we already provide the difficulty.

stratum-core/stratum-translation/src/sv2_to_sv1.rs
```diff
@@ -122,6 +122,15 @@ pub fn build_sv1_set_difficulty_from_sv2_target(target: Target) -> Result<json_r
...
+    Rounded(u64),
```
bit-aloo: Maybe we should store exponents of power of 2, so non power of two is un-representable here.

bit-aloo (commented):
some suggestions

stratum-core/stratum-translation/src/sv2_to_sv1.rs
GitGab19: Done

stratum-core/stratum-translation/src/sv2_to_sv1.rs
GitGab19: Done

bit-aloo (approved):
ACK. Thanks

PR #2293, opened 2026-08-11
speedup integration tests (on companion PRs)

plebhash:
so we can benefit from the speedups of https://github.com/stratum-mining/sv2-apps/pull/698 on companion PRs here as well

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -140,8 +140,10 @@ cargo update -p stratum-core
 echo "✅ Updated Cargo.toml to use local dependencies"
 echo "🏃 Running integration tests..."
 
-# Run the integration tests
-RUST_BACKTRACE=1 RUST_LOG=debug cargo nextest run --nocapture --verbose
+# Failure triage: re-run a single test with captured logs shown live:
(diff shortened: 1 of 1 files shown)


scripts/run-integration-tests.sh
```diff
@@ -140,8 +140,10 @@ cargo update -p stratum-core
 echo "✅ Updated Cargo.toml to use local dependencies"
 echo "🏃 Running integration tests..."
 
-# Run the integration tests
-RUST_BACKTRACE=1 RUST_LOG=debug cargo nextest run --nocapture --verbose
+# Failure triage: re-run a single test with captured logs shown live:
+#   RUST_LOG=debug cargo nextest run --manifest-path=integration-tests/Cargo.toml \
```
GitGab19:
```suggestion
#   RUST_LOG=debug cargo nextest run \
```

GitGab19 (approved):
ACK

Just a nit on the comment.

scripts/run-integration-tests.sh
GitGab19: Amended this, and force-pushed the PR.

PR #2295, opened 2026-08-13
Remove vec slice encodable binary sv2

### LINKED ISSUE #2217: Remove From<Vec<u8>> and From<Slice> from binary_sv2
More information: https://github.com/stratum-mining/stratum/pull/2200#discussion_r3481267868

bit-aloo:
closes: https://github.com/stratum-mining/stratum/issues/2217

### DIFF
--- a/fuzz/fuzz_targets/common.rs
+++ b/fuzz/fuzz_targets/common.rs
@@ -124,7 +124,7 @@ macro_rules! test_datatype_roundtrip {
             // Ensure serialization is canonical: re-encoding must match the consumed input.
             assert_eq!(
                 encoded,
-                input[..encoded.get_size()],
+                input[..encoded.len()],
                 "Serialization is not stable"
             );
         }
@@ -162,7 +162,7 @@ macro_rules! test_datatype_roundtrip {
             // reserialization must match the consumed input bytes.
             assert_eq!(
                 encoded,
-                input[..encoded.get_size()],
+                input[..encoded.len()],
                 "{}: Serialization is not stable",
                 stringify!($datatype)
             );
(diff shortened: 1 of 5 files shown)


GitGab19 (approved):
Clanker finding, non-blocking:

Remove the stale BYTES mapping from the documentation. The PR removes the implementations that made raw byte vectors encodable, but both the [crate documentation](https://github.com/stratum-mining/stratum/blob/479fcfbfd53b6cff3770a2927ef18555a3a42045/sv2/binary-sv2/src/lib.rs#L29) and [README](https://github.com/stratum-mining/stratum/blob/479fcfbfd53b6cff3770a2927ef18555a3a42045/sv2/binary-sv2/README.md#L30) still advertise [u8] <-> BYTES. Users following that documentation will encounter missing trait implementations. Remove those rows or explain that BYTES is handled by framing context rather than binary_sv2.

bit-aloo:
> Clanker finding, non-blocking:
> 
> Remove the stale BYTES mapping from the documentation. The PR removes the implementations that made raw byte vectors encodable, but both the [crate documentation](https://github.com/stratum-mining/stratum/blob/479fcfbfd53b6cff3770a2927ef18555a3a42045/sv2/binary-sv2/src/lib.rs#L29) and [README](https://github.com/stratum-mining/stratum/blob/479fcfbfd53b6cff3770a2927ef18555a3a42045/sv2/binary-sv2/README.md#L30) still advertise [u8] <-> BYTES. Users following that documentation will encounter missing trait implementations. Remove those rows or explain that BYTES is handled by framing context rather than binary_sv2.

Thanks, and also added the reason why we are removing it.

GitGab19 (approved):

PR #2303, opened 2026-08-18
`channels_sv2`: enforce share `nTime` bounds, bound `seen_shares`, and fix job-ID/chain-tip/coinbase validation gaps (plus `template_distribution_sv2` docs fix))

### LINKED ISSUE #2297: `channels_sv2`: need to bound the accepted-share dedup cache (`seen_shares`, `client` + `server`)
Follow-up to #2262 (client `rejected_shares` bound) and sibling of #2289 (job-storage retention).

The same unbounded-collection-under-attacker-controlled-messages pattern exists in the duplicate-share detection of both `ShareAccounting` implementations. `seen_shares` inserts every distinct accepted share hash and is only cleared on a chain-tip transition, which the peer controls the timing of (long-lived tip, or a stalled/withheld `SetNewPrevHash`). Memory grows linearly with accepted shares for the lifetime of a tip.

Unlike the job-storage caps in #2289, this is not a matter of picking a number and evicting past it — and the two `ShareAccounting` implementations do not want the same bound:

- On the **server**, evicting is not an option at all: forgetting a still-valid hash re-admits its duplicate, re-enabling the replay/accounting fraud the set exists to prevent. Overflow has to be an explicit failure, which in turn means the bound must be provably unreachable for any honestly-configured channel.
- On the **client**, eviction is harmless (it costs one double-counted local statistic), but the *size* is the hard constraint: the `client` module is documented as `no_std`-compatible for embedded use, and an adversarial upstream can drive the cache to its bound at message speed — so the bound is memory the smallest supported device must have spare, not a theoretical ceiling.

So this needs a design decision, not just a bound: a bound whose size is derived from peer-supplied inputs is not a bound at all on the side where the peer is the threat.

### 1. Server `seen_shares` has no bound

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/share_accounting.rs#L136-L145

`update_share_accounting` inserts unconditionally; the only clear path is `flush_seen_shares`, called on chain-tip transitions:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/standard.rs#L584

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/extended.rs#L628

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/extended.rs#L678

Trust boundary: a malicious **downstream miner**. Server channel targets derive from downstream-supplied nominal hashrate and max target; a zero-hashrate channel at the maximum target makes essentially every unique header pass `validate_share`, so the set grows at message speed with no PoW cost. Affects standard and extended server channels (`validate_share` → `is_share_seen` / `update_share_accounting` in both).

### 2. Client `seen_shares` has the same defect

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/share_accounting.rs#L193-L203

`track_validated_share` mirrors the server path; flushed only via `flush_seen_shares`:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/standard.rs#L364

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/extended.rs#L512

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/extended.rs#L562

Trust boundary: an **adversarial upstream** advertises an extremely easy channel target while holding one tip active; the attached share source then grows the validating client/proxy at message speed. Especially relevant for the documented `no_std`/embedded use case where memory is constrained.

### fix (two bounds, one per trust boundary)

The two sides do **not** want the same bound. They differ in what overflow costs (fatal vs. harmless), in who controls the inputs a bound could be derived from (pool operator vs. peer), and in how much memory the host has (a pool server vs. a documented `no_std`/embedded client). Each of those points the same way: derive on the server, use a flat constant on the client.

**1. Server — derived budget, hard limit.**

The channel already stores `expected_share_per_minute`:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/standard.rs#L99

```
budget = clamp(expected_shares_per_minute × WORST_CASE_TIP_MINUTES × MARGIN,
               MIN_SEEN_SHARES_CAP, MAX_SEEN_SHARES_CAP)
```

- `WORST_CASE_TIP_MINUTES = 600` (10 h): `seen_shares` only needs to hold shares for the lifetime of one chain tip (it is flushed on every tip transition). 10 h is far beyond any realistic block interval, so the budget is unreachable by a well-behaved channel.
- `MARGIN = 2`: Poisson variance over thousands of expected shares is ~√N, so 2× is already generous; its real job is covering the window before vardiff converges.
- `MIN_SEEN_SHARES_CAP = 4 096` / `MAX_SEEN_SHARES_CAP = 2²⁰`: the derived value is only as sane as the configured rate. Without the lower clamp, a near-zero configured rate yields a budget of 1 and closes a legitimate channel on its second accepted share; without the upper one, nothing caps a single channel's cache. Only reachable via absurd configuration, but free to bound.

Compute at construction. `expected_share_per_minute` is fixed for the channel's lifetime — `update_channel` recomputes the *target* from a new nominal hashrate precisely to preserve that expected rate — so the budget never needs refreshing.

When the budget is exhausted, `validate_share` returns a new `ShareValidationError::SeenSharesBudgetExhausted` — deliberately **without** an `error_code` string, so the existing "SHOULD lead to a client disconnection or application shutdown" documentation applies and the embedding application closes the channel. Evicting is not an option here: forgetting a still-valid hash re-admits its duplicate, re-enabling the replay/accounting fraud (double-counted shares are payout fraud) the set exists to prevent.

Deriving is worth it on this side precisely because the limit is load-bearing: it must stay unreachable for *any* honestly-configured rate, and honest rates span orders of magnitude. And the input is pool-configured, not peer-supplied, so the derivation cannot be steered by an attacker. Document the per-channel worst-case memory next to the constant (at 6 shares/min the budget is 7 200 hashes; at 100 shares/min, 120 000).

**2. Client — flat constant bound, FIFO eviction.**

```rust
/// Maximum number of accepted-share hashes a client channel retains for duplicate detection.
pub const MAX_SEEN_SHARES: usize = 4_096; // one 128 KB allocation
```

Held in a single `VecDeque<Hash>`, oldest evicted *before* each insert so the queue never exceeds the bound even transiently and its backing allocation settles at exactly `MAX_SEEN_SHARES` entries.

*Why a constant sized for the smallest host, and not a derived cap.* The `client` module is documented as `no_std`-compatible for embedded use, and an adversarial upstream advertising a trivial target can drive this cache to its bound at message speed — so **the bound is memory the smallest supported device must have spare**, not a theoretical ceiling. That rules out anything tip-sized: a cap derived from target and nominal hashrate (the approach this issue originally proposed) measured **137 MB peak RSS for a single channel** at full occupancy, which is fatal on any MCU and unpleasant anywhere.

The derivation is also unsound on this side independently of its size. Its inputs are upstream-controlled, and a bound computed from attacker-supplied parameters cannot bound the attacker; it would need clamping into an absolute range regardless, at which point the clamp is the real bound and the derivation only picks where inside it eviction begins. Worse, the inverse relation (`hash_rate_from_target`) overflows for any target easier than ~2²⁴³ — including the maximum target — which is exactly the adversarial case, so the derivation degenerates precisely where it is needed.

*Why 4 096 is enough.* A client cache is not a tip-reconciliation structure. It exists to catch a share source re-submitting work it already sent — a retransmit or a buggy loop, arriving within seconds. 4 096 hashes covers ~11 h of history for a typical 6 shares/min channel and ~7 min for a very busy 600 shares/min proxy channel, both far beyond any realistic duplicate window.

*Why one `VecDeque` and not a set plus a companion order queue.* At this size a linear scan beats a hash lookup (random hashes diverge in the first byte, so `contains` early-exits immediately), and storing each hash once instead of twice halves the footprint — the quantity that actually matters on the constrained target.

Overflow evicts rather than fails because the consequence is local only: an evicted-then-replayed hash double-counts one statistic (`validated_shares` / `validated_work_sum`); nothing is paid out, and nothing is forwarded as newly-validated that the upstream will not independently dedup.

Both server constants, `MAX_SEEN_SHARES`, and the rationale above should live in rustdoc next to their definitions, and `is_share_seen` / `flush_seen_shares` docs updated to state each side's bound and overflow behavior.

### notes

- Trust boundaries differ per side (downstream miners for 1, upstream + share source for 2) — worth keeping distinct in review, same as #2289.
- Vardiff is not a mitigation: the target is attacker-influenced at channel setup and the cache must be safe before any difficulty adjustment kicks in.

### regression tests

- **client**: flood past `MAX_SEEN_SHARES` with unique hashes and assert retention is bounded and oldest-first — the evicted hashes are the oldest, the resident ones the newest, and a chain-tip transition clears the queue. The audit PoCs' 4 097-unique-hash flood exercises exactly one eviction at this bound.
- **server**: fill to the budget and assert (a) every hash accepted along the way is *still resident* — nothing is evicted, so no duplicate can be re-admitted — and (b) the next `validate_share` fails with `SeenSharesBudgetExhausted` rather than growing the cache. Note this replaces the audit PoC's "the full 10 000-share history stays resident" expectation: with a bounded budget the 10 000th share is refused, not retained.

PoCs available on:
- https://github.com/project-loupe/audit-stratum/issues/130
- https://github.com/project-loupe/audit-stratum/issues/132
- https://github.com/project-loupe/audit-stratum/issues/118

### LINKED ISSUE #2298: `channels_sv2`: client share validation ignores each job's own `min_ntime` (`standard` + `extended`)
Follow-up to the chain-tip `min_ntime` lower bound added earlier (audit-stratum https://github.com/project-loupe/audit-stratum/issues/55 / https://github.com/project-loupe/audit-stratum/issues/71).

That fix enforces `share.ntime >= chain_tip.min_ntime()`, but an immediately-active job carries its **own** `min_ntime`, documented in-tree as the smallest nTime available for that job — and it may be later than the current chain tip's minimum.

Both client channel types validate only against the tip, so a share in the gap (`chain_tip.min_ntime <= ntime < job.min_ntime`) is hashed and returned as `Valid`/`BlockFound` whenever it meets the target.

### 1. `client/standard.rs`

`validate_share` enforces only the chain-tip bound:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/standard.rs#L423-L427

while `store_new_mining_job` explicitly matches the immediate job's `min_ntime` and discards it for validation purposes (it survives only inside the stored message):

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/standard.rs#L295-L296

The rustdoc also promises only the tip bound (`#L375`), so the doc needs updating alongside the check.

### 2. `client/extended.rs`

Same defect, same shape:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/extended.rs#L644-L648

The selected job is stored as a full `NewExtendedMiningJob`, so its `min_ntime` is already available at the check site.

### impact

Low, matching the audit severity: the miner still needs valid proof of work. But a proxy / JD-client embedding this validator as the gate for attached hashing devices can forward out-of-window work upstream, account it as locally validated, or surface a bogus block candidate that the real upstream (or the network) rejects. A pure client validating only its own work mostly self-harms.

Note the gap only exists for **immediately-active** jobs (`min_ntime` present on arrival): jobs activated from the future queue get their `min_ntime` overwritten with the `SetNewPrevHash` timestamp (`client/standard.rs#L339`, and the equivalent in `client/extended.rs`), which equals the tip minimum, so the per-job check is redundant there — and harmless.

### fix (one comparison per site)

In both `validate_share` implementations, after resolving the job and alongside the existing chain-tip check:

```rust
if let Some(job_min_ntime) = job.0.min_ntime.clone().into_inner() {
    if share.ntime < job_min_ntime {
        return Err(ShareValidationError::Invalid(
            ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
        ));
    }
}
```

Keep the chain-tip check as well — for activated-future jobs the two are equal, and for immediate jobs the effective bound becomes `max(tip, job)`, which is the documented semantics. Update the `validate_share` rustdocs on both channels to state the per-job bound.

Out of scope here: the nTime **upper** bound — tracked in #2299 (audit finding: https://github.com/project-loupe/audit-stratum/issues/123).

### regression tests

PoCs (activate a tip at nTime *t*, install an immediate job with `min_ntime > t`, assert a share in the gap is rejected; the extended PoC first re-confirms the chain-tip check still holds) available on:
- https://github.com/project-loupe/audit-stratum/issues/122
- https://github.com/project-loupe/audit-stratum/issues/124

---

One thing to verify when implementing: whether past/stale jobs on these channels also retain a usable `min_ntime` (they should, since they were once active) — the fix as written covers them for free because it reads whatever the stored message carries.

### LINKED ISSUE #2299: `channels_sv2`: need to enforce an nTime upper bound on `validate_share` (all channel types)
Follow-up to #2240, which added the lower bound (`share.ntime >= chain_tip.min_ntime()`), and complement of #2298 (per-job lower bound). No `validate_share` implementation enforces any **upper** bound: `ntime` values as large as `u32::MAX` are hashed and returned as `Valid`/`BlockFound` whenever they meet the target.

All four sites have the lower-bound check but nothing above it:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/standard.rs#L664

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/extended.rs#L788

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/standard.rs#L423

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/extended.rs#L644

### impact

Low — policy bypass, not PoW bypass. Server side: a share with an absurdly future `ntime` that meets the network target is reported as `BlockFound` with a header bitcoind will reject (consensus caps block time at ~2 h in the future) — corrupted block accounting and a forfeited solution. Client side: a proxy/JDC embedding this validator can forward protocol-invalid work or surface a bogus block candidate.

### fix (clockless, consensus-anchored)

The `SubmitShares*` docs require `ntime <= SetNewPrevHash timestamp + seconds elapsed since receipt`, but the crate is `no_std` and has no clock. Instead of injecting a time source, anchor Bitcoin's consensus allowance at tip receipt:

```rust
/// Mirrors Bitcoin Core's MAX_FUTURE_BLOCK_TIME (src/chain.h): a block
/// timestamp more than 2h in the future is consensus-invalid. Anchored at
/// chain-tip receipt, since `min_ntime` ≈ wall time when the tip arrived.
pub const MAX_FUTURE_BLOCK_TIME: u32 = 2 * 60 * 60;
```

and at each of the four sites, next to the existing lower bound:

```rust
if share.ntime > chain_tip.min_ntime() + MAX_FUTURE_BLOCK_TIME {
    return Err(ShareValidationError::Invalid(
        ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
    ));
}
```

(Use `saturating_add` to keep the bound sane near `u32::MAX`.)

Notes:

- This is deliberately **looser** than the spec's elapsed-time window (which is stricter than consensus) but strictly enforceable without a clock; embedding applications that have a time source can additionally enforce the spec-exact window. Document this in the `validate_share` rustdocs.
- The bound equals the consensus limit at tip receipt and becomes conservative as the tip ages; a false rejection requires a >2 h-old chain tip *and* a miner stamping wall time instead of rolling from the job's `min_ntime` — document as a known, negligible edge.

### regression tests

Reuse each file's existing share-validation vectors: set the chain tip at timestamp *t*, submit a share with `ntime = u32::MAX` (and one with `t + MAX_FUTURE_BLOCK_TIME + 1`) under a permissive target, assert rejection; assert `t + MAX_FUTURE_BLOCK_TIME` is still accepted. PoC (client standard) available on:

- https://github.com/project-loupe/audit-stratum/issues/123

### LINKED ISSUE #2300: `channels_sv2`: `merkle_root_from_path` accepts semantically non-coinbase transactions
Follow-up to #2258, which fixed the panic when syntactically *malformed* coinbase bytes made this function return `None`. The complementary gap remains: syntactically **valid** bytes that are not a coinbase return `Some`.

`merkle_root_from_path` treats any byte string that `bitcoin::consensus::deserialize::<Transaction>` can parse as a validated coinbase and immediately derives its txid — it never checks coinbase semantics (exactly one input whose previous outpoint is null):

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/merkle_root.rs#L34-L45

A peer can therefore supply an ordinary transaction as `coinbase_tx_prefix || extranonce || coinbase_tx_suffix`; the function returns `Some`, so callers accept and store the job instead of rejecting it, and miners work on a Merkle root whose first transaction cannot be a valid block coinbase — wasted work until the job is replaced.

All five callers route through this single function:

- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/standard.rs#L262 (`on_new_group_channel_job` — upstream-controlled prefix/suffix; returns `InvalidCoinbase` on `None`)
- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/client/extended.rs#L626 (`validate_share`)
- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/extended.rs#L770 (`validate_share`)
- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/jobs/factory.rs#L233
- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/jobs/extended.rs#L119

Trust boundaries differ per site: for clients the prefix/suffix come from the upstream mining peer (group-channel jobs); server-side the custom-job path (`SetCustomMiningJob`) carries work-selection-client-controlled bytes; the template-driven factory paths build the coinbase locally and mainly benefit as defense in depth.

### impact

Low, matching the audit severity: requires a malicious or faulty peer, and the damage is job integrity / wasted work rather than memory corruption or fund exposure.

### fix (single point, covers all callers)

`rust-bitcoin` already provides the exact semantic check — no new logic needed:

```rust
let coinbase: Transaction = match consensus::deserialize(&coinbase[..]) {
    Ok(trans) => trans,
    Err(e) => {
        error!("ERROR: {}", e);
        return None;
    }
};

if !coinbase.is_coinbase() {
    error!("ERROR: not a coinbase transaction");
    return None;
}
```

`Transaction::is_coinbase()` requires exactly one input with a null previous outpoint — the minimal coinbase invariant. Callers already handle `None` (that's what #2258 fixed), so no caller changes are needed. Update the rustdoc to state the semantic validation.

### regression test

PoC (a validly encoded transaction whose sole input spends a non-null outpoint must make `merkle_root_from_path` return `None`) available on:

- https://github.com/project-loupe/audit-stratum/issues/125

### LINKED ISSUE #2301: `channels_sv2`: server channels wedge on `SetNewPrevHash` with no queued future template (`group` + `standard`)
Per the [TDP spec §7.3](https://github.com/stratum-mining/sv2-spec/blob/main/07-Template-Distribution-Protocol.md), a Template Provider must send `SetNewPrevHash` immediately upon validating a new best block, and **must** have sent at least one future `NewTemplate` before it:

> Upon successful validation of a new best block, the server MUST immediately provide a `SetNewPrevHash` message.
> Prior to that, the server MUST send at least one, but potentially multiple `NewTemplate` messages with `future_template` flag set.

So a `SetNewPrevHash` arriving with no future template queued is a **protocol violation by the peer** — but two of the three server channel types respond to it by rejecting the message and never recording the new chain tip, which leaves the channel in a state strictly worse than the violation itself. The chain-tip fields (`prev_hash`, `n_bits`, `header_timestamp`) are self-contained and do not depend on `template_id` resolving to anything, so discarding them buys nothing and costs the channel its ability to recover.

### 1. `server/group.rs`

`on_set_new_prev_hash` unconditionally returns `TemplateIdNotFound` when no future jobs are queued, leaving `chain_tip` unset/stale (the rustdoc even documents this: "Returns an error if no matching future job is found, leaving the chain tip untouched"):

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/group.rs#L325-L328

### 2. `server/standard.rs`

Identical branch, identical consequence:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/standard.rs#L557-L559

### the model fix already exists in-tree

`server/extended.rs` handles the no-future-jobs case correctly: it demotes the previously-active job, marks past jobs stale, clears the target mappings, flushes seen shares, and **still updates the chain tip**, keeping `TemplateIdNotFound` only for the case where future jobs exist but none matches:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/extended.rs#L590-L604

### impact

Low severity, but the failure is unrecoverable and silent in one of its two forms:

- **Established channel** (chain tip already set): the tip stays on the *previous* block. Every subsequent non-future `NewTemplate` builds jobs committed to a dead `prev_hash`, and `validate_share` keeps validating against the stale tip — the channel keeps accepting and crediting work that can never form a block.
- **Channel that never received a tip**: the next valid non-future `NewTemplate` fails with `ChainTipNotSet`, wedging the job pipeline in a persistent error path.
- In both cases `TemplateIdNotFound` carries no `error_code`, so the error documentation's recommendation of shutdown for non-protocol variants applies — an embedding pool/proxy may terminate a shared process over one peer's misbehaviour.

### reachability

This requires a misbehaving or racing peer; it is not part of a conforming TDP flow. The realistic paths are:

- a TP emitting two `SetNewPrevHash` in rapid succession (fast blocks, reorg) without a future `NewTemplate` in between;
- an embedding application that does not backfill newly opened channels with the pending future template. Note the reference pool in `sv2-apps` does: `handle_open_standard_mining_channel` refuses to open a channel when no future template is cached (`PoolErrorKind::FutureTemplateNotPresent`) and otherwise feeds it `last_future_template` before replying — so this path is closed there, but it is not guaranteed by the library.

The point of the fix is therefore **recovery from a peer that broke the rules**, not conformance: a channel should not be permanently poisoned by one malformed message when the recoverable state is sitting in that same message.

### fix

Mirror the extended channel's `false` branch in both files, minus the state each channel doesn't have:

- **standard**: demote active job → `mark_past_jobs_as_stale` → clear `job_id_to_target` → `flush_seen_shares` → update `chain_tip` → `Ok(())`. The demote-then-stale ordering matters: without it a late share for the still-active old job skips the stale check in `validate_share` and panics on the `job_id_to_target` entry that was just cleared.
- **group**: group channels track no shares and (post-#2290) retain no past jobs, so the branch reduces to dropping the active job and updating `chain_tip`. Prefer dropping it outright (a new `JobStore::clear_active_job`) over routing it through `deactivate_job` + `mark_past_jobs_as_stale` — group channels never validate shares against stored jobs, so the stale set would only pin the job's retired extranonce prefix for an extra tip.

In both, `TemplateIdNotFound` remains for the true-branch case (future jobs queued, none matching). Update both rustdocs, which currently document the buggy behavior as intended, and state that the branch handles a non-conforming peer rather than a normal flow.

### related: the in-tree TDP docs are wrong

Worth fixing separately, since it is what makes this defect look like correct behavior on inspection. `sv2/subprotocols/template-distribution/src/set_new_prev_hash.rs` omits the spec's mandatory precondition and contradicts itself:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/subprotocols/template-distribution/src/set_new_prev_hash.rs#L5-L12

- struct doc: *"**If** a `NewTemplate` message has previously been sent with the `future_template` flag set, the `template_id` field **should** be set to…"* — conditional, SHOULD
- field doc: *"This **must** be identical to previously sent `NewTemplate` message"* — unconditional, MUST

Neither states that the TP MUST send at least one future template before every `SetNewPrevHash`. The conditional reading is what originally framed this finding as "a valid state transition"; it is not one.

### regression tests

- **group**: send `SetNewPrevHash` to a fresh `GroupChannel` with no queued future template; assert `Ok` and that the chain tip records the message's `prev_hash`/`header_timestamp`/`n_bits`, and that no job was activated. PoC available on https://github.com/project-loupe/audit-stratum/issues/128 (its expectation is correct; its assertion message states the superseded justification).
- **standard**: the mirrored test, plus the two properties group doesn't have — a late share for the previously active job must be rejected as `Stale` (not panic on the cleared `job_id_to_target`), and the channel must accept the next non-future `NewTemplate` instead of failing `ChainTipNotSet`.
`

### LINKED ISSUE #2302: `channels_sv2`: `JobIdFactory::next` does not handle `u32` exhaustion (panic or job-ID reuse)
`JobIdFactory` exists to guarantee unique job IDs (its rustdoc says "Ensures unique job ids"), but `next()` increments with unchecked arithmetic:

https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/jobs/factory.rs#L67-L71

After `u32::MAX` allocations the behavior depends on build flags: with overflow checks the next call **panics** (terminating the embedding server task/process); in ordinary release builds it silently **wraps to zero** and reuses identifiers. The defect is not the wrap itself but that the overflow behavior is undefined-by-build-profile and undocumented.

All three allocation sites are in this file:

- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/jobs/factory.rs#L218
- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/jobs/factory.rs#L305
- https://github.com/stratum-mining/stratum/blob/905cf73c80e1e1566b00a548fc3696d6617652c9/sv2/channels-sv2/src/server/jobs/factory.rs#L507

### impact

Low, matching the audit severity: exhaustion needs ~2³² allocations against one long-lived factory. But the terminal condition is deterministic, requires no privileged access, and the panic variant terminates the embedding server task over an event that is otherwise harmless.

### why ID reuse after a wrap is safe

Reusing an ID is only dangerous if some `HashMap<u32, Job>` store still holds an entry under it. Job stores only retain *recent* jobs: future jobs are consumed on activation, and past/stale jobs are flushed on every chain-tip transition (with retention further bounded by #2289). By the time an ID comes around again — 2³² allocations later — no store can still track its previous holder, so a collision with a live job is unreachable in any bounded deployment.

### fix

Make the wrap explicit and documented, rather than an error:

```rust
/// Increments then returns the internal state on a new ID.
///
/// Explicitly wraps to `0` after `u32::MAX`, restarting the sequence. This makes the
/// overflow behavior identical across build profiles. Reuse of an ID is safe because
/// job stores only retain recent jobs, so by the time an ID comes around again
/// (2³² allocations later), no store still tracks its previous holder.
fn next(&mut self) -> u32 {
    self.state = self.state.wrapping_add(1);
    self.state
}
```

`next()` stays infallible, so no new error variant and no channel-teardown path in embedding applications for an event that effectively never fires. Update the `JobFactory` rustdoc to qualify the uniqueness guarantee ("unique within any window of 2³² allocations") and state the wrap.

Note this deliberately diverges from the exhaustion-as-explicit-error fix sketched in the audit finding: an error variant would force every embedder to handle a condition that is safe to survive, while the wrap is safe under the retention argument above.

### regression tests

Place the private counter at `u32::MAX - 1` and require the next public job-creation calls to yield IDs `u32::MAX`, `0`, `1` — deterministic across build profiles. The test panics on overflow-checked builds without the fix, so it still catches the original defect (audit PoC: https://github.com/project-loupe/audit-stratum/issues/131, adapted from its expect-an-error form to expect the documented wrap).
`

plebhash:
close #2297
close #2298
close #2299
close #2300
close #2301
close #2302
close https://github.com/project-loupe/audit-stratum/issues/130
close https://github.com/project-loupe/audit-stratum/issues/132
close https://github.com/project-loupe/audit-stratum/issues/118
close https://github.com/project-loupe/audit-stratum/issues/122
close https://github.com/project-loupe/audit-stratum/issues/124
close https://github.com/project-loupe/audit-stratum/issues/123
close https://github.com/project-loupe/audit-stratum/issues/125
close https://github.com/project-loupe/audit-stratum/issues/128
close https://github.com/project-loupe/audit-stratum/issues/131

companion https://github.com/stratum-mining/sv2-apps/pull/733

### DIFF
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -608,7 +608,9 @@ impl ExtendedChannel {
     ///
     /// Updates channel state with the share validation result:
     /// - Prevents propagation of stale, duplicate, low-difficulty, or low-ntime shares
-    ///   (shares whose `ntime` is below the chain tip's `min_ntime`).
+    ///   (shares whose `ntime` is below the chain tip's `min_ntime` or the job's own
+    ///   `min_ntime` — an immediately-active job may carry a `min_ntime` later than the chain
+    ///   tip's minimum, so the effective lower bound is the larger of the two).
     /// - Indicates whether a block was found from the share.
     /// - Maintains local share accounting for later reconciliation with upstream acknowledgements.
     pub fn validate_share(
@@ -683,6 +685,17 @@ impl ExtendedChannel {
             return Err(ShareValidationError::Invalid(
                 ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
             ));
+        }
+
+        // an immediately-active job carries its own min_ntime, which may be later than the
+        // chain tip's minimum; jobs activated from the future queue have it overwritten with
+        // the SetNewPrevHash timestamp, making this check redundant there (and harmless)
+        if let Some(job_min_ntime) = job.0.min_ntime.clone().into_inner() {
+            if share.ntime < job_min_ntime {
+                return Err(ShareValidationError::Invalid(
+                    ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
+                ));
+            }
         }
 
         // Only BIP323 general-purpose bits may differ from the job's advertised version.
@@ -1316,7 +1329,7 @@ mod tests {
                 0, 0, 0, 0, 0, 38, 106, 36, 170, 33, 169, 237, 226, 246, 28, 63, 113, 209, 222,
                 253, 63, 169, 153, 223, 163, 105, 83, 117, 92, 105, 6, 137, 121, 153, 98, 180, 139,
                 235, 216, 54, 151, 78, 140, 249, 1, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
             ]
             .try_into()
             .unwrap(),
@@ -2335,5 +2348,97 @@ mod tests {
         }
         assert_eq!(channel.get_past_jobs_count(), 0);
         assert!(channel.get_active_job().is_none());
+    }
+
+    #[test]
+    fn test_share_validation_ntime_below_job_min_ntime() {
+        // Regression test: an immediately-active job carries its own min_ntime, which may be
+        // later than the chain tip's minimum. A share in the gap
+        // (chain_tip.min_ntime <= ntime < job.min_ntime) must be rejected.
+        let channel_id = 1;
+        let extranonce_prefix = [
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+        ]
+        .to_vec();
+
+        let mut channel = ExtendedChannel::new(
+            channel_id,
+            "user_identity".to_string(),
+            ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
+            Target::from_le_bytes([0xff; 32]),
+            1.0,
+            true,
+            8u16,
+        );
+
+        let job = |job_id: u32, min_ntime: Option<u32>| NewExtendedMiningJob {
+            channel_id,
+            job_id,
+            min_ntime: Sv2Option::new(min_ntime),
+            version: 536870912,
+            version_rolling_allowed: true,
+            coinbase_tx_prefix: vec![
+                2, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 34, 82, 0,
+            ]
+            .try_into()
+            .unwrap(),
+            coinbase_tx_suffix: vec![
+                255, 255, 255, 255, 2, 0, 242, 5, 42, 1, 0, 0, 0, 22, 0, 20, 235, 225, 183, 220,
+                194, 147, 204, 170, 14, 231, 67, 168, 111, 137, 223, 130, 88, 194, 8, 252, 0, 0, 0,
+                0, 0, 0, 0, 0, 38, 106, 36, 170, 33, 169, 237, 226, 246, 28, 63, 113, 209, 222,
+                253, 63, 169, 153, 223, 163, 105, 83, 117, 92, 105, 6, 137, 121, 153, 98, 180, 139,
+                235, 216, 54, 151, 78, 140, 249, 1, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            ]
+            .try_into()
+            .unwrap(),
+            merkle_path: vec![].try_into().unwrap(),
+        };
+
+        let share = |sequence_number: u32, job_id: u32, ntime: u32| SubmitSharesExtended {
+            channel_id,
+            sequence_number,
+            job_id,
+            nonce: 741057,
+            ntime,
+            version: 536870912,
+            extranonce: vec![1, 0, 0, 0, 0, 0, 0, 0].try_into().unwrap(),
+        };
+
+        // activate a chain tip at nTime t via a future job
+        let tip_ntime: u32 = 1745596930;
+        channel.on_new_extended_mining_job(job(1, None)).unwrap();
+        // network target: 000000000000d7c0... (hard, so no accidental BlockFound)
+        channel
+            .on_set_new_prev_hash(SetNewPrevHashMp {
+                channel_id,
+                job_id: 1,
+                prev_hash: [
+                    200, 53, 253, 129, 214, 31, 43, 84, 179, 58, 58, 76, 128, 213, 24, 53, 38, 144,
+                    205, 88, 172, 20, 251, 22, 217, 141, 21, 221, 21, 0, 0, 0,
+                ]
+                .into(),
+                nbits: 453040064,
+                min_ntime: tip_ntime,
+            })
+            .unwrap();
+
+        // re-confirm the chain-tip lower bound still holds
+        let res = channel.validate_share(share(0, 1, tip_ntime - 1));
+        assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
+
+        // install an immediately-active job whose own min_ntime is later than the tip's
+        let job_min_ntime = tip_ntime + 3;
+        channel
+            .on_new_extended_mining_job(job(2, Some(job_min_ntime)))
+            .unwrap();
+
+        // a share in the gap passes the chain-tip bound but not the job's own bound
+        let res = channel.validate_share(share(1, 2, job_min_ntime - 1));
+        assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
+
+        let res = channel.validate_share(share(2, 2, job_min_ntime));
+        assert!(matches!(res, Ok(ShareValidationResult::Valid(_))));
     }
 }
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -15,7 +15,7 @@ use crate::{
     extranonce_manager::ExtranoncePrefix,
     merkle_root::merkle_root_from_path,
     target::{bytes_to_hex, u256_to_block_hash},
-    MAX_EXTRANONCE_LEN, VERSION_ROLLING_MASK,
+    MAX_EXTRANONCE_LEN, MAX_FUTURE_BLOCK_TIME, VERSION_ROLLING_MASK,
 };
 use alloc::{collections::VecDeque, format, string::String, vec, vec::Vec};
 use binary_sv2::Sv2OptionOwned;
@@ -610,7 +610,10 @@ impl ExtendedChannel {
     /// - Prevents propagation of stale, duplicate, low-difficulty, or low-ntime shares
     ///   (shares whose `ntime` is below the chain tip's `min_ntime` or the job's own
     ///   `min_ntime` — an immediately-active job may carry a `min_ntime` later than the chain
-    ///   tip's minimum, so the effective lower bound is the larger of the two).
+    ///   tip's minimum, so the effective lower bound is the larger of the two — as well as
+    ///   shares whose `ntime` exceeds the chain tip's `min_ntime + MAX_FUTURE_BLOCK_TIME`, see
+    ///   [`MAX_FUTURE_BLOCK_TIME`] for how this clockless upper bound relates to the spec's
+    ///   elapsed-time window).
     /// - Indicates whether a block was found from the share.
     /// - Maintains local share accounting for later reconciliation with upstream acknowledgements.
     pub fn validate_share(
@@ -682,6 +685,14 @@ impl ExtendedChannel {
         let nbits: CompactTarget = CompactTarget::from_consensus(chain_tip.nbits());
 
         if share.ntime < chain_tip.min_ntime() {
+            return Err(ShareValidationError::Invalid(
+                ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
+            ));
+        }
+
+        // consensus caps block timestamps at ~2h in the future; the allowance is anchored at
+        // chain-tip receipt, since this crate has no clock (see MAX_FUTURE_BLOCK_TIME)
+        if share.ntime > chain_tip.min_ntime().saturating_add(MAX_FUTURE_BLOCK_TIME) {
             return Err(ShareValidationError::Invalid(
                 ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
             ));
@@ -2439,6 +2450,94 @@ mod tests {
         assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
 
         let res = channel.validate_share(share(2, 2, job_min_ntime));
+        assert!(matches!(res, Ok(ShareValidationResult::Valid(_))));
+    }
+
+    #[test]
+    fn test_share_validation_ntime_above_max_future_block_time() {
+        // Regression test: a share ntime beyond the chain tip's min_ntime +
+        // MAX_FUTURE_BLOCK_TIME would put a consensus-invalid timestamp in the block header,
+        // so it must be rejected; ntime exactly on the bound is still accepted.
+        let channel_id = 1;
+        let extranonce_prefix = [
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+        ]
+        .to_vec();
+
+        let mut channel = ExtendedChannel::new(
+            channel_id,
+            "user_identity".to_string(),
+            ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
+            Target::from_le_bytes([0xff; 32]),
+            1.0,
+            true,
+            8u16,
+        );
+
+        channel
+            .on_new_extended_mining_job(NewExtendedMiningJob {
+                channel_id,
+                job_id: 1,
+                min_ntime: Sv2Option::new(None),
+                version: 536870912,
+                version_rolling_allowed: true,
+                coinbase_tx_prefix: vec![
+                    2, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 34, 82, 0,
+                ]
+                .try_into()
+                .unwrap(),
+                coinbase_tx_suffix: vec![
+                    255, 255, 255, 255, 2, 0, 242, 5, 42, 1, 0, 0, 0, 22, 0, 20, 235, 225, 183,
+                    220, 194, 147, 204, 170, 14, 231, 67, 168, 111, 137, 223, 130, 88, 194, 8, 252,
+                    0, 0, 0, 0, 0, 0, 0, 0, 38, 106, 36, 170, 33, 169, 237, 226, 246, 28, 63, 113,
+                    209, 222, 253, 63, 169, 153, 223, 163, 105, 83, 117, 92, 105, 6, 137, 121, 153,
+                    98, 180, 139, 235, 216, 54, 151, 78, 140, 249, 1, 32, 0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0,
+                ]
+                .try_into()
+                .unwrap(),
+                merkle_path: vec![].try_into().unwrap(),
+            })
+            .unwrap();
+
+        let tip_ntime: u32 = 1745596930;
+        // network target: 000000000000d7c0... (hard, so no accidental BlockFound)
+        channel
+            .on_set_new_prev_hash(SetNewPrevHashMp {
+                channel_id,
+                job_id: 1,
+                prev_hash: [
+                    200, 53, 253, 129, 214, 31, 43, 84, 179, 58, 58, 76, 128, 213, 24, 53, 38, 144,
+                    205, 88, 172, 20, 251, 22, 217, 141, 21, 221, 21, 0, 0, 0,
+                ]
+                .into(),
+                nbits: 453040064,
+                min_ntime: tip_ntime,
+            })
+            .unwrap();
+
+        let share = |sequence_number: u32, ntime: u32| SubmitSharesExtended {
+            channel_id,
+            sequence_number,
+            job_id: 1,
+            nonce: 741057,
+            ntime,
+            version: 536870912,
+            extranonce: vec![1, 0, 0, 0, 0, 0, 0, 0].try_into().unwrap(),
+        };
+
+        // one second above the bound: rejected before any PoW evaluation
+        let res = channel.validate_share(share(0, tip_ntime + crate::MAX_FUTURE_BLOCK_TIME + 1));
+        assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
+
+        // u32::MAX is likewise rejected (the bound saturates instead of wrapping)
+        let res = channel.validate_share(share(1, u32::MAX));
+        assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
+
+        // exactly on the bound the share is accepted (channel target is permissive)
+        let res = channel.validate_share(share(2, tip_ntime + crate::MAX_FUTURE_BLOCK_TIME));
         assert!(matches!(res, Ok(ShareValidationResult::Valid(_))));
     }
 }
--- a/sv2/channels-sv2/src/client/mod.rs
+++ b/sv2/channels-sv2/src/client/mod.rs
@@ -35,6 +35,20 @@ pub const MAX_FUTURE_JOBS: usize = 16;
 /// small measured memory cost — see the load-test data in PR #2290.
 pub const MAX_PAST_JOBS: usize = 50;
 
+/// Maximum number of accepted-share hashes a client channel retains for duplicate detection.
+///
+/// 4 096 hashes is one 128 KB allocation, which keeps the `no_std`/embedded use case viable:
+/// the bound has to be affordable on the smallest supported device, since an adversarial
+/// upstream advertising a trivial target can drive the cache to it at message speed.
+///
+/// A client cache does not need to hold a whole chain tip's worth of shares. It exists to catch
+/// a share source re-submitting work it already sent — a retransmit or a buggy loop, which
+/// arrives within seconds — not to reconcile a tip. 4 096 covers ~11 hours of history for a
+/// typical 6 shares/min channel and ~7 minutes for a very busy 600 shares/min proxy channel,
+/// far beyond any realistic duplicate window in both cases. Overflow evicts oldest-first, and
+/// an evicted-then-replayed hash costs one double-counted local statistic.
+pub const MAX_SEEN_SHARES: usize = 4_096;
+
 // Type aliases that switch between `std::collections` and `hashbrown`
 // depending on whether the `no_std` feature is enabled.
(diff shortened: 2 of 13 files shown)


plebhash:
ready for review

bit-aloo:
clankers findings:

>   Findings

  > 1. Wrong constant in the client-side doc — sv2/channels-sv2/src/lib.rs:88
  /// Client channels bound `seen_shares` at [`MAX_SEEN_SHARES_CAP`] outright and evict oldest-first;
  Clients bound at client::MAX_SEEN_SHARES (4 096 ≈ 128 KB), not MAX_SEEN_SHARES_CAP (1 << 20 ≈ 32 MB). This contradicts the doc six lines above on MAX_SEEN_SHARES_CAP itself ("Client channels do not use
  this: they bound their cache at the far smaller client::MAX_SEEN_SHARES") and inverts the embedded-affordability argument the whole client design rests on. It's a live rustdoc link, so a reader
  following it lands on the wrong number.

  > 2. The protocol violation is now completely silent — server/standard.rs:576, server/group.rs:339
  Both no-future-jobs branches return Ok(()). The commit message argues (correctly) that the tip is usable and rejecting it wedges the channel — but the previous TemplateIdNotFound was also the only
  signal the embedding pool got that the TP is non-conforming. Now there is no return value, no log, nothing. tracing is already imported in both files (use tracing::debug), so a warn! in that branch
  costs nothing and preserves the observability the error return used to carry.

 > 3. 6d9de36b's fix stops at the client; server extended has the same gap
  Server extended jobs built from templates get min_ntime = chain_tip.min_ntime() (factory.rs:259, :352), so no gap there — but custom jobs take the client-supplied value verbatim (factory.rs:535,
  Sv2OptionOwned::new(Some(set_custom_mining_job.min_ntime))), and server/extended.rs::validate_share checks only the chain-tip bound (:824, :833), never job.min_ntime. A share in [chain_tip.min_ntime, 
  custom_job.min_ntime) is accepted and credited against a window the downstream JDC explicitly declared. Lower impact than the client case (nothing consensus-invalid results), but it's the same class,
  and the series otherwise applies its bounds uniformly across all four channel types.

  > 4. Both new bounds assume mainnet block cadence
  MAX_FUTURE_BLOCK_TIME is anchored at chain_tip.min_ntime() and WORST_CASE_TIP_MINUTES = 600 assumes a tip turns over well inside 10h. On regtest/CI or any harness where the tip doesn't advance, a
  long-lived tip both tightens the nTime window below consensus and lets an honest channel walk into SeenSharesBudgetExhausted (7 200 shares at 6/min ≈ 20h). The MAX_FUTURE_BLOCK_TIME doc already concedes
  the aging-tip edge; the seen_shares_budget doc frames hitting the budget as unambiguously "a protocol-abuse signal", which is only true under mainnet-like cadence. Worth one sentence there.

  > 5. Nits
  > - server/standard.rs:2699 — comment says 1 share/min -> 1200 hashes; the actual budget is 4 096, because MIN_SEEN_SHARES_CAP clamps it. The test uses seen_shares_budget() so it passes regardless, but
    the comment misstates the mechanism it's demonstrating.
  > - lib.rs:58 — MIN_SEEN_SHARES_CAP is documented as "4 096 hashes ≈ 128 KB" using payload math, while seen_shares_budget two constants later deliberately uses measured HashSet math ("rounds to a power of
    two at ~7/8 load"). By that same measure this one is ~270 KB (8 192 slots × 33 B). Pick one convention.
  > - client/share_accounting.rs:112 — "at this size a linear scan costs less than a hash lookup would (random hashes diverge in the first byte)". A 4 096-entry scan is ~two orders of magnitude slower than
    a SipHash lookup; early divergence makes each comparison cheap, not the scan. It doesn't matter in practice (two scans per share, tens of µs, on a client), and the memory half of the rationale —
    storing each hash once instead of set-plus-order-queue — is sound on its own. Just drop the CPU claim.

bit-aloo (approved):
ACK

bit-aloo (changes requested):
.

plebhash:
addressed all findings posted by @bit-aloo above, except 3

arguments from my clanker:

> Finding 3 doesn't hold, though the reasoning path is understandable: `validate_share` and the factory alone do suggest a gap. The missing piece is `on_set_custom_mining_job` (server/extended.rs), which maintains the invariant that makes a per-job check redundant:
> 
> 1. `self.chain_tip = Some(new_chain_tip)` runs **unconditionally** — after every accepted custom job, `chain_tip.min_ntime == job.min_ntime`. The *active* custom job can never sit above the tip.
> 2. If a custom job arrives whose `(prev_hash, nbits, min_ntime)` differs from the current tip in *any* field — including `min_ntime` alone — the `is_new_chain_tip` branch runs `deactivate_job()` + `mark_past_jobs_as_stale()`. So the only way a job with a *different* `min_ntime` can leave the active slot is into the **stale** set, and stale shares are rejected before any nTime check.
> 3. Past jobs therefore only ever contain jobs whose `min_ntime` equals the current tip's (custom) or was assigned from it (template), and the TDP `on_set_new_prev_hash` path stales everything too.
> 
> So the interval `[chain_tip.min_ntime, custom_job.min_ntime)` is empty for every job a share can still validate against. The client needed the explicit check because upstream-supplied immediately-active jobs genuinely carry `min_ntime ≠ tip`; the server enforces the same bound by construction. Added a comment at the server-extended tip check stating this invariant, so it doesn't get re-derived next review.

I inspected this and I agree with the rationale.

Additionally, `on_set_custom_mining_job` has a documented assumption that `SetCustomMiningJob` has already been validated against the chain tip before this function is called, which further reinforces the invariant that `[chain_tip.min_ntime, custom_job.min_ntime)` is always empty.

To be fair, validation of `SetCustomMiningJob.min_ntime` is a real gap on the current SRI JDS implementation. But it's already tracked via https://github.com/stratum-mining/sv2-apps/issues/500 so I think it's safe to proceed here.

bit-aloo (approved):
ACK

GitGab19 (commented):
Clanker finding, non-blocking:

---

**Low priority:** The wrap-safety argument overlooks `stale_jobs`, which remain until the next chain-tip transition. After 2³² non-future/custom jobs under one tip, a wrapped ID could collide with a retained stale job; validation checks stale jobs first, so shares for the new job would be rejected as stale. Consider rejecting exhaustion or skipping retained IDs, with a regression test covering this collision.

plebhash:
> Clanker finding, non-blocking:
> 
> **Low priority:** The wrap-safety argument overlooks `stale_jobs`, which remain until the next chain-tip transition. After 2³² non-future/custom jobs under one tip, a wrapped ID could collide with a retained stale job; validation checks stale jobs first, so shares for the new job would be rejected as stale. Consider rejecting exhaustion or skipping retained IDs, with a regression test covering this collision.

2³² non-future/custom jobs feels like an extremely unlikely edge case.

I'll skip addressing this for now.

just adding docs to make the rationale explicitly clear.

GitGab19 (approved):

PR #2307, opened 2026-08-18
`channels_sv2`: make the past-jobs cap configurable w/ a 'good' default value

gimballock:
`MAX_PAST_JOBS` is really a retention window — `cap ÷ template rate` — and the template rate is a deployment property the library can't see. The same constant buys 50s of late-share tolerance at a 1s template cadence and over 16 minutes at 20s, so I don't think one number can be right for everyone.

Adds `max_past_jobs: Option<usize>` to the server and client `ExtendedChannel`/`StandardChannel` constructors. `None` and `Some(0)` both select `MAX_PAST_JOBS`, so nothing changes for existing callers.

Treating `Some(0)` as "no opinion" rather than as a literal zero keeps a zero cap unreachable: it would evict the job that just retired, so the most common late share would come back `InvalidJobId` with no startup error to warn you.

Group channels take no parameter — client `GroupChannel` has no `past_jobs`, and server group channels use `replace_active_job`.

A second commit will set the default once we've measured it; the analysis is coming in a comment.

Follow-up to #2290.

companion https://github.com/stratum-mining/sv2-apps/pull/735

### DIFF
--- a/sv2/channels-sv2/src/client/standard.rs
+++ b/sv2/channels-sv2/src/client/standard.rs
@@ -5,7 +5,7 @@
 //! and chain tip state, enabling share validation and mining job lifecycle management.
 
 extern crate alloc;
-use super::{resolve_max_past_jobs, HashMap, MAX_FUTURE_JOBS};
+use super::{HashMap, MAX_FUTURE_JOBS, MAX_PAST_JOBS};
 use crate::{
     chain_tip::ChainTip,
     client::{
@@ -24,7 +24,6 @@ use bitcoin::{
     hashes::sha256d::Hash,
     CompactTarget, Target,
 };
-use core::num::NonZeroUsize;
 use mining_sv2::{
     NewExtendedMiningJobOwned, NewMiningJobOwned, SetNewPrevHashOwned as SetNewPrevHashMp,
     SubmitSharesStandardOwned, ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW,
@@ -48,7 +47,7 @@ pub type StandardJob = (NewMiningJobOwned, Target);
 /// - future mining jobs (indexed by job_id, activated upon [`NewMiningJob`](mining_sv2::NewMiningJob) receipt, capped at [`MAX_FUTURE_JOBS`])
 /// - active mining job
 /// - past jobs (active jobs under current chain tip, indexed by job_id, capped at
-///   [`MAX_PAST_JOBS`](super::MAX_PAST_JOBS))
+///   [`MAX_PAST_JOBS`])
 /// - stale jobs (jobs from previous chain tip, indexed by job_id)
 /// - share accounting state
 /// - chain tip state
@@ -70,7 +69,7 @@ pub struct StandardChannel {
     past_job_order: VecDeque<u32>,
     stale_jobs: HashMap<u32, StandardJob>,
     // Cap on `past_jobs` under the current chain tip, resolved from the constructor's
-    // `Option<NonZeroUsize>` against `MAX_PAST_JOBS`.
+    // `Option<usize>`; `None` and `Some(0)` both resolve to `MAX_PAST_JOBS`.
     max_past_jobs: usize,
     share_accounting: ShareAccounting,
     chain_tip: Option<ChainTip>,
@@ -79,16 +78,23 @@ pub struct StandardChannel {
 impl StandardChannel {
     /// Creates a new [`StandardChannel`] instance with provided channel parameters.
     ///
-    /// `max_past_jobs` caps the past jobs retained under the current chain tip; `None` uses
-    /// [`MAX_PAST_JOBS`](super::MAX_PAST_JOBS).
+    /// `max_past_jobs` caps the past jobs retained under the current chain tip. `None` and
+    /// `Some(0)` both select [`MAX_PAST_JOBS`].
     pub fn new(
         channel_id: u32,
         user_identity: String,
         extranonce_prefix: ExtranoncePrefix,
         target: Target,
         nominal_hashrate: f32,
-        max_past_jobs: Option<NonZeroUsize>,
+        max_past_jobs: Option<usize>,
     ) -> Self {
+        // fall back to the default when the caller has no opinion: `None`, or `Some(0)`, which
+        // would otherwise evict the just-retired job and reject the most common late share
+        let max_past_jobs = match max_past_jobs {
+            Some(cap) if cap > 0 => cap,
+            _ => MAX_PAST_JOBS,
+        };
+
         Self {
             channel_id,
             user_identity,
@@ -101,7 +107,7 @@ impl StandardChannel {
             past_jobs: HashMap::new(),
             past_job_order: VecDeque::new(),
             stale_jobs: HashMap::new(),
-            max_past_jobs: resolve_max_past_jobs(max_past_jobs),
+            max_past_jobs,
             share_accounting: ShareAccounting::new(),
             chain_tip: None,
         }
@@ -211,7 +217,7 @@ impl StandardChannel {
 
     /// Returns an iterator over all past jobs for the channel (active jobs under current chain tip).
     ///
-    /// At most [`MAX_PAST_JOBS`](super::MAX_PAST_JOBS) jobs are kept (oldest evicted first).
+    /// At most [`MAX_PAST_JOBS`] jobs are kept (oldest evicted first).
     pub fn get_past_jobs(&self) -> impl Iterator<Item = (&u32, &StandardJob)> + '_ {
         self.past_jobs.iter()
     }
@@ -223,7 +229,7 @@ impl StandardChannel {
 
     /// Returns the number of past jobs tracked by this channel.
     ///
-    /// At most [`MAX_PAST_JOBS`](super::MAX_PAST_JOBS) jobs are kept (oldest evicted first).
+    /// At most [`MAX_PAST_JOBS`] jobs are kept (oldest evicted first).
     pub fn get_past_jobs_count(&self) -> usize {
         self.past_jobs.len()
     }
@@ -305,7 +311,7 @@ impl StandardChannel {
     /// - If `min_ntime` is empty, the job is added to future jobs. At most [`MAX_FUTURE_JOBS`]
     ///   future jobs are kept: storing a new one beyond that limit evicts the oldest.
     /// - If an active job exists, it is moved to past jobs on activation. At most
-    ///   [`MAX_PAST_JOBS`](super::MAX_PAST_JOBS) past jobs are kept: retiring one beyond that limit evicts the oldest.
+    ///   [`MAX_PAST_JOBS`] past jobs are kept: retiring one beyond that limit evicts the oldest.
     pub fn on_new_mining_job(&mut self, new_mining_job: NewMiningJobOwned) {
         self.store_new_mining_job(new_mining_job);
     }
@@ -589,7 +595,6 @@ mod tests {
     };
     use binary_sv2::Sv2OptionOwned as Sv2Option;
     use bitcoin::Target;
-    use core::num::NonZeroUsize;
     use mining_sv2::{
         NewExtendedMiningJobOwned as NewExtendedMiningJob, NewMiningJobOwned as NewMiningJob,
         SetNewPrevHashOwned as SetNewPrevHashMp, SubmitSharesStandardOwned,
@@ -775,8 +780,8 @@ mod tests {
     fn test_past_jobs_respect_constructor_override() {
         // Some(cap) must override MAX_PAST_JOBS on the client standard channel's own eviction
         // path, which keeps its past jobs directly rather than delegating to a JobStore.
-        let custom_cap = NonZeroUsize::new(3).unwrap();
-        assert!(custom_cap.get() < MAX_PAST_JOBS);
+        let custom_cap = 3usize;
+        assert!(custom_cap < MAX_PAST_JOBS);
 
         let channel_id = 1;
         let extranonce_prefix = [
@@ -814,15 +819,31 @@ mod tests {
         }
 
         // bounded by the override, not by MAX_PAST_JOBS
-        assert_eq!(channel.get_past_jobs_count(), custom_cap.get());
+        assert_eq!(channel.get_past_jobs_count(), custom_cap);
 
         // the last job is active; only the newest `custom_cap` retired jobs survive
-        for job_id in 0..job_count - 1 - custom_cap.get() as u32 {
+        for job_id in 0..job_count - 1 - custom_cap as u32 {
             assert!(channel.get_past_job(job_id).is_none());
         }
-        for job_id in job_count - 1 - custom_cap.get() as u32..job_count - 1 {
+        for job_id in job_count - 1 - custom_cap as u32..job_count - 1 {
             assert!(channel.get_past_job(job_id).is_some());
         }
+
+        // `Some(0)` is not a zero cap: it means "no opinion" and selects the default
+        let mut zero_cap_channel = StandardChannel::new(
+            channel_id,
+            "user_identity".to_string(),
+            ExtranoncePrefix::from_wire(vec![0; 27]).unwrap(),
+            Target::from_le_bytes([0xff; 32]),
+            1.0,
+            Some(0),
+        );
+        for job_id in 0..MAX_PAST_JOBS as u32 + 2 {
+            let mut job = active_job.clone();
+            job.job_id = job_id;
+            zero_cap_channel.on_new_mining_job(job);
+        }
+        assert_eq!(zero_cap_channel.get_past_jobs_count(), MAX_PAST_JOBS);
     }
 
     #[test]
--- a/sv2/channels-sv2/src/client/standard.rs
+++ b/sv2/channels-sv2/src/client/standard.rs
@@ -24,6 +24,7 @@ use bitcoin::{
     hashes::sha256d::Hash,
     CompactTarget, Target,
 };
+use core::num::NonZeroUsize;
 use mining_sv2::{
     NewExtendedMiningJobOwned, NewMiningJobOwned, SetNewPrevHashOwned as SetNewPrevHashMp,
     SubmitSharesStandardOwned, ERROR_CODE_SUBMIT_SHARES_DIFFICULTY_TOO_LOW,
@@ -68,18 +69,25 @@ pub struct StandardChannel {
     // Replaced IDs move to the back; overflow evicts from the front.
     past_job_order: VecDeque<u32>,
     stale_jobs: HashMap<u32, StandardJob>,
+    // Cap on `past_jobs` under the current chain tip, resolved from the constructor's
+    // `Option<NonZeroUsize>` against `MAX_PAST_JOBS`.
+    max_past_jobs: usize,
     share_accounting: ShareAccounting,
     chain_tip: Option<ChainTip>,
 }
 
 impl StandardChannel {
     /// Creates a new [`StandardChannel`] instance with provided channel parameters.
+    ///
+    /// `max_past_jobs` caps the past jobs retained under the current chain tip; `None` uses
+    /// [`MAX_PAST_JOBS`].
     pub fn new(
         channel_id: u32,
         user_identity: String,
         extranonce_prefix: ExtranoncePrefix,
         target: Target,
         nominal_hashrate: f32,
+        max_past_jobs: Option<NonZeroUsize>,
     ) -> Self {
         Self {
             channel_id,
@@ -93,6 +101,9 @@ impl StandardChannel {
             past_jobs: HashMap::new(),
             past_job_order: VecDeque::new(),
             stale_jobs: HashMap::new(),
+            max_past_jobs: max_past_jobs
+                .map(NonZeroUsize::get)
+                .unwrap_or(MAX_PAST_JOBS),
             share_accounting: ShareAccounting::new(),
             chain_tip: None,
         }
@@ -313,7 +324,7 @@ impl StandardChannel {
         self.past_job_order.retain(|id| *id != job_id);
         self.past_job_order.push_back(job_id);
 
-        if self.past_jobs.len() > MAX_PAST_JOBS {
+        if self.past_jobs.len() > self.max_past_jobs {
             if let Some(evicted_job_id) = self.past_job_order.pop_front() {
                 self.past_jobs.remove(&evicted_job_id);
             }
@@ -604,6 +615,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -664,6 +676,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -710,6 +723,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -773,6 +787,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         let active_job = NewMiningJob {
@@ -823,6 +838,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let ntime: u32 = 1746839905;
@@ -877,6 +893,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -957,6 +974,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -1027,6 +1045,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -1103,6 +1122,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -1174,6 +1194,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -1255,6 +1276,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -1335,6 +1357,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let future_job = NewMiningJob {
@@ -1414,6 +1437,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         let merkle_root = [
@@ -1495,6 +1519,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         let merkle_root = [
@@ -1565,6 +1590,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         let merkle_root = [
@@ -1650,6 +1676,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             target,
             nominal_hashrate,
+            None,
         );
 
         let malformed_job = NewExtendedMiningJob {
@@ -1692,6 +1719,7 @@ mod tests {
             ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
             Target::from_le_bytes([0xff; 32]),
             1.0,
+            None,
         );
 
         channel.on_new_mining_job(NewMiningJob {
--- a/sv2/channels-sv2/src/client/standard.rs
+++ b/sv2/channels-sv2/src/client/standard.rs
@@ -5,7 +5,7 @@
 //! and chain tip state, enabling share validation and mining job lifecycle management.
 
 extern crate alloc;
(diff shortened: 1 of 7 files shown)


sv2/channels-sv2/src/server/jobs/job_store.rs
```diff
@@ -35,8 +38,22 @@ pub(crate) const MAX_FUTURE_JOBS: usize = 16;
 /// 50 buys ample headroom over the reachable submit depth (a miner abandons a job once the next
 /// one arrives, so late shares realistically target the last 1-2 jobs) at a small measured
 /// memory cost — see the load-test data in PR #2290.
+///
+/// This is only the default. The cap is really a retention window — `cap / template rate` — and
+/// the template rate belongs to the deployment, so channel constructors accept a `max_past_jobs`
+/// override and fall back to this value when none is given.
 pub(crate) const MAX_PAST_JOBS: usize = 50;
```
plebhash: I was under the impression that you'd try to fine-tune this to some optimal value?

from https://github.com/stratum-mining/stratum/pull/2290#pullrequestreview-4922125774

> If a continuously-hashing A/B across 16 / 50 / 300 would help pick the constant, that is something we could look into.

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -93,6 +101,7 @@ impl StandardChannel {
             past_jobs: HashMap::new(),
             past_job_order: VecDeque::new(),
             stale_jobs: HashMap::new(),
+            max_past_jobs: resolve_max_past_jobs(max_past_jobs),
```
plebhash: I don't think we need `resolve_max_past_jobs`, nor `NonZeroUsize` type

we can add some code above the `Self` declaration, where we fall back to `MAX_PAST_JOBS` constant if:
- `max_past_jobs == Some(0)`
- `max_past_jobs == None`

gimballock:
Now that the cap (`MAX_PAST_JOBS`) is configurable, here's what it should default to — and why our data doesn't yet license moving it off 50.

**Two losses, opposite fixes.** An uncredited share is either an eviction (`invalid-job-id`, within a tip) or a cross-tip loss (`stale-share`, at a prev-hash rotation). If it's eviction, `past_jobs` must stay deep and the memory is unavoidable. If it's cross-tip loss, `past_jobs` can be tiny and the stale window sized on its own. Everything below serves one question: which one are we seeing?

### Why our data can't decide

The costs are the ones already in #2290. On the memory side, at fixed connection count, cap 300 → 50 cut per-connection memory 36.5% (1.429 → 0.907 MiB, ~4.5 kB per retained job per channel). Both arms carried the same connection count within noise, so the cap moves **density, not capacity** — it buys headroom on the box rather than raising the connection ceiling. One caveat on that figure: per-connection memory isn't comparable across differing hardware, connection counts or trial lengths, and we published a 45.5% version of it that was wrong for exactly that reason. 36.5% is the corrected same-conditions result.

On the yield side, that same change cost 2.7 points on a channel-level metric we track — uncredited channels: those that opened successfully but never had a single share credited, 3.16% → 5.86%. Those channels stayed connected for the whole trial, so they were paying full per-connection memory while producing nothing creditable — a yield loss, not a connection loss. Both figures come from a gauge we've seen undercount under scrape pressure, so the 2.7-point delta is sturdier than either absolute.

Theory says the eviction path should be nearly empty. The required depth — how many past jobs the cap has to retain — is `1 + ceil(latency ÷ template period)`, which is 1 at the shipped `min_interval = 5`, and a code-level argument lands on the same 1–2 independently. But we can't attribute those 2.7 points of uncredited channels either way. Every simulated miner submitted one share seconds after connecting and then stopped, so no share was ever old enough to test the required depth — our `invalid-job-id ≈ 0` is vacuous, not reassuring. That metric and `stale-share` rose together going 300 → 50, but under a one-shot workload that's correlation, not cause.

### The experiment that decides

That uncredited-channel metric can't be reused: it only means anything because each channel submitted exactly once, and once miners resubmit after a rejection every channel eventually lands an accepted share, so the count collapses toward zero however much work is being lost. So this experiment needs to measure an actual stream of shares — rejected ÷ submitted, split by error code — under continuous hashing at fixed connection count. Those two counters are the discriminator, and previous runs simply never put load on them.

| arm | caps | purpose |
|---|---|---|
| rare-tip | 1, 2, 4, 50 | eviction alone — locates the knee, the cap below which `invalid-job-id` starts climbing |
| rotating-tip | 1, 2 | adds the cross-tip path |

Six runs. Without rotations, `mark_past_jobs_as_stale` never fires, so the rare-tip arm has no cross-tip path by construction and its losses are eviction only. The rotating arm has both. That gives two readings of the same question: the error-code split says **which** mechanism, and **the rotating-minus-rare rejection rate at equal cap says how much is cross-tip loss** — eviction appears in both arms and cancels.

That subtraction assumes eviction is equal across arms, which is close but not exact: each rotation drains past into stale, so the ring briefly can't evict while it refills, and the rotating arm suppresses eviction slightly. The bias runs against attributing loss to the past/stale coupling, so a signal that survives it is the stronger result.

The ladder is deliberately low. Eviction can only fire when a share arrives more than `cap × template_period` after retirement — 50 seconds of latency at cap 50 and a 1 s cadence, which is unreachable — so a ladder starting at 8 would report `invalid-job-id ≈ 0` at every rung and prove nothing. Caps 1 and 2 are where the model says eviction must start biting, and they're the caps worth shipping; 50 stays as the tie-back to existing data. Low caps fill in seconds, so these are the cheap arms. Rotation rate gets reported alongside every `stale-share` figure, since rotating faster than mainnet inflates cross-tip loss.

Predictions, registered before the run:

| if the loss is… | `invalid-job-id` | `stale-share` | rotation sensitivity | implication |
|---|---|---|---|---|
| eviction | rises as cap falls, knee near `1 + ceil(L/T)` | flat | low | `past_jobs` stays deep; memory unavoidable |
| cross-tip loss | flat | rises as cap falls | high | `past_jobs` → ~2; size the stale window separately |
| neither | flat | flat | — | the cap isn't the cause; the default can drop to the required depth |

The third row is a real outcome, named in advance so it isn't treated as a failed run: rejection rates flat across the whole ladder would mean the earlier 2.7 points came from the one-shot workload rather than from the cap, and that nothing on the yield side argues against a small default.

**Recommendation: keep the default at 50 until this runs.** It's the safer side of a trade-off whose downside is unquantified below it. We wouldn't argue against a smaller cap on the eviction path; we can't vouch for it on the cross-tip path.

When it does run, the knee is a floor rather than a default. It's measured against one latency distribution and one cadence, so a shipped default wants headroom above it — and since the knee moves with both, the parameter this PR adds probably matters more than whichever number we end up recommending.

Contributions:

- **Measured** job retention at ~4.5 kB per retained job per channel, the largest per-connection memory lever we've found.
- **Derived** the required depth as `1 + ceil(latency ÷ template period)`, showing depth is set by latency rather than share rate or connection count.
- **Implicated** the past/stale coupling rather than eviction, by splitting rejections per error code — a lead the experiment above is built to confirm or kill.
- **Specified** a six-run experiment that separates the two loss paths, with predictions registered in advance.

sv2/channels-sv2/src/server/jobs/job_store.rs
gimballock: yes sorry, i was slow to put together a good experiment proposal. 
See new comment below.

gimballock:
This work is progressing, I have a solid theory of what to expect from the test. I have a physical mining test running now to validate against expectation. In short I think our setting of 50 is way too high and the better value is closer to 2 (as a conservative default value). 

A lot of the analysis yesterday went into deciding if the stale share buffer and the past jobs buffer should be split into two and sized independently but eventually I realized that at realistic rates on mainnet the two buffers should be small and in approximately equal proportion.

I'll post more today once the hw results are in.

and i'm trying to get just enough of the analysis together to be informative w/o overwhelming.

gimballock:
## Hardware A/B for the default value — now running

Following up on the plan above: the run that decides the default just started on real
mining hardware. Results in a few hours.

### The question

How many past jobs must a pool retain per channel? The `channels_sv2` default is 50;
#2290 raised whether that's 25× larger than necessary. The cap costs ~4.5 kB per retained
job **per channel**, so at scale it's real memory. Retaining too few loses creditable
work: a late share whose job has been evicted is rejected `invalid-job-id` even though it
was valid.

### The design

Two pools on one host, sharing one Bitcoin node, **identical in every setting except the
cap**:

| | Arm A | Arm B |
|---|---|---|
| `max_past_jobs` | **2** | **50** |
| miner | same model, ~215 TH/s | same model, ~245 TH/s |
| firmware | cgminer-derived (bmminer lineage) | *byte-identical build* |
| shares/min, batch size, cadence | 600, 10, 6.0 s | same |

Each arm: one real SV1 miner → its own translator → its own pool. Same chain tip, same
mempool, same template timing. Both pools attest their cap in their startup log, so an arm
silently running the wrong value would be caught.

Network: testnet4, templates driven by the node over IPC with the fee threshold at 0 and a
5 s minimum interval, yielding a measured 6.0 s template cadence in both arms.

### The mechanism under test

```
cap 2 retains:   1 active job + 2 past  =  3 creditable jobs
cgminer caches:                            3 jobs
```

**Cap 2 exactly covers a 3-job cache, with zero margin.** That's the substance of the
"floor is 2" claim. The cap is really a retention *window* — `cap ÷ template rate` — so at
6.0 s cadence arm A tolerates 12 s of staleness and arm B tolerates 300 s. Anything arm A
loses is a share delayed past its cache depth by latency.

This is why real miners matter: a simulated miner that only ever submits against the
current job can never exercise the cache, which is why earlier attempts measured nothing.

### What we measure

Pool-side `invalid-job-id` per arm, as a rate per submitted share. At ~600 shares/min/arm:

| Elapsed | Shares/arm | Detects ≥ | If zero, 95% bound |
|---|---|---|---|
| 2 h | 72,000 | 0.007% | < 0.004% |
| **4 h** | 144,000 | 0.0035% | < 0.002% |

Reference point: the same units' long-run stale-share rate against their normal production
pool is ~0.145%, so 4 h resolves roughly 70× below it.

### What each outcome means

* **Zero in both** — the expected result. Cap 2 suffices at this cadence, and the bound is
  the deliverable: "0 in 144,000 shares, loss < 0.002%." Supports lowering the default.
* **Arm A > arm B** — cap 2 loses creditable work; the floor is above 2, and the size of
  the gap gives the required depth. This would contradict the current recommendation.
* **Both nonzero and equal** — not cap-related; something else is rejecting shares, and the
  run says nothing about retention.

### What it cannot tell us

One firmware build, one template cadence, testnet4, two miners. It does not establish
behaviour across other firmware, nor at production template rates — we deliberately run
~10× a typical share rate, which stresses job age *harder* than production would. That
asymmetry makes a null result stronger, and means a positive result would need a confirming
run at a normal share rate to rule out a queuing artifact.

Two known asymmetries, neither affecting job-cache behaviour: the miners sit in different
power modes (pre-existing, hence the hashrate difference), and one arm has a head start in
accumulated shares — so results will be reported as rates over the overlapping window, not
as raw totals.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

gimballock:
There are many details left out of that relating to why i think 2 is a good value but I can explain all of that if the test affirms that hypothesis, otherwise you probably don't care about my false logic.

But part of the estimate is based on our belief of how many jobs a miner can be working on in parallel.
This is partly answered in [SV2 spec §5.3.15] "this mining job is active and miner must start mining on it immediately", and anecdotally I was told by one of our firmware engineers that cgminer will cache the most recent 3 jobs. I think that modern firmwares prefer to stick to one job at a time but I don't really know. I *think* cgminer probably covers a big portion of what we will see but since spec says you should always be working on the newest job then we don't need to honor these older submissions from weird firmwares (probably old sv1)

plebhash:
thanks for the deep analysis @gimballock 

> if the test affirms that hypothesis

please let us know which conclusion you arrive to

---

now some reflection: if we conclude that 2 is the optimal value (maybe 3 or 4 to give some headroom), does it really make sense to make it customizable?

I mean, I can see the argument for making it customizable if we concluded that some arbitrary-looking value (e.g.: 16, 50 or 300) was the optimal

but 2 feels more "canonical", in the sense that there's not much room for configurability or customization

unless we can think of use-cases in which that would make sense

gimballock:
So the test passed (i'll post an ai synopsis in a second), a buffer of 2 works great on the miner I used (S21-class). So i modified the template provider to increase the template rate from 6s/template to 2s/template to try and stress the buffer more but the results of that were similar. This miner is so efficient that it doesn't ever need more than 2 slots of this buffer.

But then I scanned the source of our firmware and found some per-chip logic that affects the central buffer inside the firmware and realized that the real test is to use an older miner. Which I don't have access to at the moment.

Doing some analysis on different chip speeds and chip densities leads me to believe that on older hw a deeper buffer is absolutely helpful but I can't easily verify that at the moment in hardware. But a reasonable analysis from plugging in chip speeds read from their spec sheets suggests that the original guess of 16 is reasonable.

AI analysis post to follow...

gimballock:
## Detailed analysis behind the summary above

**A cap of 2 cost nothing in 211,215 real shares. That pins the mechanism, not the number — so the
recommendation here is to keep the default at 50 for now, ship the knob, and let one real-hardware knee
test set the value.** Our own estimate of the right cap has moved 8× across three passes over the
firmware, which is itself the argument for not shipping a number yet.

Below, in order: two completed hardware runs, what the mechanism review does and does not establish, a
completed 579-million-share passive survey - and then the one piece still in flight, a cap ladder on an
older miner that finishes tomorrow.

### The two runs

Two pools on one host sharing one Bitcoin node, identical in every setting except the cap, one real SV1
miner per arm through its own translator. Same chain tip, same mempool, same template timing. Both pools
attest their cap at startup, so an arm running the wrong value would be caught. Rejection counters carried
8 live series per arm throughout, so the zeros are measurements rather than absent instrumentation.

**Both runs are complete.** Phase 2 repeated phase 1 with one variable changed: the template cadence was
cut from 6.0 s to 2.01 s, shrinking the cap-2 retention window from 12 s to 4 s.

| | Phase 1 (complete) | | Phase 2 (complete) | |
|---|---|---|---|---|
| | Arm A | Arm B | Arm A | Arm B |
| `max_past_jobs` | 2 | 50 | 2 | 50 |
| Template cadence | 6.0 s | 6.0 s | 2.01 s | 2.01 s |
| Retention window | 12.0 s | 300 s | 4.0 s | 100 s |
| Submitted | 65,918 | 80,419 | 145,297 | 145,042 |
| `invalid-job-id` | **0** | **0** | **0** | **0** |
| `stale-share` | 1 | 0 | 1 | 2 |

Phase 1 ran 2 h 9 min (124 samples); phase 2 ran 3 h 39 min (220 samples) and finished before the run
described at the end of this comment began. Phase 2's arms match to within 0.2% on sample size.

```
cap 2, combined:  0 invalid-job-id in 211,215 shares
95% upper bound:  3/211,215  =  0.0014%
```

The `stale-share` counts are a different mechanism and confirm it: arm B, with the *larger* stale window,
recorded **more** of them (2 vs 1). That's the opposite of a cap effect, and 1-2 events across ~22
testnet4 block boundaries is what prev-hash changes alone predict.

### The mechanism: established, and not

**Established.** The structure that matters is not a job cache but a **work ring**. The miner hashes only
the newest job, yet each work item retains its own `job_id`, and a returning nonce is submitted under
*that* id rather than the current one. The only staleness gate is `clean_jobs` on the newest job - no age
check, no job-count window. Late submissions therefore happen by design, and the pool-side cap decides
whether they are credited.

**Not established: whether the ring is one space per box or one per hashboard.** This is a 3× factor on
every number below, and we have not resolved it.

* The evidence for a single device-wide space is Bitmain's own cgminer, where a returned nonce is just
  `{work_id, nonce}` with no board field and one device-wide work table, the board being recovered from
  the nonce bits. **Caveat: that is the 2013 S1-S4 generation with a 32-bit `work_id`.** The miners at
  issue are S17-S21 with a byte-sized id. We resolved the old architecture and assumed it carried forward.
* A single-chip open firmware shows the same device-wide shape, but a one-chip device cannot testify
  either way - its single table fits both readings equally.
* No modern **multi-board** driver has been read. That is the gap.

**Also not established: the ring size `N`.** 16 for modern parts and 32 for the oldest family both come
from single-chip send loops. A multi-board controller could partition the id space differently.

So the shape of the answer is known and the constants are not:

```
global reading:     required_cap ~ ceil( N / (push_rate * T) )
per-chain reading:  required_cap ~ ceil( N * boards / (push_rate * T) )
```

`push_rate` is measurable as `Local Work / Elapsed`. **With one assumption we did not verify:** that each
tick advances one job id on a device-wide counter. If ticks are per-board, or a tick is not one id
advance, the measured rate is not the ring-cycle rate.

Measured push rates, and what each reading implies:

| Unit | push_rate | N | global depth | per-chain depth | cap @6 s (global) |
|---|---|---|---|---|---|
| current-gen, 3 boards | 2.08/s | 16 | 7.7 s | 23.1 s | 2 |
| current-gen, 2 boards | 1.28/s | 16 | 12.5 s | 25.0 s | 3 |
| older 32-slot, 3 boards | 1.82/s | 32 | 17.6 s | 52.8 s | 3 |

Across the older family generally, lower push rates move the global figure to roughly 6-11 at a 6 s
cadence. Either way the mechanism floor is low, and about 3× lower under the global reading.

One correction worth recording, because we published the opposite earlier: **board count cancels out of
the search-space calibration.** `S = H_chain * refresh_chain = (H / b) * (b / rate) = H / rate`. Board
count affects ring *depth*, not `S`. And `S` is not transferable - computed per unit it comes out 1.05e14,
1.29e14 and 0.53e14 across three models, a 2.4× spread. It is a per-model number at best.

### Passive corroboration: what the tail actually looks like

`Pool Stale%` is a passive sample of the submission-age tail - a stale share *is* a late submission, one
that crossed a prev-hash boundary. With boundaries roughly Poisson at `T_block` and ages far smaller, the
fraction crossing one gives mean age directly: `mean_age ~ stale_rate * T_block`.

A connect-sweep of two /16s (130,048 addresses) identified 17,988 miners. Restricting to units with
substantial history on the **same production pool**, so cadence and block rate are held constant:

| Firmware family | Units | Shares | Stale% | **Derived mean age** |
|---|---|---|---|---|
| cgminer-derived | 3,303 | 285.2M | 0.1853 | **1.11 s** |
| different vendor | 13,209 | 293.7M | 0.0610 | **0.37 s** |

579 million shares. **Mean submission age is ~1 s** - far below any ring-depth figure above, under either
reading. That is why cap 2 came back clean, and it is the strongest evidence that the *typical* case needs
almost nothing.

It also shows **firmware family matters 3×**: the second vendor submits late shares at one third the rate.
A pool cannot assume one behaviour across the miners connecting to it.

What this cannot do is measure the **tail**, which is what a cap has to cover. Ages here are ~1 s against
~600 s boundaries, so the method is sensitive to the mean and blind to the shape between 1 s and 50 s -
exactly the band in dispute. `clean_jobs` gating also suppresses some late submissions before they can
become stale shares, so the rate is a floor. And coverage is bounded by reachability: a third range known
to hold miners returned 0 of 512 probes from our vantage point.

### What we recommend, and what would change it

**Keep the default at 50. Ship the knob.**

Three reasons, none of them "50 is optimal":

1. **Global-vs-per-chain is unconfirmed on modern multi-board hardware**, and it is a 3× factor.
2. **The tail is unmeasured.** Everything we have constrains the mean, and the mean is ~1 s. The cap
   exists for the tail.
3. **Our own estimate has swung 8× in three passes** over the firmware - 1-2, then 4-16, then 2-6. Each
   swing came from better evidence, which is healthy, but it means any value we name today is provisional.
   Shipping a smaller default on that basis trades a silent, asymmetric failure for a saving we cannot yet
   size.

The one thing we would argue for regardless: **the correct value is `window / cadence`, and cadence
belongs to the deployment, not the library.** That rule reproduces both historical constants at the
cadences each implied - a ~100 s window gives ~17 at 6 s and 50 at 2 s. Neither integer was arbitrary;
they are one rule at different template rates. That is what the per-channel parameter in this PR makes
usable, and it holds whatever the constants turn out to be.

For scale, at ~4.5 kB per retained job per channel: 50 is 225 kB/channel or ~24.8 GB across 110k channels;
16 is 72 kB and ~7.9 GB; 2 is 9 kB and ~1.0 GB. The saving is real, which is why it is worth measuring
rather than guessing.

### The run that decides it - still in flight

Everything above is finished and reported. This last section is the only part not yet measured, and it
uses **different hardware from phases 1 and 2**: those ran on current-generation miners, this one runs on
an older 32-slot part.

The scan turned up one such unit among the ~18,000 it identified - an S19j Pro, 96 TH, 3 boards, measured
push rate 1.82/s. It is mining a pool we control over SV2, and a cap ladder is under way:
**cap 1, 2, then 8 at a 2.01 s cadence**, six hours each, ~216,000 shares per rung, resolving a loss rate
to 0.0014%. **It completes tomorrow.**

Two things make it worth waiting for.

**It measures the required cap directly**, so it does not inherit the `N` or global-vs-per-chain
assumptions. If loss appears at one rung and vanishes at the next, that locates the effective window
whatever the constants are.

**The cap-8 rung discriminates the two readings.** Its 16.1 s window sits just below the global prediction
(17.6 s, marginal) and far below the per-chain one (52.8 s, clear loss). A clean cap-8 rung is evidence
for global; a lossy one is evidence for per-chain.

| Outcome | What it establishes | Effect on the default |
|---|---|---|
| Loss at cap 1, clean by cap 8 | knee measured on 32-slot silicon; global reading supported | 50 can likely come down to ~8-16 |
| **Zero at every rung** | tail stays thin even behind a deep ring | 50 can likely come down to ~2-4 |
| Loss at every rung | something other than the cap is rejecting shares | unchanged; run is uninformative |

Limits, so the result is not over-read: **n = 1**, so no statistical weight; one chip family, so silent
about other older parts; sequential arms, so chain conditions differ between rungs. One strong data point
on one older chip, not a characterisation of older hardware.

**If you have S17/S19-class hardware, the ladder is cheap to repeat and these are the predictions to
break.** A second unit would also let the arms run concurrently and remove the sequential confound.

Deliberately not attempted: derating a modern miner to emulate a deep ring. A unit found running 1 of 2
boards showed its push rate decoupling from nonce exhaustion, so an artificially slowed miner is not a
valid stand-in for a genuinely slower chip.

### Contributions

* **Measured** a cap of 2 against real hardware at two template cadences - zero `invalid-job-id` in
  211,215 shares, bounding eviction loss below 0.0014%.
* **Identified** the mechanism that creates the requirement: a work ring whose entries outlive the job
  they belong to, with `clean_jobs` as the only gate.
* **Derived** mean submission age from 579 million production shares - ~1 s, far below any ring-depth
  estimate, which is why small caps survive in practice.
* **Quantified** a 3× spread in late-submission rate between firmware vendors across 16,512 units.
* **Bounded** what a code review can settle: the ring's existence and rough size, but not whether it is
  per-box or per-board on modern multi-board hardware, and not the tail.
* **Retracted** two of our own claims - that board count enters the search-space calibration, and that
  the per-chain reading was established.
* **Withheld** a numeric recommendation, keeping 50 until the running knee test lands rather than shipping
  a figure that has already moved 8×.

Sources: [driver-bitmain.h](https://github.com/bitmaintech/cgminer/blob/master/driver-bitmain.h),
[bm1370.c](https://github.com/bitaxeorg/ESP-Miner/blob/master/components/asic/bm1370.c)

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

gimballock:
> thanks for the deep analysis @gimballock
> 
> > if the test affirms that hypothesis
> 
> please let us know which conclusion you arrive to
> 
> now some reflection: if we conclude that 2 is the optimal value (maybe 3 or 4 to give some headroom), does it really make sense to make it customizable?
> 
> I mean, I can see the argument for making it customizable if we concluded that some arbitrary-looking value (e.g.: 16, 50 or 300) was the optimal
> 
> but 2 feels more "canonical", in the sense that there's not much room for configurability or customization
> 
> unless we can think of use-cases in which that would make sense

agreed, a handful of slots doesn't need configuration.

The analysis is suggesting that measuring the needs of the oldest/slowest hw is needed to see any differences in buffer-size b/c everything modern doesn't seem to need much of a buffer. I found 1 s19j and am running a test on that but even it looks like it doesn't need much of a buffer (early read so far) but i don't have anything older on hand atm.

gimballock:
The final experiment finished last night, and interestingly the s19j had some thermal issues so it was under-clocked which should make it easier to see the buffer depth. This test finally shows some invalid-job-ids, proving the buffer cap was too small, but not many. The evidence is a bit thin but we will take it.

@plebhash The resulting conservative default recommendation has come full circle now and we are landing where we started, but with theory+evidence, to 16.

Actually it needs about 16s of time divided by whatever rate templates come in, fastest configurable is 1/second but default on mainnet is 6/second. So that's the difference between a buffer of ~3 vs 16.  So 16 is recommended as the conservative worst case option and the configuration parameter, in many/most real world use cases, can be lowered safely to increase headroom (available memory) on their servers.

I'll most the results and explanation below.

gimballock:
## Results and explanation

The ladder ran three retention windows against the under-clocked S19j Pro overnight, ~6 h and ~200k
shares each. Numbers first, then what three events do and don't support, then how 16 falls out.

| window (cap) | submitted | `invalid-job-id` | rate |
|---|---|---|---|
| 2.01 s (cap 1) | 205,989 | 1 | 0.0005% |
| 4.02 s (cap 2) | 206,268 | 2 | 0.001% |
| 16.08 s (cap 8) | 215,572 | **0** | 0% |
| **total** | **627,829** | **3** | **0.0005%** |

**A correction first, because I published the wrong figures yesterday.** I reported
`invalid-job-id = 0` at cap 1 and cap 2. That was a harness artifact: my summary read each rung's counter
from its last sample, and the miner's channel reset three times per rung, so I was only ever reading the
final fragment. The same defect undercounted volume — rung 2 read 15,665 when it was 206,268. The table
above accumulates increments across resets and is the real result.

### Why the under-clocking helped

The unit thermally throttled from ~100 to ~63–71 TH/s partway through. That works in our favour: a slower
chip takes longer to exhaust each unit of work, so its work buffer spans **more** wall-clock time and its
submission-age tail is **fatter**. This was a harsher test than the same miner at nameplate would have
been, which makes the small numbers a stronger bound rather than a weaker one.

### What three events support

The events are real and they are asymmetric across the ladder: at the two tight windows a larger cap
would have credited work that was thrown away, and at a 16 s window nothing was. So a cap of 1–2 at this
cadence demonstrably costs something, and 16 s of retention demonstrably costs nothing. That much is
direct.

The cost is also very small, and worth putting next to the other rejection classes from the same run:

```
difficulty-too-low   896
stale-share            8
invalid-job-id         3     <- the only class the cap controls
```

**Cap-driven loss is the smallest rejection class by two orders of magnitude.** That is the single most
useful number here, because it means the default should be chosen on memory cost, not on yield.

### What three events do not support

They cannot rank the rungs or locate a knee. At one to two events the counts are Poisson noise and all
three rungs are statistically consistent with the same underlying rate. And the two non-zero points fall
in the *wrong* order for eviction — the tighter 2.01 s window logged fewer (1) than the 4.02 s window (2).
So I can bound the cost as tiny at every window down to 2 s; I cannot measure a slope, and I withdraw the
"knee confirmed" reading from my earlier comment.

The more useful negative result: **the worst-case ring arithmetic over-predicts.** For this throttled unit
the ring computes to ~48 s, which should have produced loss well past a 16 s window. None appeared. The
reachable tail is far thinner than the ring depth, so the ring is an upper bound the hardware does not
visit — which is why a small cap survives at all, and why the requirement is a matter of seconds rather
than the ring's full span.

### How 16 falls out

The smallest window with zero observed loss was **16.08 s**. The required count is that window divided by
the template period:

```
required_cap  =  ceil( ~16 s / template_period )
```

| template period | cap | window given | memory/channel |
|---|---|---|---|
| one every 1 s (fastest configurable) | **16** | 16.0 s | 72 kB |
| one every 2 s (this test) | 8 | 16.0 s | 36 kB |
| one every 6 s (typical mainnet) | 3 | 18.0 s | 13.5 kB |
| one every 10 s | 2 | 20.0 s | 9 kB |

A library default cannot know the deployment's cadence, so it has to cover the fastest configurable one —
which gives **16**. An operator who knows their own template period should set `ceil(16 / T)` and take the
memory back: a pool issuing one template every 6 s can run **3** and use 13.5 kB per channel instead of 72.

**Why not lower as the default.** On yield alone 3 is defensible — the sub-16 s cost measured 0.0005%.
But the trace losses sit at exactly the low end, the failure is silent and asymmetric, and a fixed count
under an unknown cadence is where a small number breaks: a cap of 3 against one template per second is
only a 3 s window, inside the region where loss appeared.

**Why not stay at 50.** At one template every 2 s that provisions ~100 s of retention, six times past
where loss had already vanished, for **3.1× the memory** — 225 kB versus 72 kB per channel, or ~24.8 GB
versus ~7.9 GB across 110k channels. Nothing measured supports 100 s: production mean submission age came
out at ~1 s across 579 million shares, and the tail beyond 16 s produced nothing on the most adverse
hardware we could find.

### Scope

One miner, one cadence, the eviction path only. The correction fixes the defect I know about, but at three
events the result stays instrument-limited: each channel reset is a discontinuity that could itself create
or mask an event, and one mis-captured reset moves the total by 100%. Treat 0.0005% as an order of
magnitude, not a figure to three places.

### What this run establishes

* **Measured** the eviction cost of a tight cap on real 32-slot hardware — 0.0005% at windows down to
  2.01 s, on a unit running well below nameplate.
* **Placed** cap-driven loss as the smallest rejection class in the run by two orders of magnitude, which
  moves the default from a yield question to a memory one.
* **Bounded** the practical retention requirement at ~16 s, against ~48 s from the worst-case ring
  arithmetic — the model over-provisions.
* **Derived** the default as `ceil(16 s / template_period)`, giving 16 at the fastest configurable cadence
  and 3 at a typical mainnet one.
* **Corrected** yesterday's published zeros and named the harness defect behind them.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

gimballock:
Perhaps a future refactor can find a simple way to isolate the (rate/distribution of solutions from the miners) from the (rate of templates from the tp). The former is measured and stable, the latter is a deployment configuration parameter (of an external component) so we can't directly read it, we could measure it at runtime but then how do you initialize the buffers and do we even want dynamic or lazy buffers?

The default is set now to 16, which is a very conservative value that should be reduced for anyone running modern miners. Perhaps a bitaxe would be an interesting followup test..

plebhash:
thanks @gimballock 

I order to reduce the back-and-forth and expedite this getting merged, I did some minor changes to your branch and pushed to https://github.com/plebhash/stratum/tree/feat/configurable-max-past-jobs

I wasn't able to force-push them here, can you cherry-pick them in favor of the current commits?

basically I reworded some comments and set `MAX_PAST_JOBS` to 16 on the client side as well

gimballock:
Done — your branch is the head now (`e3b3ea6e`). Two things in it beyond the reword, both improvements:

- rebased onto current `main` (`941d004a`); my head was still on `aa242700`;
- `client::MAX_PAST_JOBS` moved to 16 as well, which is the part I'd missed.

You were right that the client had to move. I'd only lowered the server side, which left the client's own doc contradicting itself — it justified 50 as "matches the server-side cap" while no longer matching it. The asymmetry was residue from an unfinished change, not a decision.

Checked the reasoning rather than just the wording, and the invariant holds in the direction you state it. A client retaining **fewer** jobs than its upstream loses creditable work. A client retaining **more** gains nothing and is arguably worse than useless: the proxy validates against a job the pool has already evicted, credits the share locally, forwards it, and the pool answers `InvalidJobId`. The miner is told "accepted" for work nobody pays for. So tracking the server default is the only value that's right in both directions, and the override is what handles an upstream that has raised its own cap.

The one place I thought the client genuinely needed headroom — an upstream JDC declaring often, so the client's job rate isn't bounded by any template cadence — your text already resolves: on a JDC the upstream channel's job rate is the client's own `SetCustomMiningJob` rate, which it does know and can size against. That's the mirror of the pool-side caveat, and the two now read consistently.

Verified locally on `e3b3ea6e`: 174 `channels_sv2` tests plus 3 doc-tests pass, `cargo fmt --check` clean, no new clippy warnings.

Companion #735 updated in the same pass — see the note there; the docs I'd written for it were predicated on the 50 you just removed.

plebhash (approved):

PR #2316, opened 2026-08-25
binary-sv2: bound and redact the ValueExceedsMaxSize sample

### LINKED ISSUE #2315: Avoid cloning the full frame for an invalid length prefix
expected_length_variable does data.to_vec() on the whole remaining frame for an over-limit length prefix → allocation amplification DoS

bit-aloo:
closes: #2315 
closes: https://github.com/project-loupe/audit-stratum/issues/108
closes: https://github.com/project-loupe/audit-stratum/issues/181
closes: https://github.com/project-loupe/audit-stratum/issues/177
closes: https://github.com/project-loupe/audit-stratum/issues/167

### DIFF
--- a/sv1/src/utils.rs
+++ b/sv1/src/utils.rs
@@ -1,5 +1,5 @@
 use crate::error::{self, Error};
-use binary_sv2::{B032Owned, U256Owned, B032};
+use binary_sv2::{B032Owned, U256Owned, B032, ERROR_SAMPLE_LEN};
 use bitcoin::hashes::hex::{DisplayHex, FromHex};
 use byteorder::{BigEndian, ByteOrder, LittleEndian, WriteBytesExt};
 use serde_json::Value;
@@ -216,7 +216,7 @@ impl TryFrom<&str> for PrevHash {
                         32,
                         0,
                         0,
-                        prev_hash_stratum_order,
+                        prev_hash_stratum_order[..len.min(ERROR_SAMPLE_LEN)].to_vec(),
                         len,
                     ),
                 ))
--- a/sv2/binary-sv2/src/datatypes/mod.rs
+++ b/sv2/binary-sv2/src/datatypes/mod.rs
@@ -39,7 +39,7 @@ pub(crate) use non_copy_data_types::Inner;
 pub use non_copy_data_types::{
     B016MOwned, B0255Owned, B032Owned, B064KOwned, Mac, MacOwned, PubKey, PubKeyOwned, Seq0255,
     Seq0255Owned, Seq064K, Seq064KOwned, Signature, SignatureOwned, Str0255, Str0255Owned,
-    Sv2Option, Sv2OptionOwned, U256Owned, B016M, B0255, B032, B064K, U256,
+    Sv2Option, Sv2OptionOwned, U256Owned, B016M, B0255, B032, B064K, ERROR_SAMPLE_LEN, U256,
 };
 
 use alloc::vec::Vec;
--- a/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
+++ b/sv2/binary-sv2/src/datatypes/non_copy_data_types/inner.rs
@@ -257,6 +257,11 @@ fn expected_length_variable<
     }
 }
 
+/// Upper bound on the offending bytes retained in `Error::ValueExceedsMaxSize`.
+///
+/// Retain one hash-sized prefix for diagnostics while keeping error allocations bounded.
+pub const ERROR_SAMPLE_LEN: usize = 32;
+
 fn validate_payload<
     const ISFIXED: bool,
     const SIZE: usize,
@@ -276,7 +281,7 @@ fn validate_payload<
                 SIZE,
                 HEADERSIZE,
                 MAXSIZE,
-                value.to_vec(),
+                value[..value.len().min(ERROR_SAMPLE_LEN)].to_vec(),
                 value.len(),
             ))
         };
@@ -288,7 +293,7 @@ fn validate_payload<
             SIZE,
             HEADERSIZE,
             MAXSIZE,
-            value.to_vec(),
+            value[..value.len().min(ERROR_SAMPLE_LEN)].to_vec(),
             value.len(),
         ))
     } else {
@@ -488,7 +493,11 @@ impl<const ISFIXED: bool, const SIZE: usize, const HEADERSIZE: usize, const MAXS
     type Error = Error;
 
     fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
-        Self::new(value.to_vec())
+        validate_payload::<ISFIXED, SIZE, HEADERSIZE, MAXSIZE>(value)?;
+
+        Ok(Self {
+            data: value.to_vec(),
+        })
     }
 }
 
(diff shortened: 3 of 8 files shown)


GitGab19 (commented):
Clanker review:

---

## Review summary

Overall, the implementation looks correct and directly addresses the reported allocation-amplification issue. The decoder now retains only the 1–3-byte length prefix instead of cloning the entire remaining frame.

### Non-blocking documentation issue

The new `ValueExceedsMaxSize` documentation describes its `Vec<u8>` as a “bounded diagnostic sample.” That is only true for the decoded-length path changed here.

Other paths in `validate_payload` still use `value.to_vec()`, potentially retaining the complete caller-provided invalid value. I suggest qualifying the documentation, for example:

> The `Vec<u8>` contains the offending value when supplied directly, or only the length prefix when an overflow is detected from a declared encoded length.

Hardening the other constructors against large invalid inputs could be tracked separately; I would not consider it a blocker for this PR.

---

Can you check if other constructors have a similar issue to the one you're solving on this PR?

bit-aloo:
Other places where we use it, is for our own internal construction and not by the caller generated payload.

GitGab19:
> Other places where we use it, is for our own internal construction and not by the caller generated payload.

Clanker answer to this:

---
Since it's a small change, let's handle the construction-side paths in this PR too rather than a follow-up: in `validate_payload`, replace both `value.to_vec()` calls with a capped sample, e.g. `value[..value.len().min(32)].to_vec()`. The full length is already reported in the last field of `ValueExceedsMaxSize`, so no diagnostic info is lost — and it makes the new doc comment ("bounded diagnostic sample") true for every path that builds this error. Please add a small test like your `oversized_length_error_retains_only_the_header` one covering `Inner::new` / `InnerOwned::new` with an oversized value.

GitGab19 (commented):
## Requested changes

The new `ERROR_SAMPLE_LEN = 32` limits how many bytes of an invalid value are retained inside `ValueExceedsMaxSize`.

The value `32` is **not a protocol limit**. It is simply a diagnostic-memory cap: errors retain the first 32 bytes for debugging while the final field of `ValueExceedsMaxSize` records the complete input length.

### Validate before creating an owned copy

`TryFrom<&[u8]> for InnerOwned` still copies the complete input before validating it:

```rust
Self::new(value.to_vec())
```

A large invalid value therefore causes a large temporary allocation even though the resulting error retains only 32 bytes.

Please validate first and allocate only when the value is valid:

```rust
fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
    validate_payload::<ISFIXED, SIZE, HEADERSIZE, MAXSIZE>(value)?;

    Ok(Self {
        data: value.to_vec(),
    })
}
```

The variable-array conversions should similarly validate before allocating, preferably by delegating to the slice implementation:

```rust
fn try_from(value: [u8; N]) -> Result<Self, Self::Error> {
    Self::try_from(&value[..])
}

fn try_from(value: &[u8; N]) -> Result<Self, Self::Error> {
    Self::try_from(&value[..])
}

fn try_from(value: &mut [u8; N]) -> Result<Self, Self::Error> {
    Self::try_from(&value[..])
}
```

Please also exercise the owned-from-slice path in the regression test:

```rust
match B032Owned::try_from(oversized.as_slice()).unwrap_err() {
    Error::ValueExceedsMaxSize(false, 1, 1, 32, retained, 1_048_576) => {
        assert_eq!(retained.len(), ERROR_SAMPLE_LEN);
    }
    other => panic!("unexpected error: {other:?}"),
}
```

### Clarify the documentation

“Never a full copy” is not strictly correct because invalid values of 32 bytes or fewer are copied completely. Suggested wording:

> The `Vec<u8>` is a bounded diagnostic sample containing at most the first 32 bytes of the offending value, or only the length prefix when the overflow is detected from a declared encoded length. The final field reports the complete offending length.

The constant should also explain why 32 was selected:

```rust
// Retain one hash-sized prefix for diagnostics while keeping error allocations bounded.
const ERROR_SAMPLE_LEN: usize = 32;
```

GitGab19:
Last clanker review:

---

### Must fix

**Missing tests for the redaction path.** This PR closes audit #177 and #181, and each of those ships a PoC test (`binary_error_display_does_not_dump_embedded_payload`) that isn't included here. Nothing currently asserts that `Display` of a `ValueExceedsMaxSize` carrying a large `Vec` stays bounded — which matters because the variant is public and can be constructed with an arbitrary `Vec` by any downstream crate, independent of `ERROR_SAMPLE_LEN`. Please add the two PoC tests (codec and framing) essentially verbatim, plus one in `binary-sv2` for the `Display` impl itself.

### Nits (non-blocking)

- In `oversized_value_error_retains_only_a_capped_sample`, three asserts use a literal `32` and one uses `ERROR_SAMPLE_LEN`. The `32` in the match patterns is `B032`'s `MAXSIZE`, a different 32 that only coincidentally equals the sample cap — use the constant for all four `retained.len()` asserts so the two aren't conflated.
- `sv1/utils.rs` hardcodes `32` because `ERROR_SAMPLE_LEN` is private to `inner.rs`. Consider exporting it from `binary_sv2` and reusing it, so the enum doc's "at most the first 32 bytes" claim has a single source of truth.
- `parsers-sv2/src/error.rs:37` still formats `binary_sv2::Error` with `{e:?}`. That's tracked as a separate Loupe finding so it's fine to leave, but with the new `Display` impl it's now a one-character fix (`{e:?}` → `{e}`) if you want to fold it in.

--- 

A personal comment: by looking at the PR title, description and commit, it's not clear what this is doing, especially because we have everything in one commit, where we're touching multiple crates, even though we're closing 5 issues.

Split changes in multiple commits, with a proper description.

The clanker suggested this structure, feel free to do it differently:

1.
```
binary-sv2: retain only the length prefix in oversized-length errors

expected_length_variable cloned the entire remaining frame into
ValueExceedsMaxSize when a length prefix exceeded MAXSIZE. The slice is
the decoder's remaining input, so a peer could force an allocation and
copy proportional to the frame with a single invalid byte. Keep only the
1-3 byte prefix; the declared length is already reported separately.

Closes #2315
Closes project-loupe/audit-stratum#108
```

2.
```
binary-sv2: cap the value sample stored in ValueExceedsMaxSize

validate_payload copied the complete rejected value into the error, so
validating an untrusted slice allocated a second buffer proportional to
the input, and TryFrom<&[u8]> for InnerOwned allocated once more before
validating. Retain at most ERROR_SAMPLE_LEN bytes for diagnostics (the
final field carries the full length) and validate before allocating.

Closes project-loupe/audit-stratum#167
```

3.
```
binary-sv2: redact the diagnostic sample from Display output

codec-sv2 and framing-sv2 formatted binary_sv2::Error with Debug, which
writes every byte of a ValueExceedsMaxSize sample into logs. The variant
is public, so a downstream crate can still construct it with an
arbitrary Vec regardless of the in-crate cap. Add a Display impl that
prints only the scalar fields and switch both consumers to it.

Closes project-loupe/audit-stratum#177
Closes project-loupe/audit-stratum#181
````

4.
```
sv1: bound the prev-hash sample stored in ValueExceedsMaxSize

The rejected value comes from a hex field in a peer-supplied JSON
message, so keep the same 32-byte cap used by binary-sv2.
```

GitGab19 (approved):

Issue #2272, opened 2026-08-05
Slice derives Clone, allowing pool memory reuse while clones are alive


bit-aloo:
Slice derives Clone, but cloning does not update the pool's shared state. Dropping either copy marks the slot as free, allowing the pool to reuse the underlying memory while another clone still references it.

This can cause the surviving clone to read overwritten data or create conflicting mutable references through as_mut, resulting in memory-safety issues from safe Rust.

The Clone derive should be removed or cloning should correctly update the pool's reference tracking.

PR #2317, opened 2026-08-26
Refactor codec and framing

### LINKED ISSUE #2115: Add channel_id API to framing_sv2
The spec explicitly defines that when channel_msg is set, the first 4 bytes of the payload are a channel_id. Framing should expose this.

### LINKED ISSUE #2276: framing_sv2: refactor

### LINKED ISSUE #2277: Integer truncation in Sv2Frame::from_message bypasses the U24 length check
Sv2Frame::from_message casts message.get_size() from usize to u32 before validating that it fits in the 24-bit frame length field. On 64-bit targets, sizes larger than u32::MAX can be silently truncated to a small value and incorrectly pass the U24 check.

This can produce a frame whose header declares a much smaller payload than its actual serialized size, potentially desynchronizing the frame stream.

The size should be validated before performing the narrowing conversion.

### LINKED ISSUE #2280: HandshakeRole::Clone can cause transport nonce reuse
HandshakeRole derives Clone, allowing an in-progress handshake state, including its ephemeral key and transcript, to be duplicated. Two cloned initiators can complete using the same responder message and produce transport codecs with identical keys and nonce state.

This can cause nonce reuse when both sessions encrypt data, breaking the security assumptions of the transport encryption.

HandshakeRole should not be clonable once it contains handshake state.

### LINKED ISSUE #2288: Make codec state transition compile time invariant
Currently, codec state transitions happen at runtime, which can lead to panics and makes it harder to understand how the codec evolves from `NonInitialized` → `Handshake` → `Transport`. Make this compile time invariant.

### LINKED ISSUE #2327: Decoders allocate peer-declared frame lengths eagerly, with no cap
Both decode paths copy a peer-controlled length into the missing-bytes counter, and the next writable() passes it to a growing buffer before any payload arrives. Unencrypted: decoder.rs:398 takes  SizeHint::Missing, so the 6-byte header 00 00 00 ff ff ff reserves ~16 MiB. Noise: decoder.rs:295 assigns encrypted_payload_length(&header), ~16.8 MiB plus per-chunk MACs. A peer stalls after the header and repeats across connections. Mitigating factor worth recording: the count is returned as MissingBytes(n) / writable_len() before allocation, so a caller enforcing its own cap is safe  but the in-tree examples follow the uncapped writable() loop. Fix: a configurable maximum frame size on the decoder, or a bounded incremental read window.

### LINKED ISSUE #2342: Noise decoder reusable after payload authentication failure
After successfully decrypting a frame header, if payload decryption fails with an AeadError, the decoder clears its buffer but does not invalidate itself. Since the cipher remains at the payload nonce while the decoder resets to expecting a header, a caller that continues using the connection can end up decrypting valid payload ciphertext as header ciphertext.

An active attacker could exploit this by corrupting a payload to trigger the failure and then replaying the original payload ciphertext, potentially causing an attacker-controlled frame to be emitted. Callers that disconnect on AeadError are not affected.

### LINKED ISSUE #2370: Reject oversized frame serialization destinations
Sv2Frame::serialize does not enforce an exact destination length when serializing an in-memory payload. The header writer fills only six bytes, the payload writer fills only its encoded size, and both returned byte counts are discarded before the method returns Ok(()). If dst is larger than encoded_length(), its tail remains unchanged. A downstream relay or service that reuses a fixed-size buffer and transmits the full successful destination can therefore send stale plaintext from a prior message or tenant to its peer; an attacker able to induce a short outgoing frame can maximize the exposed tail. The in-tree encoders request exactly encoded_length(), so direct repository reachability is constrained to downstream callers of this public API and severity is Low. A fix should reject non-exact destinations, clear the tail, or return the actual written length.

bit-aloo:
closes: https://github.com/stratum-mining/stratum/issues/2276
closes: https://github.com/stratum-mining/stratum/issues/2277
closes: https://github.com/project-loupe/audit-stratum/issues/46
closes: https://github.com/stratum-mining/stratum/issues/2115
closes: https://github.com/stratum-mining/stratum/issues/2280
closes: https://github.com/project-loupe/audit-stratum/issues/60
closes: https://github.com/stratum-mining/stratum/issues/2288
closes: https://github.com/project-loupe/audit-stratum/issues/183
closes: https://github.com/project-loupe/audit-stratum/issues/133
closes: https://github.com/project-loupe/audit-stratum/issues/182
closes: https://github.com/project-loupe/audit-stratum/issues/179
closes: #2327 
closes: #2342 
closes: https://github.com/project-loupe/audit-stratum/issues/219
closes: #2370 
closes: https://github.com/project-loupe/audit-stratum/issues/220

companion https://github.com/stratum-mining/sv2-apps/pull/806


GitGab19 (commented):
First round of clanker review:

---

## Review

The direction here is good: splitting `Sv2Frame` (outgoing, message-holding) from `SerializedSv2Frame` (incoming bytes), replacing the runtime `State` enum with `Handshake<R>` / `Transport` typestates, making `EncodableFrame` a trait, and pulling `noise_sv2` out of `framing_sv2`'s dependency graph are all real improvements. The handshake wrapper routing, chunk math (65519+16 on the encoder vs 65535 on the decoder), `Transport::split` mirroring and `encrypted_payload_length` all check out against the old code, and the PR builds and passes tests under every feature combination.

Below is everything I found, split into (a) things I'd fix before merging, (b) pre-existing issues that live in code this PR rewrote, and (c) nits. Items marked **confirmed** were reproduced by execution at `7ff8e04`.

---

### A. Should fix before merge

#### 1. `WithoutNoise::encode` leaks reserved bytes on error
[`sv2/codec-sv2/src/encoder.rs#L218`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/encoder.rs#L218)

`get_writable(len)` advances the cursor, then `item.encode_into(writable)?` returns without draining. The PR added exactly this cleanup to `NoiseEncoder::encrypt_frame` but not here, while at the same time making `EncodableFrame` a public, user-implementable trait.

**Confirmed** (both buffer modes): after one failed `encode`, the next 6-byte frame comes out as 16 bytes (`[0,0,0,0,0,0,0,0,0,0, 0,0,1,0,0,0]`) with the stale region prepended; the peer's `StandardDecoder` parses two spurious empty frames before the real one and the stream is desynchronised with no error on the sender side.

#### 2. The read-window cap silently breaks the `MissingBytes(n) == writable().len()` contract
[`sv2/codec-sv2/src/decoder.rs#L420`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L420) and [`#L202`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L202)

Before this PR `writable_len()` returned `missing_b` verbatim, so `MissingBytes(n)` and `writable().len()` always agreed. Now `writable()` is capped at `SV2_FRAME_CHUNK_SIZE` while `n` still reports the full remainder. Every in-repo bench still sizes its copy from `n`:

- `benches/noise_roundtrip.rs:99` — `w[..n]`
- `benches/decoder.rs:34`, `benches/pool_lifecycle.rs:87`
- `benches/buffer_exhaustion.rs:309/340/362/409/431/470`

It's latent today only because bench payloads are < 2 KiB; any frame with payload > 65535 panics (index OOB / `copy_from_slice` length mismatch). Downstream read loops (sv2-apps) are written to the same old contract.

Suggestion: make the error carry what the decoder will actually accept next (e.g. `MissingBytes(self.writable_len())`, or `MissingBytes { total, next_read }`), or at minimum fix the benches and call out the contract change in the changelog.

#### 3. `try_encrypt_frame` panics on a foreign `EncodableFrame` shorter than a header
[`sv2/codec-sv2/src/encoder.rs#L140`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/encoder.rs#L140)

`&sv2[..SV2_FRAME_HEADER_SIZE]` (and `sv2.len() - start` at L145) trust whatever `encoded_length()` reports. The trait docs explicitly invite external impls; a `Raw(Vec<u8>)` impl with 4 bytes panics inside `encode_transport` with `range end index 6 out of range for slice of length 4` instead of returning `Err`.

#### 4. `fuzz/Cargo.lock` was not regenerated
[`fuzz/Cargo.lock#L286`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/fuzz/Cargo.lock#L286)

`fuzz` is excluded from the root workspace, so the root lock doesn't cover it. It still pins `parsers_sv2` under `fuzz-tests` and `tracing` under `codec_sv2`, both removed from the manifests by this PR.

**Confirmed:** `cd fuzz && cargo metadata --locked` → `error: cannot update the lock file fuzz/Cargo.lock because --locked was passed`. Fuzz CI stays green only because it runs without `--locked` and silently rewrites the committed file.

#### 5. New crate-level docs break rustdoc on default features
[`sv2/codec-sv2/src/lib.rs#L18`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/lib.rs#L18)

The new paragraph links `` [`state`] ``, `` [`Handshake`] ``, `` [`Transport`] ``, `` [`NoiseEncoder::encode_handshake`] ``, `` [`StandardNoiseDecoder::next_handshake_frame`] ``, `` [`NoiseEncoder::encode_transport`] `` and `` [`StandardNoiseDecoder::next_transport_frame`] ``, all `cfg(noise_sv2)`-gated.

**Confirmed:** `RUSTDOCFLAGS=-D warnings cargo doc -p codec_sv2 --no-deps` fails on default features and on `--no-default-features` (7 new unresolved links, 12 total). Docs CI only builds with `with_buffer_pool,noise_sv2` so it doesn't catch it. Gate the paragraph with `#![cfg_attr(feature = "noise_sv2", doc = ...)]` or drop the brackets.

#### 6. Fuzz harnesses became tautologies; write path is now untested
[`fuzz/fuzz_targets/deserialize_sv2frame.rs#L7`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/fuzz/fuzz_targets/deserialize_sv2frame.rs#L7), `deserialize_stdframe.rs:7`

`let serialized = frame.as_bytes().to_vec();` is the input by construction, so the final `assert_eq!` can never fail. The old harness at least round-tripped through `serialize`. Combined with:

- the deleted `prop_header_encrypted_len_calculation` quickcheck has no replacement for `codec_sv2::encrypted_payload_length`;
- no unit test covers `SerializedSv2Frame::encode_into`, `NoiseEncoder::encode_handshake`, or the U24 rejection in `Sv2Frame::from_message` (the new `from_message_rejects_a_size_that_does_not_fit_in_a_u32` uses `u32::MAX + 2`, which fails at the new `u32` check and never reaches the U24 check; the `HugeMsg` scaffolding makes a `get_size() == 16_777_216` case one assertion away).

#### 7. Doc drift from the rename
- [`sv2/parsers-sv2/src/tlv/list.rs#L91`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/parsers-sv2/src/tlv/list.rs#L91) — rewritten by this PR to say `Sv2Frame::from_bytes()`, which no longer exists; it's `SerializedSv2Frame::from_bytes`.
- [`sv2/codec-sv2/src/decoder.rs#L52`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L52), `encoder.rs:41/47/210` — still describe "generic message types (`T`)" that were removed.
- [`sv2/codec-sv2/BENCHES.md#L34`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/BENCHES.md#L34) — still documents `Encoder<T>` and "stores message as `Option<T>`"; doesn't list the new `noise/encrypted_payload_length` group.
- [`sv2/buffer-sv2/README.md#L70`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/buffer-sv2/README.md#L70) — links `framing_sv2::framing::Frame`, deleted by this PR.
- [`sv2/codec-sv2/benches/serialization.rs#L7`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/benches/serialization.rs#L7) — the `not(with_buffer_pool)` alias body was removed but its `#[cfg(not(...))]` attribute is left stacked on the `#[cfg(...)]` one, so `type Slice` is compiled out under every feature combination.

#### 8. `Sv2Frame::serialize` re-walks the message
[`sv2/framing-sv2/src/framing.rs#L126`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/framing-sv2/src/framing.rs#L126)

The new length check re-runs `message.get_size()`, which both encoders already called to size the buffer (`encoder.rs:129` and `:215`). Derived `get_size` iterates every element of `Seq0255`/`Seq064K`, so `DeclareMiningJob`-style messages pay an extra O(elements) pass per frame. `from_message` already stored the length in the header, so `encoded_length()` can be `Header::SIZE + self.header.payload_length()` (O(1)) and remove both walks.

---

### B. Pre-existing, but re-exposed by code this PR rewrote

Not regressions, but they live in lines this PR touched and the refactor is the natural moment to address them.

#### 9. `SerializedSv2Frame` derives `Clone` → pool slot double-free under `with_buffer_pool`
[`sv2/framing-sv2/src/framing.rs#L155`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/framing-sv2/src/framing.rs#L155)

`buffer_sv2::Slice::clone` copies the raw pool pointer/slot index; its `Drop` XOR-toggles the shared slot bit. Cloning and dropping the clone frees the slot while the original still aliases it; dropping the original later re-marks the slot in use forever.

**Confirmed:** `let f = decoder.next_frame()?; let g = f.clone(); drop(g);` → the next decode overwrote `f`'s payload in place (`aa aa aa aa` → `bb bb bb bb`, same address) and `drop(decoder)` spun forever in `BufferPool::drop`. The old `Sv2Frame<T, B>` had the same derive, but the new consuming `into_bytes(self)` / `encode_into(self)` API actively invites `frame.clone()`.

#### 10. The chunk cap doesn't actually bound memory
[`sv2/codec-sv2/src/decoder.rs#L306`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L306)

"Cap the decoder read window at one chunk" only changes *when* the declared bytes are reserved, not *whether*. `next_transport` still requires `noise_buffer.len() == encrypted_payload_length(header)` before decrypting anything, then `get_data_owned()`s the whole ciphertext and copies it chunk-wise into `sv2_buffer`. Observed: a 131038-byte payload reported `MissingBytes(131070)` and decrypted nothing until every byte arrived. A peer sending a max-U24 header and trickling the payload holds ~16 MiB ciphertext + ~16 MiB plaintext per connection. A `max_frame_length` check at header-parse time (here and at L400) or per-chunk decryption would make the cap load-bearing.

#### 11. Stale `NOISE_FRAME_HEADER_SIZE` after the handshake
[`sv2/codec-sv2/src/decoder.rs#L114`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L114)

`self.missing_noise_b = NOISE_FRAME_HEADER_SIZE` (2) is a leftover from the removed 2-byte noise framing; the next thing on the wire for either role is the 22-byte encrypted Sv2 header, and `WithNoise::new()` starts at 0. Observed read sizes for the first transport frame after the handshake: `[2, 20, 0]`. Correct, but every connection pays an extra round, and a `read()`-based caller that calls `writable()` before the first `next_*` gets a 0-byte window it may treat as EOF. Deriving the window from `expected - buffered` would remove the cached field entirely and, as a bonus, make `UnexpectedTrailingBytes` / `reset_after_surplus` unreachable (surplus can only arise from calling `writable()` twice).

#### 12. `MissingBytes(0)` for empty payloads
[`sv2/codec-sv2/src/decoder.rs#L299`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L299)

A header with `msg_length = 0` yields `Err(MissingBytes(0))` and a 0-byte `writable()` (observed hints `[22, 0]`). A caller treating a 0-byte read as EOF closes the connection, and the crate's own `assert!(n > 0)` in `prop_noise_decoder_handles_partial_data` is violated for such a message. Same block: `crate::encrypted_payload_length(&header)` is computed twice (L298/L299) and there is a no-op `self.sv2_buffer.as_ref();` at L294.

#### 13. Typestate doesn't hold for `step_0`
[`sv2/codec-sv2/src/state.rs#L93`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/state.rs#L93)

The module says transitions are checked at compile time, but `Handshake<Initiator>::step_0` takes `&mut self` and `noise_sv2::Initiator::step_0` calls `mix_hash`/`encrypt_and_hash` on every invocation, so a retried `step_0` (e.g. after a send failure) fails at `step_2` with `AeadError` at runtime. Making `step_0` consume `self` and return `(HandshakeFrame, Handshake<InitiatorSent>)` would make the claim true.

#### 14. `unencrypted` example deadlocks under `with_buffer_pool`
[`sv2/codec-sv2/examples/unencrypted.rs#L115`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/examples/unencrypted.rs#L115)

`receiver_side` returns the `Slice`-backed frame while its local `StandardDecoder` drops, and `BufferPool::drop` spin-waits on outstanding slices.

**Confirmed:** `cargo run --example unencrypted --features with_buffer_pool` never returns (killed after 120 s on both PR head and `main`); the `encrypted` example, whose decoder lives in `main`, completes. CI only runs this example without features.

#### 15. `State::with_transport_mode` removed with no replacement
`Transport::new` is private, so anyone who runs the Noise handshake via `noise_sv2` directly (as noise-sv2's own example does) can no longer hand the engine to the codec. Nothing in-repo needs it, but it's a removed public capability with no migration path — worth either a `pub` constructor or a changelog note.

#### 16. `Handshake::step_1` / `step_2` re-implement `noise_sv2`'s std-gated versions
[`sv2/codec-sv2/src/state.rs#L106`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/state.rs#L106), [`#L136`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/state.rs#L136)

Both inline the `SystemTime::now()...as_secs() as u32` computation (with the `.unwrap()` on a pre-epoch clock and the `u32` truncation) and `rand::thread_rng()`, which `noise_sv2::Initiator::step_2` / `Responder::step_1` already do under the same `std` gate. `self.role.step_2(message)` / `self.role.step_1(re_pub)` are drop-ins; as written two crates own the same clock/RNG logic and can drift.

#### 17. `while_handshaking` still hand-rolls what `HandshakeFrame::from_message` now does
[`sv2/codec-sv2/src/decoder.rs#L227`](https://github.com/stratum-mining/stratum/blob/7ff8e04661df388aff8fceebe3843764b529664e/sv2/codec-sv2/src/decoder.rs#L227)

`get_data_owned().as_mut().to_vec()` followed by a `#[cfg(feature = "with_buffer_pool")]` / `not` pair keyed on codec_sv2's feature, while `HandshakeFrame`'s `Slice` type is keyed on framing_sv2's. A dependency graph enabling `framing_sv2/with_buffer_pool` but depending on codec_sv2 without it fails with `expected Slice, found Vec<u8>`. Lines 227–239 collapse to `HandshakeFrame::from_message(self.noise_buffer.get_data_owned())`, which removes the cfg pair, the extra copy and the hazard.

---

### C. Nits

- `make_transport_state_pair` is byte-identical in `encoder.rs:325` and `decoder.rs:943`, with near-copies in `lib.rs::transport_halves`, `decoder.rs::make_handshake_pair`, `benches/encoder.rs::setup_noise_states` and `benches/noise_roundtrip.rs::setup_noise_engine_pair`; the `AUTHORITY_*` keys are repeated six times. A `#[cfg(test)] pub(crate) mod test_utils` (and a `common.rs` helper for benches) would collapse them. Same for the feed-`writable()`-until-`Ok` loop, which is inlined in `lib.rs:149`, `lib.rs:257`, `encoder.rs:415` although `decoder.rs::decode_noise_frame` already exists.
- `Handshake::expected_message_size(&self)` only forwards `R::EXPECTED_MESSAGE_SIZE`; the single caller is already generic over `R`.
- `NoiseEncoder::encode_handshake` returns `Result<B::Slice>` but has no failure path (and has no test/bench/example exercising it — the examples ship `frame.payload()` directly).
- The three `compile_fail` doctests in `state.rs` carry no error code, so they pass on any compile error; the "no replay" one relies on `step_1`, which is `cfg(std)`, so under `--no-default-features --features noise_sv2` it fails with E0599 instead of the intended E0382. Use `compile_fail,E0382` etc.
- `SerializedSv2Frame::encode_into` and `Sv2Frame::serialize` each hand-roll the same `dst.get_mut(..required)` / `DestinationTooShort { required, actual }` check; only the `Sv2Frame` copy is tested.
- `SerializedSv2Frame`'s `encoded_length()` is `bytes.len()` rather than `Header::SIZE + header.payload_length()`, so a frame built via the public `from_bytes_unchecked` from mismatched bytes is re-encoded (and MACed) under an inconsistent header.
- Three different buffer-sizing policies across the four constructors: `NoiseEncoder::new` switches on `with_buffer_pool`, `Encoder::new` uses `DEFAULT_BUFFER_SIZE` (512) unconditionally, both decoders use the pool size unconditionally. `BufferFromSystemMemory::new(_)` ignores its argument, so the `not(with_buffer_pool)` arm has no effect and the only place `DEFAULT_BUFFER_SIZE` matters is giving the unencrypted encoder a 512-byte pool vs 320 KiB for the others.
- `SV2_FRAME_CHUNK_SIZE - AEAD_MAC_LEN` is spelled out at `lib.rs:97`, `encoder.rs:145/148/159` and in tests, while the decoder uses bare `SV2_FRAME_CHUNK_SIZE`; a named `SV2_FRAME_PLAINTEXT_CHUNK_SIZE` next to `ENCRYPTED_SV2_FRAME_HEADER_SIZE` would keep the pair in one place.
- `Sv2Frame::from_message` still returns `Option`, discarding the `InvalidU24` the PR is now in a position to propagate; every call site ends in `.expect("Failed to create the frame")` with no length in the message.
- Forwarding an already-serialized frame through `encode_transport` copies N bytes twice (into `sv2_buffer` as scratch, then chunk-wise into `noise_buffer`) because `EncodableFrame` only exposes `encode_into(self, dst)`; slightly better than the old `swap_with_slice` path, but the trait shape now locks it in. A borrowed accessor (`fn serialized(&self) -> Option<&[u8]>`) would let the noise encoder chunk-copy straight from the bytes. Similarly `Encoder::encode` on a `SerializedSv2Frame<B::Slice>` copies when `into_bytes()` is already the zero-copy answer.

---

**Summary:** I'd block on #1, #2 (at least fix the benches), #4 and #5, and strongly suggest #3 and #9 given the new public trait surface. Everything else is take-it-or-leave-it.

bit-aloo:
Took most of them, but not all, since they were touching the same `buffer_sv2` region and will be fixed once that is fixed. Thanks!

sv2/codec-sv2/src/decoder.rs
```diff
@@ -308,7 +237,6 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
             Err(_) => {
                 // Not a no-op: the decrypt offset and the plaintext decrypted so far persist across
                 // calls, so without this the next frame is decrypted at the failing chunk's offset.
-                self.sv2_buffer.danger_set_start(0);
                 self.sv2_buffer.get_data_owned();
```
plebhash: fable clanker finding:

> This arm regressed in 8781a6b4 — the comment's claim is inverted. `get_data_owned` resets the
> *cursor* but not the decrypt offset: `start` is only ever cleared by `danger_set_start`
> (`buffer.rs:82-94`; the pool impl behaves the same), and `as_ref`/`get_writable` both offset by it.
> 9d05f32d had the reset here; the fixup removed it and kept the comment.
> 
> Concrete failure: on a multi-chunk transport frame, the payload loop runs
> `danger_set_start(decrypted_len)` before each chunk decrypt. If chunk ≥ 2 fails (one flipped
> ciphertext bit from a MITM does it), we return with `start = decrypted_len` ≥ 65525. The next
> `next_transport_frame` header round does `get_writable(22)` and then the AEAD calls
> `as_ref()` → `[..22][65525..]` → panic (`range start index 65525 out of range for slice of
> length 22`). A failed *single*-chunk frame leaves `start = 6` instead: no panic, but every
> subsequent frame decrypts the wrong 16-byte region — permanent desync.
> 
> `noise_decoder_recovers_from_a_failed_decryption` doesn't catch this because it fails on the
> header round, where `start` is still 0.
> 
> ```suggestion
>                 // `get_data_owned` resets the cursor but not the decrypt offset
>                 // (`danger_set_start`), so clear the offset here too — otherwise the next
>                 // frame is decrypted at the failing chunk's offset.
>                 self.sv2_buffer.danger_set_start(0);
>                 self.sv2_buffer.get_data_owned();
> ```
> 
> Making `get_data_owned` clear `start` in buffer-sv2 would be the root-cause fix for every
> caller, but I understand buffer-sv2 is out of scope here, so restoring the line seems right.

sv2/framing-sv2/src/header.rs
```diff
@@ -106,28 +120,6 @@ impl Header {
     pub fn channel_msg(&self) -> bool {
```
plebhash, line 120: fable clanker finding:

> The PR description says `closes #2115`, but I can't find a channel_id accessor anywhere in the
> diff — `Header` still only exposes the `channel_msg()` bit test, and the "first 4 payload bytes
> are a U32 channel_id" semantics exist only in the comment at the top of this file.
> 
> Either drop the `closes #2115` line, or add the accessor while we're here. Since the header
> alone can't see the payload, the natural home is probably
> `SerializedSv2Frame::channel_id(&self) -> Option<u32>`: `None` when `channel_msg` is unset or
> the payload is shorter than 4 bytes, otherwise `u32::from_le_bytes(payload[..4])` per spec.
>

plebhash:
why do we have 03a641b9e3160d013453b6b75702c0c49c6f06fc as a standalone commit?

sv2/codec-sv2/src/decoder.rs
bit-aloo: done

sv2/framing-sv2/src/header.rs
bit-aloo: Added back again.

plebhash:
note from dev call:

I'll have a detailed look into new `codec_sv2` and `framing_sv2` APIs being shaped on this PR

sv2/codec-sv2/src/lib.rs
```diff
@@ -9,11 +9,11 @@
...
 //!
 //! ## Usage
-//! `codec-sv2` supports both standard Sv2 frames (unencrypted) and Noise-encrypted Sv2 frames to
-//! ensure secure communication. To encode messages for transmission, choose between the
-//! [`Encoder`] for standard Sv2 frames or the [`NoiseEncoder`] for encrypted frames. To decode
-//! received messages, choose between the [`StandardDecoder`] for standard Sv2 frames or
-//! [`StandardNoiseDecoder`] to decrypt Noise frames.
+//! Encode with the [`Encoder`], or the `NoiseEncoder` under the `noise_sv2` feature.
+//!
+//! Decode with the [`StandardDecoder`], or the `StandardNoiseDecoder` under the same feature.
```
plebhash:
```suggestion
//! Encode with the [`Encoder`], or the [`NoiseEncoder`] under the `noise_sv2` feature.
//!
//! Decode with the [`StandardDecoder`], or the [`StandardNoiseDecoder`] under the same feature.
```

sv2/codec-sv2/src/decoder.rs
```diff
@@ -1,91 +1,59 @@
...
+};
 
 /// Standard Sv2 decoder with Noise protocol support.
 ///
-/// Used for decoding and decrypting generic message types (`T`) encoded in Sv2 frames and
-/// encrypted via the Noise protocol.
+/// Used for decoding Sv2 frames encrypted via the Noise protocol.
 #[cfg(feature = "noise_sv2")]
-pub type StandardNoiseDecoder<T> = WithNoise<Buffer, T>;
+pub type StandardNoiseDecoder = WithNoise<Buffer>;
```
plebhash:
```suggestion
pub type NoiseDecoder = WithNoise<Buffer>;
```

as noted here https://github.com/stratum-mining/stratum/issues/1129#issuecomment-2356904898 we should try to standardize these namings

ironically, the lowest hanging fruit here is to drop `Standard*` prefix, to match the `NoiseEncoder` counterpart

sv2/codec-sv2/src/decoder.rs
```diff
@@ -1,91 +1,59 @@
...
 
 /// Standard Sv2 decoder without Noise protocol support.
 ///
-/// Used for decoding generic message types (`T`) encoded in Sv2 frames.
-pub type StandardDecoder<T> = WithoutNoise<Buffer, T>;
+/// Used for decoding plain Sv2 frames.
+pub type StandardDecoder = WithoutNoise<Buffer>;
```
plebhash:
```suggestion
pub type Decoder = WithoutNoise<Buffer>;
```

as noted here https://github.com/stratum-mining/stratum/issues/1129#issuecomment-2356904898 we should try to standardize these namings

ironically, the lowest hanging fruit here is to drop `Standard*` prefix to match the `Encoder` counterpart

plebhash:
splitting `Sv2Frame<T, B>` into `Sv2Frame<T>` vs  `SerializedSv2Frame<B>` is a great API improvement

much cleaner and intuitive code, completely eliminating the panic footguns from the convoluted APIs of `Sv2Frame<T, B>`

sv2/codec-sv2/src/state.rs
```diff
@@ -0,0 +1,273 @@
...
+    Initiator, NoiseDecryptor, NoiseEncryptor, NoiseEngine, Responder, ELLSWIFT_ENCODING_SIZE,
+    INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE,
+};
+
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for noise_sv2::Initiator {}
+    impl Sealed for noise_sv2::Responder {}
+    impl Sealed for super::InitiatorSent {}
+}
```
plebhash: we should document this

bit-aloo:
Updated the companion, to see new APIs in action: https://github.com/stratum-mining/sv2-apps/pull/806

sv2/codec-sv2/src/decoder.rs
```diff
@@ -120,57 +75,72 @@ pub struct WithNoise<B: IsBuffer, T: Serialize + binary_sv2::GetSize> {
     //
     // Stores the decrypted data until it is ready to be processed and converted into a Sv2 frame.
     sv2_buffer: B,
+
+    missing_noise_b: usize,
```
plebhash, line 82: we should add docs here, all other struct fields above have some

sv2/codec-sv2/src/state.rs
```diff
@@ -0,0 +1,273 @@
...
+///
+/// fn initiator() -> Handshake<Initiator> { unimplemented!() }
+///
+/// let _transport =
+///     initiator().step_2_with_now([0; INITIATOR_EXPECTED_HANDSHAKE_MESSAGE_SIZE], 0);
+/// ```
+#[derive(Debug)]
+pub struct Handshake<R: HandshakeRole> {
+    role: Box<R>,
+}
```
plebhash: do we really need `Sealed` + `HandshakeRole` marker traits?

if we're aiming to leverage generics with `Handshake`, why not give each handshake role its own constructor flavor and cut off the marker trait complexity?

```rust
pub struct Handshake<R> { role: Box<R> }

impl Handshake<Initiator> {
     pub fn initiator(role: Box<Initiator>) -> Self { ... }
     ...
}

impl Handshake<InitiatorSent> {
    ...
}

impl Handshake<Responder> {
    pub fn responder(role: Box<Responder>) -> Self { ... }
    ...
}
```

sv2/codec-sv2/src/state.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl HandshakeRole for Initiator {}
+impl HandshakeRole for InitiatorSent {}
+impl HandshakeRole for Responder {}
+
+/// A handshake role that is waiting on a message from its counterpart.
+///
+/// An [`Initiator`] that has not sent its first message yet is deliberately not one: nothing is
+/// coming back until it does, so a decoder cannot be asked to read for it.
+pub trait ExpectsHandshakeMessage: HandshakeRole {
```
plebhash: I do think `ExpectsHandshakeMessage` trait is justifiable, but I have a comment about the `HandshakeRole` binding that's related to the [other one around marker traits](https://github.com/stratum-mining/stratum/pull/2317/changes#r3927802350)

even if we REALLY wanted to find an opportunity to use sealed traits (in a "hammer looking for a nail" kind of bias), I don't feel `ExpectsHandshakeMessage` really deserves one

I mean... what are we trying to prohibit bogus `ExpectsHandshakeMessage` trait implementors from? the trait only defines a constant... there's no associated types or methods to be misused

sv2/codec-sv2/src/state.rs
plebhash: actually, I think this is just unnecessary

see https://github.com/stratum-mining/stratum/pull/2317#discussion_r3927802350

sv2/codec-sv2/src/state.rs
bit-aloo: The thought process was not to let downstream to implement any variant of HandshakeRole. But I guess considering we only provide impl for concrete types like Initiator and responder. We should not care. Updated.

plebhash (approved):

bit-aloo:
Moving this to draft to accomodate: https://github.com/project-loupe/audit-stratum/issues/219

bit-aloo:
> Moving this to draft to accomodate: [project-loupe/audit-stratum#219](https://github.com/project-loupe/audit-stratum/issues/219)

Had to rework the decoder a bit. Earlier, bytes buffered past the end of a frame made the decoder drain everything, the complete frame included, and return an error. On the Noise side that left the cipher out of step, the header had already been decrypted, so our nonce had advanced, but the payload it described was dropped and the peer's nonce kept going for chunks we never decrypted. The encoder and decoder never lined up again. Now the decoder takes exactly the bytes the current frame needs and carries the rest over as the start of the next one. This applies to the plain decoder, the Noise transport rounds, and the handshake, so both decoders behave the same way. This was already the case on main. We never saw it because the apps size every read from writable, so a surplus only arises by writing into the decoder twice without decoding in between. With carry-over in place, extending the API to accept streaming becomes straightforward, can be thought in future. The transport decrypt state now goes into next_transport_frame by value and only comes back on a round that did not fail. That is what closes the replay after a failed payload tag, a state that failed to authenticate cannot be handed back to the decoder, so the rolled-back nonce can never be reused on the peer's next ciphertext. The surplus change is the second half of the same cleanup.

sv2/codec-sv2/src/decoder.rs
```diff
@@ -120,132 +74,155 @@ pub struct WithNoise<B: IsBuffer, T: Serialize + binary_sv2::GetSize> {
...
+    Frame(F),
+    Incomplete(usize),
+}
+
+#[cfg(feature = "noise_sv2")]
+#[derive(Debug)]
+pub enum Decrypted<F> {
+    Frame(F, TransportDecryptState),
+    Incomplete(usize, TransportDecryptState),
 }
```
plebhash, line 111: let's add docs for these new `enum`s

sv2/codec-sv2/src/decoder.rs
```diff
@@ -120,132 +74,155 @@ pub struct WithNoise<B: IsBuffer, T: Serialize + binary_sv2::GetSize> {
...
+    /// Attempts to decode the next encrypted frame with the decrypting half of a completed
+    /// handshake.
+    ///
+    /// On [`Decrypted::Incomplete`], resize the decoder buffer using `writable`, read another
+    /// chunk from the stream, and call this method again with the state it carries until it
+    /// returns a [`Decrypted::Frame`]. The count it carries is what the decoder will accept on
+    /// the next read, which is always [`Self::writable_len`] and never more than one chunk, not
+    /// the number of bytes left in the frame.
+    ///
+    /// Bytes buffered past the end of the frame are kept as the start of the next one.
```
plebhash, line 187:
```suggestion
    /// Bytes buffered past the end of the frame are kept as the start of the next one.
    
    /// An `Err` consumes the state. After a failed decryption the cipher's nonce no longer
    /// follows the bytes the peer keeps sending: a retransmission of the ciphertext that just
    /// failed still authenticates under the nonce the cipher stayed at, and the decoder, back to
    /// waiting for a header, would read it as one. The connection has to be torn down and a new
    /// [`crate::Handshake`] run, which the types enforce by never handing the same state back:
    ///
    /// ```compile_fail,E0382
    /// use codec_sv2::{NoiseDecoder, TransportDecryptState};
    ///
    /// fn state() -> TransportDecryptState { unimplemented!() }
    ///
    /// let mut decoder = NoiseDecoder::new();
    /// let state = state();
    /// if decoder.next_transport_frame(state).is_err() {
    ///     let _retry = decoder.next_transport_frame(state);
    /// }
    /// ```
```

plebhash (approved):

sv2/codec-sv2/src/decoder.rs
```diff
@@ -319,76 +302,64 @@ impl<'a, T: Serialize + GetSize + Deserialize<'a>, B: IsBuffer + AeadBuffer> Wit
...
-                Err(Error::MissingBytes(header.encrypted_len()))
-            }
+            let src = self.take(expected);
+            let decrypted_header = self.sv2_buffer.get_writable(expected);
+            decrypted_header.copy_from_slice(&src.as_ref()[..expected]);
+            decrypt(&mut self.sv2_buffer)?;
+            let header =
+                Header::from_bytes(self.sv2_buffer.get_data_by_ref(SV2_FRAME_HEADER_SIZE))?;
+            self.expect(crate::encrypted_payload_length(&header));
+            Ok(Decoded::Incomplete(self.writable_len()))
```
GitGab19: Fable comment:

When the decrypted header declares an empty payload, this returns Incomplete(0) and the caller has to call again without reading. Any caller that does read(writable()) sees a zero-length read, which most loops treat as EOF. The companion's try_read_frame already has that shape. Could the header branch check encrypted_payload_length(&header) == 0 and fall through to build the frame in the same round? The carry-over case at line 786 would still produce Incomplete(0), so the doc on Decoded::Incomplete should say explicitly that zero means "call again without reading".

sv2/framing-sv2/src/framing.rs
```diff
@@ -50,192 +54,194 @@ impl core::fmt::Display for SizeHint {
...
     }
+}
 
-    /// Tries to build a [`Sv2Frame`] from raw bytes.
+/// A frame carrying the serialized bytes of its header and payload.
+///
+/// This is the incoming side of an Sv2 exchange: what a decoder hands back once it has read a
+/// whole frame off the wire. Because it is built from those bytes, [`Self::payload`] always has
+/// them.
+#[derive(Debug, Clone)]
```
GitGab19, line 151: Fable comment:

Cloning this under with_buffer_pool still frees the pool slot twice, and the consuming into_bytes and encode_into API makes frame.clone() the obvious thing to reach for. Since the real fix is #2272 on Slice, could we leave a comment here referencing it so the derive gets revisited when that lands?

sv2/framing-sv2/src/framing.rs
```diff
@@ -50,192 +54,194 @@ impl core::fmt::Display for SizeHint {
...
-    pub fn from_bytes_unchecked(mut bytes: B) -> Self {
-        // Unchecked function caller is supposed to already know that the passed bytes are valid
-        let header = Header::from_bytes(bytes.as_mut()).expect("Invalid header");
-        Self {
-            header,
-            payload: None,
-            serialized: Some(bytes),
-        }
+    pub fn from_parts(header: Header, bytes: B) -> Self {
+        Self { header, bytes }
```
GitGab19, line 173: Fable comment:

Since this is public, a debug_assert! that Header::SIZE + header.payload_length() == bytes.as_ref().len() would catch a mismatched pair in tests without costing anything in release.

sv2/framing-sv2/src/framing.rs
```diff
@@ -50,192 +54,194 @@ impl core::fmt::Display for SizeHint {
...
+    ) -> Option<Self> {
+        let extension_type = update_extension_type(extension_type, channel_msg);
+        let len = u32::try_from(message.get_size()).ok()?;
+        Header::from_len(len, message_type, extension_type).map(|header| Self { header, message })
+    }
+
+    /// Serializes the frame into the first [`Sv2Frame::encoded_length`] bytes of `dst`, erroring
+    /// out if `dst` is shorter than that.
     #[inline]
     pub fn serialize(self, dst: &mut [u8]) -> Result<(), Error> {
```
GitGab19: I asked Fable how to improve names, and here a suggestion:

This is the same operation as EncodableFrame::encode_into at line 70, which the trait impl at line 79 forwards to. Two verbs for one act makes a reader look for a difference that isn't there, and serialize is binary_sv2's word for message-to-bytes, which is only the payload half of what this does. Suggest dropping the inherent method and keeping encode_into as the single spelling, next to encoded_length. See the naming thread at line 107.

sv2/framing-sv2/src/framing.rs
```diff
@@ -50,192 +54,194 @@ impl core::fmt::Display for SizeHint {
...
 ///
-/// Represents a regular Sv2 frame, used for all communication outside of the Noise protocol
-/// handshake process. It contains a [`Header`] and a message payload, which can be serialized for
-/// encoding and transmission or decoded and deserialized upon receipt.
+/// This is the outgoing side of an Sv2 exchange: a message plus the [`Header`] that describes it,
+/// built with [`Sv2Frame::from_message`] and written out with [`Sv2Frame::serialize`]. A frame
+/// read off the wire is a [`SerializedSv2Frame`] instead.
 #[derive(Debug, Clone)]
-pub struct Sv2Frame<T, B> {
+pub struct Sv2Frame<T> {
```
GitGab19: I asked Fable how to improve names, and here a suggestion:

One half of the pair carries a qualifier and the other claims the generic name, so a reader sees Sv2Frame in a signature and assumes it is the frame type, when it is specifically the not-yet-serialized one. The companion already re-aliases these as OutboundFrame / InboundFrame, which shows the pair doesn't read as a pair, and direction is the wrong axis anyway: a proxy forwards a SerializedSv2Frame, so it goes both ways.

Proposal: name both halves by representation and drop the Sv2 infix the crate path already provides.

Sv2Frame<T> → MessageFrame<T>
SerializedSv2Frame<B> → SerializedFrame<B>
Two follow-ons in the same spirit, pointed at inline below: fold Sv2Frame::serialize into the trait's encode_into, and retire the StandardSerializedFrame alias in codec_sv2. Fixup targets would be 85c90346 for framing and cf543652 for the codec alias.

sv2/codec-sv2/src/lib.rs
```diff
@@ -37,332 +37,81 @@
...
-                    TransportEncryptState { encryption },
-                    TransportDecryptState { decryption },
-                ))
-            }
-            _ => Err(Error::UnexpectedNoiseState),
-        }
-    }
-}
+/// An Sv2 frame as the decoders hand it back, carrying the bytes read off the wire.
+pub type StandardSerializedFrame = SerializedSv2Frame<<Buffer as IsBuffer>::Slice>;
```
GitGab19: Fable comment:

Last surviving Standard* name after the decoders lost the prefix in this PR, and "standard" doesn't say what it is: the serialized frame over the feature-selected buffer type. Suggest SerializedFrame here too, or dropping the alias if the companion can name the buffer type itself. See the naming thread at framing.rs:107.

sv2/framing-sv2/src/framing.rs
```diff
@@ -246,48 +252,33 @@ impl<T, B> TryFrom<Frame<T, B>> for Sv2Frame<T, B> {
 /// handshake process. Once the handshake is complete, regular Sv2 communication switches to
 /// [`Sv2Frame`] for ongoing communication.
 #[derive(Debug)]
-pub struct HandShakeFrame {
+pub struct HandshakeFrame {
```
GitGab19: Fable optional renaming:

Optional, since this only inherited the old name with the capitalization fixed: the doc at line 17 says no header is included, so this isn't a frame. HandshakeMessage would say what it is, and would also separate its from_message at line 270 from the quite different from_message on the message frame at line 116.

sv2/codec-sv2/src/decoder.rs
bit-aloo: Done

sv2/framing-sv2/src/framing.rs
bit-aloo: Will be done in buffer_sv2.

sv2/framing-sv2/src/framing.rs
bit-aloo: Added

sv2/framing-sv2/src/framing.rs
bit-aloo: Though the comment is not required considering, we will be looking at that next.

sv2/framing-sv2/src/framing.rs
bit-aloo: Done.

sv2/framing-sv2/src/framing.rs
bit-aloo: Done

sv2/codec-sv2/src/lib.rs
bit-aloo: Done

sv2/framing-sv2/src/framing.rs
bit-aloo: Done

sv2/codec-sv2/src/encoder.rs
```diff
@@ -284,89 +185,133 @@ impl<T: Serialize + GetSize> Default for WithNoise<Buffer, T> {
...
-    /// stored in the internal buffer. Otherwise, errors on a serialization failure.
-    pub fn encode(
-        &mut self,
-        item: Sv2Frame<T, B::Slice>,
-    ) -> core::result::Result<B::Slice, crate::Error> {
+    /// Serializes `item` into a byte stream. The resulting bytes are stored in the internal
+    /// `buffer`, preparing the frame for transmission. On success, the method returns a reference
+    /// to the serialized bytes stored in the internal buffer. Otherwise, errors on a
+    /// serialization failure.
+    pub fn encode<F: EncodableFrame>(&mut self, item: F) -> crate::Result<B::Slice> {
```
GitGab19: Fable non-blocking comment: 

The EncodableFrame doc states "An encoder rejects a frame that reports less than [SV2_FRAME_HEADER_SIZE]", and try_encrypt_frame does — but WithoutNoise::encode doesn't, so a foreign impl (which the trait doc invites) reporting < 6 bytes goes out as a sub-header frame and desyncs the peer with no error on either side. Mirror the noise-side guard here.

sv2/codec-sv2/src/decoder.rs
```diff
@@ -399,90 +387,102 @@ impl<T: Serialize + binary_sv2::GetSize> Default for WithNoise<Buffer, T> {
...
+                let frame = self.buffer.get_data_owned();
+                self.buffer
+                    .get_writable(carried.len())
+                    .copy_from_slice(carried);
+                self.missing_b = Header::SIZE.saturating_sub(carried.len());
+                Ok(Decoded::Frame(SerializedFrame::<B::Slice>::from_bytes(
+                    frame,
+                )?))
             }
+            Err(hint @ SizeHint::Exact) => Err(Error::UnexpectedFrameSize(hint)),
```
GitGab19: Fable non-blocking comment:

parse_header maps Equal to Ok(header), so its Err can never be Exact — yet the arm exists and Error::UnexpectedFrameSize(SizeHint::Exact) is a public value that would render as "buffered bytes do not hold exactly one frame: exactly one complete frame". A two-variant error (Missing/Surplus) for parse_header would remove both the dead arm and the impossible public state.

sv2/codec-sv2/src/lib.rs
```diff
@@ -37,332 +37,82 @@
...
 pub mod error;
+#[cfg(feature = "noise_sv2")]
+pub mod state;
+#[cfg(all(test, feature = "noise_sv2"))]
+pub(crate) mod test_utils;
 
 pub use error::{Error, Result};
 
-pub use decoder::{StandardEitherFrame, StandardSv2Frame};
+pub use framing_sv2::framing::{EncodableFrame, MessageFrame};
```
GitGab19: next_handshake_frame returns Decoded<HandshakeMessage> and Error::UnexpectedFrameSize wraps SizeHint, but neither is re-exported — a standalone codec_sv2 user must add a version-pinned framing_sv2 dependency just to name them. Re-exporting HandshakeMessage and SizeHint closes it.

sv2/codec-sv2/src/encoder.rs
GitGab19, line 206: Fable non-blocking comment:

Non-blocking: the EncodableFrame doc says "An encoder rejects a frame that reports less than [SV2_FRAME_HEADER_SIZE]", and try_encrypt_frame does (if len < SV2_FRAME_HEADER_SIZE), but the plain encode here has no such check — a foreign impl reporting < 6 bytes goes out as a sub-header frame and desyncs the peer with no error on either side. Mirroring the noise-side guard would make the doc true for both encoders.

sv2/codec-sv2/src/decoder.rs
GitGab19: Fable non-blocking comment:

Non-blocking: parse_header maps Equal to Ok(header), so its Err can never be SizeHint::Exact and this arm is unreachable — and Error::UnexpectedFrameSize(SizeHint::Exact) becomes a public value that would render as "buffered bytes do not hold exactly one frame: exactly one complete frame". A two-variant error (Missing/Surplus) for parse_header would drop both the dead arm and the impossible state. Fine as a follow-up.

sv2/codec-sv2/src/lib.rs
GitGab19: Fable non-blocking comment:

Non-blocking: next_handshake_frame returns Decoded<HandshakeMessage> and Error::UnexpectedFrameSize wraps SizeHint, but neither type is re-exported, so a standalone codec_sv2 user has to add a version-pinned framing_sv2 dependency just to name them. Re-exporting HandshakeMessage and SizeHint alongside EncodableFrame/MessageFrame would close that.

GitGab19 (approved):
I left a couple of minor nits, which are non-blocking, which I got from Fable 5.1

sv2/codec-sv2/src/encoder.rs
bit-aloo: Added.

sv2/codec-sv2/src/decoder.rs
bit-aloo: Removed Exact, was not adding much

sv2/codec-sv2/src/lib.rs
bit-aloo: done

GitGab19 (approved):

PR #2318, opened 2026-08-26
register per-channel specific custom cb suffix

lorbax:


lorbax:
sorry guys I confused the upstream!

PR #2319, opened 2026-08-27
feat(channels_sv2, sv1_api): support prefix updates and typed submit errors

GitGab19:
This PR is a prerequisite for a final PR on `sv2-apps` which is aiming to close almost all the currently tracked issues of tProxy (specifically https://github.com/stratum-mining/sv2-apps/issues/36, https://github.com/stratum-mining/sv2-apps/issues/139, https://github.com/stratum-mining/sv2-apps/issues/164, https://github.com/stratum-mining/sv2-apps/issues/386, https://github.com/stratum-mining/sv2-apps/issues/650).

It has four commits:

- `channels_sv2`: add APIs for replacing the upstream extranonce prefix while preserving local suffixes, channel allocations, bitmap state, and rollable size.
- `channels_sv2`: validate the complete extended extranonce layout transactionally on prefix updates.
- `sv1_api`: acknowledge `mining.extranonce.subscribe`, pass its `client_id` to the server, and use a mutable receiver so tProxy can record downstream support directly.
- `sv1_api`: expose typed submit outcomes for the standard SV1 rejection codes 20–23, serialized in the legacy SV1 error-array format while retaining compatibility with object-form errors.

companion https://github.com/stratum-mining/sv2-apps/pull/823

### DIFF
--- a/sv1/src/json_rpc.rs
+++ b/sv1/src/json_rpc.rs
@@ -2,13 +2,34 @@
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{fmt, fmt::Display};
 
-#[derive(Clone, Serialize, Deserialize, Debug)]
+#[derive(Clone, Serialize, Debug)]
 #[serde(untagged)]
 pub enum Message {
     StandardRequest(StandardRequest),
     Notification(Notification),
     OkResponse(Response),
     ErrorResponse(Response),
+}
+
+impl<'de> Deserialize<'de> for Message {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum MessageWire {
+            StandardRequest(StandardRequest),
+            Notification(Notification),
+            Response(Response),
+        }
+
+        Ok(match MessageWire::deserialize(deserializer)? {
+            MessageWire::StandardRequest(request) => Self::StandardRequest(request),
+            MessageWire::Notification(notification) => Self::Notification(notification),
+            MessageWire::Response(response) => response.into(),
+        })
+    }
 }
 
 impl Message {
@@ -92,6 +113,14 @@ impl fmt::Display for Response {
     }
 }
 
+/// An SV1 response error, serialized as `[code, message, data]`.
+///
+/// The three-element format is used by the [original Stratum implementation][a] and the
+/// [NiceHash extranonce extension][b]. Deserialization additionally accepts `[code, message]`
+/// with absent data, and object-form errors, for compatibility.
+///
+/// [a]: https://github.com/slush0/stratum/blob/master/stratum/protocol.py
+/// [b]: https://github.com/nicehash/Specifications/blob/master/NiceHash_extranonce_subscribe_extension.txt
 #[derive(Clone, Debug, PartialEq)]
 pub struct JsonRpcError {
     pub code: i32, // json do not specify precision which one should be used?
@@ -117,6 +146,7 @@ impl<'de> Deserialize<'de> for JsonRpcError {
         #[serde(untagged)]
         enum JsonRpcErrorWire {
             Legacy((i32, String, Option<serde_json::Value>)),
+            LegacyWithoutData((i32, String)),
             Object {
                 code: i32,
                 message: String,
@@ -124,18 +154,20 @@ impl<'de> Deserialize<'de> for JsonRpcError {
             },
         }
 
-        match JsonRpcErrorWire::deserialize(deserializer)? {
+        let (code, message, data) = match JsonRpcErrorWire::deserialize(deserializer)? {
             JsonRpcErrorWire::Legacy((code, message, data))
             | JsonRpcErrorWire::Object {
                 code,
                 message,
                 data,
-            } => Ok(Self {
-                code,
-                message,
-                data,
-            }),
-        }
+            } => (code, message, data),
+            JsonRpcErrorWire::LegacyWithoutData((code, message)) => (code, message, None),
+        };
+        Ok(Self {
+            code,
+            message,
+            data,
+        })
     }
 }
 
@@ -185,6 +217,72 @@ mod tests {
     }
 
     #[test]
+    fn deserializes_two_element_errors_and_serializes_three_elements() {
+        for code in 20..=25 {
+            let message: Message = serde_json::from_value(serde_json::json!({
+                "id": 42,
+                "result": null,
+                "error": [code, "Rejected"],
+            }))
+            .unwrap();
+
+            let Message::ErrorResponse(response) = &message else {
+                panic!("a two-element error must remain an error response");
+            };
+            assert_eq!(
+                response.error,
+                Some(JsonRpcError {
+                    code,
+                    message: "Rejected".to_string(),
+                    data: None,
+                })
+            );
+            assert_eq!(
+                serde_json::to_value(&message).unwrap(),
+                serde_json::json!({
+                    "id": 42,
+                    "result": null,
+                    "error": [code, "Rejected", null],
+                })
+            );
+            assert!(matches!(
+                crate::methods::Method::try_from(message),
+                Ok(crate::methods::Method::ErrorMessage(_))
+            ));
+        }
+    }
+
+    #[test]
+    fn legacy_error_round_trip_preserves_additional_data() {
+        let wire = serde_json::json!([20, "Rejected", {"detail": "invalid share"}]);
+        let error: JsonRpcError = serde_json::from_value(wire.clone()).unwrap();
+        assert_eq!(
+            error.data,
+            Some(serde_json::json!({"detail": "invalid share"}))
+        );
+        assert_eq!(serde_json::to_value(error).unwrap(), wire);
+    }
+
+    #[test]
+    fn rejects_legacy_errors_with_invalid_types_or_arity() {
+        for wire in [
+            serde_json::json!([]),
+            serde_json::json!([20]),
+            serde_json::json!([20, "Rejected", null, null]),
+            serde_json::json!(["20", "Rejected"]),
+            serde_json::json!(["20", "Rejected", null]),
+            serde_json::json!([20.5, "Rejected"]),
+            serde_json::json!([20, 42]),
+            serde_json::json!([20, 42, null]),
+        ] {
+            assert!(
+                serde_json::from_value::<JsonRpcError>(wire.clone()).is_err(),
+                "invalid error unexpectedly accepted: {wire}"
+            );
+        }
+    }
+
+    #[test]
     fn deserializes_json_rpc_error_object_for_compatibility() {
         let response: Response = serde_json::from_value(serde_json::json!({
             "id": 42,
@@ -205,5 +303,28 @@ mod tests {
                 data: None,
             })
         );
+    }
+
+    #[test]
+    fn message_classifies_response_from_error_field() {
+        let error: Message = serde_json::from_value(serde_json::json!({
+            "id": 42,
+            "result": null,
+            "error": [22, "Duplicate share", null],
+        }))
+        .unwrap();
+        assert!(matches!(&error, Message::ErrorResponse(_)));
+        assert!(matches!(
+            crate::methods::Method::try_from(error),
+            Ok(crate::methods::Method::ErrorMessage(_))
+        ));
+
+        let success: Message = serde_json::from_value(serde_json::json!({
+            "id": 43,
+            "result": true,
+            "error": null,
+        }))
+        .unwrap();
+        assert!(matches!(success, Message::OkResponse(_)));
     }
 }
--- a/sv1/src/json_rpc.rs
+++ b/sv1/src/json_rpc.rs
@@ -1,5 +1,5 @@
 //! https://www.jsonrpc.org/specification#response_object
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{fmt, fmt::Display};
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
@@ -92,11 +92,51 @@ impl fmt::Display for Response {
     }
 }
 
-#[derive(Clone, Serialize, Deserialize, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct JsonRpcError {
     pub code: i32, // json do not specify precision which one should be used?
     pub message: String,
     pub data: Option<serde_json::Value>,
+}
+
+impl Serialize for JsonRpcError {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        (&self.code, &self.message, &self.data).serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for JsonRpcError {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum JsonRpcErrorWire {
+            Legacy((i32, String, Option<serde_json::Value>)),
+            Object {
+                code: i32,
+                message: String,
+                data: Option<serde_json::Value>,
+            },
+        }
+
+        match JsonRpcErrorWire::deserialize(deserializer)? {
+            JsonRpcErrorWire::Legacy((code, message, data))
+            | JsonRpcErrorWire::Object {
+                code,
+                message,
+                data,
+            } => Ok(Self {
+                code,
+                message,
+                data,
+            }),
+        }
+    }
 }
 
 impl From<Response> for Message {
@@ -118,5 +158,52 @@ impl From<StandardRequest> for Message {
 impl From<Notification> for Message {
     fn from(n: Notification) -> Self {
         Message::Notification(n)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deserializes_legacy_sv1_error_array() {
+        let response: Response = serde_json::from_value(serde_json::json!({
+            "id": 42,
+            "result": null,
+            "error": [22, "Duplicate share", null],
+        }))
+        .unwrap();
+
+        assert_eq!(
+            response.error,
+            Some(JsonRpcError {
+                code: 22,
+                message: "Duplicate share".to_string(),
+                data: None,
+            })
+        );
+    }
+
+    #[test]
+    fn deserializes_json_rpc_error_object_for_compatibility() {
+        let response: Response = serde_json::from_value(serde_json::json!({
+            "id": 42,
+            "result": null,
+            "error": {
+                "code": 22,
+                "message": "Duplicate share",
+                "data": null,
+            },
+        }))
+        .unwrap();
+
+        assert_eq!(
+            response.error,
+            Some(JsonRpcError {
+                code: 22,
+                message: "Duplicate share".to_string(),
+                data: None,
+            })
+        );
     }
 }
--- a/sv1/src/lib.rs
+++ b/sv1/src/lib.rs
@@ -28,7 +28,9 @@
 //! Every response contains the following parts
 //! * message ID: same ID as in request, for pairing request-response together
 //! * result: any json-encoded result object (number, string, list, array, …)
-//! * error: null or list (error code, error message)
+//! * error: null or a three-element list (error code, error message, additional data). The third
+//!   element is null when no additional data is provided. For compatibility, deserialization also
+//!   accepts two-element lists with absent data and object-form errors.
 //!
 //! References:
 //! [https://docs.google.com/document/d/17zHy1SUlhgtCMbypO8cHgpWH73V5iUQKk_0rWvMqSNs/edit?hl=en_US#]
@@ -140,7 +142,7 @@ pub trait IsServer {
             }
             methods::Client2Server::ExtranonceSubscribe(subscribe) => {
                 self.handle_extranonce_subscribe(client_id)?;
-                Ok(Some(subscribe.respond(true)))
+                Ok(Some(subscribe.respond()))
             }
             methods::Client2Server::Submit(submit) => {
                 let has_valid_version_bits = match &submit.version_bits {
@@ -155,7 +157,11 @@ pub trait IsServer {
                 };
 
                 if !self.is_authorized(client_id, &submit.user_name)? {
-                    return Err(Error::InvalidSubmission.into());
+                    return Ok(Some(submit.respond(
+                        client_to_server::SubmitOutcome::Rejected(
+                            client_to_server::SubmitError::UnauthorizedWorker,
+                        ),
+                    )));
                 }
 
                 let outcome = if self.extranonce2_size(client_id)? != submit.extra_nonce2.len()
@@ -224,8 +230,17 @@ pub trait IsServer {
         request: &client_to_server::Authorize,
     ) -> Result<bool, Self::Error>;
 
-    /// When miner find the job which meets requested difficulty, it can submit share to the server.
-    /// Only [Submit](client_to_server::Submit) requests for authorized user names can be submitted.
+    /// Validates a share after the default request handler has checked its SV1 session fields.
+    ///
+    /// [`Self::handle_parsed_request`] calls this only when the worker is authorized, extranonce2
+    /// has the size returned by [`Self::extranonce2_size`], and the version bits match the negotiated
+    /// version-rolling mask (including whether version bits must be present).
+    ///
+    /// Unauthorized workers receive [`client_to_server::SubmitError::UnauthorizedWorker`] (24).
+    /// Invalid extranonce2 sizes or version bits receive [`client_to_server::SubmitError::Other`]
+    /// (20). These rejections bypass this method and return `Ok(Some(response))`, not a Rust error;
+    /// applications that log or count them should inspect the returned response's `error` field.
+    /// Errors returned by the session-state accessors are still propagated to the caller.
     fn handle_submit(
         &self,
         client_id: Option<usize>,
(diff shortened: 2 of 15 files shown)


GitGab19:
Integration tests on CI here should be failing, but their executions are green simply because I edited the description after opening the PR.

And the current IT workflow excludes IT execution in case there's no difference between the edits of the description, when there's no "companion" string there.

But this mechanism causes some false positive, like in this case, so it must be fixed.

For now I prefer to put this in draft, just to avoid a possible merge without its companion PR.

GitGab19:
Blocked by https://github.com/stratum-mining/stratum/pull/2320

sv1/src/lib.rs
```diff
@@ -123,7 +123,7 @@ pub trait IsServer {
                 Ok(Some(configure.respond(version_rolling, min_diff)))
             }
             methods::Client2Server::ExtranonceSubscribe(_) => {
-                self.handle_extranonce_subscribe()?;
+                self.handle_extranonce_subscribe(client_id)?;
```
plebhash, line 126: fable clanker finding:

>  `mining.extranonce.subscribe` arrives as a `StandardRequest` carrying an `id`, but this arm returns `Ok(None)`, so the server never sends any response back.
>
> until now that was harmless, but the companion (sv2-apps#823) makes this subscription load-bearing: it decides whether a miner is kept alive and sent `mining.set_extranonce` across a `SetExtranoncePrefix`, or disconnected. SV1 client stacks disagree on how to treat a missing ack: some (e.g. the cgminer `#xnsub` lineage) tolerate silence and apply `mining.set_extranonce` unconditionally, but others only enable it after a positive ack. a miner in the second group gets recorded by `handle_extranonce_subscribe` as supporting the feature, stays connected across a prefix change, and then keeps mining on the stale extranonce1 — with tProxy still answering `result: true` to every share.
>
> suggestion: reply `{"id": X, "result": true, "error": null}` instead of `Ok(None)`. that means `ExtranonceSubscribe` needs to retain the request id — it's currently a unit struct (`ExtranonceSubscribe()`), so the id is dropped at parse time. it should specifically be `result: true`: an explicit `false`/error response makes cgminer-derived clients mark the whole pool dead, which is worse than staying silent.

plebhash:
cross-referencing since there's scope overlap: https://github.com/stratum-mining/stratum/issues/2330#issuecomment-5501652427

please let me know whether you want to address #2330 here or on a separate PR

GitGab19:
I'd keep #2330 out of the scope here, and properly solve it in a separate PR.

sv1/src/lib.rs
GitGab19: Good catch, I fixed it, and now it returns the response as well.

bit-aloo (commented):
Clanker review:
# Review

Everything compiles, targeted tests pass (`channels_sv2` extranonce tests: **42/42**, `sv1_api`: **27/27**), and Clippy reports only pre-existing test-code warnings.

Findings are ordered by importance.

## Issues

### 1. SV1 client can no longer parse rejection responses emitted by its own server

`lib.rs:141-153` now responds to a rejected `mining.submit` with:

```json
{
  "result": null,
  "error": [20, "message", null]
}
```

This is correct on the wire, but the SV1 client currently can't parse it correctly.

`Message` is `#[serde(untagged)]` and defines:

```text
OkResponse(Response)
ErrorResponse(Response)
```

with `OkResponse` first. Since both variants wrap the same type, every response deserializes as `OkResponse`.

`Method::try_from` then routes the response through `GeneralResponse::try_from`, which expects `result.as_bool()` and fails because the rejection now has `result: null`.

I verified this with a small test:

```text
VARIANT: OkResponse
METHOD: Err(ParsingMethodError(MultipleError([
    ImpossibleToParseResultField ...
```

**Why this matters:** Previously, rejected submissions used `result: false`, which happened to parse correctly. The untagged-enum ordering is pre-existing, but this change exposes the problem.

**Possible fix:** Route based on `error.is_some()` either:

* in a custom `Deserialize` implementation for `Message`, or
* inside `Method::try_from`.

That would allow `ErrorResponse` to correctly become `Method::ErrorMessage`.

`GeneralResponse` / `Submit` should also tolerate `result: null` when an error is present.

---

### 2. `mining.extranonce.subscribe` gets a `true` response, but `IsClient` doesn't track its ID

`lib.rs:125-128` now sends an acknowledgement for `mining.extranonce.subscribe`.

That's the correct server behavior — a real ASIC expects the acknowledgement.

However, if anything using `IsClient` sends this request, the response reaches `update_response` as a `GeneralResponse`. Its ID matches neither authorize nor submit, so it becomes:

```text
Error::UnknownID
```

Nothing in-tree currently sends this request, so this is a **latent asymmetry rather than an immediate bug**.

**Possible fix:** Either document this limitation or add something like an `id_is_extranonce_subscribe` hook.

---

### 3. Unauthorized workers still produce a hard `Err` instead of a code-24 response

`lib.rs:141-143` still returns an error when:

```text
is_authorized == false
```

Extranonce2-size and version-bit failures were changed from `Err(InvalidSubmission)` into `[20, ...]` responses, which is good because the miner now receives an actual response.

The unauthorized-worker path still returns `Err`, meaning no JSON-RPC response reaches the miner. The caller's connection loop has to decide what happens next.

Stratum V1 already defines:

* `24` — Unauthorized worker
* `25` — Not subscribed

**Suggestion:** Add these to `SubmitError` and return the corresponding JSON-RPC response instead of a hard error.

---

### 4. `ExtendedChannel::new` doesn't enforce the invariant its setter now enforces

Client-side `ExtendedChannel::new` accepts any:

```text
prefix.len() + rollable_extranonce_size
```

while `set_extranonce_prefix` and `set_upstream_extranonce_prefix` reject values greater than 32.

That means a channel can be constructed in a state that its own setters would reject.

This is pre-existing, but the new setters make the asymmetry more visible.

**Possible options:**

* Validate the invariant in `new` and return `Result`, or
* Document that `new` assumes the caller has already validated it.

---

### 5. No server-side counterpart to `set_upstream_extranonce_prefix`

The allocator documentation describes the motivating scenario as a proxy handling `SetExtranoncePrefix` from upstream.

However, that proxy's downstream channels are:

```text
server::ExtendedChannel
server::StandardChannel
```

Their only prefix setter is:

```text
set_extranonce_prefix(AllocatedExtranoncePrefix)
```

So updating an upstream prefix currently requires a **new allocation for every downstream channel**.

Each new allocation holds two bitmap slots until old jobs become stale, which can lead to `CapacityExhausted` beyond 50% occupancy.

An in-place server-side `set_upstream_extranonce_prefix` seems safe because:

* `local_index` stays unchanged, so there's no slot-reuse hazard.
* Existing jobs already carry their own prefix bytes.
* Only the scriptSig-budget check needs to be re-run.

If this is intentionally deferred to a follow-up, that's fine. Otherwise, the new API currently covers only half of the proxy use case.

## Improvements

### Error type inconsistency

`set_extranonce_prefix` returns:

```text
ExtendedChannelError::NewExtranoncePrefixTooLarge
```

while the sibling `set_upstream_extranonce_prefix` returns:

```text
ExtranoncePrefixError::ExceedsMaxLength
```

for effectively the same condition.

It would be cleaner to map this into `ExtendedChannelError` consistently. The same applies to `StandardChannel`.

### `JsonRpcError` rejects two-element error arrays

The current legacy representation expects exactly:

```json
[20, "message", null]
```

but some servers may send:

```json
[20, "message"]
```

The latter currently fails to deserialize.

Consider deserializing into `Vec<Value>` or using a sequence visitor that accepts either two or three elements.

### `ExtranonceSubscribe::respond(is_ok: bool)`

The only current caller passes `true`, while `handle_extranonce_subscribe` returns `()`.

Either:

* let the handler return `bool` if the server may reject the subscription, or
* remove the parameter if the response is always `true`.

### Semver

This changes public API signatures:

* `IsServer::handle_submit`
* `IsServer::handle_extranonce_subscribe`
* `upstream_prefix_len`: `Option<u8>` → `u8` on three public types

`sv1_api` is currently `6.0.0` and `channels_sv2` is `8.0.0`, so both should get a major-version bump.

Worth mentioning explicitly in the PR description.

### Two-phase upstream-prefix update

`ExtranonceAllocator::set_upstream_prefix` updates the allocator, while existing allocated prefixes continue holding the old bytes until they're updated individually.

The docs mention this, but there's no way to enumerate live prefixes from the allocator.

At minimum, consider either:

* returning the new `upstream_prefix` slice from the setter, making the intended update flow more natural, or
* documenting the required ordering at the module level:

```text
1. Update allocator
2. Update channels
3. Send SetExtranoncePrefix downstream
```

## Redundancy / Cleanup

### Unnecessary `checked_add` / `checked_sub`

Prefix lengths are bounded to ≤ 32, while `rollable_extranonce_size` is `u16`.

These additions cannot overflow `usize` on any supported target, so constructs like:

```rust
checked_add(...)
    .ok_or(ExceedsMaxLength)?
```

add noise and turn an impossible integer overflow into a misleading length error.

Plain arithmetic would be clearer.

### Re-deriving preserved prefix length

The `checked_sub` in `set_upstream_extranonce_prefix` effectively checks:

```text
upstream_prefix_len <= prefix.len()
```

which is already an invariant maintained privately by `ExtranoncePrefix`.

Consider exposing something like:

```rust
local_len()
```

or:

```rust
preserved_len()
```

and using that from both channel types.

### Double bounds check

The client channel first checks:

```text
upstream + preserved + rollable <= 32
```

and then `ExtranoncePrefix::set_upstream_prefix` checks:

```text
upstream + preserved <= 32
```

The inner check cannot fail once the outer one succeeds.

It's harmless, but the layering could be clearer:

* `ExtranoncePrefix` enforces its own prefix invariant.
* The channel enforces the complete extranonce invariant.

### `AllocatedExtranoncePrefix::set_upstream_prefix`

Server channels take ownership of `AllocatedExtranoncePrefix` during construction and don't return it.

That leaves very little useful public surface for calling this setter outside the window between allocation and channel construction.

If it's mainly needed by allocator tests, consider making it `pub(crate)` or replacing its use case with the server-side channel setter mentioned above.

sv1/src/methods/client_to_server.rs
```diff
@@ -106,10 +106,47 @@ fn from_to_json_rpc(auth: Authorize) -> bool {
...
+        })
+    }
+}
+
+impl TryFrom<StandardRequest> for ExtranonceSubscribe {
+    type Error = ParsingMethodError;
+
+    fn try_from(msg: StandardRequest) -> Result<Self, Self::Error> {
+        match msg.params.as_array() {
+            Some(params) if params.is_empty() => Ok(Self { id: msg.id }),
```
plebhash: low severity, non-blocker fable clanker finding:

> this tightens `mining.extranonce.subscribe` parsing: before this PR the method was accepted regardless of `params`, now `"params": null` fails with `ValueNotAnArray` and any non-empty array fails with `WrongArgs`. in the companion (sv2-apps#823, `sv1_server/mod.rs`) a `Client2Server::try_from` error becomes a `TproxyError::disconnect`, so a miner that sends anything other than `[]` here gets dropped instead of subscribed.
> 
> the NiceHash extension defines no params and cgminer-derived firmware sends `[]`, so this is probably fine in practice, but strictness buys nothing on a method whose only payload is the id. suggestion: keep the id and ignore `params` entirely, matching the pre-PR behavior.
>

sv1/src/lib.rs
```diff
@@ -138,16 +154,23 @@ pub trait IsServer {
...
+                    || !has_valid_version_bits
+                {
+                    client_to_server::SubmitOutcome::Rejected(client_to_server::SubmitError::Other)
                 } else {
-                    Err(Error::InvalidSubmission.into())
-                }
+                    self.handle_submit(client_id, &submit)?
+                };
+
+                Ok(Some(submit.respond(outcome)))
```
plebhash, line 175: mid-severity, non-blocking fable clanker finding:

> side effect worth a deliberate call: the unauthorized-worker (24) and extranonce2-size / version-bits (20) rejections are now decided and answered inside the default method, and `handle_submit` is never invoked for them. before this PR they surfaced as `Err(Error::InvalidSubmission)`, which the implementer could log, count, or use to disconnect. now the implementer has no visibility at all, so tProxy cannot include these in its share metrics. `Error::InvalidSubmission` is also no longer produced anywhere in the crate, only its definition and `Display` arm remain in `error.rs`.
> 
> answering with a proper sv1 response instead of a transport error is the right call, the observability just should not go away silently. suggestion: at minimum say in the `handle_submit` rustdoc which rejections never reach it, and drop the dead `Error::InvalidSubmission` variant. if tProxy wants to count them, a `on_submit_rejected(client_id, &submit, SubmitError)` hook with a no-op default is the smallest addition, and this major bump is the cheap moment to add it.
>

sv1/src/json_rpc.rs
```diff
@@ -92,13 +113,53 @@ impl fmt::Display for Response {
...
+
+impl<'de> Deserialize<'de> for JsonRpcError {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum JsonRpcErrorWire {
+            Legacy((i32, String, Option<serde_json::Value>)),
```
plebhash, line 148: low-severity non-blocking fable clanker finding:

> 
> `Legacy` only matches a 3-tuple, so `"error": [20, "Other/Unknown"]` fails the whole `Message` parse with "did not match any variant of untagged enum MessageWire" (a string code like `["20", ...]` fails the same way). not a regression, the derived struct form never parsed arrays at all, and only `IsClient` users are affected, of which the in-tree example is the only one. flagging it because the module docs at the top of `lib.rs` describe the error as a "list (error code, error message)", i.e. two elements.
> 
> suggestion (fine as a follow-up): deserialize the legacy form as `Vec<serde_json::Value>` and accept 2 or 3 elements, with `data` as `None` when absent.
>

sv1/src/methods/client_to_server.rs
```diff
@@ -106,10 +106,47 @@ fn from_to_json_rpc(auth: Authorize) -> bool {
...
+    pub fn respond(self) -> Response {
+        Response {
+            id: self.id,
+            result: Value::Bool(true),
+            error: None,
+        }
+    }
+}
+
+impl From<ExtranonceSubscribe> for Message {
```
plebhash, line 131: low-severity non-blocking fable clanker finding:

> minor asymmetry: this `From` impl gives a client a way to send `mining.extranonce.subscribe`, but the `result: true` ack the server now returns parses as a `GeneralResponse`, and `IsClient::update_response` maps an id that is neither authorize nor submit to `Error::UnknownID`. nothing in-tree sends this request, so it is latent rather than a bug. suggestion: a one-line note on this struct that `IsClient` does not track the id, or leave it until a client needs it.

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -159,28 +161,61 @@ impl ExtranoncePrefix {
...
+    /// prefixes, this method replaces the complete value and keeps it wire-sourced.
+    ///
+    /// Jobs retain the prefix bytes captured when they were created; this only changes the
+    /// current prefix used by future jobs. The prefix is left unchanged if the updated value would
+    /// exceed [`MAX_EXTRANONCE_LEN`].
+    pub fn set_upstream_prefix(
+        &mut self,
+        upstream_prefix: &[u8],
+    ) -> Result<(), ExtranoncePrefixError> {
+        let preserved_bytes = &self.prefix[self.upstream_prefix_len as usize..];
```
plebhash, line 220: low-severity non-blocker fable clanker finding:

> nit: `len - upstream_prefix_len` is the locally owned tail, and the same subtraction is repeated inline in the three channel setters (`client/extended.rs:222`, `server/extended.rs:371`, `server/standard.rs:332`). a `local_len()` accessor next to `upstream_prefix_len()` would let the channels write `upstream_prefix.len() + self.extranonce_prefix.local_len() + rollable` and keep the invariant in one place.

sv2/channels-sv2/src/extranonce_manager/prefix.rs
```diff
@@ -246,15 +281,20 @@ impl AllocatedExtranoncePrefix {
...
-        self.0
-            .upstream_prefix_len()
-            .expect("AllocatedExtranoncePrefix always carries an allocation record")
+        self.0.upstream_prefix_len()
+    }
+
+    /// Replaces `upstream_prefix` while preserving `local_prefix | local_index`,
+    /// any standard-channel rollable padding, and the allocator ownership record.
+    #[inline]
+    pub fn set_upstream_prefix(
```
plebhash: low-severity non-blocking fable clanker finding:

> nit: this is `pub`, but the only callers are the allocator tests. the companion updates prefixes through the channel setters, and server channels take ownership of the `AllocatedExtranoncePrefix` at construction, so there is no window to call it afterwards. `pub(crate)` unless there is a planned external user.

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -150,22 +150,34 @@ impl StandardChannel {
...
 
+    /// Sets the upstream-assigned region of this channel's extranonce prefix.
+    ///
+    /// For an allocator-produced prefix, `local_prefix | local_index`, rollable padding, and its
+    /// allocation are preserved. For a wire-sourced prefix, the entire prefix is
+    /// `upstream_prefix` and is replaced. Jobs received before this call retain their captured
+    /// prefix bytes; new jobs use the updated prefix. Returns
+    /// [`StandardChannelError::NewExtranoncePrefixTooLarge`] without changing the channel if the
+    /// resulting prefix would exceed [`MAX_EXTRANONCE_LEN`].
+    pub fn set_upstream_extranonce_prefix(
```
plebhash, line 256: low-severity non-blocker fable clanker finding:

> nit: this is the only one of the four new channel setters without a test. the other three assert the preserved suffix, the allocation count and the transactional failure; a copy of `set_upstream_extranonce_prefix_enforces_full_extranonce_size_transactionally` from `client/extended.rs`, adapted to `allocate_standard`, would close the gap.

sv2/channels-sv2/src/extranonce_manager/allocator.rs
```diff
@@ -203,6 +203,49 @@ impl ExtranonceAllocator {
...
+    /// channels, update group-channel size where applicable, and finally send
+    /// `SetExtranoncePrefix` downstream. If the transition cannot be completed, retain the old
+    /// layout or enter the application's fallback path. Jobs created before a successful update
+    /// continue using the prefix bytes captured when they were created.
+    ///
+    /// The allocator is left unchanged if the resulting full extranonce would
+    /// exceed [`MAX_EXTRANONCE_LEN`].
+    pub fn set_upstream_prefix(
+        &mut self,
+        upstream_prefix_bytes: Vec<u8>,
```
plebhash, line 233: low-severity non-blocker fable clanker finding:

> nit: the allocator takes `Vec<u8>` while every other `set_upstream_prefix` / `set_upstream_extranonce_prefix` takes `&[u8]`, so the companion clones the same bytes once for the allocator and borrows them for each channel. taking `&[u8]` here too and copying internally keeps the four signatures uniform.

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -188,12 +191,15 @@ impl ExtendedChannel {
     /// Jobs created before this call retain the previous extranonce prefix,
```
plebhash, line 217: low-severity non-blocking fable clanker finding:

> ~~not asking to solve #2330 here (agreed it is a separate PR), but with this PR the crate ships two prefix-update paths with different guarantees: this one keeps the promise that pre-rotation jobs stay valid while dropping the old lease on line 206 (the #2330 part-1 hazard), and `set_upstream_extranonce_prefix` below keeps the lease attached. suggestion: a one-line pointer in this doc saying that callers with live jobs under an allocator-minted prefix should prefer `set_upstream_extranonce_prefix`, so the two are not picked interchangeably until #2330 lands.~~

plebhash: after human inspection of the finding above, I realized there's some overlap with #2336

so the clanker finding is reworded into:

> heads-up on #2336, which closes #2330 by making the client channels' `set_extranonce_prefix` retain the rotated-out lease until no live job references it. once that lands the two prefix-update paths have the same liveness guarantee, so nothing to document here. whichever of the two merges second needs a rebase; i ran a scratch merge of both heads and the only textual conflicts are in `client/extended.rs`, all mechanical: the `ExtendedChannel::new` rustdoc and its first check (both PRs make the constructor fallible for a different reason, both checks must survive), the new `retired_extranonce_prefixes` field init, the `set_extranonce_prefix` rustdoc (keep both paragraphs), and the test-module imports. one semantic conflict on top: #2336 turns `ExtendedJob` into a struct, so `upstream_prefix_update_only_affects_subsequent_jobs` must read `.extranonce_prefix` instead of `.1` at lines 921-922. on the sv2-apps side, sv2-apps#823 and sv2-apps#826 overlap on the translator's `channel_manager/{mod,mining_message_handler}.rs`, `error.rs`, `monitoring.rs` and JDC's `upstream_message_handler.rs`, so the companion of the second one needs the same treatment.

plebhash: this scope overlap had been flagged here: https://github.com/stratum-mining/stratum/issues/2330#issuecomment-5501652427

sv1/src/methods/client_to_server.rs
GitGab19: Fixed, now `mining.extranonce.subscribe` keeps the request ID and ignores the params, preserving the previous parsing behavior

sv1/src/lib.rs
GitGab19: Added some more docs about the checks performed before the `handle_submit`, and removed the stale `InvalidSubmission` variant

sv1/src/json_rpc.rs
GitGab19: Done, deserialization now accepts both `[code, message]` and `[code, message, data]`

sv1/src/methods/client_to_server.rs
GitGab19: Added the note

sv2/channels-sv2/src/extranonce_manager/prefix.rs
GitGab19: Added a `preserved_len()` helper for this

sv2/channels-sv2/src/extranonce_manager/prefix.rs
GitGab19: Removed it

sv2/channels-sv2/src/client/standard.rs
GitGab19: Added the test for this

sv2/channels-sv2/src/extranonce_manager/allocator.rs
GitGab19: Kept `Vec<u8>` because the allocator stores that buffer directly, allowing callers with an owned buffer to transfer it without another copy

sv2/channels-sv2/src/client/extended.rs
GitGab19: Just rebased against main, incorporating the changes you made on #2336

GitGab19:
@bit-aloo, about your clanker review:

1. Fixed response parsing, now errors are identified through the error field and correctly reach `Method::ErrorMessage`.
2. Added a note about the `IsClient` limitation. I'd keep the routing changes out of scope here.
3. Added codes 24 and 25 to `SubmitError`. The default handler uses 24 for unauthorized workers, while tProxy uses 25 for submits received before subscription.
4. Added the full `extranonce-size` check to the constructor.
5. Added the upstream-prefix setters to both server channel types, preserving the allocation and the prefix captured by existing jobs.
For the other points:
- Normalized the channel errors and simplified the length arithmetic.
- Added support for two-element error arrays, while still serializing three elements.
- Removed the boolean argument from `ExtranonceSubscribe::respond` and the forwarding setter on `AllocatedExtranoncePrefix`.
- Added `preserved_len()` and expanded the docs about coordinating allocator and channel updates.

About the versions, they are already bumped compared to the published ones, so another major bump isn't needed.

plebhash (approved):

sv1/src/lib.rs
```diff
@@ -122,9 +140,9 @@ pub trait IsServer {
                 let (version_rolling, min_diff) = self.handle_configure(client_id, &configure)?;
                 Ok(Some(configure.respond(version_rolling, min_diff)))
             }
-            methods::Client2Server::ExtranonceSubscribe(_) => {
-                self.handle_extranonce_subscribe()?;
-                Ok(None)
+            methods::Client2Server::ExtranonceSubscribe(subscribe) => {
+                self.handle_extranonce_subscribe(client_id)?;
+                Ok(Some(subscribe.respond()))
```
bit-aloo, line 145: Clanker findings:

> Now that the server acks mining.extranonce.subscribe, the client side needs to be able to recognise that ack. update_response funnels every untracked general-response id into Server2ClientResponse::Submit (line 377), so this response will be delivered to handle_submit_response as a phantom accepted share — corrupting whatever share accounting or logging the application hangs off that handler. Could IsClient track the extranonce-subscribe id the same way it tracks authorize ids, so update_response can route it to its own arm? Documenting the gap on ExtranonceSubscribe isn't enough given this crate now produces the message on the server side.

sv1/src/json_rpc.rs
```diff
@@ -11,6 +11,27 @@ pub enum Message {
     ErrorResponse(Response),
```
bit-aloo, line 11: Seeing all the serde gymnastics we are doing just to parse `Response` as either `Error` or `Ok`, I wonder if it would be cleaner to model them as variants of a single `Response` type and provide something like an `is_error` method on it. I think that would make the code quite a bit easier to reason about.

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -327,6 +327,44 @@ impl StandardChannel {
...
+    /// prefix.
+    ///
+    /// The channel is left unchanged if the resulting prefix exceeds
+    /// [`MAX_EXTRANONCE_LEN`] or the assembled coinbase `scriptSig` budget.
+    ///
+    /// Old prefix bytes share ownership of the same allocator slot while any future, active or
+    /// past job uses them. A later `set_extranonce_prefix` rotation cannot release that slot
+    /// prematurely. Once those jobs are stale or evicted, only the current prefix (if it still
+    /// uses this allocation) keeps the slot reserved. This update consumes no additional slot.
+    pub fn set_upstream_extranonce_prefix(
```
bit-aloo, line 350: Clanker review:

> When upstream_prefix.len() differs from the current upstream region, this changes the channel's full_extranonce_size, which is the invariant GroupChannel enforces across its members. The only way to fix the group afterwards is set_full_extranonce_size, which drops every member channel id. Two options: give this method its own doc paragraph spelling out that a length-changing update requires rebuilding the group, or add a GroupChannel test covering the transition so the cost is visible. The README clause at allocator.rs:224 is easy to miss for someone reading this method.

bit-aloo (commented):
Few suggestions, clanker pointed out nits, but not important

sv1/src/lib.rs
GitGab19: Added `id_is_extranonce_subscribe` with its dedicated response variant

sv1/src/json_rpc.rs
GitGab19: I agree that the two public variants are somewhat redundant, but the response is already deserialized only once and then classified from its `error` field

sv2/channels-sv2/src/server/standard.rs
GitGab19: Expanded the docs about it

bit-aloo (approved):
ACK

PR #2320, opened 2026-08-27
fix(ci): prevent false green IT skip when no companion is present

GitGab19:
When a PR has no companion line and the body is edited, the old logic compared empty strings, found them equal, and skipped the test run with a green check — masking the prior failing run.

Only skip on edited when a companion line actually exists and was unchanged by the edit. If there's no companion line at all, re-run the tests.

TEST BY EDITING THE DESCRIPTION --> IT SHOULD BE TRIGGERED AND NOT SKIP


bit-aloo (approved):
ACK

PR #2328, opened 2026-08-31
Acknowledge field and struct visibility in parser

### LINKED ISSUE #2168: Binary_sv2(proc_macro): Make visibility intentional
We only support pub and private fields only, and reject pub(crate) / pub(super). We should incorporate these visibility rules.

bit-aloo:
closes: #2168

### DIFF
--- a/sv2/binary-sv2/derive_codec/src/lib.rs
+++ b/sv2/binary-sv2/derive_codec/src/lib.rs
@@ -37,6 +37,11 @@
 //! Parses and captures a struct’s name, generics, and field data, enabling custom encoding and
 //! decoding functionality.
 //!
+//! ### Visibility
+//! The generated `{Name}Owned` struct is emitted next to the source struct and carries the
+//! visibility written on it (`pub`, `pub(crate)`, `pub(super)`, private, ...); each of its fields
+//! likewise carries the visibility of the corresponding source field.
+//!
 //! ### Custom Implementations
 //! The `Encodable` macro generates an `EncodableField` implementation by serializing each field,
 //! while `Decodable` constructs the struct from binary data. Both macros provide support for
@@ -221,6 +226,8 @@ enum ParserState {
 struct ParsedStruct {
     // Name of the struct.
     pub name: String,
+    // Visibility of the struct, verbatim (empty when private).
+    pub vis: String,
     // Generics associated with the struct, if any.
     pub generics: String,
     // List of fields within the struct.
@@ -247,6 +254,8 @@ struct ParsedField {
     name: String,
     // Span of the field name.
     span: Span,
+    // Visibility of the field, verbatim (empty when private).
+    vis: String,
     // Type of the field.
     type_: String,
     // Generics associated with the field, if any.
@@ -263,6 +272,7 @@ impl ParsedField {
         ParsedField {
             name: "".to_string(),
             span: Span::call_site(),
+            vis: "".to_string(),
             type_: "".to_string(),
             generics: "".to_string(),
         }
@@ -464,10 +474,15 @@ fn get_struct_properties(item: TokenStream) -> Result<ParsedStruct, MacroError> 
     let item = remove_attributes(item);
     let mut stream = item.into_iter();
 
-    // Check if the stream is a struct
+    // Check if the stream is a struct, collecting its visibility on the way
+    let mut struct_vis = "".to_string();
     loop {
         match stream.next() {
             Some(TokenTree::Ident(i)) if i.to_string() == "struct" => break,
+            Some(TokenTree::Ident(i)) if i.to_string() == "pub" => struct_vis = i.to_string(),
+            Some(TokenTree::Group(g)) if !struct_vis.is_empty() => {
+                struct_vis = format!("{struct_vis}{g}")
+            }
             Some(_) => continue,
             None => {
                 return Err(MacroError {
@@ -533,6 +548,7 @@ fn get_struct_properties(item: TokenStream) -> Result<ParsedStruct, MacroError> 
 
     Ok(ParsedStruct {
         name: struct_name,
+        vis: struct_vis,
         generics: struct_generics,
         fields,
     })
@@ -572,7 +588,7 @@ fn parse_struct_fields(group: Vec<TokenTree>) -> Result<Vec<ParsedField>, MacroE
         match (token, &field_parser_state) {
             (TokenTree::Ident(i), ParserState::Name) => {
                 if i.to_string() == "pub" {
-                    continue;
+                    field_.vis = i.to_string();
                 } else {
                     field_.name = i.to_string();
                     field_.span = i.span();
@@ -632,6 +648,9 @@ fn parse_struct_fields(group: Vec<TokenTree>) -> Result<Vec<ParsedField>, MacroE
                         field_.generics = format!("{}{}", field_.generics, p);
                     }
                 }
+            }
+            (TokenTree::Group(g), ParserState::Name) if !field_.vis.is_empty() => {
+                field_.vis = format!("{}{}", field_.vis, g);
             }
             (token, _) => {
                 return Err(MacroError {
@@ -815,9 +834,9 @@ pub fn decodable(item: TokenStream) -> TokenStream {
         };
         let owned_struct_field = format!(
             "
-            pub {}: {},
+            {} {}: {},
             ",
-            f.name, owned_ty
+            f.vis, f.name, owned_ty
         );
         owned_struct_fields.push_str(&owned_struct_field);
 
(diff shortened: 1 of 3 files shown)


plebhash:
https://github.com/stratum-mining/stratum/pull/2328/commits/48b38a1879307407d9791e899e08ade2e3e3eaad

`add tewst`?

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -464,10 +474,15 @@ fn get_struct_properties(item: TokenStream) -> Result<ParsedStruct, MacroError>
     let item = remove_attributes(item);
     let mut stream = item.into_iter();
 
-    // Check if the stream is a struct
+    // Check if the stream is a struct, collecting its visibility on the way
+    let mut struct_vis = "".to_string();
     loop {
         match stream.next() {
             Some(TokenTree::Ident(i)) if i.to_string() == "struct" => break,
+            Some(TokenTree::Ident(i)) if i.to_string() == "pub" => struct_vis = i.to_string(),
```
plebhash: fable clanker finding:

> **Regression:** visibility that reaches the derive wrapped in a `Delimiter::None` group — which is what a `macro_rules!` `$vis:vis` fragment produces — matches neither of these arms: the `pub` arm needs a bare `Ident`, and the `Group` arm requires `struct_vis` to already be non-empty. It falls into `Some(_) => continue` and is silently dropped, so the generated `{Name}Owned` comes out private even though the source struct is `pub`:
>
> ```rust
> macro_rules! def {
>     ($v:vis $n:ident) => {
>         #[derive(Clone, Deserialize, Serialize, PartialEq, Debug)]
>         $v struct $n<'decoder> { pub d: B0255<'decoder> }
>     };
> }
> def!(pub MacroMsg); // in mod inner; inner::MacroMsgOwned used from the parent
> ```
>
> This compiles on `main` (Owned was unconditionally `pub use`-exported) and fails with E0603 on this branch — verified by differential compile against both trees.
>
> Since `remove_attributes` has already stripped attributes and rustc guarantees derive input is a valid item, only visibility tokens can precede `struct` — so collecting them verbatim handles every form (`pub`, `pub(crate)`, `pub(in path)`, and `$vis:vis` None-groups, which `Display` as their contents) with no arm-ordering coupling:
>
> ```rust
> Some(TokenTree::Ident(i)) if i.to_string() == "struct" => break,
> Some(t) => struct_vis.push_str(&t.to_string()),
> ```
>
> (Enum/union input still ends at the same "Expected a struct definition" error — the collected string is discarded.)

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -464,10 +474,15 @@ fn get_struct_properties(item: TokenStream) -> Result<ParsedStruct, MacroError>
     let item = remove_attributes(item);
```
plebhash, line 475: fable clanker finding:

> **Second, independent path that loses the visibility:** `remove_attributes` never resets `is_attribute` after skipping an attribute's `[...]` group. When a doc comment or attribute precedes a `$vis:vis` fragment, the None-delimited vis group that follows is swallowed as if it were still part of the attribute — the vis never reaches the vis-collection loop at all, so fixing that loop alone won't cover this.
>
> Struct-level: macro emits `/// docs` then `$v struct M<'decoder> {...}` with `$v = pub` → `MOwned` is private (E0603 at use sites); the same code compiles on `main`. Field-level is worse because it's silent: `/// field docs` + `$fv f: B0255<'decoder>` with `$fv = pub` → the derive *succeeds* but the Owned field is private (E0616 from outside the module), instead of the loud "Unexpected token" error you'd get without the doc comment.
>
> Fix: reset `is_attribute = false` after consuming the bracket group in `remove_attributes`.

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -633,6 +649,9 @@ fn parse_struct_fields(group: Vec<TokenTree>) -> Result<Vec<ParsedField>, MacroE
                     }
                 }
             }
+            (TokenTree::Group(g), ParserState::Name) if !field_.vis.is_empty() => {
```
plebhash: fable clanker finding:

> Field-level `$vis:vis` with no preceding attribute still hard-errors here: the None-delimited group fails the `!field_.vis.is_empty()` guard and falls to the "Unexpected token" arm. Pre-existing (same error on `main`), but the README and module docs added in this PR now claim each Owned field "carries the visibility written on the source field", which isn't true for macro-generated fields. The same None-group handling that fixes the struct level applies here — or scope the docs claim down.

sv2/binary-sv2/tests/test.rs
```diff
@@ -747,3 +747,47 @@ mod test_to_writer_len {
...
+                &self.private
+            }
+        }
+    }
+
+    use core::convert::TryInto;
+    use message::{Message, MessageOwned};
+
+    #[test]
+    fn test_owned_struct_keeps_source_visibility() {
```
plebhash, line 821: fable clanker finding:

> This test only pins the permissive direction — every access it makes is also legal under the old always-`pub` emission. Mutation-tested: reverting the macro to emit `pub struct MessageOwned` with all-`pub` fields (the exact pre-PR behavior) still passes. So the PR's headline guarantee — visibility is *not widened* — is unpinned, and a regression re-widening `pub(crate)`/private to `pub` would ship green.
>
> Pinning the restrictive direction needs a compile-fail probe, e.g. `trybuild` accessing `owned.private` (expect E0616) or naming `MessageOwned` outside its `pub(crate)` scope (expect E0603).

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -37,6 +37,11 @@
 //! Parses and captures a struct’s name, generics, and field data, enabling custom encoding and
 //! decoding functionality.
 //!
+//! ### Visibility
+//! The generated `{Name}Owned` struct is emitted next to the source struct and carries the
```
plebhash, line 41: fable clanker finding:

> The `Decodable` docstring example further down (~line 756) still shows the layout this section replaces: `TestOwned`/`into_owned`/`as_owned` generated for a struct *without* generics, and everything inside `mod impl_parse_decodable_test`. Actual behavior: the Owned mirror only exists for generic structs, and after this PR it's emitted next to the source struct with source visibility. Worth updating in the same PR since it now contradicts this section.

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -911,16 +930,24 @@ pub fn decodable(item: TokenStream) -> TokenStream {
...
-    pub struct {}Owned {{
+    {} struct {}Owned {{
         {}
-    }}
+    }}",
+            parsed_struct.vis, parsed_struct.name, owned_struct_fields,
+        )
+    };
 
+    let owned_impl = if parsed_struct.generics.is_empty() {
```
plebhash: fable clanker finding:

> nit: duplicates the identical `if parsed_struct.generics.is_empty()` guard 13 lines up. If the emission condition ever changes, updating one but not the other emits an Owned struct without impls (or impls referencing a nonexistent struct), failing confusingly at expansion time in consumer crates. One conditional keeps them in lockstep:
>
> ```rust
> let (owned_struct, owned_impl) = if parsed_struct.generics.is_empty() {
>     (String::new(), String::new())
> } else {
>     (format!(/* struct */), format!(/* impls */))
> };

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -973,33 +998,21 @@ pub fn decodable(item: TokenStream) -> TokenStream {
...
     use ::binary_sv2::{{decodable::DecodableField, decodable::FieldMarker, encodable::EncodableField, Decodable, Error, GetSize, SizeHint}};
     use super::*;
 
     {}
     {}
-    }}
-    {}",
+    }}",
+        owned_struct,
         // imports
```
plebhash: fable clanker finding:

> nit: stale comment — it labeled the deleted imports slot and now sits above `parsed_struct.name.to_lowercase()` (the module-name arg), mislabeling it. Delete it; mismapping these positional args is exactly the mistake it invites (an empty ```suggestion``` block on this line does it).

sv2/binary-sv2/derive_codec/src/lib.rs
```diff
@@ -572,7 +588,7 @@ fn parse_struct_fields(group: Vec<TokenTree>) -> Result<Vec<ParsedField>, MacroE
         match (token, &field_parser_state) {
             (TokenTree::Ident(i), ParserState::Name) => {
                 if i.to_string() == "pub" {
-                    continue;
+                    field_.vis = i.to_string();
```
plebhash: fable clanker finding:

> nit: `to_string()` runs twice on the same ident (guard on 590, then here); same pattern at line 482. Bind it once, or assign `"pub".to_string()` since the guard proved equality. Related: this `pub` + optional `(...)` micro-grammar now exists twice in the file (here and 482–485) as two differently-shaped state machines — the verbatim-collect shape from comment 1 would make a good shared helper.

plebhash (approved):

PR #2336, opened 2026-09-02
`channels_sv2`: harden share dedup, extranonce-prefix rotation, BIP141 witness framing, and job/target validation

### LINKED ISSUE #2329: `channels_sv2`: `seen_shares` dedup can be invalidated while its shares are still creditable (server flush on unchanged `prev_hash` + client FIFO eviction)
Follow-up to #2297, which bounded the accepted-share dedup cache and made server-side exhaustion a hard error. "Three aspects of the dedup-cache lifecycle remain: two unintended gaps (parts 1–2) plus a deliberately accepted trade-off on the client (part 3). The first two are the same defect on the two server channel types (the flush is keyed to "a tip/job message arrived" instead of "`prev_hash` actually changed"); the third is the client-side analog exposed by the #2297 bound itself.

### 1. `server/extended.rs` — min-`ntime`-only custom job change flushes dedup

`on_set_custom_mining_job` treats any change in `{prev_hash, nbits, min_ntime}` as a tip transition and flushes `seen_shares`:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/extended.rs#L685-L701

But `min_ntime` is not committed into the block header — it is only a lower bound on the share's `ntime` — and neither is the job ID. A downstream custom-job peer can keep `prev_hash`, nbits, coinbase, extranonce, nonce, version and header `ntime` identical, advance only `min_ntime`, and resubmit the exact same proof under the new job ID: the flush erased its hash, so it is credited a second time instead of returning `DuplicateShare`. Repeating min-time advances multiplies credit for one proof.

### 2. all `on_set_new_prev_hash` sites — repeated `SetNewPrevHash` flushes dedup

`on_set_new_prev_hash` flushes unconditionally on every channel type, without establishing that `prev_hash` changed:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/standard.rs#L623-L624

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/extended.rs#L644-L645

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/extended.rs#L599-L600

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/standard.rs#L394-L395

(the JDC-only `client/extended.rs::on_chain_tip_update` at L548 has the same pattern but interprets a locally supplied tip rather than a peer message — include it for symmetry.)

A non-conforming or malicious Template Distribution (server) / upstream mining (client) peer can queue an otherwise-identical future template under a new `template_id` and repeat the same `SetNewPrevHash` fields. Template ID and job ID are not part of the block header, so the replacement job commits to the same header — and the previously accepted share becomes creditable again. Each repetition of the pair re-arms the replay.

### 3. client — bounded FIFO eviction is the documented, accepted behavior

The client cache evicts oldest-first at `MAX_SEEN_SHARES`:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/share_accounting.rs#L217-L223

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/mod.rs#L50

After 4 096 accepted hashes under one still-active tip, the oldest hash is forgotten, so a replay can re-pass `is_share_seen` and re-forward upstream. This is **deliberate**, and this section exists to document why eviction beats the server-style hard failure on clients:

- **Targets are upstream-controlled.** Server channels derive a dedup budget from the pool-configured expected share rate (`seen_shares_budget`, clamped to `[MIN_SEEN_SHARES_CAP, MAX_SEEN_SHARES_CAP]`). A client cannot derive a bound against an upstream that picks the target, so the flat device-safe `MAX_SEEN_SHARES` (≈ 0.3 MB, `VecDeque`) is the only honest response to an untrusted rate input.
- **The double-count is a local statistic.** Client-side `validated_shares`/`validated_work_sum`/`blocks_found` are pre-upstream numbers; the upstream's own server dedup still rejects the replayed share, so nothing is paid out on it. The in-tree rustdocs call the evicted-then-replayed hash "one double-counted local statistic" — eviction is exactly the priced tradeoff.
- **Hard failure is a DoS gift.** Failing validation at the bound (audit #170's fix, mirroring #2297's server semantics) hands a hostile upstream advertising a trivial target a one-message channel-kill vector. Eviction keeps the channel alive at the cost of a bounded, upstream-rejected replay.

### impact

Parts 1–2 (audit severity medium for 1, low for 2): accepted-share accounting can be inflated by replaying a single proof. Exploitation needs a protocol-violating TP (2) or control of the custom-job stream (1) plus a miner willing to replay. Part 3 documents deliberately accepted-replay risk (audit severity medium, closed as wontfix); it does not alter the findings.

### fix

- **Server (both channels)**: flush only when `prev_hash` actually changes. In `server/extended.rs::on_set_custom_mining_job`, split `is_new_chain_tip` so the stale/target bookkeeping still runs on any tip-field change but `flush_seen_shares` is gated on `prev_hash` inequality — retaining hashes across an nbits-only change is harmless because a different nbits yields a different header hash. For every `on_set_new_prev_hash` site (both server channels plus both client channels), gate the flush the same way on the incoming `prev_hash`; update the rustdocs accordingly and state the retained-hashes invariant.
- **Client (part 3)**: no behavior change on eviction — the audit finding (https://github.com/project-loupe/audit-stratum/issues/170) documents a deliberate, accepted trade-off. The `validate_share` and `ShareAccounting` rustdocs should explicitly state the replay window and why hard failure was rejected (see section 3 above). Mark the audit issue as accepted-risk.

### regression tests

- **extended**: accept a share, apply a min-`ntime`-only `SetCustomMiningJob` (same `prev_hash`), resubmit the identical proof under the new job ID — must be `DuplicateShare`. PoC: https://github.com/project-loupe/audit-stratum/issues/176
- **standard**: accept a share, repeat the same tip under a new `template_id`, resubmit under the replacement job ID — must be `DuplicateShare` (or the repeated tip rejected). PoC: https://github.com/project-loupe/audit-stratum/issues/180
- **client**: rustdoc statement, no behavior change

### LINKED ISSUE #2330: `channels_sv2`: extranonce prefix rotation lifecycle gaps (client lease released with live jobs; server retired-prefix bloat)
Follow-up to #2260, which fixed the server-side half of this class by retaining retired prefix leases in `JobStore` while jobs reference their bytes. Two gaps remain at opposite ends of the same mechanism: client channels still release the lease too early, and the server-side retention can be bloated indefinitely with dead tokens.

### 1. client channels release the allocation while pre-rotation jobs are live

`ExtranoncePrefix`'s `Drop` unconditionally clears the allocator bitmap slot:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/extranonce_manager/prefix.rs#L276-L283

and both client channel types replace the owning prefix in `set_extranonce_prefix` while documenting that pre-rotation jobs stay valid:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/extended.rs#L176-L187

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/standard.rs#L126

Jobs retain only a copy of the prefix bytes, so the old lease drops immediately and its slot can be reissued to another live channel — while an old job still validates shares in the same extranonce space. Per-channel duplicate tracking gives no cross-channel protection, so identical-template shares can be replayed across the two channels.

### 2. server `JobStore` accumulates dead, byte-identical retired tokens

`retire_extranonce_prefix` appends every retired token whose bytes match a live job:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/jobs/job_store.rs#L296-L301

and `prune_retired_extranonce_prefixes` retains by byte equality:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/jobs/job_store.rs#L308-L322

so N byte-identical tokens all match the same live job and all survive pruning. Repeated prefix rotations with freshly allocated but byte-identical prefixes (reachable via the server channels' `set_extranonce_prefix`) grow the vector once per update — even when the allocator behind each token has been dropped and the token no longer reserves any slot. A peer/control-plane actor that can induce allocator recreation while keeping one matching job alive (and withholding a tip transition) grows heap usage without bound.

### impact

1 — medium (audit): overlapping extranonce space across channels and cross-channel share replay. Exact routing depends on the embedding proxy, but the in-tree API explicitly supports locally allocated client prefixes with live pre-rotation jobs.
2 — low (audit): memory exhaustion; requires inducing repeated identical rotations.

### fix

- **Client**: mirror #2260's server-side retention — keep the retired prefix alive in channel state until no stored job references its bytes. Both client channels need it; their job stores are per-channel state, not the server `JobStore`.
- **Server**: make retention conditional on the token still being live — e.g. a `has_expired_allocation()` helper on `ExtranoncePrefix` (weak-count check on the allocator bitmap), skipping expired tokens at retire time and pruning expired entries before pushing. Byte-identical rotations then retain at most one effective token.

### regression tests

- **client**: two-slot allocator; store an active job under slot 0, rotate to slot 1, assert slot 0 is not reissuable while the old job is live. PoC: https://github.com/project-loupe/audit-stratum/issues/171
- **server**: retire 10 000 dead byte-identical tokens against one live job; assert retired storage stays bounded. PoC: https://github.com/project-loupe/audit-stratum/issues/174

### LINKED ISSUE #2331: `channels_sv2`: `try_strip_bip141` does not validate witness framing (accepts any nonzero flag; never checks count/length)
Complement of the now-fixed short-slice panics in the same function (audit-stratum https://github.com/project-loupe/audit-stratum/issues/469, https://github.com/project-loupe/audit-stratum/issues/742, https://github.com/project-loupe/audit-stratum/issues/1094, https://github.com/project-loupe/audit-stratum/issues/2308). Two validation gaps remain in `try_strip_bip141`, both letting wire-sourced `NewExtendedMiningJob` coinbase bytes be rewritten into a job that downstream miners hash, instead of being rejected.

### 1. any nonzero flag is treated as the current witness format

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/bip141.rs#L52-L58

BIP141's extensible envelope requires flag `0x01` for the currently defined format; flags `0x02..=0xff` may gate additional serialized sections with a different layout. HEAD accepts them all and deletes a present-day fixed-size witness tail.

### 2. witness count/length framing bytes are never checked

After recognizing marker/flag, the function assumes the 34 bytes before locktime are `count=1 || length=32 || reserved_value` and blindly subtracts the fixed size:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/bip141.rs#L70-L81

A peer can send marker `00`, flag `01`, and a suffix whose alleged witness declares multiple items or a non-32-byte item; the helper still returns `Ok(Some(..))`. Per BIP141, a coinbase carrying the witness commitment must have exactly one 32-byte reserved-value item.

### impact

Low (audit severity): requires a malicious or faulty upstream controlling the job stream; the damage is job integrity/availability — extended-channel callers and the SV2→SV1 translator accept and propagate work derived from an invalid or differently interpreted transaction.

### fix (single function, covers all callers)

Reject `flag != 0x01` at the classification site, and validate the two framing bytes before stripping:

```rust
if has_bip141_marker_and_flag && flag != 0x01 {
    return Err(StripBip141Error::FailedToDeserializeCoinbaseInputs);
}
// ...
if coinbase_tx_suffix[witness_position] != 1
    || coinbase_tx_suffix[witness_position + WITNESS_COUNT_LEN] != WITNESS_DATA_LEN as u8
{
    return Err(StripBip141Error::FailedToDeserializeCoinbaseWitness);
}
```

Update the rustdoc's `# Errors` section for both new conditions.

### regression tests

- flag `0x02` with an otherwise well-formed suffix must be rejected. PoC: https://github.com/project-loupe/audit-stratum/issues/169
- count `2` and length `31` variants must be rejected with `FailedToDeserializeCoinbaseWitness`. PoC: https://github.com/project-loupe/audit-stratum/issues/168

### LINKED ISSUE #2332: `channels_sv2`: server `StandardChannel::validate_share` ignores the group job's own `min_ntime`
Server-side complement of #2298, which fixed the same defect on the client channels. A `StandardJob` installed via `on_group_channel_job` carries the group job's own advertised `min_ntime` — `into_standard_job` preserves it:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/standard.rs#L533-L540

but `validate_share` enforces only the chain-tip bound:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/standard.rs#L713

If the embedding application installs a newly advanced group job before updating this standard channel's `chain_tip` (or otherwise supplies a group job whose minimum is later), a share with `chain_tip.min_ntime() <= ntime < job.min_ntime` is accepted and accounted — and can be returned as `BlockFound` — despite violating the job's advertised lower bound.

### impact

Low (audit severity): the miner still needs valid proof of work, and reachability depends on caller ordering around group-job fanout — no in-tree production caller establishes the required invariant.

### fix

Same shape as #2298: reject `share.ntime < job.min_ntime` alongside the existing chain-tip check, making the effective bound `max(tip, job)`. Update the `validate_share` rustdoc to state the per-job bound.

### regression test

Set the channel tip at nTime *t*, install a group job with `min_ntime > t` via `on_group_channel_job`, assert a share in the gap is rejected. PoC: https://github.com/project-loupe/audit-stratum/issues/178

### LINKED ISSUE #2333: `channels_sv2`: zero target is accepted and poisons share accounting with non-finite difficulty (all channel types)
`Target::difficulty_float()` is `TARGET_MAX_F64 / target`, and its rustdoc states it returns `f64::INFINITY` for a zero target. Every `validate_share` implementation credits `job_target.difficulty_float()` into `ShareAccounting` without any check that the target is nonzero — so a channel operating at a zero target poisons cumulative work with infinity on the first accepted share, and the wire-facing `get_last_batch_work_sum()` (`self.last_batch_work_sum as u64`) saturates to `u64::MAX`.

No constructor or setter rejects a zero target:

- `server/extended.rs` — the downstream-selected `max_target` is only clamped against (`let target = target.min(max_target)`), never validated; jobs record `self.target` per job ID:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/extended.rs#L221-L224

- `server/standard.rs` — same clamp-only pattern for `requested_max_target`:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/standard.rs#L206-L209

- `client/extended.rs` / `client/standard.rs` — the target is upstream-controlled and installed verbatim via `set_target`:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/extended.rs#L210-L214

The accounting sink in all four is the same `update_share_accounting(job_target.difficulty_float(), ...)` call — e.g.:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/server/extended.rs#L905-L920

Note the ordering in `validate_share`: the block-found branch runs **before** the job-target check, so on a zero-target channel an ordinary (non-block) share is rejected as `DoesNotMeetTarget`, but a hash meeting the *network* target is accounted at `difficulty_float() == INFINITY`. A miner can therefore open a zero-target channel, mine at the real network target, and poison accounting or obtain a grossly inflated share acknowledgement with a single block-valid submission.

### impact

Medium, matching the audit severity: on mainnet the trigger requires finding a real block, which bounds exploitability; on easier configured networks the precondition is cheaper. Server side the corruption reaches payout-adjacent accounting (`get_last_batch_work_sum` is the wire value sent in `SubmitSharesSuccess`); client side a hostile upstream can inflate locally validated statistics. Concrete payout use and any caller-side rejection of zero `max_target` live in embedding applications, outside this crate.

### fix

Reject the unrepresentable value at the trust boundary rather than special-casing accounting:

- **Server channels**: return an error from the constructors and `update_channel` when `max_target`/`requested_max_target` is zero (a new `ExtendedChannelError`/`StandardChannelError` variant, e.g. `OpenChannelInvalidMaxTarget`/`UpdateChannelInvalidMaxTarget` carrying the spec error-code string, mirroring the existing `*InvalidNominalHashrate` variants).
- **Client channels**: reject at the trust boundary as well, mirroring the server channels — no clamping (a clamped minimum still admits no share and still yields an astronomical `difficulty_float()`), and no special-casing in `validate_share` (that keeps an unattainable target in channel state and turns the condition into an endless stream of `invalid-share` rejections that the embedding application cannot tell apart from bad shares). Both client constructors and `set_target` return a new `InvalidTarget` variant of the client `ExtendedChannelError`/`StandardChannelError` for a zero target and leave the channel unchanged, so no job can ever carry a target no share can meet. This makes `ExtendedChannel::new`/`StandardChannel::new` and `set_target` fallible on the client side; the embedding applications (in sv2-apps: the translator's and JDC's client-channel construction and `SetTarget` handlers) have to handle the error, which is the point — an upstream advertising an unattainable target should trigger fallback/disconnect, not silent rejection of every share.
- Update the affected rustdocs to state the zero-target rejection.

### regression test

Open a channel with an all-zero `max_target` (server) / install a zero target via `set_target` (client); assert rejection at the boundary. PoC (server extended, easy regtest-style `nbits` so a block is found deterministically): https://github.com/project-loupe/audit-stratum/issues/175

### LINKED ISSUE #2334: `channels_sv2`: client `validate_share` rejects a live job as stale when the upstream reuses its `job_id` (`standard` + `extended`)
On both client channel types, `validate_share` checks `stale_jobs` **before** the active and past stores and returns early:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/extended.rs#L635-L641

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/standard.rs#L429-L435

But a job ID can be simultaneously stale and active. In `on_set_new_prev_hash`, the displaced active job is inserted into `stale_jobs` by ID:

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/extended.rs#L587-L592

https://github.com/stratum-mining/stratum/blob/941d004aaa034efd1773109fd1794812b32605c3/sv2/channels-sv2/src/client/standard.rs#L382-L387

If the future job being activated reuses that same `job_id`, the new job ends up in `active_job` *and* in `stale_jobs` under the same key — and every share for it is rejected as `stale-share` without hashing, even though it matches the channel's current job.

The in-tree message definitions do not rule the collision out: `NewExtendedMiningJob::job_id` is documented only as "Upstream's identification of the mining job" with no uniqueness invariant, IDs are `u32`, and jobs can originate from either an individual channel or its group. A malicious upstream can intentionally reuse the current ID at a tip transition to suppress all local share propagation; independent per-scope ID allocation can produce the same failure accidentally, silently wasting mining work.

### impact

Low, matching the audit severity: requires an upstream that reuses a live job ID (maliciously or through buggy ID allocation); the damage is availability — all shares for the new tip are dropped locally until the next job arrives.

### fix

Let the live stores win: only treat the job as stale when it is not also active or past.

```rust
if is_stale_job && !is_active_job && !is_past_job {
    return Err(ShareValidationError::Stale(
        ERROR_CODE_SUBMIT_SHARES_STALE_SHARE,
    ));
}
```

Apply symmetrically in both client channels (the server side resolves jobs through `JobStore`, where an ID maps to exactly one collection, so it does not have this ambiguity). Note the client also keeps the same job in two maps; a more thorough alternative is to drop the stale entry when activation reuses its ID, but the guard above is the minimal correct fix.

### regression test

Activate a next-tip future job whose `job_id` equals the current active job's, then submit a structurally valid share under a permissive target — must be `Valid`, not `Stale`. PoC (client extended): https://github.com/project-loupe/audit-stratum/issues/172

### LINKED ISSUE #2339: `channels_sv2`: client `set_chain_tip` replaces the tip without retiring the jobs built under the previous one (`standard` + `extended`)
Follow-up to #2254, whose part 2 made the client `ExtendedChannel::on_chain_tip_update` retire the active job on a tip change. Both client channels also expose a plain `set_chain_tip` setter that performs none of that lifecycle work, and the standard client has no retiring counterpart at all:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/client/standard.rs#L131-L134

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/client/extended.rs#L180-L183

`validate_share` resolves the job by `job_id` from the active/past stores and then builds the header from the *channel's* current tip:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/client/standard.rs#L435-L438

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/client/standard.rs#L464-L470

So once `set_chain_tip` replaces an existing tip, the old active job still passes `is_active_job` and its share is hashed against the new `prev_hash` with the old job's merkle root — work the miner was never assigned. Under a permissive channel target it comes back `Valid` and is credited to `validated_shares`/`validated_work_sum`; it can also be misreported as `BlockFound`. `seen_shares` is not flushed either, so the dedup cache keeps hashes from the wrong tip.

The message-driven paths get this right — `on_set_new_prev_hash` (both clients) and the extended `on_chain_tip_update` retire the displaced active job into `stale_jobs`, clear future/past jobs and flush `seen_shares` (gated on `prev_hash` inequality by #2329):

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/client/standard.rs#L388-L411

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/client/extended.rs#L539-L567

The setter's rustdoc ("Sets the `ChainTip`") does not say it is initialization-only, so an embedding proxy that converts an externally sourced tip into a `ChainTip` and uses the setter for updates inherits the bug.

### impact

Low (audit severity): requires an embedding application to update an *existing* tip through the setter. No in-tree or sv2-apps production caller does — the translator only seeds a freshly created downstream client channel with it (`chain_tip == None`), and JDC goes through `on_chain_tip_update`. Where it is used that way, the damage is phantom `Valid`/`BlockFound` credit for late work on a header the miner was never assigned, plus a dedup cache carried across tips.

### fix

Make the setter perform the transition whenever it *replaces* a tip, so the lifecycle contract does not depend on which entry point the caller picked. Same shape as the extended `on_chain_tip_update`, on both clients: when `self.chain_tip` is already set and the incoming tip differs, clear future jobs, move the active job and all past jobs into `stale_jobs`, clear past jobs, and flush `seen_shares` — gated on `prev_hash` inequality per #2329's invariant, not on any field change (the audit's suggested patch flushes on any of `{prev_hash, nbits, min_ntime}` differing, which would re-open the replay window #2329 closes). A first-time set (`chain_tip == None`) stays a plain assignment so initialization sequences keep working.

Concretely: let `ExtendedChannel::set_chain_tip` delegate to `on_chain_tip_update` when a tip is already present, and give `StandardChannel` the equivalent routine (today its only retiring path is `on_set_new_prev_hash`, which requires a matching future job). The minimal alternative — documenting the setter as initialization-only — leaves the trap in the public API; retiring in the setter is preferred because the setter is the only tip entry point the standard client offers to TDP-driven embedders.

Update both rustdocs to state the retirement semantics.

### regression test

Set an initial tip, install a job, validate a share for it, replace the tip via `set_chain_tip` with a different `prev_hash`, resubmit the same share — must be `Stale`, not `Valid`. Apply on both clients. PoC (client standard): https://github.com/project-loupe/audit-stratum/issues/214

### LINKED ISSUE #2340: `channels_sv2`: `ChainTip` drops the TDP `SetNewPrevHash.target`, so server `validate_share` classifies `BlockFound` on `nBits` alone (`standard` + `extended`)
`ChainTip` keeps only `{prev_hash, nbits, min_ntime}`; both Template Distribution `From<SetNewPrevHash>` impls discard the message's `target`:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/chain_tip.rs#L16-L20

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/chain_tip.rs#L48-L56

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/chain_tip.rs#L68-L76

The in-tree message definition says that field *is* the block-validity threshold and may be stricter than what `n_bits` implies:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/subprotocols/template-distribution/src/set_new_prev_hash.rs#L40-L43

Both server channels derive the block threshold solely from the tip's `nbits` and enter the `BlockFound` branch whenever that compact target is met:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/extended.rs#L914-L944

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/standard.rs#L780-L810

A share whose hash lies between the two thresholds is therefore reported as a block, credited in share accounting, increments `blocks_found`, and yields a coinbase — even though the template side declared it outside the valid range. Nothing else in the crate (or in sv2-apps) reads the TDP target, so the information is lost at the conversion.

Client channels are not affected: they only ever see the Mining Protocol `SetNewPrevHash`, which carries no `target`, so `nbits` is the only threshold available to them.

### impact

Low (audit severity): with the bitcoind template provider the explicit target equals the `n_bits` target, so mainnet behavior is unaffected today. The gap only bites with a TP that advertises a stricter target (weak-block propagation is the in-spec example), where the pool over-reports blocks and accounts/pays on them.

### fix

Carry the target in `ChainTip` and require it for block classification:

- add a `target: Target` field. `ChainTip::new(prev_hash, nbits, min_ntime)` keeps its signature and derives the field from `nbits` (`Target::from_compact(CompactTarget::from_consensus(nbits))`), so the Mining Protocol conversions, `on_set_custom_mining_job`, and every existing caller are unchanged;
- the two TDP `From` impls take the field from the message's `target` bytes (little-endian, as `Target::from_le_bytes` expects) instead;
- expose `ChainTip::target()`;
- in both server `validate_share` implementations, classify `BlockFound` only when the hash meets *both* `network_target` and `chain_tip.target()`:

```rust
if network_target.is_met_by(share_hash) && chain_tip.target().is_met_by(share_hash) {
```

Keeping `network_target` in the conjunction means a TP that sends a *looser* target cannot loosen the consensus bound; the explicit target can only tighten it.

Update the `ChainTip` and `validate_share` rustdocs.

### regression test

Activate a future job with easy regtest-style `n_bits` (`0x207fffff`) and an all-zero explicit `target`; iterate nonces at max channel target and assert no submission is reported as `BlockFound`. Apply to both server channels. PoC (server extended): https://github.com/project-loupe/audit-stratum/issues/215

### LINKED ISSUE #2341: `channels_sv2`: `GroupChannel` hardcodes `version_rolling_allowed = true`, so a group job imported via `ExtendedChannel::on_group_channel_job` overrides the channel's version-rolling policy
Sibling of #2332 in the same import path. `GroupChannel`'s private constructor hardcodes version rolling into its `JobFactory`, so every broadcast job advertises `version_rolling_allowed: true` regardless of what the pool configured on the channels that will consume it:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/group.rs#L132-L138

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/jobs/factory.rs#L345

`ExtendedChannel::new_for_pool`/`new_for_job_declaration_client` do take a `version_rolling_allowed` policy — but only forward it into the channel's own `JobFactory`, and `on_group_channel_job` rewrites just the imported job's extranonce prefix, trusting everything else:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/extended.rs#L246

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/extended.rs#L584-L604

`validate_share` then consults the *job's* flag:

https://github.com/stratum-mining/stratum/blob/d3a05cfa8b51c6923df81f42bdac7f76b2147830/sv2/channels-sv2/src/server/extended.rs#L869-L897

So in the pool's fan-out an extended channel opened with `version_rolling_allowed = false` gets its first job from its own factory (rolling forbidden) and every later template through the group job (rolling allowed): the channel's own jobs contradict each other, and from the second template on a miner can roll the general-purpose version bits on a channel where the server intended exact-version enforcement, without ever seeing `VersionRollingNotAllowed`.

Standard channels are not affected: `into_standard_job` drops the flag (`NewMiningJob` carries none; version rolling is always allowed for standard jobs), so the group's flag only ever reaches extended channels.

### impact

Low (audit severity): the accepted changes stay within the general-purpose version mask, so this is a policy/authorization bypass, not invalid blocks. No production exposure today — the sv2-apps pool and JDC both open every extended channel with `true`, matching the group's hardcode. It does mean the `version_rolling_allowed` parameter on the `ExtendedChannel` constructors is not honoured for grouped jobs, so a pool cannot actually run a no-rolling policy.

### fix

Fix the source, then guard the sink:

- **Source**: stop hardcoding the flag. `GroupChannel::new_for_pool` and `new_for_job_declaration_client` take `version_rolling_allowed: bool` and pass it to `JobFactory::new`, mirroring the `ExtendedChannel` constructors, so the embedding application builds the group and its extended channels from one policy value. Companion sv2-apps change: the pool (`pool-apps/pool/src/lib/channel_manager/mod.rs`) and JDC (`miner-apps/jd-client/src/lib/channel_manager/mod.rs`) pass the same `true` they already use for their extended channels.
- **Sink**: `on_group_channel_job` refuses a job that is *looser* than the channel's policy, mirroring the `min_ntime` check from #2332 and leaving the channel unchanged:

```rust
if extended_job.version_rolling_allowed() && !self.version_rolling_allowed {
    return Err(ExtendedChannelError::GroupJobVersionRollingNotAllowed);
}
```

`ExtendedChannel` has to keep its policy for that (today it only lives inside the `JobFactory`). A job that is *stricter* than the channel can still be imported: its flag is what the miner was told on the wire, and `validate_share` enforcing it is correct. With the guard in place every stored job carries a flag no looser than the channel's policy, so `validate_share` keeps reading `job.version_rolling_allowed()` unchanged, and what is enforced always matches what was advertised. (The audit's validation-side `self && job` conjunction would instead keep serving a job that told the miner it may roll while rejecting every rolled share.)

Update the `GroupChannel` constructor, `on_group_channel_job` and `validate_share` rustdocs.

### regression tests

- construct a `GroupChannel` with `version_rolling_allowed = false` and assert its jobs advertise `false`;
- construct an extended channel with `false`, import an active job from a group built with `true` — must be rejected with the new error and leave the channel without that job. Adapted from the PoC: https://github.com/project-loupe/audit-stratum/issues/216

plebhash:
close #2329
close #2330
close #2331
close #2332
close #2333
close #2334
close #2339
close #2340
close #2341
close https://github.com/project-loupe/audit-stratum/issues/176
close https://github.com/project-loupe/audit-stratum/issues/180
close https://github.com/project-loupe/audit-stratum/issues/170
close https://github.com/project-loupe/audit-stratum/issues/171
close https://github.com/project-loupe/audit-stratum/issues/174
close https://github.com/project-loupe/audit-stratum/issues/168
close https://github.com/project-loupe/audit-stratum/issues/169
close https://github.com/project-loupe/audit-stratum/issues/178
close https://github.com/project-loupe/audit-stratum/issues/175
close https://github.com/project-loupe/audit-stratum/issues/172
close https://github.com/project-loupe/audit-stratum/issues/217
close https://github.com/project-loupe/audit-stratum/issues/216
close https://github.com/project-loupe/audit-stratum/issues/215
close https://github.com/project-loupe/audit-stratum/issues/214

companion https://github.com/stratum-mining/sv2-apps/pull/826

### DIFF
--- a/sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
+++ b/sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
@@ -775,6 +775,11 @@ impl<'a, T: 'a> Sv2Option<'a, T> {
             _ => unreachable!(),
         }
     }
+
+    /// Borrows the inner value of Sv2Option, if present, like [`Option::as_ref`].
+    pub fn as_ref(&self) -> Option<&T> {
+        self.0.first()
+    }
 }
 
 impl<T> Sv2OptionOwned<T> {
@@ -807,6 +812,11 @@ impl<T> Sv2OptionOwned<T> {
             _ => unreachable!(),
         }
     }
+
+    /// Borrows the inner value of Sv2OptionOwned, if present, like [`Option::as_ref`].
+    pub fn as_ref(&self) -> Option<&T> {
+        self.0.first()
+    }
 }
 
 impl<T: GetSize> GetSize for Sv2Option<'_, T> {
@@ -832,6 +842,16 @@ impl<T: GetSize> GetSize for Sv2OptionOwned<T> {
 #[cfg(test)]
 mod test {
     use crate::{Decodable, Seq064K};
+
+    #[test]
+    fn sv2_option_as_ref_borrows_the_inner_value() {
+        use super::{Sv2Option, Sv2OptionOwned};
+
+        assert_eq!(Sv2OptionOwned::new(Some(7u32)).as_ref(), Some(&7));
+        assert_eq!(Sv2OptionOwned::<u32>::new(None).as_ref(), None);
+        assert_eq!(Sv2Option::new(Some(7u32)).as_ref(), Some(&7));
+        assert_eq!(Sv2Option::<u32>::new(None).as_ref(), None);
+    }
 
     #[test]
     fn seq_of_b032_encodes() {
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -623,13 +623,12 @@ impl ExtendedChannel {
     /// Validates a share prior to submission upstream.
     ///
     /// Updates channel state with the share validation result:
-    /// - Prevents propagation of stale, duplicate, low-difficulty, or low-ntime shares
-    ///   (shares whose `ntime` is below the chain tip's `min_ntime` or the job's own
-    ///   `min_ntime` — an immediately-active job may carry a `min_ntime` later than the chain
-    ///   tip's minimum, so the effective lower bound is the larger of the two — as well as
-    ///   shares whose `ntime` exceeds the chain tip's `min_ntime + MAX_FUTURE_BLOCK_TIME`, see
-    ///   [`MAX_FUTURE_BLOCK_TIME`] for how this clockless upper bound relates to the spec's
-    ///   elapsed-time window).
+    /// - Prevents propagation of stale, duplicate, low-difficulty, or out-of-window shares
+    ///   (shares whose `ntime` is outside `[min_ntime, min_ntime + MAX_FUTURE_BLOCK_TIME]`,
+    ///   where `min_ntime` is the referenced job's: the `SetNewPrevHash` timestamp for a job
+    ///   activated from the future queue, or the value its own message advertised for an
+    ///   immediately-active job; see [`MAX_FUTURE_BLOCK_TIME`] for how this clockless upper
+    ///   bound relates to the spec's elapsed-time window).
     /// - Indicates whether a block was found from the share.
     /// - Maintains local share accounting for later reconciliation with upstream acknowledgements.
     pub fn validate_share(
@@ -700,29 +699,31 @@ impl ExtendedChannel {
         let prev_hash = chain_tip.prev_hash();
         let nbits: CompactTarget = CompactTarget::from_consensus(chain_tip.nbits());
 
-        if share.ntime < chain_tip.min_ntime() {
+        // the share's ntime is bounded by the min_ntime of the job it references: a job
+        // activated from the future queue carries the SetNewPrevHash timestamp that activated
+        // it (which is also the chain tip's), an immediately-active job the value its own
+        // message advertised. Every active or past job carries one: a job without it is a
+        // future job, which is never mined on.
+        let job_min_ntime = job
+            .0
+            .min_ntime
+            .as_ref()
+            .copied()
+            .expect("active and past jobs carry a min_ntime");
+
+        if share.ntime < job_min_ntime {
             return Err(ShareValidationError::Invalid(
                 ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
             ));
         }
 
-        // consensus caps block timestamps at ~2h in the future; the allowance is anchored at
-        // chain-tip receipt, since this crate has no clock (see MAX_FUTURE_BLOCK_TIME)
-        if share.ntime > chain_tip.min_ntime().saturating_add(MAX_FUTURE_BLOCK_TIME) {
+        // consensus caps block timestamps at ~2h in the future; the allowance is anchored at the
+        // receipt of the message that supplied min_ntime, since this crate has no clock (see
+        // MAX_FUTURE_BLOCK_TIME)
+        if share.ntime > job_min_ntime.saturating_add(MAX_FUTURE_BLOCK_TIME) {
             return Err(ShareValidationError::Invalid(
                 ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
             ));
-        }
-
-        // an immediately-active job carries its own min_ntime, which may be later than the
-        // chain tip's minimum; jobs activated from the future queue have it overwritten with
-        // the SetNewPrevHash timestamp, making this check redundant there (and harmless)
-        if let Some(job_min_ntime) = job.0.min_ntime.clone().into_inner() {
-            if share.ntime < job_min_ntime {
-                return Err(ShareValidationError::Invalid(
-                    ERROR_CODE_SUBMIT_SHARES_INVALID_SHARE,
-                ));
-            }
         }
 
         // Only BIP323 general-purpose bits may differ from the job's advertised version.
@@ -2564,9 +2565,20 @@ mod tests {
             .on_new_extended_mining_job(job(2, Some(job_min_ntime)))
             .unwrap();
 
-        // a share in the gap passes the chain-tip bound but not the job's own bound
+        // a share in the gap (at or above the tip's min_ntime, below the job's) is rejected
         let res = channel.validate_share(share(1, 2, job_min_ntime - 1));
         assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
+
+        // the upper bound is anchored at the job's min_ntime as well: one second past it is
+        // rejected, exactly on it is accepted
+        let res = channel.validate_share(share(
+            3,
+            2,
+            job_min_ntime + crate::MAX_FUTURE_BLOCK_TIME + 1,
+        ));
+        assert!(matches!(res.unwrap_err(), ShareValidationError::Invalid(_)));
+        let res = channel.validate_share(share(4, 2, job_min_ntime + crate::MAX_FUTURE_BLOCK_TIME));
+        assert!(matches!(res, Ok(ShareValidationResult::Valid(_))));
 
         let res = channel.validate_share(share(2, 2, job_min_ntime));
         assert!(matches!(res, Ok(ShareValidationResult::Valid(_))));
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -63,7 +63,10 @@ pub type ExtendedJob = (NewExtendedMiningJobOwned, Vec<u8>, Target);
 /// - Past jobs (previously active under the current chain tip, indexed by `job_id`, capped at
 ///   [`MAX_PAST_JOBS`]).
 /// - Stale jobs (previously active and past jobs under the previous chain tip, indexed by
-///   `job_id`).
+///   `job_id`). Upstream job IDs carry no uniqueness guarantee, so a job may be installed under
+///   an ID a stale job holds; an ID names either a live job or a stale one, never both, and the
+///   stale namesake is dropped. A late share for it is then validated against the live job, as
+///   the channel cannot tell the two apart.
 /// - Share accounting for the channel (as tracked by the client).
 /// - The channel's current chain tip.
 #[derive(Debug)]
@@ -359,6 +362,8 @@ impl ExtendedChannel {
                 if let Some(active_job) = self.active_job.take() {
                     self.retire_job_to_past(active_job);
                 }
+                // an ID names either a live job or a stale one, never both
+                self.stale_jobs.remove(&new_extended_mining_job.job_id);
                 self.active_job = Some((
                     new_extended_mining_job,
                     self.extranonce_prefix.as_bytes().to_vec(),
@@ -494,6 +499,8 @@ impl ExtendedChannel {
         if let Some(active_job) = self.active_job.take() {
             self.retire_job_to_past(active_job);
         }
+        // an ID names either a live job or a stale one, never both
+        self.stale_jobs.remove(&new_extended_mining_job.job_id);
         self.active_job = Some((
             new_extended_mining_job,
             self.extranonce_prefix.as_bytes().to_vec(),
@@ -607,6 +614,10 @@ impl ExtendedChannel {
             self.stale_jobs
                 .insert(previously_active_job.0.job_id, previously_active_job);
         }
+
+        // the activated job may reuse the ID of a job that just went stale; an ID names either
+        // a live job or a stale one, never both
+        self.stale_jobs.remove(&set_new_prev_hash.job_id);
 
         // clear past jobs, as we're no longer going to propagate shares for them
         self.past_jobs.clear();
@@ -2671,5 +2682,126 @@ mod tests {
         // exactly on the bound the share is accepted (channel target is permissive)
         let res = channel.validate_share(share(2, tip_ntime + crate::MAX_FUTURE_BLOCK_TIME));
         assert!(matches!(res, Ok(ShareValidationResult::Valid(_))));
+    }
+
+    #[test]
+    fn test_reused_job_id_resolves_to_the_live_job() {
+        // Upstream job IDs carry no uniqueness guarantee: a job may be installed under the ID of
+        // a job that went stale, both when a future job is activated and when an immediately-
+        // active job arrives. An ID names either a live job or a stale one, never both, so the
+        // stale namesake is dropped and shares for the ID validate against the live job.
+        let channel_id = 1;
+        let extranonce_prefix = [
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+        ]
+        .to_vec();
+
+        let mut channel = ExtendedChannel::new(
+            channel_id,
+            "user_identity".to_string(),
+            ExtranoncePrefix::from_wire(extranonce_prefix).unwrap(),
+            Target::from_le_bytes([0xff; 32]),
+            1.0,
+            true,
+            8u16,
+            None,
+        );
+
+        let job = |min_ntime: Option<u32>| NewExtendedMiningJob {
+            channel_id,
+            job_id: 1,
+            min_ntime: Sv2Option::new(min_ntime),
+            version: 536870912,
+            version_rolling_allowed: true,
+            coinbase_tx_prefix: vec![
+                2, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 34, 82, 0,
+            ]
+            .try_into()
+            .unwrap(),
+            coinbase_tx_suffix: vec![
+                255, 255, 255, 255, 2, 0, 242, 5, 42, 1, 0, 0, 0, 22, 0, 20, 235, 225, 183, 220,
+                194, 147, 204, 170, 14, 231, 67, 168, 111, 137, 223, 130, 88, 194, 8, 252, 0, 0, 0,
+                0, 0, 0, 0, 0, 38, 106, 36, 170, 33, 169, 237, 226, 246, 28, 63, 113, 209, 222,
+                253, 63, 169, 153, 223, 163, 105, 83, 117, 92, 105, 6, 137, 121, 153, 98, 180, 139,
+                235, 216, 54, 151, 78, 140, 249, 1, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+            ]
+            .try_into()
+            .unwrap(),
+            merkle_path: vec![].try_into().unwrap(),
+        };
+
+        // job 1 is active under the current tip, and upstream reuses its ID for the next tip's
+        // future job
+        channel
+            .on_new_extended_mining_job(job(Some(1745596970)))
+            .unwrap();
+        channel.on_new_extended_mining_job(job(None)).unwrap();
+
+        // network target: 000000000000d7c0... (hard, so no accidental BlockFound)
+        channel
+            .on_set_new_prev_hash(SetNewPrevHashMp {
+                channel_id,
+                job_id: 1,
+                prev_hash: [
+                    200, 53, 253, 129, 214, 31, 43, 84, 179, 58, 58, 76, 128, 213, 24, 53, 38, 144,
+                    205, 88, 172, 20, 251, 22, 217, 141, 21, 221, 21, 0, 0, 0,
+                ]
+                .into(),
+                nbits: 453040064,
+                min_ntime: 1745596980,
+            })
+            .unwrap();
+
+        // ID 1 names the activated job only: its displaced namesake is gone from the stale set
+        assert_eq!(channel.get_active_job().unwrap().0.job_id, 1);
+        assert!(channel.get_stale_job(1).is_none());
+        assert_eq!(channel.get_stale_jobs_count(), 0);
+
+        // a share for the live job is validated (channel target is permissive)
+        let share = |sequence_number: u32, ntime: u32| SubmitSharesExtended {
+            channel_id,
+            sequence_number,
+            job_id: 1,
+            nonce: 0,
+            ntime,
+            version: 536870912,
+            extranonce: vec![1, 0, 0, 0, 0, 0, 0, 0].try_into().unwrap(),
+        };
+        assert!(matches!(
+            channel.validate_share(share(0, 1745596980)),
+            Ok(ShareValidationResult::Valid(_))
+        ));
+
+        // job 1 goes stale on the next tip transition, and upstream then reuses its ID for an
+        // immediately-active job
+        let mut future_job = job(None);
+        future_job.job_id = 2;
+        channel.on_new_extended_mining_job(future_job).unwrap();
+        channel
+            .on_set_new_prev_hash(SetNewPrevHashMp {
+                channel_id,
+                job_id: 2,
+                prev_hash: [
+                    154, 124, 239, 231, 221, 122, 160, 173, 164, 175, 87, 33, 74, 214, 191, 107,
+                    73, 34, 0, 162, 227, 16, 44, 40, 33, 73, 0, 0, 0, 0, 0, 0,
+                ]
+                .into(),
+                nbits: 453040064,
+                min_ntime: 1745596990,
+            })
+            .unwrap();
+        assert!(channel.get_stale_job(1).is_some());
+        channel
+            .on_new_extended_mining_job(job(Some(1745596990)))
+            .unwrap();
+
+        assert_eq!(channel.get_active_job().unwrap().0.job_id, 1);
+        assert!(channel.get_stale_job(1).is_none());
+        assert!(matches!(
+            channel.validate_share(share(1, 1745596990)),
+            Ok(ShareValidationResult::Valid(_))
+        ));
     }
 }
--- a/sv2/channels-sv2/src/client/extended.rs
+++ b/sv2/channels-sv2/src/client/extended.rs
@@ -12,7 +12,7 @@ use crate::{
         error::ExtendedChannelError,
         share_accounting::{ShareAccounting, ShareValidationError, ShareValidationResult},
     },
-    extranonce_manager::ExtranoncePrefix,
+    extranonce_manager::{prefix::RetiredExtranoncePrefixes, ExtranoncePrefix},
     merkle_root::merkle_root_from_path,
     target::{bytes_to_hex, u256_to_block_hash},
     MAX_EXTRANONCE_LEN, MAX_FUTURE_BLOCK_TIME, VERSION_ROLLING_MASK,
@@ -69,6 +69,8 @@ pub type ExtendedJob = (NewExtendedMiningJobOwned, Vec<u8>, Target);
 ///   the channel cannot tell the two apart.
 /// - Share accounting for the channel (as tracked by the client).
 /// - The channel's current chain tip.
+/// - Extranonce prefixes rotated out of the channel that live jobs were created under (see
+///   [`set_extranonce_prefix`](Self::set_extranonce_prefix)).
 #[derive(Debug)]
 pub struct ExtendedChannel {
     channel_id: u32,
@@ -96,6 +98,9 @@ pub struct ExtendedChannel {
     max_past_jobs: usize,
     share_accounting: ShareAccounting,
     chain_tip: Option<ChainTip>,
+    // extranonce prefixes rotated out of the channel that are still referenced by at least one
+    // job that can accept shares, so that their allocator slots stay reserved
+    retired_extranonce_prefixes: RetiredExtranoncePrefixes,
 }
 
 impl ExtendedChannel {
@@ -138,6 +143,7 @@ impl ExtendedChannel {
             max_past_jobs,
             share_accounting: ShareAccounting::new(),
             chain_tip: None,
+            retired_extranonce_prefixes: RetiredExtranoncePrefixes::default(),
         }
     }
 
@@ -191,6 +197,13 @@ impl ExtendedChannel {
     /// Jobs created before this call retain the previous extranonce prefix,
     /// and share validation is performed accordingly.
     ///
+    /// Because of that, a previous prefix minted by a local
+    /// [`ExtranonceAllocator`](crate::extranonce_manager::ExtranonceAllocator) (e.g. by a proxy
+    /// sub-allocating an upstream-assigned extranonce space) is not released here: its slot
+    /// stays reserved until no future, active or past job created under it remains, so that the
+    /// allocator cannot hand the same extranonce space to another live channel while those jobs
+    /// still validate shares. Wire-sourced prefixes hold no slot and are simply dropped.
+    ///
     /// Returns an error if the new extranonce prefix is too large.
     pub fn set_extranonce_prefix(
         &mut self,
@@ -200,7 +213,16 @@ impl ExtendedChannel {
             return Err(ExtendedChannelError::NewExtranoncePrefixTooLarge);
         }
 
-        self.extranonce_prefix = new_extranonce_prefix;
+        let retired_extranonce_prefix =
+            core::mem::replace(&mut self.extranonce_prefix, new_extranonce_prefix);
+        self.retired_extranonce_prefixes.retire(
+            retired_extranonce_prefix,
+            self.future_jobs
+                .values()
+                .chain(self.active_job.iter())
+                .chain(self.past_jobs.values())
+                .map(|job| job.1.as_slice()),
+        );
 
         Ok(())
     }
@@ -359,16 +381,18 @@ impl ExtendedChannel {
 
         match new_extended_mining_job.min_ntime.clone().into_inner() {
             Some(_min_ntime) => {
-                if let Some(active_job) = self.active_job.take() {
-                    self.retire_job_to_past(active_job);
-                }
                 // an ID names either a live job or a stale one, never both
                 self.stale_jobs.remove(&new_extended_mining_job.job_id);
-                self.active_job = Some((
+                // the new job is installed before the displaced one is retired: retirement may
+                // prune retired extranonce prefixes, and the new job is a live user of its bytes
+                let displaced_job = self.active_job.replace((
                     new_extended_mining_job,
                     self.extranonce_prefix.as_bytes().to_vec(),
                     self.target,
                 ));
+                if let Some(displaced_job) = displaced_job {
+                    self.retire_job_to_past(displaced_job);
+                }
             }
             None => {
                 let job_id = new_extended_mining_job.job_id;
@@ -390,6 +414,11 @@ impl ExtendedChannel {
                         self.future_jobs.remove(&evicted_job_id);
                     }
                 }
+
+                // a replaced or evicted job may have been the last one holding a retired
+                // extranonce prefix alive; release such slots now rather than at the next chain
+                // transition, which the upstream can withhold
+                self.prune_retired_extranonce_prefixes();
             }
         }
 
@@ -496,16 +525,18 @@ impl ExtendedChannel {
             merkle_path: set_custom_mining_job.merkle_path,
         };
 
-        if let Some(active_job) = self.active_job.take() {
-            self.retire_job_to_past(active_job);
-        }
         // an ID names either a live job or a stale one, never both
         self.stale_jobs.remove(&new_extended_mining_job.job_id);
-        self.active_job = Some((
+        // the new job is installed before the displaced one is retired: retirement may prune
+        // retired extranonce prefixes, and the new job is a live user of its bytes
+        let displaced_job = self.active_job.replace((
             new_extended_mining_job,
             self.extranonce_prefix.as_bytes().to_vec(),
             self.target,
         ));
+        if let Some(displaced_job) = displaced_job {
+            self.retire_job_to_past(displaced_job);
+        }
 
         Ok(())
     }
@@ -527,6 +558,24 @@ impl ExtendedChannel {
                 self.past_jobs.remove(&evicted_job_id);
             }
         }
+
+        // a replaced or evicted job may have been the last one holding a retired extranonce
+        // prefix alive; release such slots now rather than at the next chain transition, which
+        // the upstream can withhold
+        self.prune_retired_extranonce_prefixes();
+    }
+
+    // Drops every retired extranonce prefix that no future, active or past job still references,
+    // releasing its allocator slot. Stale jobs are not live: shares against them are rejected,
+    // so they hold no prefix.
+    fn prune_retired_extranonce_prefixes(&mut self) {
+        self.retired_extranonce_prefixes.prune(
+            self.future_jobs
+                .values()
+                .chain(self.active_job.iter())
+                .chain(self.past_jobs.values())
+                .map(|job| job.1.as_slice()),
+        );
     }
 
     /// Handles a [`ChainTip`] update.
@@ -572,6 +621,10 @@ impl ExtendedChannel {
         // clear past jobs, as we're no longer going to propagate shares for them
         self.past_jobs.clear();
         self.past_job_order.clear();
+
+        // the jobs that just went stale can no longer accept shares, so any retired extranonce
+        // prefix they were the last reference to is now releasable
+        self.prune_retired_extranonce_prefixes();
 
         // hashes are retained while prev_hash is unchanged, see ShareAccounting::flush_seen_shares
         if is_new_prev_hash {
@@ -632,6 +685,10 @@ impl ExtendedChannel {
         // clear past jobs, as we're no longer going to propagate shares for them
         self.past_jobs.clear();
         self.past_job_order.clear();
+
+        // the jobs that just went stale can no longer accept shares, so any retired extranonce
+        // prefix they were the last reference to is now releasable
+        self.prune_retired_extranonce_prefixes();
 
         // hashes are retained while prev_hash is unchanged, see ShareAccounting::flush_seen_shares
         if self
@@ -869,7 +926,7 @@ mod tests {
             share_accounting::{ShareValidationError, ShareValidationResult},
             MAX_FUTURE_JOBS, MAX_PAST_JOBS,
         },
-        extranonce_manager::ExtranoncePrefix,
+        extranonce_manager::{ExtranonceAllocator, ExtranonceAllocatorError, ExtranoncePrefix},
     };
     use binary_sv2::Sv2OptionOwned as Sv2Option;
     use bitcoin::Target;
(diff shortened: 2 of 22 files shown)


bit-aloo (commented):
Initial clanker review:

### 1. Share `ntime` bounds regressed from `max(tip, job)` to job-only

Locations: `client/extended.rs:814-835`, `client/standard.rs:585-605`, `server/extended.rs:887-905`, `server/standard.rs:761-780`

HEAD enforced `ntime >= chain_tip.min_ntime()`, and on the client additionally `ntime >= job.min_ntime`, with the upper bound at `chain_tip.min_ntime() + 2h`.

The diff replaces the tip anchor with the job's `min_ntime` instead of combining the two. The spec rule is tip-anchored: `SubmitShares.ntime` must be >= the latest `SetNewPrevHash.header_timestamp`.

This regresses in both directions:

* **Lower bound is loosened.** If an immediately active client job, or a fanned-out server group job, has `job.min_ntime < tip.min_ntime`, shares in `[job.min_ntime, tip.min_ntime)` are now accepted locally. On the client, these get hashed, counted in `validated_shares` / `seen_shares`, and forwarded to a pool that will reject them, so local accounting can drift from acknowledgements.
* **Upper bound is narrowed.** The same backdated job now caps shares at `job.min_ntime + 2h`. For example, with `min_ntime = T - 7300`, a wall-clock-stamped share is rejected as `Invalid` even though it's valid relative to the tip. The `lib.rs` doc's "negligible edge" previously required a >2h-old tip; now it can happen whenever the job's `min_ntime` lags the tip by >2h.

**Fix:** use:

```rust
let min_ntime = chain_tip.min_ntime().max(job_min_ntime);
```

for both bounds. This should keep the new group-job tests passing, since they exercise `job > tip`, while restoring the tip floor.

### 2. Server `set_target` bypasses the new zero-target invariant

Locations: `server/extended.rs:373`, `server/standard.rs:317`

The diff guards `max_target` in `new` / `update_channel`, but the public `set_target` still writes `self.target` unchecked. Every subsequent job then copies that value into `job_id_to_target`.

That's the exact sink the new guard is intended to protect: `difficulty_float()` → `INFINITY` → `share_work_sum`.

The new tests themselves also use this API.

**Fix:** either validate in `set_target` and return the same error, or enforce the invariant once at the point where the target is stored.

## Worth a look

### 3. Live-wins job-ID resolution leaves `stale_jobs` inconsistent

Locations: `client/extended.rs:756-762`, `client/standard.rs:551-557`

The fix itself is correct, but after activation we can have:

```rust
get_stale_job(1).is_some()
```

while job `1` is also active. The new test actually enshrines this state. Any future path that checks `stale_jobs` first could reintroduce the same shadowing bug.

It seems simpler to avoid storing the namesake in the first place. In `on_set_new_prev_hash`, either skip the stale insert when:

```rust
previously_active_job.0.job_id == set_new_prev_hash.job_id
```

or remove the activated ID from `stale_jobs`.

There's also an unavoidable behavior change worth mentioning in the commit message: an in-flight share for the old job `1` will now be checked against the new job `1` and reported as `Invalid` / `DifficultyTooLow` rather than `Stale`.

### 4. Server channels still allow a stale ID to shadow a live one

Locations: `server/extended.rs:806`, `server/standard.rs:~706`

The server still does:

```rust
if is_stale_job {
    return Stale;
}
```

before checking active/past jobs.

Server job IDs are monotonic only per factory. A channel mixing group-channel jobs from `on_group_channel_job` with jobs from its own factory can therefore collide.

Probably low likelihood, but it's the same bug class the client side now fixes. It would be good to make the behavior consistent.

### 5. Pruning runs while the incoming job is outside every collection

Locations: `client/extended.rs:542-563`, `client/standard.rs:353-374`

`store_new_*_job` currently does:

```text
active_job.take()
→ retire_job_to_past() // may prune
→ assign new active
```

Retention is keyed by prefix bytes, so this only matters if the current prefix is byte-identical to a retired allocated prefix. That's probably an edge case, but the server `JobStore` already avoids this window.

The fix is small:

```rust
let prev = self.active_job.replace(new);
if let Some(prev) = prev {
    self.retire_job_to_past(prev);
}
```

### 6. Server share hot path now allocates per share

Locations: `server/extended.rs:887-890`, `server/standard.rs:761-764`

`Sv2OptionOwned<T>` is backed by `Vec<T>`, so:

```rust
get_min_ntime().into_inner()
```

does a clone/allocation/free on every share. HEAD only copied a `u32` from the tip.

A `&self -> Option<&T>` accessor using `self.0.first()`, or caching `min_ntime: Option<u32>` on the job, would avoid this. The client already had this cost, but this introduces it into the server hot path as well.

### 7. Error-code/documentation drift

`update_channel.rs:64` still only lists `invalid-nominal-hashrate` and `invalid-channel-id`; it should also include `max-target-out-of-range`.

In `bip141.rs`, an unsupported flag is reported as `FailedToDeserializeCoinbaseInputs`, while bad witness framing becomes `FailedToDeserializeCoinbaseWitness`. The docs now have to explain these double meanings, which feels like a sign that dedicated variants would be clearer, e.g.:

```rust
UnsupportedBip141Flag(u8)
UnexpectedWitnessFraming
```

Also worth noting: `sv2_to_sv1.rs` is the only external caller, and it now errors where it previously stripped this best-effort. That may be intended, but it is an observable cross-crate behavior change.

## Simplification / duplication

A few places could be simplified:

* `RetiredExtranoncePrefixes::retire` uses a different retention predicate from `prune` (`slot-only` vs `slot && in-use`). Something like `retire = prune(&is_in_use); if holds && in_use { push }` would use one predicate and remove the need for the paragraph explaining the asymmetry.
* `dropped_job` bookkeeping is duplicated four times and only gates a prune that's effectively a length check on an empty `Vec`. Calling prune unconditionally at the end of both insert paths seems simpler and correct.
* `is_extranonce_prefix_in_use` + `prune_retired_extranonce_prefixes` are duplicated across client extended, client standard, and `JobStore`. If `retire` / `prune` accepted an iterator of live prefix bytes instead of a closure, these helpers could collapse into a `.chain(..).map(|j| j.1.as_slice())` at each call site. Full `JobStore` reuse on the client is blocked by `std::collections` vs the client's `hashbrown` alias, so that can probably be a follow-up.
* `chain_tip.as_ref().is_some_and(|t| t.prev_hash() != x)` appears six times and depends on being called before `self.chain_tip = Some(..)`. A `ShareAccounting::on_prev_hash(&U256Owned)` that tracks the previous hash and flushes when it changes would make that ordering impossible to get wrong. `on_set_custom_mining_job` also evaluates the option twice for `is_new_prev_hash` / `is_new_chain_tip`.

There's also quite a bit of test-fixture duplication: the new `job` / `future_job` closures in `client/extended.rs` re-embed the 143-byte coinbase fixture already present in `active_job_template`; `client/standard.rs`'s new `job` closure is effectively `job_template(1, min_ntime)`; and both new server tests repeat the full `NewTemplate` + reward-outputs setup.

## Comment volume

One final cleanup point: the diff adds a lot of repeated rationale in comments.

The zero-target rationale appears in many places across error docs, constructor docs, inline guards, and test preambles, and the wording is already starting to drift. The `prev_hash` flush rationale is also repeated at each call site even though it already lives on `flush_seen_shares`. `RetiredExtranoncePrefixes`, `holds_allocator_slot`, the `ShareAccounting` replay-window section, and most new tests also have fairly long explanatory blocks.

I think keeping each rationale in one canonical place — e.g. the error variant, `flush_seen_shares`, or the commit message — and leaving only short comments at the call sites would make the diff quite a bit smaller and easier to maintain.

plebhash:
review point 1 above lacks context from https://github.com/stratum-mining/sv2-spec/pull/211/commits/acc2530150fd97ceb871daf82c4e43b18f29b87e of PR https://github.com/stratum-mining/sv2-spec/pull/211

(my bad for not making it clear on PR description)

fixing it slightly different from what was suggested via commits 4e5a5ed2 and 0153c5ef

---

all other review points were addressed... not necessarily exactly as reported

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
```diff
@@ -775,6 +775,11 @@ impl<'a, T: 'a> Sv2Option<'a, T> {
             _ => unreachable!(),
         }
     }
+
+    /// Borrows the inner value of Sv2Option, if present.
+    pub fn as_option(&self) -> Option<&T> {
+        self.0.first()
+    }
```
bit-aloo, line 782: Why this?

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
```diff
@@ -807,6 +812,11 @@ impl<T> Sv2OptionOwned<T> {
             _ => unreachable!(),
         }
     }
+
+    /// Borrows the inner value of Sv2OptionOwned, if present.
+    pub fn as_option(&self) -> Option<&T> {
+        self.0.first()
+    }
```
bit-aloo, line 819: Same for this?

sv2/channels-sv2/src/client/extended.rs
```diff
@@ -520,6 +582,38 @@ impl ExtendedChannel {
...
+        past_jobs: &'a HashMap<u32, ExtendedJob>,
+    ) -> impl Fn(&[u8]) -> bool + 'a {
+        move |extranonce_prefix: &[u8]| {
+            future_jobs
+                .values()
+                .chain(active_job.iter())
+                .chain(past_jobs.values())
+                .any(|job| job.1 == extranonce_prefix)
+        }
+    }
```
bit-aloo: We have a closure inside a method? We can make this a member function of Self, and the call `self.is_extranonce_prefix_in_use()` ?

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -33,8 +33,14 @@ use mining_sv2::{
...
-/// A type alias representing a standard mining job tied to a specific `target`.
-pub type StandardJob = (NewMiningJobOwned, Target);
+/// A type alias representing a standard mining job tied to a specific `extranonce_prefix` and
+/// target.
+///
+/// Each job includes:
+/// - A [`NewMiningJob`](mining_sv2::NewMiningJob) message
+/// - The `extranonce_prefix` in use when the job was created
+/// - The target of the job
+pub type StandardJob = (NewMiningJobOwned, Vec<u8>, Target);
```
bit-aloo: Can we have this as struct, hard to identify which positional field is what?

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -333,42 +439,48 @@ impl StandardChannel {
...
+        future_jobs: &'a HashMap<u32, StandardJob>,
+        active_job: &'a Option<StandardJob>,
+        past_jobs: &'a HashMap<u32, StandardJob>,
+    ) -> impl Fn(&[u8]) -> bool + 'a {
+        move |extranonce_prefix: &[u8]| {
+            future_jobs
+                .values()
+                .chain(active_job.iter())
+                .chain(past_jobs.values())
+                .any(|job| job.1 == extranonce_prefix)
```
bit-aloo: Same points as the extended one.

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
plebhash: point 6 on clanker review above

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
plebhash: point 6 on clanker review above

sv2/channels-sv2/src/client/standard.rs
plebhash: this pattern predates the PR and exists in many other places of the crate

it's only in the diff because we're expanding the tuple

I'd rather not blow up scope... our goal here is to address the new batch of Loupe issues, not refactor the crate with minor readability improvements

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
bit-aloo: Ah shoot! I wanted to remove this from points. Sorry for inconvenience.

sv2/channels-sv2/src/client/standard.rs
bit-aloo: Ok, for not doing in this PR.

sv2/channels-sv2/src/client/extended.rs
plebhash: yeah the `extranonce_manager`-related code is getting pretty twisted

> That shape was a workaround. A `&self` predicate can't be called there: it runs inside `RetiredExtranoncePrefixes::prune`/`retire`, which hold `&mut self.retired_extranonce_prefixes`, so the method would borrow the whole struct while one field is mutably borrowed. The associated function over the three job collections kept the borrows disjoint, but it was ugly.
>
> Reworked it instead of moving it: `retire`/`prune` now take the live prefix bytes as an `impl Iterator<Item = &[u8]> + Clone`, and each call site builds it in place from its future, active and past jobs (`self.future_jobs.values().chain(self.active_job.iter()).chain(self.past_jobs.values()).map(..)`). Disjoint field borrows make that compile without a helper, so `is_extranonce_prefix_in_use` is gone from both client channels and from `JobStore`, which had the same shape since #2260. Folded into the "retain rotated-out client extranonce prefixes" commit.

sv2/channels-sv2/src/client/standard.rs
plebhash: well, I forgot this kind of refactor is pretty cheap these days

there you go 1cb7fa61fe4ee1a26a486b752c2157889e698abb

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
plebhash: do you think it's a bad addition to the `binary_sv2` API suface?

looks fine to me tbh

plebhash:
new loupe batch covering `channels_sv2`

already covered on this PR:
- https://github.com/project-loupe/audit-stratum/issues/217

not yet covered on this PR:
- https://github.com/project-loupe/audit-stratum/issues/216 -> https://github.com/stratum-mining/stratum/issues/2339
- https://github.com/project-loupe/audit-stratum/issues/215 -> https://github.com/stratum-mining/stratum/issues/2340
- https://github.com/project-loupe/audit-stratum/issues/214 -> https://github.com/stratum-mining/stratum/issues/2341

adding new commits to cover them

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
bit-aloo: hmm, then lets name it `as_ref`?

sv2/channels-sv2/src/client/standard.rs
```diff
@@ -128,15 +153,73 @@ impl StandardChannel {
...
+            .as_ref()
+            .is_some_and(|previous| previous.prev_hash() != chain_tip.prev_hash());
         self.chain_tip = Some(chain_tip);
+
+        // all other future jobs are now useless
+        self.future_jobs.clear();
+        self.future_job_order.clear();
+
+        // mark all past jobs as stale, so that shares are not propagated
+        self.stale_jobs = self.past_jobs.clone();
```
bit-aloo: considering we are clearning past_job below anyway, we can just take it instead of cloning

```suggestion
        self.stale_jobs = mem::take(&mut self.past_job)
```

sv2/channels-sv2/src/server/standard.rs
```diff
@@ -565,6 +598,19 @@ impl StandardChannel {
...
+                        .chain_tip
+                        .as_ref()
+                        .is_some_and(|chain_tip| min_ntime < chain_tip.min_ntime())
+                    {
+                        return Err(StandardChannelError::JobMinNtimeBelowChainTip);
+                    }
+                }
+
                 // associate the new active job with the current target
                 self.job_id_to_target
```
bit-aloo: Not part of this PR, or set of changes here:

Clanker pointed out this:

```rust
1. Confirmed bug: an evicted past job can drop the new active job's target mapping
Commit 1dd63ee0 establishes that a server channel is fed by two ID factories (its own via on_new_template/on_set_custom_mining_job, and its group channel's via on_group_channel_job), both counting from 1, so a job ID can repeat within one channel. JobStore::add_active_job was hardened for the live-vs-stale case — but the callers weren't.

All four of these sites insert the new job's target before calling add_active_job, then remove the evicted past job's ID:

[server/extended.rs:593-601](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/extended.rs#L593-L601)
[server/extended.rs:665-671](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/extended.rs#L665-L671)
[server/standard.rs:561-569](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/standard.rs#L561-L569)
[server/standard.rs:614-622](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/standard.rs#L614-L622)
When the new job's ID equals the ID of the past job that eviction pops, remove(&evicted_job_id) deletes the entry that was just inserted for the new active job. validate_share then falls through the job_id_to_target lookup at [server/extended.rs:894](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/extended.rs#L894) and rejects every share against the current active job as InvalidJobId until the next job replaces it.

I verified this with a temporary test (max_past_jobs = 2; group jobs 1, 2, 3 via on_group_channel_job, then on_new_template mints own-factory job 1):


thread '...' panicked at server/extended.rs:4632:
the active job lost its target mapping to the eviction of the same-ID past job
The file is restored — the tree is clean.

The fix pattern already exists in this same file: on_set_custom_mining_job at [server/extended.rs:810-818](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/extended.rs#L810-L818) does add_active_job → remove(evicted) → insert(new), in that order, and is not affected. The four sites above just need the same ordering.

2. add_active_job clears a stale namesake but not a past one
[job_store.rs:206-215](vscode-webview://1a3pno24nqql9ntakk0tc01g7b9g5uue6ov273c3g2re8fr7ogcb/sv2/channels-sv2/src/server/jobs/job_store.rs#L206-L215) removes the stale job sharing the new job's ID, but leaves a past job with that ID in place. validate_share checks is_active_job first, so that past job becomes permanently unreachable while still consuming a max_past_jobs slot. The JobStore doc comment states the live-vs-stale invariant but is silent on this. Removing the past namesake there too would make the invariant uniform and would also close finding #1 independently of the caller ordering.
```

bit-aloo (commented):
Should be good to go after these.

sv2/channels-sv2/src/server/standard.rs
plebhash: should be addressed now, thanks

sv2/channels-sv2/src/client/standard.rs
plebhash: done

sv2/binary-sv2/src/datatypes/non_copy_data_types/seq_inner.rs
plebhash: done

bit-aloo (approved):
ACK

PR #2343, opened 2026-09-07
Binary_sv2 combined loupe fixes

### LINKED ISSUE #2344: Prevent overflow in fixed-size decode error offsets
The blanket SizeHint impl for Fixed types computes offset + Self::SIZE in its error branch with plain addition. Since Fixed is public and does not bound SIZE, that sum can overflow. Use saturating addition so the error path cannot panic or report a wrapped length.

### LINKED ISSUE #2345: Redact retained payloads from binary error Debug output
Error derived Debug, so formatting ValueExceedsMaxSize with {:?} printed the retained byte sample. Display already redacts it, but logging and panics use Debug. Implement Debug by hand so the sample is replaced with [redacted] while the size fields are kept.

### LINKED ISSUE #2346: Prevent 32-bit EncodableField size wraparound
GetSize for EncodableField::Struct sums child sizes with unchecked addition. On 32-bit targets a field built from many zero-copy B064K values sharing one buffer can reach 2^32 encoded bytes, which panics in debug and wraps to zero in release. Accumulate with saturating addition

### LINKED ISSUE #2347: Bound recursive EncodableField traversal to prevent stack exhaustion
encode, get_size, and the generated destructor of EncodableField all recurse once per nesting level. A deeply nested value is cheap to build on the heap but overflows the stack in each of them, which aborts the process. Walk the tree with an explicit stack instead.

### LINKED ISSUE #2348: Validate element widths before allocating Seq064K markers
The sequence get_structure impls check only that the declared element count is no larger than the remaining byte count, then allocate a FieldMarker per declared element before checking whether those elements can fit. A 65,537-byte Seq064K<U256> payload declaring 0xffff elements allocates ~65,537 markers (roughly 1.5–2 MiB) and is only rejected later by the caller's size_hint pass. Preflight each element's width against the input before materializing the marker vector.

### LINKED ISSUE #2349: Avoid per-element AST allocation when decoding nested Seq064K
When a sequence is a field of a derived struct, FieldMarker::decode builds a DecodableField for every element before the final Vec<T> is assembled. A valid maximum Seq064K<u16> of about 128 KiB causes a multi-megabyte intermediate allocation. Decode the sequence directly from its bytes instead of expanding it element by element.

### LINKED ISSUE #2359: `B064K` and `B0255` Display hex-encode the full value with no cap
`Seq064K<B016M>` truncates at `MAX_DISPLAY_BYTES` (#2282), but `B064K`'s Display and `B0255::as_hex` still hex-encode the entire value into an intermediate String. A peer-supplied 65,535-byte field (SubmitSolution.coinbase_tx, NewTemplate.coinbase_tx_outputs, RequestTransactionDataSuccess.excess_data, DeclareMiningJob blobs) therefore expands to over 131 KiB of log text per message, plus a same-sized transient allocation.

Fix: apply the same prefix-and-"<truncated N chars>" formatting used for B016M to B064K and
B0255, encoding only the displayed prefix.

### LINKED ISSUE #2360: Derived `Debug` on `Inner` prints every byte as a decimal list
`Inner` and `InnerOwned` derive Debug, so any message holding a B064K or B016M renders the whole payload as `[1, 2, 3, …]` under `{:?}`. A wire-valid 16 MiB B016M in RequestTransactionDataSuccess expands to roughly 80 MiB of text; every derived-Debug message and the parser enums inherit this. Debug is what tracing, `unwrap` panics and error context use.

Fix: hand-written Debug for `Inner` and `InnerOwned` that prints the type name, length, and a bounded hex prefix.

bit-aloo:
closes: #2344 
closes: #2345 
closes: #2346 
closes: #2347 
closes: #2348 
closes: #2349
closes: https://github.com/stratum-mining/stratum/issues/2359
closes: https://github.com/stratum-mining/stratum/issues/2360
closes: https://github.com/project-loupe/audit-stratum/issues/212
closes: https://github.com/project-loupe/audit-stratum/issues/211
closes: https://github.com/project-loupe/audit-stratum/issues/208
closes: https://github.com/project-loupe/audit-stratum/issues/166
closes: https://github.com/project-loupe/audit-stratum/issues/165
closes: https://github.com/project-loupe/audit-stratum/issues/164
closes: https://github.com/project-loupe/audit-stratum/issues/159
closes: https://github.com/project-loupe/audit-stratum/issues/194
closes: https://github.com/project-loupe/audit-stratum/issues/196
closes: https://github.com/project-loupe/audit-stratum/issues/200
closes: https://github.com/project-loupe/audit-stratum/issues/201
closes: https://github.com/project-loupe/audit-stratum/issues/203
closes: https://github.com/project-loupe/audit-stratum/issues/231
closes: https://github.com/project-loupe/audit-stratum/issues/197
closes: https://github.com/project-loupe/audit-stratum/issues/202
closes: https://github.com/project-loupe/audit-stratum/issues/204
closes: https://github.com/project-loupe/audit-stratum/issues/228
closes: https://github.com/project-loupe/audit-stratum/issues/229
closes: https://github.com/project-loupe/audit-stratum/issues/232
closes: https://github.com/project-loupe/audit-stratum/issues/235

### DIFF
--- a/sv2/binary-sv2/src/codec/decodable.rs
+++ b/sv2/binary-sv2/src/codec/decodable.rs
@@ -80,8 +80,9 @@ pub enum PrimitiveMarker {
 
 /// Recursive enum representing data structure fields.
 ///
-/// A `FieldMarker` can either be a primitive or a nested structure. The marker helps the decoder
-/// understand the layout and type of each field in the data, guiding the decoding process.
+/// A `FieldMarker` is a primitive, a nested structure, or a raw span of bytes that the owning
+/// type decodes itself. The marker helps the decoder understand the layout and type of each field
+/// in the data, guiding the decoding process.
 #[derive(Debug, Clone)]
 pub enum FieldMarker {
     /// A primitive data type.
@@ -89,6 +90,9 @@ pub enum FieldMarker {
 
     /// A structured type composed of multiple fields, allowing for nested data.
     Struct(Vec<FieldMarker>),
+
+    /// A span of the given size whose owner decodes it directly from the bytes.
+    Raw(usize),
 }
 
 /// Trait for retrieving the [`FieldMarker`] associated with a type.
@@ -142,6 +146,9 @@ pub enum DecodableField<'a> {
 
     /// Structured field, allowing for nested data structures.
     Struct(Vec<DecodableField<'a>>),
+
+    /// Undecoded bytes, handed to the owning type's `from_bytes`.
+    Raw(&'a mut [u8]),
 }
 
 impl SizeHint for PrimitiveMarker {
@@ -193,6 +200,13 @@ impl SizeHint for FieldMarker {
                 }
                 Ok(size)
             }
+            Self::Raw(size) => {
+                if data.len().saturating_sub(offset) >= *size {
+                    Ok(*size)
+                } else {
+                    Err(Error::ReadError(data.len(), offset.saturating_add(*size)))
+                }
+            }
         }
     }
 }
@@ -240,6 +254,7 @@ impl<'a> From<DecodableField<'a>> for Vec<DecodableField<'a>> {
         match v {
             DecodableField::Primitive(p) => vec![DecodableField::Primitive(p)],
             DecodableField::Struct(ps) => ps,
+            DecodableField::Raw(bytes) => vec![DecodableField::Raw(bytes)],
         }
     }
 }
@@ -367,6 +382,12 @@ impl FieldMarker {
                     decodeds.push(p.decode(head)?);
                 }
                 Ok(DecodableField::Struct(decodeds))
+            }
+            Self::Raw(size) => {
+                if data.len() < *size {
+                    return Err(Error::ReadError(data.len(), *size));
+                }
+                Ok(DecodableField::Raw(&mut data[..*size]))
             }
         }
     }
--- a/sv2/binary-sv2/src/codec/encodable.rs
+++ b/sv2/binary-sv2/src/codec/encodable.rs
@@ -162,38 +162,83 @@ pub enum EncodableField<'a> {
     Struct(Vec<EncodableField<'a>>),
 }
 
-impl EncodableField<'_> {
+impl<'a> EncodableField<'a> {
     /// The `encode` method serializes a field into the destination buffer `dst`, starting
-    /// at the provided `offset`. If the field is a structure, it recursively encodes
-    /// each contained field. If the buffer is too small or encoding fails, the method
-    /// returns an error.
-    pub fn encode(&self, dst: &mut [u8], mut offset: usize) -> Result<usize, Error> {
-        match (self, dst.len() >= offset) {
-            (Self::Primitive(p), true) => p.encode(&mut dst[offset..]),
-            (Self::Struct(ps), true) => {
-                let mut result = 0;
-                for p in ps {
-                    let encoded_bytes = p.encode(dst, offset)?;
-                    offset += encoded_bytes;
-                    result += encoded_bytes;
-                }
-                Ok(result)
+    /// at the provided `offset`. If the field is a structure, it encodes each contained
+    /// field in order. If the buffer is too small or encoding fails, the method returns an
+    /// error.
+    pub fn encode(&self, dst: &mut [u8], offset: usize) -> Result<usize, Error> {
+        if dst.len() < offset {
+            return Err(Error::WriteError(offset, dst.len()));
+        }
+        let mut written = 0;
+        for p in self.primitives() {
+            let at = offset + written;
+            if dst.len() < at {
+                return Err(Error::WriteError(at, dst.len()));
             }
-            (_, false) => Err(Error::WriteError(offset, dst.len())),
+            written += p.encode(&mut dst[at..])?;
+        }
+        Ok(written)
+    }
+
+    fn primitives(&self) -> Primitives<'_, 'a> {
+        // Descending into a top level struct keeps flat messages from allocating a stack.
+        let current = match self {
+            Self::Struct(ps) => ps.iter(),
+            other => core::slice::from_ref(other).iter(),
+        };
+        Primitives {
+            current,
+            stack: Vec::new(),
+        }
+    }
+}
+
+// Yields the primitives of a field in encoding order without recursing, so structural depth
+// cannot exhaust the stack.
+struct Primitives<'s, 'a> {
+    current: core::slice::Iter<'s, EncodableField<'a>>,
+    stack: Vec<core::slice::Iter<'s, EncodableField<'a>>>,
+}
+
+impl<'s, 'a> Iterator for Primitives<'s, 'a> {
+    type Item = &'s EncodablePrimitive<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            match self.current.next() {
+                Some(EncodableField::Primitive(p)) => return Some(p),
+                Some(EncodableField::Struct(ps)) => self
+                    .stack
+                    .push(core::mem::replace(&mut self.current, ps.iter())),
+                None => self.current = self.stack.pop()?,
+            }
         }
     }
 }
 
 impl GetSize for EncodableField<'_> {
     fn get_size(&self) -> usize {
-        match self {
-            Self::Primitive(p) => p.get_size(),
-            Self::Struct(ps) => {
-                let mut size = 0;
-                for p in ps {
-                    size += p.get_size();
+        self.primitives()
+            .fold(0usize, |size, p| size.saturating_add(p.get_size()))
+    }
+}
+
+/// Drops nested structs iteratively, so structural depth cannot exhaust the stack.
+///
+/// Note that this makes [`EncodableField`] impossible to destructure by move: consumers have
+/// to match its variants by reference. It also means drop-check requires an [`EncodableField`]
+/// to go out of scope before the data it borrows from, so a field cannot be declared before the
+/// buffer it points into.
+impl Drop for EncodableField<'_> {
+    fn drop(&mut self) {
+        if let Self::Struct(ps) = self {
+            let mut pending = core::mem::take(ps);
+            while let Some(mut field) = pending.pop() {
+                if let Self::Struct(inner) = &mut field {
+                    pending.append(inner);
                 }
-                size
             }
         }
     }
--- a/sv2/binary-sv2/src/datatypes/non_copy_data_types/mod.rs
+++ b/sv2/binary-sv2/src/datatypes/non_copy_data_types/mod.rs
@@ -28,6 +28,7 @@ use core::fmt::Write as _;
 pub(crate) mod inner;
 mod seq_inner;
 
+use inner::HexPrefix;
 pub use inner::ERROR_SAMPLE_LEN;
 pub(crate) use inner::{Inner, InnerOwned};
 pub use seq_inner::{Seq0255, Seq0255Owned, Seq064K, Seq064KOwned, Sv2Option, Sv2OptionOwned};
@@ -79,8 +80,6 @@ fn bytes_to_hex<'a>(bytes: impl IntoIterator<Item = &'a u8>) -> String {
     hex
 }
 
-const MAX_DISPLAY_BYTES: usize = 250;
-
 impl fmt::Display for Sv2Option<'_, u32> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let inner = self.to_owned().into_inner();
@@ -103,15 +102,13 @@ impl fmt::Display for Sv2OptionOwned<u32> {
 
 impl B0255<'_> {
     pub fn as_hex(&self) -> String {
-        let inner = bytes_to_hex(self.as_bytes());
-        format!("B0255({inner})")
+        format!("B0255({})", HexPrefix(self.as_bytes()))
     }
 }
 
 impl B0255Owned {
     pub fn as_hex(&self) -> String {
-        let inner = bytes_to_hex(self.as_bytes());
-        format!("B0255({inner})")
+        format!("B0255({})", HexPrefix(self.as_bytes()))
     }
 }
 
@@ -137,15 +134,13 @@ impl Str0255Owned {
 
 impl fmt::Display for B064K<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let inner = bytes_to_hex(self.as_bytes());
-        write!(f, "B064K({inner})")
+        write!(f, "B064K({})", HexPrefix(self.as_bytes()))
     }
 }
 
 impl fmt::Display for B064KOwned {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let inner = bytes_to_hex(self.as_bytes());
-        write!(f, "B064K({inner})")
+        write!(f, "B064K({})", HexPrefix(self.as_bytes()))
     }
 }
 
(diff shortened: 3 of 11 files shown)


sv2/binary-sv2/src/codec/decodable.rs
plebhash: > either be a primitive or a nested structure

this is outdated now

plebhash:
commits have titles but no message bodies

as described on [this link](https://chris.beams.io/posts/git-commit/) listed on `CONTRIBUTING.md`, we should start enforcing this on all commits, so that we always have in-tree documented rationales for all code changes

plebhash:
the following issues are asking us to redact tokens from debug logs:

- https://github.com/project-loupe/audit-stratum/issues/197
- https://github.com/project-loupe/audit-stratum/issues/228
- https://github.com/project-loupe/audit-stratum/issues/232

we are not doing that in this PR, which I believe is the correct thing to do, since I don't really see this as a meaningful security issue, and printing tokens in logs can be very useful for monitoring and debugging

---

but if we don't do anything, Loupe is going to continue flagging it

so I think we should add a commit with some comments explaining why this is ok:

```diff
diff --git a/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs b/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs
index 4a984042..d5ac1115 100644
--- a/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs
+++ b/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs
@@ -46,6 +46,11 @@ pub struct AllocateMiningJobTokenSuccess<'decoder> {
     pub request_id: u32,
     /// A token that makes the JDC eligible for committing a mining job for approval/transactions
     /// declaration or for identifying custom mining job on mining connection.
+    ///
+    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
+    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
+    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
+    /// the value is what correlates a declaration across JDC, JDS and pool logs.
     pub mining_job_token: B0255<'decoder>,
     /// Bitcoin transaction outputs added by JDS.
     pub coinbase_outputs: B064K<'decoder>,
diff --git a/sv2/subprotocols/job-declaration/src/declare_mining_job.rs b/sv2/subprotocols/job-declaration/src/declare_mining_job.rs
index 99eff46e..ceaa2f4d 100644
--- a/sv2/subprotocols/job-declaration/src/declare_mining_job.rs
+++ b/sv2/subprotocols/job-declaration/src/declare_mining_job.rs
@@ -16,6 +16,11 @@ pub struct DeclareMiningJob<'decoder> {
     /// Used for pairing request/response.
     pub request_id: u32,
     /// Token received previously through [`crate::AllocateMiningJobTokenSuccess`] message.
+    ///
+    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
+    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
+    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
+    /// the value is what correlates a declaration across JDC, JDS and pool logs.
     pub mining_job_token: B0255<'decoder>,
     /// Header version field.
     pub version: u32,
@@ -83,6 +88,11 @@ pub struct DeclareMiningJobSuccess<'decoder> {
     /// to start mining on a non declared job. If the token is different (irrespective of if the
     /// downstream is already mining using it), the downstream **must** send a `SetCustomMiningJob`
     /// message on each connection which wishes to mine using the declared job.
+    ///
+    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
+    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
+    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
+    /// the value is what correlates a declaration across JDC, JDS and pool logs.
     pub new_mining_job_token: B0255<'decoder>,
 }

diff --git a/sv2/subprotocols/mining/src/set_custom_mining_job.rs b/sv2/subprotocols/mining/src/set_custom_mining_job.rs
index a27b149e..d25df8ff 100644
--- a/sv2/subprotocols/mining/src/set_custom_mining_job.rs
+++ b/sv2/subprotocols/mining/src/set_custom_mining_job.rs
@@ -24,6 +24,11 @@ pub struct SetCustomMiningJob<'decoder> {
     pub request_id: u32,
     /// Provide the information for the upstream to authorize the custom job that has been or will
     /// be negotiated between the Job Declarator Client and Job Declarator Server.
+    ///
+    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
+    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
+    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
+    /// the value is what correlates a declaration across JDC, JDS and pool logs.
     pub token: B0255<'decoder>,
     /// Version field that reflects the current network consensus.
     ///
```

bit-aloo:
> the following issues are asking us to redact tokens from debug logs:
> 
> * [Bound DeclareMiningJob derived Debug output project-loupe/audit-stratum#197](https://github.com/project-loupe/audit-stratum/issues/197)
> * [Bound allocation-success derived Debug output project-loupe/audit-stratum#228](https://github.com/project-loupe/audit-stratum/issues/228)
> * [Bound custom mining job derived Debug output project-loupe/audit-stratum#232](https://github.com/project-loupe/audit-stratum/issues/232)
> 
> we are not doing that in this PR, which I believe is the correct thing to do, since I don't really see this as a meaningful security issue, and printing tokens in logs can be very useful for monitoring and debugging
> 
> but if we don't do anything, Loupe is going to continue flagging it
> 
> so I think we should add a commit with some comments explaining why this is ok:
> 
> ```diff
> diff --git a/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs b/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs
> index 4a984042..d5ac1115 100644
> --- a/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs
> +++ b/sv2/subprotocols/job-declaration/src/allocate_mining_job_token.rs
> @@ -46,6 +46,11 @@ pub struct AllocateMiningJobTokenSuccess<'decoder> {
>      pub request_id: u32,
>      /// A token that makes the JDC eligible for committing a mining job for approval/transactions
>      /// declaration or for identifying custom mining job on mining connection.
> +    ///
> +    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
> +    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
> +    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
> +    /// the value is what correlates a declaration across JDC, JDS and pool logs.
>      pub mining_job_token: B0255<'decoder>,
>      /// Bitcoin transaction outputs added by JDS.
>      pub coinbase_outputs: B064K<'decoder>,
> diff --git a/sv2/subprotocols/job-declaration/src/declare_mining_job.rs b/sv2/subprotocols/job-declaration/src/declare_mining_job.rs
> index 99eff46e..ceaa2f4d 100644
> --- a/sv2/subprotocols/job-declaration/src/declare_mining_job.rs
> +++ b/sv2/subprotocols/job-declaration/src/declare_mining_job.rs
> @@ -16,6 +16,11 @@ pub struct DeclareMiningJob<'decoder> {
>      /// Used for pairing request/response.
>      pub request_id: u32,
>      /// Token received previously through [`crate::AllocateMiningJobTokenSuccess`] message.
> +    ///
> +    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
> +    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
> +    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
> +    /// the value is what correlates a declaration across JDC, JDS and pool logs.
>      pub mining_job_token: B0255<'decoder>,
>      /// Header version field.
>      pub version: u32,
> @@ -83,6 +88,11 @@ pub struct DeclareMiningJobSuccess<'decoder> {
>      /// to start mining on a non declared job. If the token is different (irrespective of if the
>      /// downstream is already mining using it), the downstream **must** send a `SetCustomMiningJob`
>      /// message on each connection which wishes to mine using the declared job.
> +    ///
> +    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
> +    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
> +    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
> +    /// the value is what correlates a declaration across JDC, JDS and pool logs.
>      pub new_mining_job_token: B0255<'decoder>,
>  }
> 
> diff --git a/sv2/subprotocols/mining/src/set_custom_mining_job.rs b/sv2/subprotocols/mining/src/set_custom_mining_job.rs
> index a27b149e..d25df8ff 100644
> --- a/sv2/subprotocols/mining/src/set_custom_mining_job.rs
> +++ b/sv2/subprotocols/mining/src/set_custom_mining_job.rs
> @@ -24,6 +24,11 @@ pub struct SetCustomMiningJob<'decoder> {
>      pub request_id: u32,
>      /// Provide the information for the upstream to authorize the custom job that has been or will
>      /// be negotiated between the Job Declarator Client and Job Declarator Server.
> +    ///
> +    /// Not a secret. The token is a capability scoped to one job declaration, and presenting it
> +    /// requires an authenticated connection as the JDC it was issued to, so a copy recovered from
> +    /// a log grants no access on its own. It is therefore printed in full rather than redacted:
> +    /// the value is what correlates a declaration across JDC, JDS and pool logs.
>      pub token: B0255<'decoder>,
>      /// Version field that reflects the current network consensus.
>      ///
> ```

Adding this guardrail would be helpful. The suggestion was to redact it, though that doesn’t really make sense here. I added an upper bound to `Debug` so we don’t waste memory. I will add the guardrail separately so it doesn’t probe it again. Thanks!

bit-aloo:
> commits have titles but no message bodies
> 
> as described on [this link](https://chris.beams.io/posts/git-commit/) listed on `CONTRIBUTING.md`, we should start enforcing this on all commits, so that we always have in-tree documented rationales for all code changes

Oh! I would always remove them because they’re super verbose. But I will keep them from now on. I will update my other PR with this as well, then.

plebhash (approved):

sv2/binary-sv2/src/datatypes/non_copy_data_types/mod.rs
```diff
@@ -103,15 +102,13 @@ impl fmt::Display for Sv2OptionOwned<u32> {
...
-        format!("B0255({inner})")
+        format!("B0255({})", HexPrefix(self.as_bytes()))
     }
 }
 
 impl B0255Owned {
     pub fn as_hex(&self) -> String {
-        let inner = bytes_to_hex(self.as_bytes());
-        format!("B0255({inner})")
+        format!("B0255({})", HexPrefix(self.as_bytes()))
```
GitGab19, line 111: Fable non-blocking comment:

B0255 is bounded at 255 bytes, so the old output was at most 510 chars and there is nothing to defend against here. With the cap at 250 a full-length value now loses its last 5 bytes in logs, which hurts debugging for zero benefit. Either leave as_hex unbounded for B0255 or bump MAX_DISPLAY_BYTES to 255 so it never truncates. The b0255_as_hex_is_bounded test would change accordingly.

sv2/binary-sv2/src/codec/encodable.rs
```diff
@@ -162,38 +162,69 @@ pub enum EncodableField<'a> {
...
+impl<'s, 'a> Iterator for Primitives<'s, 'a> {
+    type Item = &'s EncodablePrimitive<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            match self.current.next() {
+                Some(EncodableField::Primitive(p)) => return Some(p),
+                Some(EncodableField::Struct(ps)) => self
+                    .stack
+                    .push(core::mem::replace(&mut self.current, ps.iter())),
```
GitGab19, line 214: Fable non-blocking comment:

This push allocates on the first Struct, which is every derived message, so get_size and encode now each heap-allocate once per call where the recursive version allocated nothing. Since EncodableField depth is fixed by the static type nesting, a peer cannot deepen it, so this is defensive rather than a fix for a reachable stack overflow. If we keep the iterative walk, a small inline array for the stack with a Vec fallback would keep the common case allocation-free.

sv2/binary-sv2/src/codec/encodable.rs
```diff
@@ -162,38 +162,69 @@ pub enum EncodableField<'a> {
...
-            Self::Struct(ps) => {
-                let mut size = 0;
-                for p in ps {
-                    size += p.get_size();
+        self.primitives()
+            .fold(0usize, |size, p| size.saturating_add(p.get_size()))
+    }
+}
+
+impl Drop for EncodableField<'_> {
```
GitGab19, line 234: Fable non-blocking comment:

Worth noting in the doc comment that this Drop impl means EncodableField can no longer be destructured by move, so consumers must match by reference. Nothing downstream does that today, but it is an API constraint we are adding on a public type.

GitGab19 (commented):
A couple of nits, non-blocking, by Fable:

```
The body of c94b379e56 says the tree can be "built from peer input", but EncodableField is only constructed from local typed values, so its depth is fixed at compile time. Could the message drop that claim so the in-tree rationale stays accurate? The change is still fine as hardening.
```

sv2/binary-sv2/src/datatypes/non_copy_data_types/mod.rs
bit-aloo: Makes sense, done.

sv2/binary-sv2/src/codec/encodable.rs
bit-aloo: Done

sv2/binary-sv2/src/codec/encodable.rs
bit-aloo: Done

sv2/binary-sv2/src/codec/encodable.rs
```diff
@@ -162,38 +162,78 @@ pub enum EncodableField<'a> {
...
-                for p in ps {
-                    let encoded_bytes = p.encode(dst, offset)?;
-                    offset += encoded_bytes;
-                    result += encoded_bytes;
-                }
-                Ok(result)
+    /// at the provided `offset`. If the field is a structure, it encodes each contained
+    /// field in order. If the buffer is too small or encoding fails, the method returns an
+    /// error.
+    pub fn encode(&self, dst: &mut [u8], offset: usize) -> Result<usize, Error> {
```
GitGab19, line 170: Tiny behavior change: an empty Struct encoded at an offset past dst.len()
used to be a WriteError and is now Ok(0). Probably fine, but if we want to
keep the old contract an up-front `if dst.len() < offset` check restores it.

sv2/binary-sv2/src/codec/encodable.rs
```diff
@@ -162,38 +162,78 @@ pub enum EncodableField<'a> {
...
-                    size += p.get_size();
+        self.primitives()
+            .fold(0usize, |size, p| size.saturating_add(p.get_size()))
+    }
+}
+
+/// Drops nested structs iteratively, so structural depth cannot exhaust the stack.
+///
+/// Note that this makes [`EncodableField`] impossible to destructure by move: consumers have
+/// to match its variants by reference.
```
GitGab19: One more constraint worth recording here: because this is a manual Drop on a
type that borrows, drop-check now requires an EncodableField to be dropped
before anything it borrows from. A local declared before its source, or a
struct holding both the owner and the field, will no longer compile. Nothing
downstream does that today.

GitGab19 (approved):
A couple of non-blocking comments by Fable:

```
The body of c03c8e7e still says the field tree can be "built from peer input".
EncodableField is only constructed from local typed values, so its depth is
fixed at compile time. Could you drop that clause so the in-tree rationale
stays accurate? The change itself is fine as hardening.
```

sv2/binary-sv2/src/codec/encodable.rs
bit-aloo: Added

sv2/binary-sv2/src/codec/encodable.rs
bit-aloo: Appended.

PR #2350, opened 2026-09-07
add generators to fuzz target

### LINKED ISSUE #2338: add generators to fuzz targets
I've been running the fuzz targets for several months now, and it looks like fuzzing with random inputs has reached a plateau. Code coverage hasn't increased for quite some time, and I haven't been able to find any new interesting inputs or behaviors in a long while.

I think this is a good time to introduce generators. I could have done this from the beginning, but I wanted to first see what we could uncover using random bytes alone, without adding any structure or assumptions about the input format.

Now that the random-input approach seems to be yielding diminishing returns, adding generators should allow us to explore more structured parts of the input space and potentially reach code paths that are difficult to trigger with completely random data.

ref:https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html

lucasbalieiro:
closes #2338
this PR add generators for datatypes and protocol messages


fuzz/fuzz_targets/generators.rs
lucasbalieiro, line 1: an AI review of this file would be very welcome.

Most of what I'm doing here is fairly mechanical: implementing fuzz generators for the datatypes defined in the spec. The main thing to verify is that each generator matches the corresponding datatype definition and encoding in the spec.

lucasbalieiro:
This is ready for review. I'll update the corpus after the merge

lucasbalieiro:
I'm gonna put this PR on draft again, because I've found some crashing inputs that needs to be solved before merging this, they are being tracked as sub-issues of https://github.com/stratum-mining/stratum/issues/2338;

Also, we have a few other PR open on the specs repo that might affect some assertions I'm doing here;

Also, I started incorporating some insights from @brunoerg (thanks by the way);

PR #2369, opened 2026-09-10
fix(sv1_api): serialize configure min-bit-count as integer

cfilipescu:
`mining.configure` currently encodes `version-rolling.min-bit-count` as a hexadecimal string: `HexU32Be(16)` becomes `"00000010"`. [BIP 310](https://github.com/bitcoin/bips/blob/master/bip-0310.mediawiki#extension-version-rolling) requires a non-negative JSON integer, so pools that validate the parameter type reject this request.

Serialize the underlying `u32` in both branches that include the minimum bit count. The mask remains a hexadecimal string, and the public API and support for parsing legacy string-valued inputs are preserved. Extend the existing configure test to check outgoing parameter types with and without a mask; the test fails against the old encoder with `String("00000010") != 16`.

This is an API-compatible serialization fix with no new dependencies. The crate is already bumped to the unpublished `6.0.0` on `main`, so no additional version bump is needed. A backport to `5.0.x` would also let existing `stratum-core 0.5.x` consumers adopt the fix without a major dependency upgrade.

Validation:

- `CC= CPATH= cargo test --locked -p sv1_api` — 21 tests passed on Rust 1.75.0.
- `CC= CPATH= cargo +1.89.0 clippy --locked --all-features -- -D warnings` — workspace check using the CI Clippy toolchain.
- `cargo fmt --all -- --check --config-path /dev/null` — Rust 1.75.0; empty configuration matches repository defaults and avoids inherited machine-specific formatting rules.
- `git diff --check`.

Build commands used `TMPDIR` pointing to a writable temporary directory because this machine's `/tmp` quota is exhausted.


cfilipescu:
is there anything missing from this pr?

PR #2388, opened 2026-09-14
enforce version bumps of modified crates via CI

### LINKED ISSUE #2191: enforce versioning rules via CI
# intro

- `stratum` is a monorepo with SRI low-level crates
- `sv2-apps` is a monorepo with SRI app-level crates

all crates across these two repositories have a well-defined dependency hierarchy (see #2192 for a visual representation via `mermaid` diagram)

`CONTRIBUTING.md` establishes the main algorithm for version control:
https://github.com/stratum-mining/stratum/blob/a15c224eed3f5933dd70e955b9c98626d2206b37/CONTRIBUTING.md?plain=1#L104-L130

unfortunately, this algorithm is non-trivial and invites human-error.

# proposal

as discussed in https://github.com/stratum-mining/stratum/pull/2187#issuecomment-4685810154, I propose we create a shell script that enforces the rules established on `CONTRIBUTING.md` 

I should clarify that I'm not pushing for us to remove human responsability from the loop here. Ultimately, PR author is responsible for making sure their contribution abides by repo rules.

but I think we are at a place where we could confidently enforce these rules via github CI.

plebhash:
close #2191


scripts/check-version-bumps.sh
```diff
@@ -0,0 +1,63 @@
...
+# origin/main) but its version is not above the highest version ever published
+# on crates.io. Crates are only published during global releases, so a crate
+# modified since its last release must already carry its version bump.
+
+set -eu
+
+cd "$(git rev-parse --show-toplevel)"
+
+BASE="${1:-origin/main}"
+CHANGED=$(git diff --name-only "$(git merge-base "$BASE" HEAD)" HEAD)
```
GitGab19, line 16: Fable finding:

`git diff` runs rename detection by default, and with `--name-only` a detected rename is listed only under its new path. So a file moved from one crate into another (with mostly unchanged content) is charged only to the destination crate, and the crate it came from is never checked, even though it lost a file. If that file was a private module, `cargo semver-checks` has nothing to say either, so the source crate can end up modified without a bump, which is the case this script exists to catch.

`--no-renames` makes both sides show up (verified: a move with a one-line edit lists just the destination today, both paths with the flag). Files rewritten during a move already come out as delete + add, so behaviour only changes for the near-identical moves.

```suggestion
# Rename detection would list a moved file only under its new path, but the
# crate it left has changed too and must be checked as well.
CHANGED=$(git diff --no-renames --name-only "$(git merge-base "$BASE" HEAD)" HEAD)
```

plebhash: should be addressed now

plebhash:
after this gets approved/merged here, I'll submit the same to `sv2-apps`

apparently https://github.com/stratum-mining/sv2-apps/pull/865 could have been prevented by this

GitGab19 (approved):

PR #2392, opened 2026-09-16
Add noise unauthenticated initiator docs

### LINKED ISSUE #2391: `noise_sv2::Initiator::without_pk()` yields an unauthenticated session indistinguishable from a verified one
When the initiator is built with `Initiator::without_pk()` or `Initiator::new(None)`, `SignatureNoiseMessage::verify_with_now()` takes the `else` branch and returns `true` without checking the certificate, its validity window, or the responder's static key (`signature_message.rs:144-145`). The handshake then completes normally and `step_2()` returns `Ok(NoiseCodec)`, so the caller gets an encrypted channel to a peer whose identity was never authenticated — with no error, warning, or state flag distinguishing it from a verified session.

The doc comment (`initiator.rs:215-220`) says the certificate is not validated and that the mode is "intended for use when both the initiator and responder are within the same network", but nothing in the API enforces that scope and the name `without_pk` does not state the consequence. The mode itself is intentional and legitimate (Noise NX without a pinned authority key); the defect is that it is easy to select by accident and impossible to detect afterwards. The deployment-level consequence — a role configured without an authority key accepting an impersonated Template Provider — is tracked in sv2-apps [#54](https://github.com/stratum-mining/sv2-apps/issues/54) and [#59](https://github.com/stratum-mining/sv2-apps/issues/59).

Fix: make the unauthenticated mode explicit and hard to select accidentally. At minimum, rename `without_pk()` and the `no_std` `without_pk_with_rng()` variant to state the consequence (e.g. `without_server_authentication()`) and add an explicit security warning to the doc comment. The rename is a breaking change for a 2.x crate, so either deprecate the current names in favor of the new ones for one release cycle or take the MAJOR bump. Preferably also have `verify_with_now()`/`step_2()` surface a tri-state result (verified / no certificate required / failed) or require an explicit opt-in instead of `Option::None` silently disabling verification, so callers can log or reject unauthenticated sessions.

cc @bit-aloo

bit-aloo:
closes: https://github.com/stratum-mining/stratum/issues/2391

We are just adding doc guardrails, so the API users are explicitly made aware of this fact.


plebhash:
@bit-aloo the original issue had this suggestion:

> Fix: make the unauthenticated mode explicit and hard to select accidentally. At minimum, rename `without_pk()` and the `no_std` `without_pk_with_rng()` variant to state the consequence (e.g. `without_server_authentication()`) and add an explicit security warning to the doc comment.

I crated these commits:
- ee29d838144e1a99c5c9dbcd46f43e4cfa256870
- 261e8408041a59bf9651dc39dce044ec36e8ad6d

if you agree on cherry-picking them here, I can create a companion PR for `sv2-apps` with the corresponding changes

plebhash (approved):
also fine with proceeding with the PR as is

bit-aloo:
I did see the suggestion. TBH, I am a bit reluctant to make this change right before the release. I think there’s room to improve some of the other APIs around Noise as well, and I would rather think through those first and adapt the all the noise API accordingly.

I am ok to go with the above suggestion (on keeping change localized to unauthenticated APIs), if we have consensus. cc: @GitGab19

plebhash:
sounds good @bit-aloo let's proceed as is

can you rebase against `main`?

bit-aloo:
> sounds good @bit-aloo let's proceed as is
> 
> can you rebase against `main`?

rebased

